PR #14, opened 2022-08-17
Don't create soci directory on `soci create`

rdpsin:

*Issue #, if available:*

*Description of changes:*
This commit removes the creation of a `soci/` directory on `soci
create`, since it is a remnant of the old artifacts store, and not
needed anymore.

The creation of the directory interferes with `soci create`, since the renamed `soci` binary conflicts with the CLI's attempt to create a `soci` directory. 

### DIFF
--- a/soci/soci_index.go
+++ b/soci/soci_index.go
@@ -23,9 +23,7 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
 
-	"github.com/awslabs/soci-snapshotter/fs/config"
 	"github.com/containerd/containerd/content"
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/log"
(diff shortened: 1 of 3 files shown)


vkuzniet (approved):

kzys (approved):

soci/soci_index.go
```diff
@@ -255,7 +253,7 @@ func buildSociLayer(ctx context.Context, cs content.Store, desc ocispec.Descript
 		Digest:         ztocDesc.Digest.String(),
 		OriginalDigest: desc.Digest.String(),
 		Type:           ArtifactEntryTypeLayer,
-		Location:       path.Join(config.SociIndexDirectory, desc.Digest.String()),
+		Location:       desc.Digest.String(),
```
Kern--, line 256: Does location make sense anymore? The only place we're using this metadata is in the CLI which assumes the content is in the containerd content store.

rdpsin: Yeah, I was hesitant to remove it entirely in case it broke something. I will create another PR to remove it.

PR #20, opened 2022-08-17
t.Fatal() doesn't take printf-like specifiers

kzys:
It has to be Fatalf() and %w is only for Errorf.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/fs/span-manager/span_manager.go
+++ b/fs/span-manager/span_manager.go
@@ -303,10 +303,10 @@ func (m *SpanManager) getSpanFromCache(spanId string, offset, size soci.FileSize
 }
 
 func (m *SpanManager) verifySpanContents(compressedData []byte, id soci.SpanId) error {
-	actualDigest := digest.FromBytes(compressedData)
-	expectedDigest := m.ztoc.ZtocInfo.SpanDigests[id]
-	if actualDigest != expectedDigest {
-		return ErrIncorrectSpanDigest
+	actual := digest.FromBytes(compressedData)
+	expected := m.ztoc.ZtocInfo.SpanDigests[id]
+	if actual != expected {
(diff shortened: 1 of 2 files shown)


vkuzniet (approved):

wmesard:
> t.Fatal() doesn't take printf-like specifiers

It's easy enough to read the [documentation](https://pkg.go.dev/testing#T.Logf) and infer otherwise, but yeah. If it's really important (e.g., to capture wrapped errors), you an always pass it through `fmt` first, like:
```
t.Fatal(fmt.Errorf("oopsie: %w", err))
```

wmesard (approved):

fs/span-manager/span_manager.go
```diff
@@ -303,10 +303,10 @@ func (m *SpanManager) getSpanFromCache(spanId string, offset, size soci.FileSize
 }
 
 func (m *SpanManager) verifySpanContents(compressedData []byte, id soci.SpanId) error {
-	actualDigest := digest.FromBytes(compressedData)
-	expectedDigest := m.ztoc.ZtocInfo.SpanDigests[id]
-	if actualDigest != expectedDigest {
-		return ErrIncorrectSpanDigest
+	actual := digest.FromBytes(compressedData)
```
sbuckfelder, line 306: @kzys 
Could you explain why made these changes?  Variable names?  Change in the error?

We'd all really appreciate your thoughts on idiomatic go

kzys: Previously the error only tells that the digests are different. Including the digests in the error helps troubleshooting the issue. For example, if SpanDigests are somehow not initialized, `expected` could be `""`.

Changing the names is probably more about my preference and I'm fine to revert. The function is short and only handling digests. So `Digest` suffix wouldn't help readers understand the code better.

sbuckfelder (approved):

PR #21, opened 2022-08-17
Add document on pull modes

vkuzniet:

*Issue #, if available:*

*Description of changes:*
Document the snapshotter's modes during pull of a container image.
This is re-open of https://github.com/awslabs/soci-snapshotter/pull/13 with the latest comments addressed.


sbuckfelder (approved):

docs/pull-modes.md
```diff
@@ -0,0 +1,30 @@
+## Pull modes of soci-snapshotter
+
+Soci-snapshotter is a remote snapshotter. It is able to lazily load the contents of a container image when a *SOCI index* is present in the remote registry. If a SOCI index is not found, it will download and uncompress the layers at launch time, just like the default snapshotter does. SOCI indices can also be "sparse", meaning that any individual layer may not be indexed. In that case, that layer will be downloaded at launch time, while the indexed layers will be lazily loaded.
```
hanyuel, line 3: Why not be more specific just saying overlay snapshotter not " default snapshotter" which I imagine would raise a question of what is the default?

vkuzniet: Soci-snapshotter does not define which snapshotter will be used for non-soci layers. It downloads/decompresses the layer blob and then gives control to container runtime. That said, the "default snapshotter" should be a better definition here.

docs/pull-modes.md
```diff
@@ -0,0 +1,30 @@
...
+Every layer which is being lazily loaded gets mounted as a fuse mountpoint. Every other layer gets mounted as an overlay layer.
```
wmesard, line 4: "[FUSE](https://en.wikipedia.org/wiki/Filesystem_in_Userspace)". All caps.

docs/pull-modes.md
```diff
@@ -0,0 +1,30 @@
...
+During the pull phase there can be the following cases:
+1. For lazy loading the snapshotter expects the user to pass `--soci-index-digest`, which should be the sha256 of the SOCI index manifest. If it’s not passed, the snapshotter will download and decompress the layers just as the default snapshotter does. It will then print `unable to find SOCI index digest for %v: the container image will be pulled in non-SOCI mode`. It will print `using SOCI index digest: %v` otherwise.
+2. If 1 goes well, the launch is a SOCI use case. On the first layer mount, there will be an attempt to download and parse the SOCI manifest. If this doesn’t go well, there will be the following error in the log: `unable to fetch SOCI artifacts: `, which will indicate that the container image will not be lazily loaded.
```
wmesard, line 7: Nit: "SOCI *index* manifest"

wmesard (approved):
I left a couple of comments about nits. But this looks good to me. If you do decide to create another rev, please format the text down to reasonable number of characters per line. (Obviously, the log snippet is fine as it is.) Markdown files are supposed to be human-readable even when they are unformatted.

PR #22, opened 2022-08-17
Don't create 'soci' directory on `soci create`

rdpsin:
This commit removes the creation of a soci/ directory on soci create, since it is a remnant of the old artifacts store, and not
needed anymore.

### Testing

- `make && make test && make integration`
- Successfully ran `sudo ./soci create $IMAGE`. No `soci` directory was created.

### DIFF
--- a/soci/soci_index.go
+++ b/soci/soci_index.go
@@ -23,9 +23,7 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
 
-	"github.com/awslabs/soci-snapshotter/fs/config"
 	"github.com/containerd/containerd/content"
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/log"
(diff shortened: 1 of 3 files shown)


hanyuel (approved):

soci/soci_index.go
```diff
@@ -255,7 +253,7 @@ func buildSociLayer(ctx context.Context, cs content.Store, desc ocispec.Descript
 		Digest:         ztocDesc.Digest.String(),
 		OriginalDigest: desc.Digest.String(),
 		Type:           ArtifactEntryTypeLayer,
-		Location:       path.Join(config.SociIndexDirectory, desc.Digest.String()),
+		Location:       desc.Digest.String(),
```
Kern--, line 256: Should we just remove location now?

rdpsin: I'm not sure about that, since it will require changes to the artifacts.db schema which will be breaking change. I can submit another PR to do the same.

sbuckfelder:
Could you also include how you tested this?  I'm guessing manual testing.  Maybe put in the commands that you used.....

rdpsin:
@sbuckfelder Added.

vkuzniet (approved):

PR #27, opened 2022-08-19
Update default span size to 4MiB

rdpsin:
*Description of changes:*
Early benchmarks show that setting the default span size as 4MiB
performs better than the current default of 1MiB.

Also updated the `TestSociCreateSparseIndex` integration test since it
was failing.

[Span Size Raw Data.pdf](https://github.com/awslabs/soci-snapshotter/files/9383251/Span.Size.Raw.Data.pdf)

*Testing performed:*
- `make && make check && make test && make integration`
- Manual testing: Able to run containers using the new default value. 

### DIFF
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -42,8 +42,8 @@ var CreateCommand = cli.Command{
 	Flags: []cli.Flag{
 		cli.Int64Flag{
 			Name:  "span-size",
-			Usage: "span size of index. Default is 1 MiB",
-			Value: 1 << 20,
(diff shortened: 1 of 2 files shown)


wmesard (commented):
Suggest you attach a PDF of the spreadsheet with those performance numbers. Our future selves are definitely going to want to refer to that when looking at this change.

Also, Process nit: separable changes should go in separate commits, but no need to fix it this time.

rdpsin:
@wmesard  Done.

vkuzniet (approved):

wmesard (approved):

PR #36, opened 2022-08-23
Move CI checks to new job

Kern--:

*Issue #, if available:*
N/A

*Description of changes:*
This change moves CI checks to a separate job so that its more obvious
whether a CI failure is due to a build/test issue or a policy issue. It
also makes CI fail faster when there are policy issues.

*Testing performed:*
`make check` 

The workflow can be verified in my fork:
https://github.com/Kern--/soci-snapshotter/actions/runs/2913977728

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,14 +7,20 @@ on:
     branches: [ main ]
 
 jobs:
-  build:
+  check:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 20
+      - run: make install-check-tools
+      - run: make check-ltag
+      - run: make check-dco
+      - run: make check-lint
+  test:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
       - run: make
       - run: make test
-      - run: make install-check-tools
-      - run: make check
 
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ CMD=soci-snapshotter-grpc soci
 
 CMD_BINARIES=$(addprefix $(OUTDIR)/,$(CMD))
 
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -64,8 +64,10 @@ install-zlib:
 	@rm -rf zlib-1.2.12
 	@rm -f zlib-1.2.12.tar.gz
 
-# "check" depends "build". out/libindexer.a seems needed to process cgo directives
-check: build check-ltag check-dco
+check: check-ltag check-dco check-lint
+
+# "check-lint" depends "build". out/libindexer.a seems needed to process cgo directives
+check-lint: build
```
rdpsin, line 70: Instead of `build`, can we just use `pre-build`?

.github/workflows/build.yml
```diff
@@ -7,14 +7,20 @@ on:
...
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 20
+      - run: make install-check-tools
+      - run: make check-ltag
+      - run: make check-dco
+      - run: make check-lint
+  test:
```
rdpsin, line 20: This runs in parallel. Does it maybe make more sense to only trigger this if `check` passes?

Kern--: I think it's a better experience for contributors to be able to see all of the automated blockers up front. If you have both a missing license header and a broken test, it would be pretty annoying to have to discover that serially.

rdpsin: Fair enough.

Makefile
Kern--: Yeah, that works and saves 30s-1minute.

rdpsin (approved):

kzys (dismissed):

sbuckfelder (dismissed):

rdpsin (approved):

sbuckfelder (approved):

PR #39, opened 2022-08-25
Fix the size of span content returned from cache

hanyuel:
resolveSpanFromCache should return the span content of requested size not
the content from offset to the end of the span.

*Issue #, if available: resolveSpanFromCache returns span content that is more than needed. It should return the span starting from the offset with the given size, not from the offset to the end of the span. Sometimes we see this error message for on-demand fetch:

> "error":"file.Read: unexpected copied data size for on-demand fetch. read = 4096, expected = 446"

*Description of changes: 
1. fixed the offsets of the span buffer returned for resolveSpanFromCache
2. updated the related Span Manager test

*Testing performed: make test, make integration

### DIFF
--- a/fs/span-manager/span_manager.go
+++ b/fs/span-manager/span_manager.go
@@ -364,7 +364,7 @@ func (m *SpanManager) resolveSpanFromCache(s *span, offsetStart, size soci.FileS
 		if err != nil {
 			return nil, err
 		}
-		return bytes.NewReader(uncompSpanBuf[offsetStart:]), nil
+		return bytes.NewReader(uncompSpanBuf[offsetStart:offsetStart+size]), nil
 	}
 	return nil, ErrSpanNotAvailable
 }
--- a/fs/span-manager/span_manager_test.go
+++ b/fs/span-manager/span_manager_test.go
@@ -135,35 +135,43 @@ func TestSpanManagerCache(t *testing.T) {
 	cache := cache.NewMemoryCache()
 	defer cache.Close()
 	m := New(ztoc, r, cache)
-	m.addSpanToCache("spanId", content)
+	spanID := 0
+	err = m.ResolveSpan(soci.SpanId(spanID), r)
+	if err != nil {
+		t.Fatalf("failed to resolve span 0: %v", err)
+	}
 
 	testCases := []struct {
 		name   string
 		offset soci.FileSize
+		size   soci.FileSize
 	}{
 		{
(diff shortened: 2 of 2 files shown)


fs/span-manager/span_manager_test.go
```diff
@@ -314,6 +322,10 @@ func TestValidateState(t *testing.T) {
 	}
 }
 
+// func TestGetSpanContent(t *testing.T) {
```
sbuckfelder, line 325: Why the commented function?  Is this still to be added?

fs/span-manager/span_manager_test.go
```diff
@@ -135,35 +135,43 @@ func TestSpanManagerCache(t *testing.T) {
 	cache := cache.NewMemoryCache()
 	defer cache.Close()
 	m := New(ztoc, r, cache)
-	m.addSpanToCache("spanId", content)
+	spanID := 0
+	err = m.ResolveSpan(soci.SpanId(spanID), r)
```
sbuckfelder, line 139: any reason on why the move from `addSpanToCache` to `ResolveSpan`?

sbuckfelder (approved):
lgtm, just a few questions for my own knowledge

wmesard (approved):

fs/span-manager/span_manager_test.go
hanyuel: Because ResolveSpan changes the span states which are needed for testing expected behavior of `resolveSpanFromCache`, while `addSpanToCache` just adds a span to cache and doesn't change span states.

fs/span-manager/span_manager_test.go
hanyuel: I will move this out...

fs/span-manager/span_manager.go
```diff
@@ -364,7 +364,7 @@ func (m *SpanManager) resolveSpanFromCache(s *span, offsetStart, size soci.FileS
 		if err != nil {
 			return nil, err
 		}
-		return bytes.NewReader(uncompSpanBuf[offsetStart:]), nil
+		return bytes.NewReader(uncompSpanBuf[offsetStart:offsetStart+size]), nil
```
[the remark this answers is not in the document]
hanyuel: I don't think there's a strong reason of doing that given how `offsetStart` and `size` are calculated.

sbuckfelder (approved):

PR #46, opened 2022-08-29
Use `uint8_t` instead of `int` for `bits`

### LINKED ISSUE #35: Don't use `int` for `bits` field in `gzip_index_point`
Currently, the `bits` field in `gzip_index_point` has the type `int` which is 4 bytes on most implementations. However, it only needs to contain values for 0-7. The problem is compounded because we only serialize/deserialize 1 byte out of the 4. Changing it to use a smaller sized integer which shave off a few bytes, but more importantly, help get rid of pesky bugs like https://github.com/awslabs/soci-snapshotter/issues/24.

Use something like `uint8_t` or `int8_t` to represent bits. 

N/A 

**Additional context**
N/A

rdpsin:

*Issue #, if available:*
Fixes #35 

*Description of changes:*
The `bits` field can only contain values between 0-7. Replacing `int` with `uint8_t` to avoid any potential bugs.

*Testing performed:*

### DIFF
--- a/c/indexer.c
+++ b/c/indexer.c
@@ -58,7 +58,7 @@ void free_index(struct gzip_index *index)
 
 /* Add an entry to the access point list.  If out of memory, deallocate the
    existing list and return NULL. */
-static struct gzip_index *addpoint(struct gzip_index *index, int bits,
+static struct gzip_index *addpoint(struct gzip_index *index, uint8_t bits,
     off_t in, off_t out, unsigned left, unsigned char *window)
 {
     struct gzip_index_point *next;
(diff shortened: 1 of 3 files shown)


c/indexer.c
```diff
@@ -590,7 +590,6 @@ struct gzip_index* blob_to_index(void* buf)
         cur += 8;
         memcpy(&pt->out, cur, 8);
         cur += 8;
-        memset(&pt->bits, 0, sizeof(int));
         memcpy(&pt->bits, cur, 1);
```
wmesard, line 593: Or the less obfuscated
```
pt->bits = *(uint8_t *)cur;
```
and similarly for `pt->in` and `pt->out`, above. But okay.

wmesard (approved):

c/indexer.c
sbuckfelder: Given these recommended changes are unrelated to the PR I'd prefer this to be done in a separate PR just incase it introduces something.  If we even want this.

sbuckfelder (approved):

PR #47, opened 2022-08-29
Improve zTOC build errors

Kern--:

*Issue #, if available:*

N/A

*Description of changes:*

Before this change, when trying to build a zTOC for an uncompressed
docker image layer, `soci` returned an unhelpful error:

```
soci: could not get index: -3
```

This is returning the raw error from `inflate` which fails because
the file is not compressed.

This change improves the error messages when zTOC creation fails by:
1. Checking layer compression type before attempting to build a zTOC
2. Including the layer digest in error messages (since zTOC builds are run in multiple
 goroutines, it is not always clear)
3. Adding additional context around gzip errors

*Testing performed:*

```
```

Manual tests:
Using this dockerfile:
```
FROM alpine:latest

RUN touch /var/soci
```

```
$ docker build -t example .
$ docker image save -o example.tar example
$ sudo ctr i import ./example.tar
$ sudo soci create docker.io/library/example:latest
soci: could not build zTOC for sha256:f5e8680d2a16fbf0f463b28376c15d72d2478b5003f35a8d60b7d7f7ac022f09: only layers compressed with gzip are supported
```

---

For anyone who comes across this as an actual error, the way to avoid it is to comrpess blobs when importing:
```
$ sudo ctr i import --compress-blobs ./example.tar
```

If you've already imported without compression, you will need to purge both the image and the content before importing again. E.g.
```
$ sudo ctr i rm docker.io/library/example:latest
$ sudo ctr content rm sha256:f5e8680d2a16fbf0f463b28376c15d72d2478b5003f35a8d60b7d7f7ac022f09
```


rdpsin (dismissed):

soci/soci_index.go
```diff
@@ -160,7 +160,7 @@ func BuildSociIndex(ctx context.Context, cs content.Store, img images.Image, spa
 		eg.Go(func() error {
 			desc, err := buildSociLayer(ctx, cs, l, spanSize, store, &config)
 			if err != nil {
-				return err
+				return fmt.Errorf("could not build zTOC for %s: %w", l.Digest.String(), err)
```
sbuckfelder, line 163: could this cause a runtime error based on whether `l` has `l.Digest.String()`?

sbuckfelder (commented):
one small question - but lgtm.  Only reason not hitting approve is because integration test is failling?  Flaky?

soci/soci_index.go
Kern--: No, `l.Digest` is not a pointer so it must be present. 

It also wouldn't be a valid descriptor, so it's not clear that it would even be possible to get containerd to accept a manifest like that. And this also isn't the first access to `l.Digest`. `buildSociLayer` calls `contentStore.ReaderAt` which uses the digest to find open the file on disk.

Kern--:
> one small question - but lgtm. Only reason not hitting approve is because integration test is failling? Flaky?

Yes. https://github.com/awslabs/soci-snapshotter/pull/48

kzys (dismissed):

rdpsin (approved):

sbuckfelder (approved):

PR #51, opened 2022-09-01
Add OCI Reference Types support

### LINKED ISSUE #42: Implement Reference Types alongside ORAS

This project currently relies on the image and distribution specifications from the [ORAS](https://github.com/oras-project) project. But that project is being deprecated in favor of the standards out of the Reference Types working group [image-spec](https://github.com/opencontainers/image-spec) and [distribution-spec](https://github.com/opencontainers/distribution-spec), respectively.

* soci-snapshotter and `soci create` will support both ORAS and Reference Types. Which one is used will be selected at runtime by *one of*: a config file option, command line option, or environment variable. (Implementor's choice. But the default should probably change to Reference Types as part of this work.
* References to ORAS in the documentation are updated, accordingly. For example, `docs/GETTING_STARTED.md` should describe how to set up a local Reference Types distro.

## Note: This is a breaking change, since the SOCI index manifest format will be changing slightly.
We should get this work done early, so we break as few people as possible.

We considered just ripping out ORAS outright. That would be less work and more satisfying. But we decided to support both for a time to make the transition smoother.

**Additional context**
N/A

rdpsin:
This commit adds support for OCI Reference Types artifacts. `soci
create` now defaults to creating an ORT-compatible manifest for the
index. Added a flag `--create-oras-manifest` to create an
ORAS-compatible manifest.

Note that it is still not possible to push ORT-compatible indexes. `soci
push` will return an error.

*Issue #, if available:*
Partially resolves #42 

*Description of changes:*
- `soci create` will now create ORT-compatible SOCI indexes. Use the flag `--create-oras-manifest` to create an ORAS-compatible manifest.
- Can't push an ORT manifest to remote; `soci push` will return an error. The ability to push and pull ORT-compatible indexes will be done later.

*Testing performed:*

- Create an index for an image

```
% sudo ./soci create docker.io/library/alpine:latest
layer sha256:213ec9aee27d8be045c6a92b7eac22c9a64b44558193775a1a7f626352392b49 -> ztoc sha256:3038eb20acfde160f329f806a57a1ba2b4c006cdddc3865d550766210953fd2d
```

- View the list of indices
```
% sudo ./soci index list 
DIGEST                                                                     SIZE    IMAGE REF                          PLATFORM       ORAS ARTIFACT
sha256:c0693e27b471979f0df8ab3f3039fd26da77ff393225419e97c673c7062b7ec8    758     docker.io/library/alpine:latest    linux/amd64    false 
```

- You can view the index

```
% sudo ./soci index info sha256:c0693e27b471979f0df8ab3f3039fd26da77ff393225419e97c673c7062b7ec8

{
  "mediaType": "application/vnd.oci.artifact.manifest.v1+json",
  "artifactType": "application/vnd.amazon.soci.index.v1+json",
  "blobs": [
    {
      "mediaType": "application/zstd",
      "digest": "sha256:3038eb20acfde160f329f806a57a1ba2b4c006cdddc3865d550766210953fd2d",
      "size": 29899,
      "annotations": {
        "com.amazon.soci.image-layer-digest": "sha256:213ec9aee27d8be045c6a92b7eac22c9a64b44558193775a1a7f626352392b49",
        "com.amazon.soci.image-layer-mediaType": "application/vnd.oci.image.layer.v1.tar+gzip"
      }
    }
  ],
  "refers": {
    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
    "digest": "sha256:1304f174557314a7ed9eddb4eab12fed12cb0cd9809e4c28f29af86979a3c870",
    "size": 528
  },
  "annotations": {
    "com.amazon.soci.build-tool-identifier": "AWS SOCI CLI",
    "com.amazon.soci.build-tool-version": "0.1"
  }
}
```

- Create an ORAS manifest

```
% sudo ./soci create docker.io/library/alpine:latest --create-oras-manifest                          
layer sha256:213ec9aee27d8be045c6a92b7eac22c9a64b44558193775a1a7f626352392b49 -> ztoc sha256:3038eb20acfde160f329f806a57a1ba2b4c006cdddc3865d550766210953fd2d
```

- 
```
% sudo ./soci index list                                                                             
DIGEST                                                                     SIZE    IMAGE REF                          PLATFORM       ORAS ARTIFACT
sha256:c0693e27b471979f0df8ab3f3039fd26da77ff393225419e97c673c7062b7ec8    758     docker.io/library/alpine:latest    linux/amd64    false    
sha256:dd1b7bb54b210e88f26092d5f0b51c38c682ed910587823431bd4bd64131ab65    765     docker.io/library/alpine:latest    linux/amd64    true 
```

- 
```
% sudo ./soci index info sha256:dd1b7bb54b210e88f26092d5f0b51c38c682ed910587823431bd4bd64131ab65  | jq
{
  "mediaType": "application/vnd.cncf.oras.artifact.manifest.v1+json",
  "artifactType": "application/vnd.amazon.soci.index.v1+json",
  "blobs": [
    {
      "mediaType": "application/zstd",
      "digest": "sha256:3038eb20acfde160f329f806a57a1ba2b4c006cdddc3865d550766210953fd2d",
      "size": 29899,
      "annotations": {
        "com.amazon.soci.image-layer-digest": "sha256:213ec9aee27d8be045c6a92b7eac22c9a64b44558193775a1a7f626352392b49",
        "com.amazon.soci.image-layer-mediaType": "application/vnd.oci.image.layer.v1.tar+gzip"
      }
    }
  ],
  "subject": {
    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
    "digest": "sha256:1304f174557314a7ed9eddb4eab12fed12cb0cd9809e4c28f29af86979a3c870",
    "size": 528
  },
  "annotations": {
    "com.amazon.soci.build-tool-identifier": "AWS SOCI CLI",
    "com.amazon.soci.build-tool-version": "0.1"
  }
}
```


soci/artifacts.go
```diff
@@ -57,6 +57,7 @@ var (
 	bucketKeyPlatform       = []byte("platform")
 	bucketKeyLocation       = []byte("location")
 	bucketKeyType           = []byte("type")
+	bucketKeyORASArtifact   = []byte("oras_artifact")
```
Kern--, line 60: Can we store/display the the media type instead? 

If we expect ORAS to be replaced with ORT, then we probably don't want to anchor our code and CLI to ORAS. And, while I don't expect there to be a third media type in this vein, storing this info as a boolean implicitly ties us to exactly 2 options, one of which is ORAS. It's not unfixable, but we don't need to bake in that inflexibility.

cmd/soci/commands/push.go
```diff
@@ -92,6 +92,9 @@ if they are available in the snapshotter's local content store.
 		}
 
 		indexDesc := indexDescriptors[len(indexDescriptors)-1]
+		if !indexDesc.IsORASManifest {
```
Kern--, line 95: What's missing that prevents pushing ORT manifests? Lack of library support?

soci/soci_index.go
```diff
@@ -57,26 +56,47 @@ var (
...
+	/// Returns an optional reference to an existing manifest.
+	/// For OCI Reference Types, it returns the Refers field.
+	/// For ORAS, it returns the Subject field.
+	GetRefers() *ocispec.Descriptor
+
+	/// Returns the manifest annotations.
+	GetAnnotations() map[string]string
+
+	/// Converts the index to a JSON byte array.
+	ToJSON() ([]byte, error)
```
Kern--, line 81: nit: this suggestion is idiomatic go
```suggestion
	JSON() ([]byte,error)
```
https://go.dev/doc/effective_go#interface-names

soci/soci_index.go
```diff
@@ -57,26 +56,47 @@ var (
...
+
+	/// Returns whether the index is ORAS Manifest.
+	IsORASManifest() bool
```
Kern--, line 84: Is this necessary? Shouldn't this be determined either by inspecting the media type or by type assertion?

soci/soci_index.go
```diff
@@ -57,26 +56,47 @@ var (
...
+	Index       Index
 	ImageDigest digest.Digest
 	Platform    ocispec.Platform
 }
 
-func GetIndexDescriptorCollection(ctx context.Context, cs content.Store, img images.Image) ([]ocispec.Descriptor, error) {
-	descriptors := []ocispec.Descriptor{}
+type IndexDescriptorInfo struct {
+	ocispec.Descriptor
+	IsORASManifest bool
```
Kern--, line 95: Can we use the descriptor's media type instead of wrapping it?

soci/soci_index.go
```diff
@@ -133,7 +157,14 @@ func WithBuildToolVersion(version string) BuildOption {
 	}
 }
 
-func BuildSociIndex(ctx context.Context, cs content.Store, img images.Image, spanSize int64, store orascontent.Storage, opts ...BuildOption) (*SociIndex, error) {
+func WithCreateORASManifest(val bool) BuildOption {
```
Kern--, line 160: Similar to other commens, I would prefer if we took a specific "index type". 

Maybe something like this:
```
type ManifestType string

const (
    ORASManifestType = ManifestType("oras")
    OCIManifestType = ManifestType("oci")
)

func WithManifestType(val ManifestType) BuildOption
```

I'm not totally sold on this exact interface. From the library's perspective, MediaType is the important piece. But requiring a CLI user to type out the full MediaType would be a terrible UX. Maybe there should be a conversion between the CLI name and the real MediaType somewhere in the CLI?

soci/soci_index.go
```diff
@@ -57,26 +56,47 @@ var (
...
-	// descriptors of ztocs
-	Blobs []ocispec.Descriptor `json:"blobs,omitempty"`
-	// descriptor of image manifest
-	Subject ocispec.Descriptor `json:"subject,omitempty"`
+/// Index represents an interface to interact with ORAS/OCI Ref Types.
+/// In the fullness of time, ORAS support will be removed from the snapshotter
+/// and this interface will not be needed.
+type Index interface {
+	/// Returns the media type of the artifact manifest.
+	GetMediaType() string
```
Kern--, line 64: nit: this suggestion is idiomatic go
```suggestion
	MediaType() string
```
This applies to all the getters here.
https://go.dev/doc/effective_go#Getters

cmd/soci/commands/push.go
rdpsin: Yes.

soci/soci_index.go
rdpsin: The reason I chose `GetX` is because having a field and method with the same name causes a conflict. https://go.dev/play/p/g82cyp7RcXu

soci/soci_index.go
rdpsin: I can do something like:

```
manifestType := soci.OCIManifestType
if cliContext.Bool("--create-oras-manifest") {
    manifestType := soci.ORASManifestType
}
BuildSociIndex(..., WithManifestType(manifestType))
```

soci/soci_index.go
rdpsin: True. I put it in there for it to be a convenient method (instead of always doing `index.MediaType() == ExpectedMediaType`, but I can remove it if people feel strongly against it.

soci/soci_index.go
rdpsin: Sure, I'll update it.

soci/soci_index.go
rdpsin: Sure. I'll remove it. I'll keep `IndexDescriptorInfo` struct though, as it can be helpful if we ever want to pass additional info from the library to the CLI (for e.g., index creation time).

soci/artifacts.go
rdpsin: Fair point. I'll update it.

soci/soci_index.go
Kern--: Is there a reason to export the fields if we have getters?

rdpsin: JSON marshaling won't work with unexported fields. There are ways around it though. (For e.g, [this](https://stackoverflow.com/questions/11126793/json-and-dealing-with-unexported-fields))

kzys: If exposing the fields is fine, do we need getters?

rdpsin: The main reason for having an interface with getters instead of a concrete type is because ORAS and ORT have different names for the field that points to the manifest it's referring. It's called `subject` in ORAS, and `refers` in ORT.  

I could do something like:

```
struct Index {
    ...
    // Deserialize into whichever field that manifest supports
    Refers `json:"refers,omitempty"` // For ORT
    Subject `json:"subject, omitempty"` // For ORAS
    ...
}
```
but I'm not a fan of this solution.

kzys: How about using ORT throughout the snapshotter and only converting them to ORAS around registry interactions?

soci/artifacts.go
rdpsin: Done.

soci/soci_index.go
rdpsin: Done.

soci/soci_index.go
rdpsin: I've updated it so that `Index` is no longer an interface. Instead, it is now a `struct` with both `Refers` and `Subject` field. During deserialization, the appropriate field is selected. 

I chose this method because: 
- It is the path of least resistance when we eventually phase out ORAS support, since it will only require removing a single field from the `Index` struct. 
- The snapshotter doesn't need to Subject/Refers field when running the container. The only time we utilize the field is when creating the index during ztoc creation.

soci/soci_index.go
rdpsin: Done.

soci/soci_index.go
```diff
@@ -330,32 +387,23 @@ func WriteSociIndex(ctx context.Context, indexWithMetadata IndexWithMetadata, st
 
 	log.G(ctx).WithField("digest", dgst.String()).Debugf("soci index has been written")
 
+	var refers *ocispec.Descriptor
+	if indexWithMetadata.Index.MediaType == OCIRefTypeManifestMediaType {
+		refers = indexWithMetadata.Index.Refers
+	} else {
+		refers = indexWithMetadata.Index.Subject
+	}
```
kzys, line 395: How about also checking invalid combinations here (e.g. MediaaType is ORT but having Subject)?

rdpsin: Hmm, I don't think it's necessary. This function is called after creating the index, so as long as the index creation logic is correct (which we verify through unit tests), this shouldn't be a problem. I've added a check for `refers == nil` though, which should return an errors on invalid combinations.

soci/soci_index.go
```diff
@@ -51,32 +49,60 @@ const (
 	IndexAnnotationBuildToolIdentifier = "com.amazon.soci.build-tool-identifier"
 	// index annotation for build tool version
 	IndexAnnotationBuildToolVersion = "com.amazon.soci.build-tool-version"
+	// media type for OCI Reference Types manifest
+	OCIRefTypeManifestMediaType = "application/vnd.oci.artifact.manifest.v1+json"
```
wmesard, line 53: We here in 2022 know what "OCIRefType" and "OCI Reference Types" means, because we remember the name of the working group that produced it. But the future versions of ourselves will be confused, and will have trouble googling that phrase to get unconfused. Let's use the term that [the spec](https://github.com/opencontainers/image-spec/blob/ff0257194e20f7fd17ca21c67a2e725f21c4d4f2/artifact.md#oci-artifact-manifest-specification) uses: `OCI Artifact`

cmd/soci/commands/create.go
```diff
@@ -41,15 +45,19 @@ var CreateCommand = cli.Command{
 	ArgsUsage: "[flags] <image_ref>",
 	Flags: []cli.Flag{
 		cli.Int64Flag{
-			Name:  "span-size",
```
sbuckfelder, line 44: why was this changed?

cmd/soci/commands/index/list.go
```diff
@@ -103,10 +103,11 @@ var listCommand = cli.Command{
 
 func writeArtifactEntry(w io.Writer, ae *soci.ArtifactEntry, imageRef string) {
 	w.Write([]byte(fmt.Sprintf(
-		"%s\t%d\t%s\t%s\t\n",
+		"%s\t%d\t%s\t%s\t%v\t\n",
 		ae.Digest,
 		ae.Size,
 		imageRef,
 		ae.Platform,
+		ae.MediaType == soci.ORASManifestMediaType,
```
sbuckfelder, line 111: Is it an ORASManifestMediaType even for the Referrers API?

cmd/soci/commands/push.go
```diff
@@ -133,7 +136,7 @@ if they are available in the snapshotter's local content store.
 			return nil
 		}
 
-		err = oraslib.CopyGraph(context.Background(), src, dst, indexDesc, options)
+		err = oraslib.CopyGraph(context.Background(), src, dst, indexDesc.Descriptor, options)
```
sbuckfelder, line 139: what is happening here?

fs/artifact_fetcher.go
```diff
@@ -165,7 +166,7 @@ func (f *artifactFetcher) Store(ctx context.Context, desc ocispec.Descriptor, re
 	return nil
 }
 
-func FetchSociArtifacts(ctx context.Context, imageRef, indexDigest string, store content.Storage) (*soci.SociIndex, error) {
+func FetchSociArtifacts(ctx context.Context, imageRef, indexDigest string, store content.Storage) (*soci.Index, error) {
```
sbuckfelder, line 169: Are the changes in this file related to the OCI references type support?  If so, how?

fs/fs.go
```diff
@@ -194,7 +194,7 @@ type filesystem struct {
 	metricsController     *layermetrics.Controller
 	attrTimeout           time.Duration
 	entryTimeout          time.Duration
-	sociIndex             *soci.SociIndex
+	sociIndex             *soci.Index
```
sbuckfelder, line 197: what is this change for?

cmd/soci/commands/create.go
rdpsin: I've redefined these flags as `const`s up top. I think it's cleaner than just having bare strings.

cmd/soci/commands/index/list.go
rdpsin: Yes.

cmd/soci/commands/push.go
rdpsin: We pass the index descriptor. It traverses `index.Blobs` array, pushes them to remote, before finally pushing the index itself. [Link](https://pkg.go.dev/oras.land/oras-go/v2#CopyGraph)

fs/artifact_fetcher.go
rdpsin: Not really. As far as the artifact fetcher is concerned, it doesn't really need to know whether it's ORAS or OCI manifest, since the `Blob` field is the same in both of them. The changes are QOL changes made to avoid unnecessary copies.

fs/fs.go
rdpsin: Just a rename. `soci.SociIndex` seemed redundant. And the linter (and me) gets slightly annoyed. Hence this comment: https://github.com/awslabs/soci-snapshotter/blob/main/soci/soci_index.go#L60

soci/soci_index.go
rdpsin: Done.

cmd/soci/commands/create.go
```diff
@@ -29,6 +29,10 @@ import (
 const (
 	buildToolIdentifier = "AWS SOCI CLI"
 	buildToolVersion    = "0.1"
+
+	spanSizeFlag           = "span-size"
+	minLayerSizeFlag       = "min-layer-size"
+	createORASManifestFlag = "create-oras-manifest"
```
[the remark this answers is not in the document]
rdpsin: Sure, done.

soci/soci_index.go
```diff
@@ -330,32 +387,27 @@ func WriteSociIndex(ctx context.Context, indexWithMetadata IndexWithMetadata, st
 
 	log.G(ctx).WithField("digest", dgst.String()).Debugf("soci index has been written")
 
+	var refers *ocispec.Descriptor
+	if indexWithMetadata.Index.MediaType == OCIArtifactManifestMediaType {
```
sbuckfelder, line 391: maybe add a helper function to `Index` to get refers so that we can try to contain the logic in one place behind a function.  Something like : 

```
func (index Index) getRefers() ocispec.Descriptor {
  switch index.MediaType:
    case OCIArtifactManifestMediaType :
.....
}
```

soci/soci_index.go
```diff
@@ -182,19 +218,40 @@ func BuildSociIndex(ctx context.Context, cs content.Store, img images.Image, spa
...
+		MediaType:   imgManifestDesc.MediaType,
+		Digest:      imgManifestDesc.Digest,
+		Size:        imgManifestDesc.Size,
+		Annotations: imgManifestDesc.Annotations,
+	}
+	return NewIndex(ztocsDesc, refers, annotations, config.manifestType), nil
+}
+
+// Returns a new index.
+func NewIndex(blobs []ocispec.Descriptor, subject *ocispec.Descriptor, annotations map[string]string, manifestType ManifestType) *Index {
```
sbuckfelder, line 231: Maybe break this into two different functions and have `NewIndex` switch between them based on `config.Type?`? `NewIndexForOCI` and `NewIndexForORAS` while it dupes logic a bit it should simplify and keep some code together (easier to rip out or extend then an if statement)

rdpsin: Sure.

soci/soci_index.go
rdpsin: Done.

vkuzniet (approved):

sbuckfelder (approved):

cmd/soci/commands/push.go
```diff
@@ -92,6 +92,9 @@ if they are available in the snapshotter's local content store.
 		}
 
 		indexDesc := indexDescriptors[len(indexDescriptors)-1]
+		if indexDesc.MediaType == soci.OCIArtifactManifestMediaType {
```
kzys, line 95: Why 7:15?

kzys (approved):

cmd/soci/commands/push.go
rdpsin: Just the first 7 characters of the digest 'sha256:abcdefg...'

Issue #43, opened 2022-08-26
Define SOCI terminology


Kern--:
**Is your feature request related to a problem? Please describe.**
SOCI introduces a handful of new terms. Some, such as SOCI itself, are defined in the README. Others, such as zTOC, are mentioned in various places, but never explicitly defined. Still others, e.g. SOCI index and SOCI index manifest, are neither defined nor obvious in their distinction. 

We should add documentation with specific definitions for these terms both to help new users/contributors understand the project and to avoid talking past each other by using terms inconsistently. 

**Describe the solution you'd like**
A document (possibly the readme) with specific definitions for:
```
SOCI
SOCI index
SOCI index manifest
zTOC
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #52, opened 2022-09-02
Add glossary

Kern--:

*Issue #, if available:*
Closes https://github.com/awslabs/soci-snapshotter/issues/43

*Description of changes:*
This change adds a project glossary to make it easier to navigate the project.

*Testing performed:*
N/A


GLOSSARY.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Terminology 
+
+__SOCI__: Seekable OCI (pronounced so-CHEE). SOCI combines an unmodified [OCI Image](https://github.com/opencontainers/image-spec/blob/v1.0.2/spec.md) (or Docker v2 image) with a SOCI index to enable the SOCI snapshotter to lazily pull the image at runtime.
+
+__SOCI Index__: The combined set of metadata that enables lazy loading of an unmodified OCI image. Conceptually, the SOCI Index is the mapping between files and their locations within a compressed image. In practice, a SOCI Index is made up of a SOCI Index Manifest and zero or more zTOCs. Within a SOCI Index there may be zero or one zTOCs per image layer.
+
+__SOCI Index Manifest__: An [ORAS manfest](https://github.com/oras-project/artifacts-spec/blob/v1.0.0-rc.2/artifact-manifest.md) (soon to be an [OCI Reference Types manifest](https://github.com/opencontainers/wg-reference-types/blob/256c257cc8b725fd324722ee40ead6925b1c8ad8/docs/proposals/PROPOSAL_E.md)) containing the list of zTOCs in the SOCI Index as well as a reference to the image for which the manifest was generated.
```
rdpsin, line 9: np: 'manfest' -> 'manifest'

GLOSSARY.md
```diff
@@ -0,0 +1,25 @@
...
+
+__SOCI Index__: The combined set of metadata that enables lazy loading of an unmodified OCI image. Conceptually, the SOCI Index is the mapping between files and their locations within a compressed image. In practice, a SOCI Index is made up of a SOCI Index Manifest and zero or more zTOCs. Within a SOCI Index there may be zero or one zTOCs per image layer.
+
+__SOCI Index Manifest__: An [ORAS manfest](https://github.com/oras-project/artifacts-spec/blob/v1.0.0-rc.2/artifact-manifest.md) (soon to be an [OCI Reference Types manifest](https://github.com/opencontainers/wg-reference-types/blob/256c257cc8b725fd324722ee40ead6925b1c8ad8/docs/proposals/PROPOSAL_E.md)) containing the list of zTOCs in the SOCI Index as well as a reference to the image for which the manifest was generated. 
+
+__zTOC__: A Table of Contents for compressed data. A zTOC contains a set of spans of data, filesystem metadata, and locations of files within spans. In SOCI, zTOCs are many-to-one with image layers, that is, there may be many different zTOCs for a single image layer.
```
rdpsin, line 11: zTOCs don't contain spans. Spans are lazily-loaded from remote. zTOC contains dictionaries, where each dictionary contains the data the individually decompress spans.

Kern--: Yeah, I was being a bit lazy with spans. I used the term "span metadata" when describing how we use zTOCs to extract files. Would that be clear enough here?

I should also highlight that we pull individual spans in the example below as well.

rdpsin: I'm fine with "span metadata". If I'm being pedantic, "span metadata" and "dictionary" would be different terms. But maybe that distinction should be part of separate "Deep Dive" document.

GLOSSARY.md
```diff
@@ -0,0 +1,25 @@
...
+
+zTOCs are used by the SOCI snapshotter to extract individual files from an image layer using the following algorithm:
+1) Use the file locations from the zTOC to determine which spans contain the file
```
rdpsin, line 14: np: 'locations' -> 'metadata'

GLOSSARY.md
```diff
@@ -0,0 +1,25 @@
...
+2) Use the span metadata to determine where these spans exist in the compressed layer
+3) Use the compression context from the span metadata as well as the information from step 2 to decompress those spans out of the compressed layer
```
rdpsin, line 16: Maybe add 'Fetch the span from remote if not present locally' before 3?

GLOSSARY.md
wmesard: I would explicitly call out that the zTOC is composed of two separate parts:
1. A table of contents containing file metadata and its offset in the tar file. (That the "TOC".)
2. A collection of "snapshots" of the decompression dictionary at various points in the gzip file. (That's the "z".

GLOSSARY.md
```diff
@@ -0,0 +1,25 @@
...
+4) Use the file locations to get the subset of bytes within the spans that correspond to the file
+5) Combine the file data with the filesystem metadata to reconstruct the full file
```
wmesard, line 18: Does this belong in a glossary? We should write a theory of operation document at some point. I think this discussion should go there.

GLOSSARY.md
```diff
@@ -0,0 +1,25 @@
...
+
+__span__: A chunk of data that can be independently decompressed. A zTOC contains periodic checkpoints of compression state from which a process can resume decompression. The chunk of data between two checkpoints is a span.
+
+
+## Anti-terminology
+
+__SOCI Image__: We generally avoid the term "SOCI Image" because within SOCI, the image is an unmodified OCI image. A single image may have many SOCI indices with different parameters such as span size, layers indexed, etc. The precise way to refer to an image that has a SOCI index is to refer to the index itself.
```
wmesard, line 25: Suggest saying it more succinctly and whimsically:
> We generally avoid the term "SOCI Image" because there is no such thing! The image is an unmodified OCI image. Also, a single image may have...

wmesard (changes requested):
How about some `*` bullets to help make the start of the definition stand out?

Glossaries should be alphabetize, as is tradition. So "span" above "zTOC".

GLOSSARY.md
Kern--: Decomposing the z and TOC is a good suggestion.

I was trying to avoid getting in to a technical specification and keep this at the conceptual level. In particular, I don't think gzip or dictionaries have a place in the *definition* of a zTOC. Gzip is definitely not required and while my understanding of the theory of compression is pretty surface-level, I don't think dictionary coders are either.

GLOSSARY.md
Kern--: Fair point. Removed.

Kern--:
I'm not entirely sure what's going on with the failed tests, but it looks like we're doing subtests incorrectly somewhere:

```
=== CONT  TestLazyPullWithSparseIndex/Soci
[11823](https://github.com/awslabs/soci-snapshotter/runs/8164278961?check_suite_focus=true#step:5:11824)
    testing.go:1169: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
````

File: README.md [hunk 7]
```diff
@@ -90,3 +90,6 @@ the changes were fundamental enough that the decision was made to create soci-sn
 project.  Soci-snapshotter builds on stargz's success and innovative ideas.  Long term, this project intends
 and hopes to join [containerd](https://github.com/containerd/containerd) as a non-core project and intends to
 follow CNCF best practices.
+
+## Terminology
```
sbuckfelder, line 94: If getting started ever gets added this section would be slightly different as I've changed the README to include a learning more section

GLOSSARY.md
```diff
@@ -0,0 +1,18 @@
...
+
+* __span__: A chunk of data that can be independently decompressed. A zTOC contains periodic "snapshots" of compression state from which a process can resume decompression. The chunk of data between two checkpoints is a span.
+
+* __zTOC__: A Table of Contents for compressed data. A zTOC is composed of 2 parts. 1) a table of contents containing file metadata and its offset in the decompressed archive file (the TOC). 2) A collection of "snapshots" of the decompression state at various points in the compressed file. (the "z")
+
+
+## Anti-terminology
+
+* __SOCI Image__: We generally avoid the term "SOCI Image" because there is no such thing! The image is an unmodified OCI image. Also, a single image may have many SOCI indices with different parameters such as span size, layers indexed, etc. The precise way to refer to an image that has a SOCI index is to refer to the index itself.
```
sbuckfelder, line 18: 👍

sbuckfelder:
I realize I'm late to this party, but I personally don't think we need to adhere to alphabetical order for such a small list or terms.  Alphabetical order is used to help a reader identify a word in a large list of words.   Since our list is small we _can_ (and in my opinion _should_) order by the priority that we think will help the reader the most.

Just a passing opinion - definitely not a blocker

GLOSSARY.md
wmesard: 1/ I was thinking that for clarity it was okay to be gzip-centric for now. That is, more concrete language would make it easier for the reader to understand what's going on. When and if we add zstd support, we could revise this definition.

But if you feel differently, that's fine too. How about something like:
> A collection of "snapshots" of the state of the compression engine at various points in the layer.

2/ I feel pretty comfortable betting that we will only ever support TAR archives. The only thing that will displace TAR is OCI v2.0. And I feel equally comfortable betting that it will have intrinsic support for lazy-loading, and we can all retire.

Kern--: I would prefer not to be gzip centric here. A doc explaining the technical details of how the snapshotter works can lean on the concreteness of gzip.

wmesard (approved):

rdpsin (dismissed):

sbuckfelder (approved):

wmesard (approved):

Issue #49, opened 2022-08-30
Re-enable nginx test in `TestOptimizeConsistentSociArtifact`


rdpsin:
**Is your feature request related to a problem? Please describe.**
#48 disables the nginx test in `TestOptimizeConsistentSociArtifact` due to random failure caused due to how gob encodes FileMetadata.Xattrs (a map). Once we move away from `gob`, we should re-enable that test. This issue a placeholder so that the TODO isn't lost in the comments.

**Describe the solution you'd like**
nginx test in `TestOptimizeConsistentSociArtifact` is re-enabled.

PR #55, opened 2022-09-08
Ztoc serialization using Flatbuffers

vkuzniet:

*Issue #, if available:*

*Description of changes:*
This PR introduces the change in the serialization format for Ztoc to allow for portability. It's not serialized using Go's `gob` encoder anymore, but rather uses `flatbuffers` (see https://google.github.io/flatbuffers/), which has extensive support for multiple languages and platforms. Flatbuffers allow for extremely fast deserialization times without sacrificing the size of the ztoc.

*Testing performed:*
1. `make check && make test && make integration` pass
2. Manually tested the creation of the index, lazy pulling and successful run of the following container images: rethinkdb, drupal, rabbitmq, tomcat, redis


vkuzniet:
> I think we should use an int for `ztoc.Version` and set it to `1`. Then, any updates to Ztoc, we increment the version number. Maybe we should also add an annotation to the blob descriptor about which version of ztoc it is?

Keeping it as string allows to set version as `major.minor`, which allows for better visibility into changes. Increment in minor version will tell that this is not a breaking change. Increment in major version may mean a breaking change.

vkuzniet:
> Can you re-enable the nginx test (#49)?

Done.

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,47 @@
+namespace ztoc;
```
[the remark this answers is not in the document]
vkuzniet: All set. Updated to use tabs.

soci/ztoc_builder.go
```diff
@@ -69,7 +71,7 @@ func BuildZtoc(gzipFile string, span int64, cfg *buildConfig) (*Ztoc, error) {
 	}
 
 	return &Ztoc{
-		Version:              "0.1",
+		Version:              "0.9",
```
rdpsin, line 74: 0.9?

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,47 @@
...
+
+table Xattr {
+	key : string;
+	value : string;
+}
+
+table FileMetadata {
```
kzys, line 8: - Are they all signed? https://google.github.io/flatbuffers/flatbuffers_guide_writing_schema.html

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,47 @@
...
+	spanEnd : int;
+	firstSpanHasBits : bool;    // Flag for if there is partial uncompressed data that is stored in the previous byte
+	linkname : string;          // Target name of link (valid for TypeLink or TypeSymlink)
+	mode : long;                // Permission and mode bits
+	uid : int;                  // User ID of owner
+	gid : int;                  // Group ID of owner
+	uname : string;             // User name of owner
+	gname : string;             // Group name of owner
+
+	modTime  : string;          // Modification time
```
kzys, line 23: We should include the expected datetime format here.

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,47 @@
...
+
+	xattrs : [Xattr];
+}
+
+table ZtocInfo {
+	spanDigests : [string];
+}
+
+table Ztoc {
+	version : string;
```
kzys, line 35: Likewise, we should mention its expected format.

soci/ztoc_builder.go
```diff
@@ -81,34 +83,113 @@ func BuildZtoc(gzipFile string, span int64, cfg *buildConfig) (*Ztoc, error) {
...
-	if err != nil {
-		return nil, ocispec.Descriptor{}, fmt.Errorf("cannot create zstd writer: %w", err)
+func ztocToFlatbuffer(ztoc *Ztoc) []byte {
+	builder := flatbuffers.NewBuilder(0)
+	version := builder.CreateString(ztoc.Version)
+	buildToolIdentifier := builder.CreateString(ztoc.BuildToolIdentifier)
+
+	metadataOffsetList := make([]flatbuffers.UOffsetT, len(ztoc.Metadata))
+	for i := len(ztoc.Metadata) - 1; i >= 0; i-- {
+		me := ztoc.Metadata[i]
```
kzys, line 103: Better to extract a few functions from the loop.

soci/fbs/ztoc.fbs
vkuzniet: Yep. I'm duplicating what we already have for Ztoc struct. `FileSize` and `SpanId` are signed integers and they translate to `int64` and `int32` respectively.

soci/ztoc_builder.go
vkuzniet: Honestly, I was reluctant to put 1.0 just yet. The ztoc doesn't change, e.g. we do not add/change any fields. The only change is how it is serialized. We can put 1.0 upon concluding on the ztoc format.

soci/fbs/ztoc.fbs
vkuzniet: Done.

soci/fbs/ztoc.fbs
vkuzniet: Done.

soci/ztoc_builder.go
```diff
@@ -81,34 +83,113 @@ func BuildZtoc(gzipFile string, span int64, cfg *buildConfig) (*Ztoc, error) {
...
+
+		off := ztoc_flatbuffers.FileMetadataEnd(builder)
+		metadataOffsetList[i] = off
 	}
-
-	if _, err := zs.Write(serializedBuf.Bytes()); err != nil {
-		return nil, ocispec.Descriptor{}, fmt.Errorf("cannot compress ztoc: %w", err)
+	ztoc_flatbuffers.ZtocStartMetadataVector(builder, len(ztoc.Metadata))
+	for i := len(metadataOffsetList) - 1; i >= 0; i-- {
+		builder.PrependUOffsetT(metadataOffsetList[i])
```
sbuckfelder, line 161: any reason why this cannot be in the loop above?

vkuzniet: For flatbuffers, you first need to prepare the list of the offsets, then proceed with the placement. You cannot start vector and then start preparing offsets and placing them.

soci/ztoc_builder.go
vkuzniet: Done.

soci/ztoc_builder.go
sbuckfelder: :+1:

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,48 @@
+namespace ztoc;
+
+table Xattr {
```
rdpsin, line 3: Maybe we can use `struct` here? (https://google.github.io/flatbuffers/md__schemas.html)

wmesard:
> > I think we should use an int for `ztoc.Version` and set it to `1`. Then, any updates to Ztoc, we increment the version number. Maybe we should also add an annotation to the blob descriptor about which version of ztoc it is?
> 
> Keeping it as string allows to set version as `major.minor`, which allows for better visibility into changes. Increment in minor version will tell that this is not a breaking change. Increment in major version may mean a breaking change.

Flatbuffers schemas have explicit support for [versioning and compatibility](http://google.github.io/flatbuffers/flatbuffers_guide_writing_schema.html#autotoc_md34). So if we do it right, this field will only be for human consumption. `maj.min` makes sense to me. It will presumably be tracking an "official" schema definition for the file format.

rdpsin (approved):

kzys:
+1 for having `maj.min`. While we will do our best to keep zTOC backward-compatible, there might be time we want to break the compatibility. So I want to have the way to do that. For example, if a payload's major version != the decoder's major version, the decoder could ignore the payload.

soci/fbs/ztoc.fbs
kzys: >Use this for simple objects where you are very sure no changes will ever be made

I'd stay with `table`. We need to support whatever filesystems supports and we have less control on the schema.

While extended attributes in popular OSes are all key-value pairs, apparently Haiku has types in addition to. So unless it is becoming the peformance bottleneck, I'd like to keep the flexibility still.

https://www.haiku-os.org/docs/userguide/en/attributes.html

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,48 @@
...
+
+table Ztoc {
+	version : string;		// The version of the Ztoc in format <major_version>.<minor_version>, e.g. 1.0
+	buildToolIdentifier : string;
+
+	metadata : [FileMetadata];
+
+	compressedFileSize : long;
+	uncompressedFileSize : long;
+	maxSpanId : int;		// The total number of spans in Ztoc - 1
```
wmesard, line 43: Let's move this in to the ZtocInfo (CompressionInfo) struct.

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,48 @@
...
+
+	modTime  : string;		// Modification time (is formatted in RFC 3339 format,
+					// with sub-second precision added if present: "2006-01-02T15:04:05.999999999Z07:00")
+	devmajor : long;		// Major device number (valid for TypeChar or TypeBlock)
+	devminor : long;		// Minor device number (valid for TypeChar or TypeBlock)
+
+	xattrs : [Xattr];
+}
+
+table ZtocInfo {
```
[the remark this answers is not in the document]
wmesard: As a bit of future-proofing, let's add an enum for compression algorithm. We only support gzip today. But Zstandard is coming.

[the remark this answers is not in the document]
wmesard: Let's rename this so that it is clear that it relates to compression. We've gotten ourselves in trouble by using a cutsey name like "zTOC", because it's left us without a good name to refer specifically to the compression half of the format.

Let's fix that once and for all. Either double down on the cutsiness, and call this `ZInfo` or something like that; call it `CompressionInfo`. I prefer the latter. Either way, let's only ever use the four letters z-t-o-c when we are referring to the entire thing; both halves. And let's only ever use the three letters t-o-c when we are referring to the tarball table of contents.

soci/fbs/ztoc/FileMetadata.go
```diff
@@ -0,0 +1,283 @@
+// Code generated by the FlatBuffers compiler. DO NOT EDIT.
```
wmesard, line 1: Why are we checking in generated code? Why isn't this a 3 line Makefile change?

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,48 @@
...
+	version : string;		// The version of the Ztoc in format <major_version>.<minor_version>, e.g. 1.0
+	buildToolIdentifier : string;
+
+	metadata : [FileMetadata];
+
+	compressedFileSize : long;
+	uncompressedFileSize : long;
+	maxSpanId : int;		// The total number of spans in Ztoc - 1
+	ztocInfo : ZtocInfo;
+	indexByteData : [ubyte];
```
wmesard, line 45: Let's describe what this is? Also, shouldn't it live with in the CompressionInfo struct?

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+table Ztoc {
+	version : string;		// The version of the Ztoc in format <major_version>.<minor_version>, e.g. 1.0
+	buildToolIdentifier : string;
+
+	metadata : [FileMetadata];
+
+	compressedFileSize : long;
+	uncompressedFileSize : long;
```
wmesard, line 42: I'd like to see the schema organized into three sections:
1. Meta information that applies to both sections. I think the two size fields belong there. So let's move these to be just below the buildToolIdentifier.
2. The TOC
3. The z (CompressionInfo)

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,48 @@
...
+
+table ZtocInfo {
+	spanDigests : [string];
+}
+
+table Ztoc {
+	version : string;		// The version of the Ztoc in format <major_version>.<minor_version>, e.g. 1.0
+	buildToolIdentifier : string;
+
+	metadata : [FileMetadata];
```
wmesard, line 39: The compression info got its own struct. Why not the file metadata?

wmesard (changes requested):

soci/fbs/ztoc/FileMetadata.go
kzys: It could be a Makefile change, but I generally prefer to have them in repos. So that I can see the effect of upgrading code generators.

kzys: FYI containerd is having .pb.go files like https://github.com/containerd/containerd/tree/main/api

wmesard: > It could be a Makefile change, but I generally prefer to have them in repos. So that I can see the effect of upgrading code generators.

Then also check it in to a reference directory. I like that idea. It may be useful to some future git archeologist, or someone chasing a compatibility bug. We could even write a test to ensure that the _actual_ generated code matches the reference code.

But it is an anti-pattern to check in machine generated code and then actually build it into the product. Eventually, some chaotic neutral code-slinger is going to hand-modify it. And they'll probably screw it up in some subtle way. And even if they don't screw it up, we will have still taken on technical debt, because a future someone will have to figure out what to do when it _does_ come time to modify the schema.

soci/fbs/ztoc.fbs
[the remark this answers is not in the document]
vkuzniet: Done.

soci/fbs/ztoc.fbs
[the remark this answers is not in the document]
vkuzniet: No way to use `struct` there, since it allows for only scalar values and other structs.  

> As a bit of future-proofing, let's add an enum for compression algorithm. We only support gzip today. But Zstandard is coming.

Done.

> Let's fix that once and for all. Either double down on the cutsiness, and call this `ZInfo` or something like that; call it `CompressionInfo`. I prefer the latter. Either way, let's only ever use the four letters z-t-o-c when we are referring to the entire thing; both halves. And let's only ever use the three letters t-o-c when we are referring to the tarball table of contents.

Done.

soci/fbs/ztoc.fbs
vkuzniet: Done.

soci/fbs/ztoc.fbs
vkuzniet: Done.

soci/fbs/ztoc.fbs
vkuzniet: Done.

soci/fbs/ztoc.fbs
vkuzniet: Done.

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,48 @@
...
+	type : string;
+	uncompressedOffset : long;
+	uncompressedSize : long;
+	spanStart : int;
+	spanEnd : int;
+	firstSpanHasBits : bool;	// Flag for if there is partial uncompressed data that is stored in the previous byte
+	linkname : string;		// Target name of link (valid for TypeLink or TypeSymlink)
+	mode : long;			// Permission and mode bits
+	uid : int;			// User ID of owner
+	gid : int;			// Group ID of owner
```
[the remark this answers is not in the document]
vkuzniet: Done.

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+table FileMetadata {
+	name : string;
+	type : string;
+	uncompressedOffset : long;
+	uncompressedSize : long;
+	spanStart : int;
+	spanEnd : int;
+	firstSpanHasBits : bool;	// Flag for if there is partial uncompressed data that is stored in the previous byte
```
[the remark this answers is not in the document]
vkuzniet: This belongs to file metadata, since this value will be unique for every instance of FileMetadata.

soci/fbs/ztoc/FileMetadata.go
kzys: Sorry. I wasn't giving the full picture of the files above.

containerd/containerd and other contained packages usually

- Have auto-generated files committed due to the reasons above
- Have a GitHub Action to check these files are reproducible from the tools we expect (e.g. https://github.com/containerd/containerd/blob/f4095a6876c62dd4cd5e3749d2435b2daa9eaff8/.github/workflows/ci.yml#L105-L107)

That way, we can check the code changes due to tool changes and we can make sure nobody can manually edit these files.

vkuzniet: Added `install-flatc`, `flatc`, `check-flatc` actions in Makefile.

vkuzniet: `check-flatc` will become enabled in the next PR.

Makefile
```diff
@@ -51,6 +51,18 @@ pre-build:
...
 
+install-cmake:
+	@wget https://github.com/Kitware/CMake/releases/download/v3.24.1/cmake-3.24.1-Linux-x86_64.sh -O cmake.sh
+	@sh cmake.sh --prefix=/usr/local/ --exclude-subdir
+	@rm -rf cmake.sh
+
+install-flatc:
+	@wget https://github.com/google/flatbuffers/archive/refs/tags/v2.0.8.tar.gz -O flatbuffers.tar.gz
+	@tar xzvf flatbuffers.tar.gz
+	@cd flatbuffers-2.0.8 && cmake -G "Unix Makefiles" && make && make install
```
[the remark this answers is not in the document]
vkuzniet: Done.

Makefile
```diff
@@ -30,7 +30,7 @@ CMD=soci-snapshotter-grpc soci
 
 CMD_BINARIES=$(addprefix $(OUTDIR)/,$(CMD))
 
-.PHONY: all build pre-build check check-ltag check-dco check-lint install-check-tools add-ltag install install-zlib uninstall clean test integration
+.PHONY: all build pre-build check check-ltag check-dco check-lint install-check-tools add-ltag install install-cmake install-zlib uninstall clean test integration
```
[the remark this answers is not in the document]
vkuzniet: Done.

Makefile
```diff
@@ -60,6 +72,15 @@ install-zlib:
 
 check: check-ltag check-dco check-lint
 
+flatc:
+	@flatc -o $(CURDIR)/soci/fbs -g $(CURDIR)/soci/fbs/ztoc.fbs
+
+check-flatc: flatc ## check if protobufs needs to be generated again
```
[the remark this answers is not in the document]
vkuzniet: Done.

rdpsin (approved):

Makefile
```diff
@@ -51,14 +51,35 @@ pre-build:
 	@ar rvs ${OUTDIR}/libindexer.a ${OUTDIR}/indexer.o
 	@rm -f ${OUTDIR}/indexer.o
 
+install-cmake:
+	@wget https://github.com/Kitware/CMake/releases/download/v3.24.1/cmake-3.24.1-Linux-x86_64.sh -O cmake.sh
+	@sh cmake.sh --prefix=/usr/local/ --exclude-subdir
+	@rm -rf cmake.sh
+
+install-flatc:
```
kzys, line 59: How about using compiled binaries from https://github.com/google/flatbuffers/releases instead? Installing cmake for a Go project seems too much to me.

integration/pull_test.go
```diff
@@ -121,15 +121,10 @@ func TestOptimizeConsistentSociArtifact(t *testing.T) {
...
-		/*
-			The following test is disabled for now, since it randomly fails due to
-			the way gob encodes FileMetadata.Xattrs.
-			TODO(rdpsin): Re-enable this test once gob is no longer used to encode FileMetadata.
-			{
-				name:           "soci artifact is consistently built for nginx",
-				containerImage: "nginx:latest",
-			},
-		*/
+		{
```
kzys, line 124: Yay!

soci/fbs/ztoc/FileMetadata.go
wmesard: Sounds good to me. Best of both worlds.

wmesard (approved):
Left a couple of nits. And @kzys 's comment is worthy of consideration. Other than that LGTM

Makefile
```diff
@@ -51,14 +51,35 @@ pre-build:
...
+	@wget https://github.com/google/flatbuffers/archive/refs/tags/v2.0.8.tar.gz -O flatbuffers.tar.gz
+	@tar xzvf flatbuffers.tar.gz
+	@cd flatbuffers-2.0.8 && cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release && make && make install
+	@rm -f flatbuffers.tar.gz
+	@rm -rf flatbuffers-2.0.8
```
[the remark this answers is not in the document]
vkuzniet: Done.

soci/fbs/ztoc.fbs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+table FileMetadata {
+	name : string;
+	type : string;
+	uncompressed_offset : long;
+	uncompressed_size : long;
+	span_start : int;
+	span_end : int;
+	first_span_has_bits : bool;// Flag for if there is partial uncompressed data that is stored in the previous byte
```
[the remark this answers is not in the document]
vkuzniet: Done.

rdpsin (approved):

wmesard (approved):

rdpsin (approved):

Makefile
```diff
@@ -51,14 +51,35 @@ pre-build:
 	@ar rvs ${OUTDIR}/libindexer.a ${OUTDIR}/indexer.o
 	@rm -f ${OUTDIR}/indexer.o
 
+install-cmake:
+	@wget https://github.com/Kitware/CMake/releases/download/v3.24.1/cmake-3.24.1-Linux-x86_64.sh -O cmake.sh
+	@sh cmake.sh --prefix=/usr/local/ --exclude-subdir
+	@rm -rf cmake.sh
```
wmesard, line 57: `@` :-(

wmesard (approved):

PR #61, opened 2022-09-12
Remove flags.go from soci cli

Kern--:

*Issue #, if available:*
N/A

*Description of changes:*
This change removes `cmd/soci/commands/flags.go` as it was unused

*Testing performed:*


Kern--:
`make check-dco` fails because it hard-coded a specific commit to start the range, but the GHA only checks out the last 20. Since there are more than 20 commits since then, the check fails because it can't find the first commit in the range. 

Fixed by https://github.com/awslabs/soci-snapshotter/pull/62

rdpsin (approved):

vkuzniet (approved):

Issue #6, opened 2022-08-15
Fix make check-dco commit range


Kern--:
When initially adding the `make check-dco` target, we set the validation commit range to ignore the first couple of commits that didn't have the signed-off line, but otherwise verify all commits up to HEAD. This will work for now, but we may want to limit the check to the last N commits so that we don't have to fetch the entire history in pull request workflows. E.g. https://github.com/awslabs/soci-snapshotter/pull/5 only pulls the last 20 commits. 

The line in question:
https://github.com/awslabs/soci-snapshotter/blob/48e5b05ea0ad989beb06ecdbf0b7a661691db1bc/Makefile#L77

PR #62, opened 2022-09-12
Check last 20 commits in `make check-dco`

Kern--:

*Issue #, if available:*
Closes https://github.com/awslabs/soci-snapshotter/issues/6

*Description of changes:*
Rather than checking for the DCO sign of in every commit in the repository, we only check a subset. Ideally, this would be only the new commits in each PR, however that number is unknowable, so therefore we just choose 20 somewhat arbitrarily.

Before this commit, we had a couple commits merged into main that were either missing the DCO sign off or had an incorrect format, so we hardcoded the start of the range. Now that we have more commits, we can safely switch back to checking just the last 20.

*Testing performed:*
`make check`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          fetch-depth: 20
+          # check-dco will check the last 20 commits, but commit ranges
+          # exclude the start commit in the result, but need that commit
+          # in order to calculate the range. i.e. HEAD~20..HEAD includes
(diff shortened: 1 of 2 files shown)


rdpsin (approved):

vkuzniet (approved):

estesp:
A static set of commits is probably fine, but if you want to get rid of the extra fetching, you can check the commit range associated with a pull request with a bit of magic pulling details from the GitHub JSON data available to an action; we do it in containerd here: https://github.com/containerd/project-checks/blob/main/action.yml#L67-L78

PR #65, opened 2022-09-14
Add Referrers API support

### LINKED ISSUE #44: soci-snapshotter should optionally use the Referers API to discover the remote SOCI index

Today, soci-snapshotter has to be [explicitly told](https://github.com/awslabs/soci-snapshotter/blob/aef69ad5ab220144efe9a66ffa78d0fc625fa6aa/docs/GETTING_STARTED.md#preparing-soci-snapshotter) the digest of the SOCI index that is associated with the image in the remote repository.

That's fine for some installations, where an outside agent will be doing that query anyway (via the Referrers API) to figure out which snapshotter to use. But for some installations--e.g., anyone running containerd themselves--forcing them to feed us the SOCI index digest (using `soci rpull`) is...um...stupid. [Referrers API](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers) was developed specifically to relieve clients of this burden.

So let's fix it.

The SOCI index digest should be optional in both the `client.Pull` call and the `soci rpull` command line. If it is not specified, then the snapshotter itself will use [Referrers API](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers) to try to look it up.

If Referrers API returns more than one result, the snapshotter will use the first one in the list. (For now. We can iterate on this to later.)

If I were running the world, containerd would be allowed to have multiple snapshotters installed, and its snapshotter plugin interface have a new call which would happen before Prepare to do dynamic snapshotter selection. That is, containerd would ask each configured snapshotter "Hey, do you want to handle this image?" In that world, soci-snapshotter would do the Referrers API call at that time.

Alas, that is not the world in which we live. It could be. [containerd #6657](https://github.com/containerd/containerd/issues/6657) is trying to solve a similar problem.

**Additional context**

As part of this work, references to the `soci rpull` command will be removed from the *GETTING_STARTED.md* guide. In other words, the only thing you need to do on the runtime host, is launch the snapshotter, and then `ctr i run` the image. No `soci` CLI commands needed. That way, the demo is more wicked cool, and it is more obvious how SOCI can be used for production workloads.

If this fix is implemented after [soci-snapshotter #42](https://github.com/awslabs/soci-snapshotter/issues/42), then it would be fine if it only works for Reference Types and not ORAS. ORAS is going away, so there's no point in writing throwaway code.

rdpsin:
This commit introduces ORAS Referrers API support to the snapshotter. `--soci-index-digest` is no longer necessary as the snapshotter will invoke the Referrers API.

*Issue #, if available:*
Resolves #44 

*Description of changes:*

*Testing performed:*

`make && make check && make test && make integration`

1) Start the snapshotter
```shell
$ sudo ./soci-snapshotter-grpc &> logs &
```

2) `rpull` the image, but don't provide `--soci-index-digest`

```shell
$ sudo ./soci image rpull --user "$REGISTRY_USER:$REGISTRY_PASS" $IMAGE
fetching sha256:3fd97d24... application/vnd.docker.distribution.manifest.v2+json
fetching sha256:aec97901... application/vnd.docker.container.image.v1+json
```

3) Check `mount`

```shell
$ mount 
/local/home/rdpsin/soci/out/soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/1/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
/local/home/rdpsin/soci/out/soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/2/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
/local/home/rdpsin/soci/out/soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/3/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
...
```

4) Check the logs to make sure the referrers API is being called

```
{"key":"default/1/extract-48576544-22Re sha256:08249ce7456a1c0613eafe868aed936a284ed9f1d6144f7d2d08c514974a2af9","level":"info","mountpoint":"/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/1/fs","msg":"index digest not provided, making a Referrers API call to fetch list of indices","parent":"","time":"2022-09-15T14:22:01.056566326Z"}
{"digest":"sha256:2acf14636fb5afd8f862b7dff02fbbd053eead67b9aa0baf69e9ca5c233a2115","key":"default/1/extract-48576544-22Re sha256:08249ce7456a1c0613eafe868aed936a284ed9f1d6144f7d2d08c514974a2af9","level":"info","mountpoint":"/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/1/fs","msg":"fetching SOCI artifacts using index descriptor","parent":"","time":"2022-09-15T14:22:01.674535703Z"}
```


wmesard (dismissed):
If this PR is successful, `rpull` should now be optional. Can you 
1. Indicate that you've tested that in the *Testing performed* section.
2. As part of this change, modify [that section](https://github.com/awslabs/soci-snapshotter/blob/main/docs/GETTING_STARTED.md?plain=1#L237-L268) of the Getting Started guide to say that it's optional. I thought about suggesting removing that section entirely. But I guess there's value to leaving it there, since it also contains a good discussion of what's going on behind the scenes. So maybe just add sentence or two here:
> Once the snapshotter is running we can call the rpull command from the SOCI CLI. This command reads the manfiest from the registry and mounts FUSE filesystems for each layer.

like

> **Note:** This step is optional. If you don't explicitly provide the SOCI Index digest, the snapshotter will use the [Referrers API](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers) to discover it.

vkuzniet:
> If this PR is successful, `rpull` should now be optional. Can you
> 
>     1. Indicate that you've tested that in the _Testing performed_ section.
> 
>     2. As part of this change, modify [that section](https://github.com/awslabs/soci-snapshotter/blob/main/docs/GETTING_STARTED.md?plain=1#L237-L268) of the Getting Started guide to say that it's optional. I thought about suggesting removing that section entirely. But I guess there's value to leaving it there, since it also contains a good discussion of what's going on behind the scenes. So maybe just add sentence or two here:
> 
> 
> > Once the snapshotter is running we can call the rpull command from the SOCI CLI. This command reads the manfiest from the registry and mounts FUSE filesystems for each layer.
> 
> like
> 
> > **Note:** This step is optional. If you don't explicitly provide the SOCI Index digest, the snapshotter will use the [Referrers API](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers) to discover it.

Are we sure this will actually happen? Looking at https://github.com/awslabs/soci-snapshotter/blob/main/cmd/soci/commands/image/rpull.go#L150, there's a bunch of labels, which are passed to containerd client during `rpull` to do the job.

wmesard:
> Are we sure this will actually happen?

I don't think this PR fully resolves #44 unless this happens. That's why I'm asking.

cmd/soci/commands/image/rpull.go
```diff
@@ -81,15 +81,7 @@ command.
 			return fmt.Errorf("please provide an image reference")
 		}
 
-		sociIndexDigest := context.String("soci-index-digest")
```
sbuckfelder, line 84: why delete?

fs/artifact_fetcher.go
```diff
@@ -68,7 +67,7 @@ func newArtifactFetcher(refspec reference.Spec, localStore, remoteStore content.
 	}, nil
 }
 
-func newRemoteStore(refspec reference.Spec) (content.Storage, error) {
+func newRemoteStore(refspec reference.Spec) (*remote.Repository, error) {
```
sbuckfelder, line 70: why the change in return type?

vkuzniet:
> > Are we sure this will actually happen?
> 
> I don't think this PR fully resolves #44 unless this happens. That's why I'm asking.

Yeah, this is true. In order to do that, we need to ensure that snapshotter is supplied with the labels from https://github.com/awslabs/soci-snapshotter/blob/main/fs/source/source.go#L180. Those labels are used to infer layer descriptor during `fs.Mount` and are now being passed to containerd client during `rpull`.

fs/artifact_fetcher.go
rdpsin: It is used to [resolve the image ref](https://github.com/awslabs/soci-snapshotter/pull/65/files#diff-c7b54bb28e1654d0ca47364a2c97456c124ee0d3a65854db18d116c509d89136R232) in `fs.go`. 

Returning `content.Storage` would have meant doing something like 

```
repo, _ := remoteStore.(*remote.Repository)
```

I wanted to avoid that.

cmd/soci/commands/image/rpull.go
rdpsin: Since we are not checking `sociIndexDigest == ""` anymore, avoided creating a new variable for it.

rdpsin:
@wmesard Done and done. As discussed offline, making `rpull` optional will now not be part of this PR.

fs/fs.go
```diff
@@ -204,7 +211,53 @@ type filesystem struct {
...
+			retErr = err
+			return
+		}
+
+		indexDesc := ocispec.Descriptor{
+			Digest: digest.Digest(indexDigest),
+		}
+
+		if indexDigest == "" {
+			log.G(ctx).Infof("index digest not provided, making a Referrers API call to fetch list of indices")
```
kzys, line 231: It could be Info, not Infof.

fs/fs.go
```diff
@@ -204,7 +211,53 @@ type filesystem struct {
...
+					retErr = err
+					return
+				}
+				manifestDesc = desc
+			}
+
+			refClient := NewORASReferrersClient(remoteStore)
+			desc, err := refClient.Get(ctx, manifestDesc, defaultIndexSelectionPolicy)
+			if err != nil {
+				retErr = fmt.Errorf("cannot fetch list of referrers: %v", err)
```
kzys, line 253: Please use %w instead of %v.

fs/ref_client.go
```diff
@@ -0,0 +1,111 @@
...
+		MediaType:   ar.MediaType,
+		Digest:      ar.Digest,
+		Size:        ar.Size,
+		Annotations: ar.Annotations,
+		URLs:        ar.URLs,
+	}
+}
+
+// Retrieves the manifest list from store and returns a manifest descriptor for the current platform
+func getManifestFromIndex(ctx context.Context, store content.Storage, desc ocispec.Descriptor, platform ocispec.Platform) (ocispec.Descriptor, error) {
```
Kern--, line 94: `soci.GetImageManifestDescriptor` is very similar to this. We should have a follow up to dedupe them.

fs/ref_client.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// Responsible for making Referrers API calls to remote registry to fetch list of referrers.
+type ReferrersClient interface {
+	/// Takes in an manifest descriptor and IndexSelectionPolicy and returns a single artifact descriptor.
+	/// Returns an error (ErrNoReferrers) if the manifest descriptor has no referrers.
+	Get(context.Context, ocispec.Descriptor, IndexSelectionPolicy) (ocispec.Descriptor, error)
+}
+
+type orasReferrersClient struct {
```
Kern--, line 51: This abstraction feels weird where we create a new client that wraps the old client just to make index manifest selection a little cleaner. 

If we want this to be a wrapper client, I would prefer we embed the original `*remote.Repository` in the wrapper and then use the wrapper everywhere.

fs/fs.go
```diff
@@ -204,7 +211,53 @@ type filesystem struct {
...
+			if err != nil {
+				retErr = fmt.Errorf("unable to resolve image ref %v into a descriptor: %v", imageRef, err)
+				return
+			}
+
+			// If the ref points to an index (manifest list), fetch it from the remote store and return the manifest descriptor
+			// for the current platform.
+			if manifestDesc.MediaType == dockerManifestListMediaType || manifestDesc.MediaType == ocispec.MediaTypeImageIndex {
+				log.G(ctx).Infof("image ref points to an index; retrieving a manifest for current platform")
+				desc, err := getManifestFromIndex(ctx, remoteStore, manifestDesc, platforms.DefaultSpec())
```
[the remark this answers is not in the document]
rdpsin: Okay, I can add an optional flag to `rpull` that passes that platform, and default to `DefaultSpec` if platform is not provided. Does that seem right?

fs/fs.go
rdpsin: Done.

fs/fs.go
rdpsin: Done.

fs/fs.go
```diff
@@ -204,7 +211,53 @@ type filesystem struct {
...
+			log.G(ctx).Infof("index digest not provided, making a Referrers API call to fetch list of indices")
+			manifestDesc, err := remoteStore.Resolve(ctx, imageRef)
+			if err != nil {
+				retErr = fmt.Errorf("unable to resolve image ref %v into a descriptor: %v", imageRef, err)
+				return
+			}
+
+			// If the ref points to an index (manifest list), fetch it from the remote store and return the manifest descriptor
+			// for the current platform.
+			if manifestDesc.MediaType == dockerManifestListMediaType || manifestDesc.MediaType == ocispec.MediaTypeImageIndex {
```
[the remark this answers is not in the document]
rdpsin: Done.

fs/ref_client.go
rdpsin: Created a new `orasClient` struct that embeds `*remote.Repository`.

fs/fs.go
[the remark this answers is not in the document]
rdpsin: As discussed offline, leaving it like this for now. Will revisit this we have a discussion about multi-platform images.

fs/client.go
```diff
@@ -0,0 +1,88 @@
...
+	*remote.Repository
+}
+
+func NewORASClient(remote *remote.Repository) *OrasClient {
+	return &OrasClient{
+		Repository: remote,
+	}
+}
+
+func (c *OrasClient) Get(ctx context.Context, desc ocispec.Descriptor, fn IndexSelectionPolicy) (ocispec.Descriptor, error) {
```
vkuzniet, line 58: can you get a unit test for Get?

rdpsin: Done.

vkuzniet (dismissed):

vkuzniet (dismissed):

fs/client.go
```diff
@@ -0,0 +1,98 @@
...
+	Inner
+}
+
+func NewORASClient(inner Inner) *OrasClient {
+	return &OrasClient{
+		Inner: inner,
+	}
+}
+
+func (c *OrasClient) Get(ctx context.Context, desc ocispec.Descriptor, fn IndexSelectionPolicy) (ocispec.Descriptor, error) {
```
Kern--, line 68: These 3 methods all sounds like they should be doing the same thing.

`OrasClient.Get` -> gets a descriptor for a single referer to an input descriptor
`OrasClient.Fetch` -> gets an io.Reader for an input descriptor
`OrasClient.fetch` -> gets a list of descriptors for all referrers to an input descriptor


I think it might be clearer to rename `Get` to `SelectReferrer` and `fetch` to `allReferrers` or something similar.

fs/fs.go
```diff
@@ -263,13 +302,13 @@ func (fs *filesystem) MountLocal(ctx context.Context, mountpoint string, labels
 	return nil
 }
 
-func (fs *filesystem) getSociContext(ctx context.Context, imageRef, indexDigest string) (*sociContext, error) {
+func (fs *filesystem) getSociContext(ctx context.Context, imageRef, indexDigest, imageDigest string) (*sociContext, error) {
 	cAny, _ := fs.sociContexts.LoadOrStore(indexDigest, &sociContext{})
```
Kern--, line 306: This will be broken if `indexDigest` is empty. 

I had a suggestion to replace the map with `imageDigest` here: https://github.com/awslabs/soci-snapshotter/issues/44#issuecomment-1255771735

Is there a use-case where we would want an image to be pullable using 2 different indices? Do we support namespacing?

fs/fs.go
```diff
@@ -187,10 +192,44 @@ type sociContext struct {
 	imageLayerToSociDesc map[string]ocispec.Descriptor
 }
 
-func (c *sociContext) Init(ctx context.Context, imageRef, indexDigest string, store orascontent.Storage) error {
+func (c *sociContext) Init(ctx context.Context, imageRef, indexDigest, imageDigest string, store orascontent.Storage) error {
```
Kern--, line 195: nit: can we rename `imageDigest` to `imageManifestDigest` to be clear that we're talking about a specific iamge manifest and not an image index?

fs/fs.go
rdpsin: I can move the Referrers API call to just before `LoadOrStore`.

Kern--: That means we have a referrers API call for every layer. Are we willing to pay that price?

rdpsin: Currently, it is not possible for an image to be pulled multiple times using different indices. So maybe using the image manifest digest for `sociContexts` is the way to go about it.

fs/client.go
rdpsin: Done.

fs/client_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (f *fakeInner) Push(ctx context.Context, expected ocispec.Descriptor, content io.Reader) error {
+	return nil
+}
+
+func (f *fakeInner) Referrers(ctx context.Context, desc ocispec.Descriptor, artifactType string, fn func(referrers []artifactsspec.Descriptor) error) error {
+	return fn(f.descs)
+}
+
+func TestORASClientGet(t *testing.T) {
```
vkuzniet, line 59: nit: would be good to rename this to `TestORASClient.SelectReferrer`

vkuzniet (approved):

PR #66, opened 2022-09-15
indexer uses int64_t instead of off_t

vkuzniet:

*Issue #, if available:*

*Description of changes:*
`off_t` type in indexer.c is not platform agnostic. To improve that, the new type `offset_t` is defined, which is set to `int64_t`, so it's always 64 bits. This change is the part of the effort to make the decompression dictionary part of zTOC platform agnostic.

*Testing performed:*

### DIFF
--- a/c/indexer.h
+++ b/c/indexer.h
@@ -50,6 +50,7 @@
 #include <zlib.h>
 
 typedef unsigned char uchar;
+typedef int64_t offset_t;
 
 /* Since gzip is compressed with 32 KiB window size, WINDOW_SIZE is fixed */
 #define WINSIZE 32768U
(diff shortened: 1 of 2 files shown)


vkuzniet:
> Does it also requires changes to Go files from where we call these functions?

Go files consistently use the `FileSize`, which is `int64`, so no changes are needed.

rdpsin (approved):

kzys (approved):

c/indexer.h
```diff
@@ -50,6 +50,7 @@
 #include <zlib.h>
 
 typedef unsigned char uchar;
+typedef int64_t offset_t;
```
kzys, line 53: Since I didn't know them without googling :)

`off_t` is in [POSIX](https://pubs.opengroup.org/onlinepubs/009696899/basedefs/sys/types.h.html) and [GNU libc](https://www.gnu.org/software/libc/manual/html_node/File-Position-Primitive.html).

int64_t is in [C99](http://port70.net/~nsz/c/c99/n1256.html#7.18.1.1), not exactly though.

sbuckfelder (approved):

c/indexer.h
vkuzniet: Yep, the reason to use `int64_t` is that it's certain that its size is 8 bytes for any platform. The size of the `off_t` is not guaranteed though.

PR #71, opened 2022-09-22
Incapsulate indexer in a separate Go component

vkuzniet:
Signed-off-by: Viktor Kuznietsov <vkuzniet@amazon.com>

*Issue #, if available:*

*Description of changes:*
This change introduces the `GzipIndexer` component, which incapsulates the interaction with C indexer component. After this change is merged, the Go code must only talk to indexer through GzipIndexer.

This change also adds small fixes pointed by linter after removing `import "C"` in span manager and ztoc building logic.

*Testing performed:*
- `make test && make check && make integration` pass.
- deployed snapshotter and cli to an EC2 instance and created an index and ran container workload in lazy loading mode for `rabbitmq`, `drupal`, `tomcat` images and observed successful execution to ready line.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #72, opened 2022-09-22
Incapsulate indexer in a separate Go component

vkuzniet:

*Issue #, if available:*

*Description of changes:*
This change introduces the `GzipIndexer` component, which incapsulates the interaction with C indexer component. After this change is merged, the Go code must only talk to indexer through `GzipIndexer`.

This change also adds small fixes pointed by linter after removing `import "C"` in span manager and ztoc building logic.

Note: this is the re-submission of https://github.com/awslabs/soci-snapshotter/pull/71, since the feature branch got diverged and merge commit has become unavoidable.

*Testing performed:*
- deployed snapshotter and cli to an EC2 instance and created an index and ran container workload in lazy loading mode for `rabbitmq`, `drupal`, `tomcat` images and observed successful execution to ready line.

### DIFF
--- a/soci/gzip_index.go
+++ b/soci/gzip_index.go
@@ -1,0 +1,141 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package soci
+
+// #cgo CFLAGS: -I${SRCDIR}/../c/
+// #cgo LDFLAGS: -L${SRCDIR}/../out -lindexer -lz
+// #include "indexer.h"
+// #include <stdlib.h>
+// #include <stdint.h>
+import "C"
+
+import (
+	"fmt"
+	"unsafe"
+)
+
+type GzipIndex struct {
+	index *C.struct_gzip_index
+}
+
+// NewGzipIndex creates a new instance of `GzipIndex` from index byte blob on zTOC.
+func NewGzipIndex(indexData []byte) (*GzipIndex, error) {
+	index := C.blob_to_index(unsafe.Pointer(&indexData[0]))
+	if index == nil {
+		return nil, fmt.Errorf("cannot convert blob to gzip_index")
+	}
+	return &GzipIndex{
+		index: index,
+	}, nil
+}
+
+// NewGzipIndexFromFile creates a new instance of `GzipIndex` given gzip file name and span size.
+func NewGzipIndexFromFile(gzipFile string, spanSize int64) (*GzipIndex, error) {
+	cstr := C.CString(gzipFile)
+	defer C.free(unsafe.Pointer(cstr))
+
+	var index *C.struct_gzip_index
+	ret := C.generate_index(cstr, C.off_t(spanSize), &index)
+	if int(ret) < 0 {
+		return nil, fmt.Errorf("could not generate gzip index. gzip error: %v", ret)
+	}
+
+	return &GzipIndex{
+		index: index,
+	}, nil
+}
+
+// Close calls `C.free` on the pointer to `C.struct_gzip_index`.
+func (i *GzipIndex) Close() {
+	if i.index != nil {
+		C.free(unsafe.Pointer(i.index))
+	}
+}
+
+// Bytes returns the byte slice containing the index.
+func (i *GzipIndex) Bytes() ([]byte, error) {
+	blobSize := C.get_blob_size(i.index)
+	bytes := make([]byte, uint64(blobSize))
+	if bytes == nil {
+		return nil, fmt.Errorf("could not allocate byte array of size %d", blobSize)
+	}
+
+	ret := C.index_to_blob(i.index, unsafe.Pointer(&bytes[0]))
+	if int(ret) <= 0 {
+		return nil, fmt.Errorf("could not serialize gzip index to byte array; gzip error: %v", ret)
+	}
+	return bytes, nil
+}
+
+// MaxSpanID returns the max span ID.
+func (i *GzipIndex) MaxSpanID() SpanID {
+	return SpanID(i.index.have - 1)
+}
+
+// UncompressedOffsetToSpanID returns the ID of the span containing the data pointed by uncompressed offset.
+func (i *GzipIndex) UncompressedOffsetToSpanID(offset FileSize) SpanID {
+	return SpanID(C.pt_index_from_ucmp_offset(i.index, C.long(offset)))
+}
+
+// HasBits wraps `C.has_bits` and returns true if any data is contained in the previous span.
+func (i *GzipIndex) HasBits(spanID SpanID) bool {
+	return C.has_bits(i.index, C.int(spanID)) != 0
+}
+
+// SpanIDToCompressedOffset wraps `C.get_comp_off` and returns the offset for the span in the compressed stream.
+func (i *GzipIndex) SpanIDToCompressedOffset(spanID SpanID) FileSize {
+	return FileSize(C.get_comp_off(i.index, C.int(spanID)))
+}
+
+// SpanIDToUncompressedOffset wraps `C.get_uncomp_off` and returns the offset for the span in the uncompressed stream.
+func (i *GzipIndex) SpanIDToUncompressedOffset(spanID SpanID) FileSize {
+	return FileSize(C.get_ucomp_off(i.index, C.int(spanID)))
+}
+
+// ExtractDataFromBuffer wraps the call to `C.extract_data_from_buffer`, which takes in the compressed bytes
+// and returns the decompressed bytes.
+func (i *GzipIndex) ExtractDataFromBuffer(compressedBuf []byte, uncompressedSize, uncompressedOffset FileSize, spanID SpanID) ([]byte, error) {
+	bytes := make([]byte, uncompressedSize)
+	ret := C.extract_data_from_buffer(
+		unsafe.Pointer(&compressedBuf[0]),
+		C.off_t(len(compressedBuf)),
+		i.index,
+		C.off_t(uncompressedOffset),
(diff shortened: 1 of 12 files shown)


soci/gzip_indexer.go
```diff
@@ -0,0 +1,147 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package soci
+
+// #cgo CFLAGS: -I${SRCDIR}/../c/
```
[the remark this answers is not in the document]
vkuzniet: If we decide to go in the direction of defining SOCI as a module, then yes. Right now `soci` is just a package within soci-snapshotter project.

soci/gzip_indexer.go
```diff
@@ -0,0 +1,147 @@
...
+// #include <stdlib.h>
+// #include <stdint.h>
+import "C"
+
+import (
+	"fmt"
+	"unsafe"
+)
+
+type GzipIndexer struct {
```
rdpsin, line 31: Let's rename this to 'GzipIndex'. 'Indexer' sounds like a struct whose sole purpose it to generate the index, but this a just a wrapper for the C struct.

soci/gzip_indexer.go
```diff
@@ -0,0 +1,147 @@
...
+	index *C.struct_gzip_index
+}
+
+// GenerateIndex wraps `C.generate_index` and should be used for index generation instead of relying on C code.
+// Returns the byte slice containing the `gzip_index`.
+func GenerateIndex(gzipFile string, spanSize int64) ([]byte, error) {
```
rdpsin, line 37: The API is confusing to me. If I want to just a create a `GzipIndex`, I have to call this function to a byte array and then call `NewGzipIndexer` to convert it back to an index. 

Maybe have `GenerateIndex` just return a `GzipIndex` and have a `func (g *GzipIndex) Bytes() []byte`. This would be useful in cases where people want to just create the index and use it directly without converting into byte data.

soci/gzip_indexer.go
```diff
@@ -0,0 +1,147 @@
...
+		return 0, 0, fmt.Errorf("cannot get the span indices for file with start and end offset: %d, %d; gzip error: %v", startOffset, endOffset, ret)
+	}
+	return indexStart, indexEnd, nil
+}
+
+func NewGzipIndexer(indexData []byte) (*GzipIndexer, error) {
+	index := C.blob_to_index(unsafe.Pointer(&indexData[0]))
+	if index == nil {
+		return nil, fmt.Errorf("cannot convert blob to gzip_index")
+	}
```
rdpsin, line 143: Would also recommend adding a finalizer here that will call `Close` when the object here is GC-ed.

soci/gzip_indexer.go
```diff
@@ -0,0 +1,147 @@
...
+	var indexStart SpanID
+	var indexEnd SpanID
+	ret := C.span_indices_for_file(i.index, C.off_t(startOffset), C.off_t(endOffset), unsafe.Pointer(&indexStart), unsafe.Pointer(&indexEnd))
+	if int(ret) <= 0 {
+		return 0, 0, fmt.Errorf("cannot get the span indices for file with start and end offset: %d, %d; gzip error: %v", startOffset, endOffset, ret)
+	}
+	return indexStart, indexEnd, nil
+}
+
+func NewGzipIndexer(indexData []byte) (*GzipIndexer, error) {
```
rdpsin, line 139: I'm wondering if `GenerateIndex` should be called `NewGzipIndexer`, and this function should be called `NewGzipIndexerFromBytes`, just to keep naming consistent.

soci/gzip_indexer.go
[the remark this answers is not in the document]
rdpsin: It doesn't. It still uses `libindexer` functions. If we really want to make it available as a package others can use, we need to remove all `libindexer` code and rewrite it in Go. I don't know if we want that as part of this PR. OTOH, since we are creating a Go interface for the indexer, it should be part of this PR.

rdpsin (changes requested):

soci/gzip_indexer.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+}
+
+// GetMaxSpanID returns the max span ID.
+func (i *GzipIndexer) GetMaxSpanID() SpanID {
+	return SpanID(i.index.have - 1)
+}
+
+// GetSpanIDByUncompressedOffset returns the ID of the span containing the data pointed by uncompressed offset.
+func (i *GzipIndexer) GetSpanIDByUncompressedOffset(offset FileSize) SpanID {
```
wmesard, line 75: Naming nit: "For" not "By". `GetSpanIDForUncompressedOffset`. Alternatively, I would avoid "Get" in all of these function names, since they are converters, not getters. So
* UncompressedOffsetToSpanID
* SpanIDToCompressedOffset
* SpanIDToUncompressedOffset

wmesard (changes requested):
```
$ g co main
$ ag -l #cgo
soci/ztoc_builder.go
fs/span-manager/span_manager.go
$ g co origin/indexer-incapsulation
$ ag -l #cgo
soci/gzip_indexer.go
$ 
```
Nice!

soci/gzip_indexer.go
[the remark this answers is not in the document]
vkuzniet: This PR's intention is to provide an easily usable wrapper, so nobody deals with calling C code throughout the snapshotter. I will create an issue to track rewriting efforts and this will happen in another code change.

soci/gzip_indexer.go
vkuzniet: Not the best idea. Tests show that this causes panics due to access memory violation. That is why the API was designed in the way to let the caller do `defer indexer.Close()`. Moreover, it's already the pattern with IO operations.

rdpsin: Yes, but most IO operations are scoped to a function, so it makes sense to `defer`. The Indexer might not.

soci/gzip_indexer.go
vkuzniet: Done.

soci/gzip_indexer.go
vkuzniet: Done.

soci/gzip_indexer.go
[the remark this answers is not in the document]
vkuzniet: Done.

soci/gzip_indexer.go
```diff
@@ -0,0 +1,147 @@
...
+	ret := C.extract_data(cstr, i.index, C.off_t(uncompressedOffset), unsafe.Pointer(&bytes[0]), C.int(uncompressedSize))
+	if ret <= 0 {
+		return nil, fmt.Errorf("unable to extract data; return code = %v", ret)
+	}
+
+	return bytes, nil
+}
+
+// GetSpanIndicesForFile wraps `C.span_indices_for_file` and returns IDs of starting and ending given start and end offsets.
+func (i *GzipIndexer) GetSpanIndicesForFile(startOffset, endOffset FileSize) (SpanID, SpanID, error) {
```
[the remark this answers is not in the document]
vkuzniet: Removed it in favor of using `UncompressedOffsetToSpanID`.

soci/gzip_indexer.go
```diff
@@ -0,0 +1,147 @@
...
+	bytes := make([]byte, uncompressedSize)
+	ret := C.extract_data(cstr, i.index, C.off_t(uncompressedOffset), unsafe.Pointer(&bytes[0]), C.int(uncompressedSize))
+	if ret <= 0 {
+		return nil, fmt.Errorf("unable to extract data; return code = %v", ret)
+	}
+
+	return bytes, nil
+}
+
+// GetSpanIndicesForFile wraps `C.span_indices_for_file` and returns IDs of starting and ending given start and end offsets.
```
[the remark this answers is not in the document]
vkuzniet: Not relevant anymore, since the method has been removed.

soci/gzip_indexer.go
```diff
@@ -0,0 +1,147 @@
...
+func (i *GzipIndexer) GetSpanIndicesForFile(startOffset, endOffset FileSize) (SpanID, SpanID, error) {
+	var indexStart SpanID
+	var indexEnd SpanID
+	ret := C.span_indices_for_file(i.index, C.off_t(startOffset), C.off_t(endOffset), unsafe.Pointer(&indexStart), unsafe.Pointer(&indexEnd))
+	if int(ret) <= 0 {
+		return 0, 0, fmt.Errorf("cannot get the span indices for file with start and end offset: %d, %d; gzip error: %v", startOffset, endOffset, ret)
```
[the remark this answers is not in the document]
vkuzniet: Not relevant anymore, since the method has been removed.

soci/gzip_indexer.go
vkuzniet: There are two use cases. Generate the index for Ztoc - you pass the file name and span size and get the index. Another use case, you have Ztoc and create Index from byte slice. And snapshotter anticipates using the latter use case within span manager. `NewGzipIndex` will be using ztoc-s IndexByteData, and `NewGzipIndexFromFile` will using tar.gz file to create `*GzipIndex`

soci/gzip_indexer.go
vkuzniet: Responded in another thread.

soci/gzip_index.go
```diff
@@ -0,0 +1,141 @@
...
+
+// Close calls `C.free` on the pointer to `C.struct_gzip_index`.
+func (i *GzipIndex) Close() {
+	if i.index != nil {
+		C.free(unsafe.Pointer(i.index))
+	}
+}
+
+// IndexToBlob returns the byte slice containing the index.
+func (i *GzipIndex) IndexToBlob() ([]byte, error) {
```
rdpsin, line 70: I would just rename this method to `Bytes`.

vkuzniet: Done.

rdpsin (approved):

wmesard (approved):
This moves us in the right direction, so I'm saying shipit. But we are still implicitly leaking details about the underlying implementation. This should be addressed in future PRs before we attempt to define a real gzip-neutral `interface` (that is, before we add Zstd support).

1. `NewGzipIndex` and `NewGzipIndexFromFile` would both be taking *Reader*s if the underlying implementation were in native Go.
2. *HasBits* is an implementation detail of Gzip, and probably this implementation of the gzip-ztoc technology. So we should not be exposing this concept to the caller.

PR #80, opened 2022-09-27
Integer fields in gzip_index are serialized independently to endianness

vkuzniet:

*Issue #, if available:*

*Description of changes:*
`gzip_index`'s `have`, `size` and `span_size` as well as `gzip_index_point`'s `in`, `out` fields are converted to Big Endian representations for serialization and are converted back to the hosts architecture upon consumption. This ensures that `IndexByteData` byte sequence is consistent across multiple architectures.

*Testing performed:*
- deployed `soci` and `soci-snapshotter-grpc` to EC2 instance and ran container workload for `rabbitmq` in lazy loading mode and got successful execution.

All the layers on rabbitmq were lazily loaded.
```
soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/1/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
fusectl on /sys/fs/fuse/connections type fusectl (rw,relatime)
soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/2/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/3/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/4/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/5/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/6/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/7/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/8/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/9/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
```

The workload completed successfully.
```
2022-09-27 18:23:09.431972+00:00 [info] <0.779.0> started TCP listener on [::]:5672
 completed with 3 plugins.
2022-09-27 18:23:09.647003+00:00 [info] <0.708.0> Server startup complete; 3 plugins started.
2022-09-27 18:23:09.647003+00:00 [info] <0.708.0>  * rabbitmq_prometheus
2022-09-27 18:23:09.647003+00:00 [info] <0.708.0>  * rabbitmq_web_dispatch
2022-09-27 18:23:09.647003+00:00 [info] <0.708.0>  * rabbitmq_management_agent
```

### DIFF
--- a/c/indexer.c
+++ b/c/indexer.c
@@ -42,11 +42,29 @@
 
 #include "indexer.h"
 
+#include <endian.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
 #define CHUNK (1 << 14) // file input buffer size
+
+offset_t encode_offset(offset_t source) {
+    return htole64(source);
+}
+
+offset_t decode_offset(offset_t source) {
+    return le64toh(source);
+}
+
+int32_t encode_int32(int32_t source) {
+    return htole32(source);
+}
+
+int32_t decode_int32(int32_t source) {
+    return le32toh(source);
+}
 
 void free_index(struct gzip_index *index)
 {
@@ -90,8 +108,8 @@ static struct gzip_index *addpoint(struct gzip_index *index, uint8_t bits,
     /* fill in entry and increment how many we have */
     next = index->list + index->have;
     next->bits = bits;
-    next->in = in;
-    next->out = out;
+    next->in = encode_offset(in);
+    next->out = encode_offset(out);
     if (left)
         memcpy(next->window, window + WINSIZE - left, left);
     if (left < WINSIZE)
@@ -194,9 +212,12 @@ int generate_index_fp(FILE* in, offset_t span, struct gzip_index** idx)
     (void)inflateEnd(&strm);
     index->list = realloc(index->list, sizeof(struct gzip_index_point) * index->have);
     index->size = index->have;
-    index->span_size = span;
+    index->have = encode_int32(index->have);
+    int32_t sz = index->size;
+    index->size = encode_int32(index->size);
+    index->span_size = encode_offset(span);
     *idx = index;
-    return index->size;
+    return sz;
 
     /* return error */
   build_index_error:
@@ -208,7 +229,7 @@ int generate_index_fp(FILE* in, offset_t span, struct gzip_index** idx)
 
 int has_bits(struct gzip_index* index, int point_index)
 {
-    if (point_index >= index->have)
+    if (point_index >= decode_int32(index->have))
     {
         return 0;
     }
@@ -222,12 +243,12 @@ static uint8_t get_bits(struct gzip_index* index, int point_index)
 
 offset_t get_ucomp_off(struct gzip_index* index, int point_index)
 {
-    return index->list[point_index].out;
+    return decode_offset(index->list[point_index].out);
 }
 
 offset_t get_comp_off(struct gzip_index* index, int point_index)
 {
-    return index->list[point_index].in;
+    return decode_offset(index->list[point_index].in);
 }
 
 static int min(int lhs, int rhs)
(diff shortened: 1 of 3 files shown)


c/indexer.c
```diff
@@ -42,12 +42,30 @@
...
+}
+
+int32_t store_size(int32_t source) {
+    return htobe32(source);
+}
+
+int32_t retrieve_size(int32_t source) {
+    return be32toh(source);
+}
+
```
wmesard, line 68: These are not storing or retrieving anything. They should not have *store* or *retrieve* in their names.
Suggest `encode_offset`, `decode_offset`, `encode_int32`, `decode_int32`

wmesard (changes requested):
I hate to say this, but the endian wars are over. The bad guys won. Quoting the FlatBuffers documentation:
> Each scalar is also always represented in little-endian format, as this corresponds to all commonly used CPUs today. FlatBuffers will also work on big-endian machines, but will be slightly slower because of additional byte-swap intrinsics.

We should canonicalize to little-endian, not big-endian. That way these will all be nops on (Linux, Windows) x (x86, ARM).

But speaking of FlatBuffers, this problem would go away if we were modeling the gzip_index data in the fbs file instead of treating it as a binary blob. Should we do that?

c/indexer.c
vkuzniet: Updated the names and also changed the encoding to little-endian.

vkuzniet:
> > I hate to say this, but the endian wars are over. The bad guys won. Quoting the FlatBuffers documentation:
> > > Each scalar is also always represented in little-endian format, as this corresponds to all commonly used CPUs today. FlatBuffers will also work on big-endian machines, but will be slightly slower because of additional byte-swap intrinsics.
> > 
> > 
> > We should canonicalize to little-endian, not big-endian. That way these will all be nops on (Linux, Windows) x (x86, ARM).
> > But speaking of FlatBuffers, this problem would go away if we were modeling the gzip_index data in the fbs file instead of treating it as a binary blob. Should we do that?
> 
> I think, long term this is the right approach. We would anyway needed to do that at the time we re-write the indexer to Go, so this work seems to be inevitable.

Created an issue https://github.com/awslabs/soci-snapshotter/issues/81 to address that. Once it's implemented, encode/decode methods from this PR will be obsolete.

c/indexer.c
```diff
@@ -90,8 +108,8 @@ static struct gzip_index *addpoint(struct gzip_index *index, uint8_t bits,
     /* fill in entry and increment how many we have */
     next = index->list + index->have;
     next->bits = bits;
-    next->in = in;
-    next->out = out;
+    next->in = encode_offset(in);
+    next->out = encode_offset(out);
     if (left)
         memcpy(next->window, window + WINSIZE - left, left);
```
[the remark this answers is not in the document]
vkuzniet: Since the element type for the dictionary is the unsigned char, which size is 1 byte precisely, it does not need to be encoded. BE / LE placement goes by bytes.

Kern-- (approved):

rdpsin (approved):

Issue #75, opened 2022-09-23
Ztoc struct has to be consistent with the structure defined in ztoc.fbs


vkuzniet:
**Is your feature request related to a problem? Please describe.**
Ztoc struct has to be consistent with the structure defined in `ztoc.fbs`. We have two ztoc's and serialization logic copying data from Ztoc in ztoc.go to fbs-generated Ztoc. In order to avoid any bugs, those must be consistent. The fbs ztoc has more logical structure we all agreed on, so Ztoc struct must follow the lead.

**Describe the solution you'd like**
Ztoc struct in ztoc.go needs to be updated to reflect the structural changes agreed on when defining the serialization format in .fbs. 

**Describe alternatives you've considered**

**Additional context**

vkuzniet:
The PR https://github.com/awslabs/soci-snapshotter/pull/82 is merged.

PR #82, opened 2022-09-28
Ztoc format is consistent with fbs

vkuzniet:

*Issue #, if available:*
#75 

*Description of changes:*
Ztoc struct is modified to be consistent with the flatbuffers schema defined in `ztoc.fbs`.

*Testing performed:*
- `make test & make check & make integration` pass
- deployed `soci` and `soci-snapshotter-grpc` to the ec2 instance and successfully executed `rabbitmq` container workload in lazy loading mode.

### DIFF
--- a/soci/ztoc.go
+++ b/soci/ztoc.go
@@ -54,30 +54,32 @@ type FileMetadata struct {
 }
 
 type Ztoc struct {
-	Version             string
-	BuildToolIdentifier string
-
-	Metadata []FileMetadata
-
-	CompressedFileSize   FileSize
-	UncompressedFileSize FileSize
-	MaxSpanID            SpanID //The total number of spans in Ztoc - 1
-	ZtocInfo             ztocInfo
-	IndexByteData        []byte
+	Version                 string
+	BuildToolIdentifier     string
+	CompressedArchiveSize   FileSize
+	UncompressedArchiveSize FileSize
+	TOC                     TOC
+	CompressionInfo         CompressionInfo
 }
 
-type ztocInfo struct {
-	SpanDigests []digest.Digest
+type CompressionInfo struct {
+	MaxSpanID     SpanID //The total number of spans in Ztoc - 1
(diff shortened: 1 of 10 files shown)


soci/ztoc.go
```diff
@@ -54,20 +54,22 @@ type FileMetadata struct {
...
 
 type Ztoc struct {
-	Version             string
-	BuildToolIdentifier string
-
-	Metadata []FileMetadata
-
+	Version              string
+	BuildToolIdentifier  string
 	CompressedFileSize   FileSize
```
rdpsin, line 59: Should change this to `CompressedArchiveSize`. Same with `UncompressedArchiveSize`. 

https://github.com/awslabs/soci-snapshotter/blob/main/soci/fbs/ztoc.fbs#L46

vkuzniet: Good catch. All set.

soci/ztoc.go
```diff
@@ -54,30 +54,32 @@ type FileMetadata struct {
...
+	TOC                     TOC
+	CompressionInfo         CompressionInfo
+}
 
-	CompressedFileSize   FileSize
-	UncompressedFileSize FileSize
-	MaxSpanID            SpanID //The total number of spans in Ztoc - 1
-	ZtocInfo             ztocInfo
-	IndexByteData        []byte
+type CompressionInfo struct {
```
Kern--, line 65: Should compression info contain the algorithm?

rdpsin (approved):

soci/ztoc.go
vkuzniet: It didn't make any sense to add it just yet, since we don't have anything but gzip here (fbs has it already and this is the final destination for serialization). Before putting algorithm in the compression info in Ztoc struct, it's better to devise similarities and differences between Ztocs for gzip and zstd and model how will the new reality look like.

Kern-- (approved):

Issue #78, opened 2022-09-27
Remove libindexer


vkuzniet:
**Is your feature request related to a problem? Please describe.**
soci snapshotter has indexer C component, which at the moment needs to be built separately into `libindexer.a` and then statically linked during snapshotter's compilation. This shouldn't be this way. 

**Describe the solution you'd like**
`gzip_indexer.go` and C files need to be located in the same directory. This will allow `cgo` to search for C files within the same directory where `gzip_index.go` is located and compile/link them on the fly without creating the additional artifact.

**Describe alternatives you've considered**

**Additional context**

wmesard:
Benefit of doing this work: Eliminates a step in the build process. Simplifies the Makefile and the #cgo flags in gzip_index.go

vkuzniet:
Already in main.

PR #86, opened 2022-09-29
Remove dependency on libindexer

vkuzniet:

*Issue #, if available:*
#78 

*Description of changes:*
Removed linking to `libindexer.a`. Also moved `gizp_index.go` and C files to `soci/index`. 
`gzip_index.go` also gets the unmarshaling of gzip_index C struct, which is used in ztoc tests.

*Testing performed:*
- deployed `soci-snapshotter-grpc` and `soci` on ec2 instance and successfully ran the workload for tomcat in lazy loading mode.


soci/index/gzip_index.go
```diff
@@ -14,10 +14,10 @@
    limitations under the License.
 */
 
-package soci
+package index
```
Kern--, line 17: I'm not so sure about this structure. I think what we really want is a full ztoc package.

vkuzniet: Package soci contains soci index and ztoc definition, building and accessing - this is a full ztoc package. I moved `gzip_index.c` under `index` package to better group it with the C code. The ztoc building logic relies on `gzip_index`. 

Another option would be to simply keep C code under soci, but that bothers a little bit, since we place C files along with Go files, which don't even use the C implementations directly.

Kern--: Index here is confusing. You rename it `sociindex` when importing, but this isn't about soci indices, it's about gzip indices within a zTOC. I think we had discussed calling this `zindex` at one point which is a little more clear. This change also pushes `FileSize` down into `index`, but even gzip zindices don't *really* care about file sizes, that just happens to be the type we're using to represent offsets. I think it would be more clear to create a `ztoc` package that contains all the zTOC code, including the gzip code. 

Overall, I think we should be moving towards separating zTOC from SOCI. zTOCs are the technology that enables SOCI, but they make sense as a concept on their own.

vkuzniet:
Closing this PR in favor of #120

Issue #79, opened 2022-09-27
Unit tests fail using Go 1.19 because of data race


rdpsin:
**Describe the bug**
The unit tests inside the `soci/` directory fail due a data race. This only happens when using Go 1.19.x, not the previous versions of Go. 

**Steps To Reproduce**
```shell
$ go version
go version go1.19 linux/amd64

$ make test 
...
==================
WARNING: DATA RACE
Write at 0x00c001509038 by goroutine 55:
  runtime.racewriterange()
      <autogenerated>:1 +0x29
  internal/poll.(*FD).Pread()
      /usr/local/go/src/internal/poll/fd_unix.go:193 +0x169
  os.(*File).pread()
      /usr/local/go/src/os/file_posix.go:40 +0x335
  os.(*File).ReadAt()
      /usr/local/go/src/os/file.go:136 +0x2de
  io.(*SectionReader).ReadAt()
      /usr/local/go/src/io/io.go:552 +0x205
  github.com/awslabs/soci-snapshotter/soci.ExtractFile.func1()
      /home/ec2-user/soci-snapshotter/soci/ztoc.go:141 +0x1db
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /home/ec2-user/go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57 +0x91

Previous write at 0x00c001509039 by goroutine 56:
  runtime.racewriterange()
      <autogenerated>:1 +0x29
  internal/poll.(*FD).Pread()
      /usr/local/go/src/internal/poll/fd_unix.go:193 +0x169
  os.(*File).pread()
      /usr/local/go/src/os/file_posix.go:40 +0x335
  os.(*File).ReadAt()
      /usr/local/go/src/os/file.go:136 +0x2de
  io.(*SectionReader).ReadAt()
      /usr/local/go/src/io/io.go:552 +0x205
  github.com/awslabs/soci-snapshotter/soci.ExtractFile.func1()
      /home/ec2-user/soci-snapshotter/soci/ztoc.go:141 +0x1db
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /home/ec2-user/go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57 +0x91

Goroutine 55 (running) created at:
  golang.org/x/sync/errgroup.(*Group).Go()
[...]

PR #88, opened 2022-09-30
Fix race condition in ztoc.ExtractFile

vkuzniet:

*Issue #, if available:*
#79 

*Description of changes:*
Reading to buffer in `ztoc.ExtractFile` is now guarded by mutex. The reason the race condition was happening is due to the fact that two or more intersecting ranges of the slice can be written simultaneously. The single mutex should be OK for that case, since all the data accessed via `ztoc.ExtractFile` is local. For remote data, the span manager is used with more sophisticated locking.

*Testing performed:*
- `make test` pass with Go 1.19
- `make test && make check && make integration` pass with the current configuration.

### DIFF
--- a/soci/ztoc.go
+++ b/soci/ztoc.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"sync"
 	"time"
 
 	"github.com/opencontainers/go-digest"
@@ -130,6 +131,7 @@ func ExtractFile(r *io.SectionReader, config *FileExtractConfig) ([]byte, error)
 
 	buf := make([]byte, bufSize)
 	eg, _ := errgroup.WithContext(context.Background())
+	var mu sync.Mutex
 
 	// Fetch all span data in parallel
 	for i = 0; i < numSpans; i++ {
@@ -140,6 +142,8 @@ func ExtractFile(r *io.SectionReader, config *FileExtractConfig) ([]byte, error)
 			if j == 0 && firstSpanHasBits {
 				rangeStart--
 			}
+			mu.Lock()
+			defer mu.Unlock()
 			n, err := r.ReadAt(buf[rangeStart-start:rangeEnd-start+1], int64(rangeStart)) // need to convert rangeStart to int64 to use in ReadAt
 			if err != nil {
 				return err


soci/ztoc.go
```diff
@@ -140,6 +142,8 @@ func ExtractFile(r *io.SectionReader, config *FileExtractConfig) ([]byte, error)
 			if j == 0 && firstSpanHasBits {
 				rangeStart--
 			}
+			mu.Lock()
```
rdpsin, line 145: I'm a little confused by the need for this mutex. Theoretically, writes to the slice ranges should not overlap.

Kern-- (approved):

soci/ztoc.go
vkuzniet: When `firstSpanHasBits` holds true for the certain range, its `rangeStart`, which was initially `starts[<span_id>]` gets decremented (https://github.com/awslabs/soci-snapshotter/blob/main/soci/ztoc.go#L141). Since every `ReadAt` operation goes in its own goroutine, there is a likelihood that `rangeStart` with `firstSpanHasBits` will also be intersecting with preceding interval.

rdpsin: But we only decrement it for the first span. No other spans' `rangeStart` gets decremented.

vkuzniet: This can happen if you are extracting multiple files from the same `tar.gz` as described in TestDecompress. In this case there can be the situation like this, where the span ranges for the current and the following file intersect: 
```
[vkuzniet] spanId=0, rangeStart=3940936, rangeEnd=3953290
[vkuzniet] spanId=1, rangeStart=3953290, rangeEnd=3965604
[vkuzniet] spanId=2, rangeStart=3965604, rangeEnd=3977923
[vkuzniet] spanId=6, rangeStart=4014846, rangeEnd=4021196
[vkuzniet] spanId=4, rangeStart=3990223, rangeEnd=4002533
[vkuzniet] spanId=3, rangeStart=3977923, rangeEnd=3990223
[vkuzniet] spanId=5, rangeStart=4002533, rangeEnd=4014846
[vkuzniet] spanId=0, rangeStart=4014845, rangeEnd=4021196
--- FAIL: TestDecompress (8.39s) testing.go:1319: race detected during execution of test
```

rdpsin (approved):

PR #96, opened 2022-10-05
Remove dubious Makefile actions for `check-flatc`

wmesard:
*Description of changes:*

What we want this rule to do is update the machine-generated code if necessary. What it was doing was that plus also failing the build if any other files were staged or unstaged. That's funny, but it causes `make check` to fail, which is less funny.

*Testing performed:*

1. Make a trivial change: `sed -i 's/Soci/SOCI/' version/version.go`
2. Run check:
```
$ make check-flatc
flatc -o /local/home/wsm/src/awslabs/soci-snapshotter/soci/fbs -g /local/home/wsm/src/awslabs/soci-snapshotter/soci/fbs/ztoc.fbs
 check-flatc
 M version/version.go
diff --git a/version/version.go b/version/version.go
index bb54b5b..75abb0b 100644
--- a/version/version.go
+++ b/version/version.go
@@ -1,5 +1,5 @@
 /*
-   Copyright The Soci Snapshotter Authors.
+   Copyright The SOCI Snapshotter Authors.

    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
 please run 'make flatc' when making changes to fbs files
make: *** [check-flatc] Error 1
```

With this fix applied:

```
$ g s --short
 M version/version.go
$ make check-flatc
flatc -o /local/home/wsm/src/wmesard/soci/soci/fbs -g /local/home/wsm/src/wmesard/soci/soci/fbs/ztoc.fbs
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,10 +76,6 @@ flatc:
 	flatc -o $(CURDIR)/soci/fbs -g $(CURDIR)/soci/fbs/ztoc.fbs
 
 check-flatc: flatc ## check if protobufs needs to be generated again
-	@echo "$(WHALE) $@"
-	@test -z "$$(git status --short | grep ".go" | tee /dev/stderr)" || \
-		((git diff | cat) && \
-		(echo "$(ONI) please run 'make flatc' when making changes to fbs files" && false))
 
 # "check-lint" depends "pre-build". out/libindexer.a seems needed to process cgo directives
 check-lint: pre-build


vkuzniet (approved):
Thanks for doing that. Created a bug https://github.com/awslabs/soci-snapshotter/issues/97 to make it right.

Makefile
```diff
@@ -76,10 +76,6 @@ flatc:
 	flatc -o $(CURDIR)/soci/fbs -g $(CURDIR)/soci/fbs/ztoc.fbs
 
 check-flatc: flatc ## check if protobufs needs to be generated again
```
rdpsin, line 78: If I understand this correctly, `make flatc` will actually generate new files for any changes to `*.fbs` files. So, `make check-flatc` isn't doing any checking, just updating the flatbuffer go files. 

Example:

```
% sed -i 's/CompressionAlgorithm/CompressionAlgorithms/' soci/fbs/ztoc.fbs


% git status
On branch check-flatc
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   soci/fbs/ztoc.fbs


% make check-flatc
flatc -o /local/home/rdpsin/soci/soci/fbs -g /local/home/rdpsin/soci/soci/fbs/ztoc.fbs


% git status
On branch check-flatc
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   soci/fbs/ztoc.fbs
	modified:   soci/fbs/ztoc/CompressionInfo.go

```

I would expect `make check-flatc` to only check for changes, not generate new files.  Unless that's what we actually want to happen.

wmesard: > I would expect make check-flatc to only check for changes, not generate new files. Unless that's what we actually want to happen.

That's what was happening before. Changing that should require an issue so that it can be groomed/discussed/whatever. So I'm saying that's beyond the scope of this PR, which is just to fix something that is self-evidently broken.

But since you bring it up... the current behavior is a little surprising. `flatc` does not have a `--dryrun` option, so we would have to do something janky like run it with `-o /tmp/$$` and then compare the resulting files to the existing files, and error out if they differ.

Kern--: I don't think that `check-flatc` is fundamentally broken, the broken part is running it outside of CI. We *do* want this check to fail in CI because that means that someone committed a change to the flatbuffer files without also committing the generated files.

Containerd does this same process for protobufs, but it specifically only intended for failing CI:
https://github.com/containerd/containerd/blob/main/Makefile#L183

rdpsin (approved):

Makefile
wmesard: The containerd rule is looking for modified `pb.go` rules anywhere in the workspace. The soci-snapshotter rule is looking for `.go` files anywhere in the workspace. In the context of CI, those are both circumstantial indicators that they need to be regenerated. 

In the containerd case, it is benign when run in a dev workspace and will catch the problem if the workspace is in just the right state. In the soci-snapshotter case, it breaks `check` in dev workspaces. That seems fundamentally broken to me.

A better test would be the one I describe above (generate the files to a tmpdir and compare the results), since that can be run at any time regardless of the state the files in git, the build environment, or the stage of the build. This is better, because then we wouldn't have to wait until a change hits the CI pipeline to find the problem.

As @rdpsin mentioned above, it is surprising (and I would argue, fundamentally broken) for a `check-*` rule to depend on another rule (`flatc`) that modifies the workspace.

Also, I see that `check-flatc` is actually _not_ being run in [workflows](https://github.com/awslabs/soci-snapshotter/blob/main/.github/workflows/build.yml) today. So that should also be fixed.

See #101

PR #98, opened 2022-10-05
Improve the naming consistency around two zTOC concepts

wmesard:
*Description of changes:*

We’ve struggled to use consistent naming for the second half of the zTOC file, and for the captured state of the compression engine. This commit attempts to remedy that by using “Gzip compression info” and “Checkpoint”, respectively.

*Notes*:
* This change was generated by a script (attached below).
* If someone prefers different terminology, let's discuss. "Compression info" is a bit of a mouthful. But that's what the Flatbuffer schema used, and I can't think of anything better.
* If you see additional things that should be renamed, let me know and I will add them to the list and rerun the script.
* In some cases, the “gzip” is redundant (because the code in question is obviously gzip-specific), or incorrect (because the code in question has no business knowing about the implementation details of the compression algorithm). I left that as is for now, because that would require a judgement call, and I wanted to keep this commit entirely scripted. 

*Testing performed:*
```
go clean -testcache && make clean && make && make test 
make check
```

### DIFF
--- a/c/zinfo.h
+++ b/c/zinfo.h
@@ -40,8 +40,8 @@
   from the original.
 */
 
-#ifndef INDEXER_H
-#define INDEXER_H
+#ifndef ZINFO_H
+#define ZINFO_H
 
 #include <stdbool.h>
 #include <stdint.h>
@@ -58,13 +58,13 @@ typedef int64_t offset_t;
 
 enum 
 {
-    GZIP_INDEXER_OK = 0,
-    GZIP_INDEXER_FILE_NOT_FOUND = -80,
-    GZIP_INDEXER_INDEX_NULL = -81,
-    GZIP_INDEXER_CANNOT_ALLOC = -82,
+    GZIP_ZINFO_OK = 0,
+    GZIP_ZINFO_FILE_NOT_FOUND = -80,
+    GZIP_ZINFO_INDEX_NULL = -81,
+    GZIP_ZINFO_CANNOT_ALLOC = -82,
 };
 
-struct gzip_index_point
+struct gzip_checkpoint
 {
     offset_t out;          /* corresponding offset in uncompressed data */
     offset_t in;           /* offset in input file of first full byte */
@@ -72,11 +72,11 @@ struct gzip_index_point
     unsigned char window[WINSIZE];  /* preceding 32K of uncompressed data */    
 };
 
-struct gzip_index 
+struct gzip_zinfo 
 {
     int32_t have;           /* number of list entries filled in */
     int32_t size;           /* number of list entries allocated */
-    struct gzip_index_point *list; /* allocated list */
+    struct gzip_checkpoint *list; /* allocated list */
     offset_t span_size;
 };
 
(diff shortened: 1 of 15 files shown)


wmesard:
Oops. GitHub won't let you attach shell scripts to PRs. Here it is inline:
```
#!/bin/zsh

git reset --hard
while read old new; do
    sed -i "s/${old}/${new}/g" **/*.{c,h,go,fbs}
done <<EOF
gzip_index_point gzip_checkpoint
gzip_index       gzip_compression_info
gzipIndexPoint   gzipCheckpoint
GzipIndex        GzipCompressionInfo
gzipIndex        gzipCompressionInfo
IndexByteData    Checkpoints
indexByteData    checkpoints
index_byte_data  checkpoints
indexData        checkpoints
EOF


for d in $(git ls-files -m | xargs dirname  | sort -u); do
    pushd $d
    go fmt
    popd
done

sed -i 's/"snapshots"/*checkpoints*/g' docs/GLOSSARY.md
```

c/indexer.c
```diff
@@ -66,7 +66,7 @@ int32_t decode_int32(int32_t source) {
     return le32toh(source);
 }
 
-void free_index(struct gzip_index *index)
+void free_index(struct gzip_compression_info *index)
```
Kern--, line 69:
```suggestion
void free_compression_info(struct gzip_compression_info *index)
```

c/indexer.c
```diff
@@ -76,16 +76,16 @@ void free_index(struct gzip_index *index)
 
 /* Add an entry to the access point list.  If out of memory, deallocate the
    existing list and return NULL. */
-static struct gzip_index *addpoint(struct gzip_index *index, uint8_t bits,
+static struct gzip_compression_info *addpoint(struct gzip_compression_info *index, uint8_t bits,
```
Kern--, line 79:
```suggestion
static struct gzip_compression_info *add_checkpoint(struct gzip_compression_info *index, uint8_t bits,
```

Kern--: The variable is still named `index` here and throughout this file. Is changing that in scope?

c/indexer.c
```diff
@@ -123,12 +123,12 @@ static struct gzip_index *addpoint(struct gzip_index *index, uint8_t bits,
 
 
 /* Pretty much the same as from zran.c */
-int generate_index_fp(FILE* in, offset_t span, struct gzip_index** idx)
+int generate_index_fp(FILE* in, offset_t span, struct gzip_compression_info** idx)
```
Kern--, line 126: I'm not entirely sure what this name means, but it doesn't look right given the renaming.

c/indexer.c
wmesard: The reason I didn't go there is because
1. I wasn't sure where it would end. Do I also rename the C files?
2. I wasn't sure what to rename it *to*. As I said, "compression_info" is a mouthful. It's one thing in types and function names. As a variable name, it really starts to harm readability.

To be honest, I was kind of hoping that "compression_info" was a bad enough choice that it would spur one of the reviewers here to come up with a better alternative. Seeing none, let me throw this out there: What do you think about **zinfo**? "z" clearly connotes compression, so it's meaning is still pretty clear. Plus, it's fun to say. It's got about the same level of whimsy as "TOC" (which is the other component of the ztoc). And it's short enough that I would feel comfortable using it in file and variable names.

Thoughts 👍 or 👎 ?

c/indexer.c
wmesard: Agreed

c/indexer.c
Kern--: I think that's ok. If we end up moving towards a `compression` package in go, I would prefer we call it `compression.Info` there instead of `compression.ZInfo`, but zinfo works pretty well as a variable name.

I wish we had a more descriptive name than "info" but I've spent too much time in front of a thesaurus with nothing better to block on that.

wmesard:
This second rev addresses the comments above. It also contains separate commits to update the glossary. 
Notes:
* There are still a few places where the old terms are used. I decided to draw the line, rather than get overly dogmatic about it.
* The changes in *CompressionInfo.go* were made by hand, because it was safer that way. All the other code changes were done using an updated version of the shell script:
```
#!/bin/zsh

git reset --hard
while read old new; do
    sed -i "s/${old}/${new}/g" **/*.{c,h,go,fbs}
done <<EOF
INDEXER          ZINFO
point_index      checkpoint
gzip_index_point gzip_checkpoint
gzip_index       gzip_zinfo
gzipIndexPoint   gzipCheckpoint
GzipIndex        GzipZinfo
gzipIndex        gzipZinfo
blob_to_index    blob_to_zinfo
IndexByteData    Checkpoints
indexByteData    checkpoints
index_byte_data  checkpoints
indexData        checkpoints
free_index       free_zinfo
addpoint         add_checkpoint
generate_index_fp  generate_zinfo
indexer.h        zinfo.h
EOF


for d in $(git ls-files -m | xargs dirname  | sort -u); do
    pushd $d
    go fmt
    popd
done

# Finally, clean up any stray references to "index" in the C code.
cd c
mv indexer.c zinfo.c
mv indexer.h zinfo.h
git add zinfo.[ch]
cd ..

sed -i 's/indexer/zinfo/g' Makefile soci/gzip_index.go

# Also replaced the "index" variables and struct member by hand

mv soci/gzip_index.go soci/gzip_zinfo.go

git add -A
```

Kern-- (approved):

c/zinfo.h
```diff
@@ -40,8 +40,8 @@
   from the original.
 */
 
-#ifndef INDEXER_H
-#define INDEXER_H
+#ifndef ZINFO_H
```
vkuzniet, line 43: Maybe instead `zinfo.h`, rename it `gzip_zinfo.h`? This way, we will not need to rename it when there will be the implementation to build Zinfo for other compression algorithms.

wmesard: The scope of this change was strictly about clarifying the naming around zinfo and checkpoints. I resisted the urge to add or remove the GZIP-ness from names of things, even in cases like this where--as you point out--it would improve the code to do so.

I think that that is going to be part of the work of adding the missing interfaces around compression to make the world safe for adding ZStandard support. That is going to be a much larger effort involving the creation of new `interface`s and moving source files--including these--to new directories.

vkuzniet: Sounds good.

vkuzniet (approved):

Issue #101, opened 2022-10-06
Improve `check-flatc` rule and its invocation


wmesard:
**Is your feature request related to a problem? Please describe.**

As discussed in PR #96 , the `check-flatc` rule in the Makefile could use some fixing. Also, it should be invoked by the build workflow.

**Describe the solution you'd like**
1. It is surprising (and I would argue, fundamentally broken) for a `check-*` rule to depend on another rule (`flatc`) that modifies the workspace. It should stop doing that.
1. `flatc` does not have a `--dryrun` option. So to detect if the generated files need to be updated, `check-flatc` could run it with `-o /tmp/$$` and then compare the resulting files to the existing files, and error out if they differ.
1. `check-flatc` is not being invoked in CI today. Workflows should invoke `make check` instead individual `check-*` rules, to fix this problem and prevent it from occurring for future `check-*` rules.

**Additional context**

See discussion in #96.

rdpsin:
Closing since #102 and #104  have been merged.

PR #104, opened 2022-10-07
Add `make check-flatc` to build workflow

rdpsin:
*Issue #, if available:*
#101 

*Description of changes:*

*Testing performed:*

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,10 +17,13 @@ jobs:
           # in order to calculate the range. i.e. HEAD~20..HEAD includes
           # 20 commits, but including HEAD it needs 21 commits.
           fetch-depth: 21
+      - run: wget https://github.com/google/flatbuffers/releases/download/v22.9.29/Linux.flatc.binary.g++-10.zip
+      - run: unzip Linux.flatc.binary.g++-10.zip
       - run: make install-check-tools
       - run: make check-ltag
       - run: make check-dco
       - run: make check-lint
+      - run: PATH=$PATH:$(pwd) make check-flatc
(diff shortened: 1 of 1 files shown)


wmesard (dismissed):

.github/workflows/build.yml
```diff
@@ -17,10 +17,13 @@ jobs:
           # in order to calculate the range. i.e. HEAD~20..HEAD includes
           # 20 commits, but including HEAD it needs 21 commits.
           fetch-depth: 21
+      - run: wget https://github.com/google/flatbuffers/releases/download/v22.9.29/Linux.flatc.binary.g++-10.zip
+      - run: unzip Linux.flatc.binary.g++-10.zip
       - run: make install-check-tools
       - run: make check-ltag
       - run: make check-dco
       - run: make check-lint
+      - run: PATH=$PATH:$(pwd) make check-flatc
```
wmesard, line 26: Why does cwd have to be in PATH? If it does, that seems like a bug that should be fixed in the Makefile.

rdpsin: `cwd` was a mistake on my part. Changed it to `pwd`. The `flatc` binary is installed in `pwd`, since (unlike the Makefile), we are downloading the pre-built binary in the Workflow. (I didn't really want to do `make install-cmake && make install-flatc` on every run.)

wmesard: I see. I might have installed it to /tmp and then added that to PATH. (I really dislike how much we scribble in the source directory during build.) But okay fine.

vkuzniet (approved):

PR #105, opened 2022-10-10
Set `backgroundFetchSilencePeriod` to 0

rdpsin:
This commit sets the backgroundFetchSilencePeriod for the backgroundTaskManager to 0. Additionally, add a new field to the config to make it configurable.

*Issue #, if available:*

*Description of changes:*
Once the container is started, the prefetcher fetches spans in the background. It acquires the span's lock (in the span manager) and releases it once done. If an on-demand fetch comes in, the bg tasks are cancelled and restarted, but only after a 'silence period' of 5s. But the prefetcher is still holding onto the lock, which means on-demands fetches for those particular spans will have to wait 5+ seconds until the prefetcher releases the lock. 

A short-term solution is to the `silence period` to 0s, which is what this PR is doing. 

Longer term solution is to rewrite the prefetcher, so that it yields the lock when cancelled.

*Testing performed:*

### DIFF
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -53,15 +53,16 @@ const (
 )
 
 type Config struct {
-	HTTPCacheType       string `toml:"http_cache_type"`
-	FSCacheType         string `toml:"filesystem_cache_type"`
-	ResolveResultEntry  int    `toml:"resolve_result_entry"`
-	NoBackgroundFetch   bool   `toml:"no_background_fetch"`
-	Debug               bool   `toml:"debug"`
-	AllowNoVerification bool   `toml:"allow_no_verification"`
-	DisableVerification bool   `toml:"disable_verification"`
-	MaxConcurrency      int64  `toml:"max_concurrency"`
-	NoPrometheus        bool   `toml:"no_prometheus"`
+	HTTPCacheType                    string `toml:"http_cache_type"`
+	FSCacheType                      string `toml:"filesystem_cache_type"`
+	ResolveResultEntry               int    `toml:"resolve_result_entry"`
+	NoBackgroundFetch                bool   `toml:"no_background_fetch"`
+	Debug                            bool   `toml:"debug"`
+	AllowNoVerification              bool   `toml:"allow_no_verification"`
+	DisableVerification              bool   `toml:"disable_verification"`
+	MaxConcurrency                   int64  `toml:"max_concurrency"`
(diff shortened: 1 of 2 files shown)


fs/config/config.go
```diff
@@ -53,15 +53,16 @@ const (
...
+	HTTPCacheType                    string `toml:"http_cache_type"`
+	FSCacheType                      string `toml:"filesystem_cache_type"`
+	ResolveResultEntry               int    `toml:"resolve_result_entry"`
+	NoBackgroundFetch                bool   `toml:"no_background_fetch"`
+	Debug                            bool   `toml:"debug"`
+	AllowNoVerification              bool   `toml:"allow_no_verification"`
+	DisableVerification              bool   `toml:"disable_verification"`
+	MaxConcurrency                   int64  `toml:"max_concurrency"`
+	NoPrometheus                     bool   `toml:"no_prometheus"`
+	BackgroundFetchSilencePeriodMSec int    `toml:"background_fetch_silence_period_msec"`
```
wmesard, line 65: If we just specified all our configurable time intervals in milliseconds, we wouldn't have to have this awkward naming. But okay.

Also, task manager calls this `prioritizedTaskSilencePeriod.` I think I prefer that name, because it's clear what we're being silent in relation to. The trouble with `BackgroundFetchSilencePeriod`, is that someone might think it's a startup thing that prevents background tasks from running until X seconds post-launch. We do need such a config parameter as well. What would you call that, that would clearly disambiguate it from this?

wmesard (commented):
1. Okay. Sure. I guess. But the real problem here is that the current prefetcher is a placeholder. We need a prefetcher that is actually designed and measured. Before I can shipit this PR, I would like to see a Spike issue added to the [backlog](https://github.com/orgs/awslabs/projects/62/views/1), to do that design work.
2. Please update the Testing section or the Description to explain why this change is beneficial.

rdpsin:
Add a comment in the Description section to explain the change.

fs/config/config.go
rdpsin: Renamed to `PrioritizedTaskSilencePeriodMSec`.

rdpsin:
Created an issue https://github.com/awslabs/soci-snapshotter/issues/106 to track prefetcher redesign.

wmesard (approved):

vkuzniet (approved):

PR #107, opened 2022-10-11
Dedupe integ test mirror/dockerhub functions

Kern--:

*Issue #, if available:*

*Description of changes:*
Before this change, many tests defined the same set of helper functions. This change extracts those into reusable objects.

*Testing performed:*

### DIFF
--- a/integration/create_test.go
+++ b/integration/create_test.go
@@ -36,10 +36,6 @@ func TestSociCreateSparseIndex(t *testing.T) {
 	defer done()
 
 	rebootContainerd(t, sh, "", "")
-
-	dockerhub := func(name string) imageInfo {
-		return imageInfo{dockerLibrary + name, "", false}
-	}
 	tests := []struct {
 		name         string
 		minLayerSize int64
@@ -139,10 +135,6 @@ func TestSociCreate(t *testing.T) {
 	defer done()
 
 	rebootContainerd(t, sh, "", "")
-
-	dockerhub := func(name string) imageInfo {
-		return imageInfo{dockerLibrary + name, "", false}
-	}
 
 	tests := []struct {
 		name           string
--- a/integration/pull_test.go
+++ b/integration/pull_test.go
@@ -49,20 +49,10 @@ import (
 )
 
 const (
-	defaultContainerdConfigPath  = "/etc/containerd/config.toml"
-	defaultSnapshotterConfigPath = "/etc/soci-snapshotter-grpc/config.toml"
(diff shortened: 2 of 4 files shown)


rdpsin:
As long we're making these changes, maybe change `util_test.go` to just `utils.go`? I'm being pedantic, but `util_test.go` sounds like we're testing a `utils.go`.

vkuzniet:
> As long we're making these changes, maybe change `util_test.go` to just `utils.go`? I'm being pedantic, but `util_test.go` sounds like we're testing a `utils.go`.

Maybe `testutils.go`? This will be consistent with other places within the snapshotter and will tell the reader what is it for.

integration/util.go
```diff
@@ -88,13 +90,74 @@ func isDirExists(sh *shell.Shell, dir string) bool {
 	return sh.Command("test", "-d", dir).Run() == nil
 }
 
+func dockerhub(name string) imageInfo {
```
vkuzniet, line 93: can you move `imageInfo` to here? it's still defined in `pull_test.go`.

rdpsin (dismissed):

Kern--:
The `testutil.go` file is already 400 LOC and a mix of a bunch of unrelated bits. I'm planning to create a new `testutils` package under `integration` so we can break these up into easier to digest pieces.

Something like
```
integration/
    testutil/
        constants.go
        registry.go
        ...
```

For this PR, i'll probably leave `testutil.go` and move just the registry config stuff into a the package. Then I'll have follow up PRs to move everything else. Objections?

rdpsin:
SGTM.

Kern--:
I started trying to break up my changes into a new package, but these utils are pretty interdependent. I'd rather do the refactor into a separate package as a separate PR.

vkuzniet (approved):

wmesard (approved):
I'm approving this, because it's a nice cleanup. But test-only code really should be named `*_test.go` so that it doesn't inadvertently get compiled into the production artifacts.

In this particular case, it doesn't matter because no other packages import from the *integration* package. I just worry that naming it `testutils.go` here will establish a pattern that will bite us elsewhere. (And, yes, I know that we already have this problem elsewhere.)

See https://stackoverflow.com/a/40949505/7020851

Kern--:
I will be moving more things around in future PRs and will rename this back to `utils_test.go` with a new shiny comment at the top of the file so someone else (hopefully) doesn't feel tempted to go through this dance.

PR #92, opened 2022-10-04
[BENCHMARKS 1/x] Add base framework and example

sbuckfelder:
Signed-off-by: Scott Buckfelder <buckscot@amazon.com>

This is the first of a series of PRs that builds out performance benchmark testing for soci-snapshotter.

*Description of changes:*
This PR does 5 main things:
- it adds a main file with an example benchmark test for benchmark that is used to compile a benchmark binary.  These examples will be removed in future PRs, but do provide an example of functionality like passing in arguments (imagine different test images)
- adds a framework package (later envisioned to live in containerd upstream project)
- adds a Makefile target to build and run the benchmark binary
- updates `go.mod` to include a statistics package and the framework package
- updates `.gitignore` to exclude the benchmarking output

The way this is envisioned to be used is that the `framework` package will include anything for the benchmark framework or containerd related.  All of the soci specific logic will live in the top level benchmark directory / package.  The main concepts here are the **BenchmarkTestDriver**, this is a struct that houses the function you are benchmarking and stores all of the results for that function, and the **BenchmarkFramework**, this is the struct that has the list of  BenchmarkTestDrivers and runs them.  The benchmarks are run using the golang testing package, but the key difference is that the tests are run individually.  That allows us to run custom statistics on them (the default golang benchmarking statistics only includes averages).  Finally, everything is written out to a json file.

*Testing performed:*
`make && make test && make check`
`make benchmarks` results -->
```
{
 "commit": "474b26470659476af010cd515b7088d76c930e17",
 "benchmarkTests": [
  {
   "testName": "TwoSecondSleep",
   "numberOfTests": 10,
   "testsRun": 0,
   "stdDev": 0.0005865096459530768,
   "mean": 2.0006229670000004,
   "min": 2.000099263,
   "pct25": 2.0000997145,
   "pct50": 2.00015568,
   "pct75": 2.001075953,
   "pct90": 2.001147467,
   "max": 2.001834204
  },
  {
   "testName": "TwoSecondRandSleep",
   "numberOfTests": 10,
   "testsRun": 0,
   "stdDev": 0.5001335439728879,
   "mean": 0.5001444039,
   "min": 0.000010809,
   "pct25": 0.000010888,
   "pct50": 0.000011036,
   "pct75": 1.0000986754999999,
   "pct90": 1.000149617,
   "max": 1.000995927
  },
  {
   "testName": "ThreeSecondSleep",
   "numberOfTests": 10,
   "testsRun": 0,
   "stdDev": 0.001095211523873395,
[...]

PR #116, opened 2022-10-18
Add benchmarking framework and tests

sbuckfelder:
This adds a benchmark framework and tests for the soci snapshotter and overlayfs.

### High Level Description of the Benchmarking Framework
Each unit of work that we are benchmarking is a `BenchmarkTestDriver`.  These are then collected in a `BenchmarkFramework`object, that then runs each test driver and writes out statistics.  The benchmark framework relies on the go testing library.  So each `BenchmarkTestDriver` has a struct field `TestFunction  func(*testing.B)` that contains the actual test. 

The `BenchmarkTestDriver`'s are defined in `benchmark/main.go` and are built up of from the different lifecycle actions defined in containerd.  All of these lifecycle actions are accessible via helper functions in `benchmark/soci_utils.go`, `benchmark/framework/containerd_utils.go`, and `benchmark/framework/aws_utils.go`.

### Details of Changes
The benchmarker adds a new binary build which runs benchmarking tests.  This is called via the make target _benchmarks_ (e.g. run `make benchmarks`).  It takes in a .csv file as its input.  Each line of the csv file is a different workload that every test is run against. 

**benchmark/main.go** - this lists the tests that are being run, sets the 'global variables', and takes care of the csv parsing.  Also the main func for the binary.
**benchmark/benchmarkTests.go** - this defines all of the individual tests. 
**benchmark/framework/framework.go** - this contains benchmarking framework ideas such as TestDrivers, etc..  It also the main loop that runs the collection of tests.
**benchmark/soci_utils.go** - adds soci specific lifecycle actions and objects
**benchmark/framework/aws_utils.go** - adds aws specific lifecycle actions and objects, important for ECR
**benchmark/framework/containerd_utils.go** - adds containerd specific lifecycle actions and objects
**benchmark/soci_config.toml** - soci config file
**benchmark/containerd-config.toml** - containerd config file
**.gitignore** - ignores the artifcats created in benchmarking
**Makefile** - adds the make benchmarks target
**go.mod** - updated due to adding a stats library
**go.sum** - updated due to adding a stats library

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ GO_LD_FLAGS=-ldflags '-s -w -X $(PKG)/version.Version=$(VERSION) -X $(PKG)/versi
 SOCI_SNAPSHOTTER_PROJECT_ROOT ?= $(shell pwd)
 LTAG_TEMPLATE_FLAG=-t ./.headers
 FBS_FILE_PATH=$(CURDIR)/soci/fbs/ztoc.fbs
+COMMIT=$(shell git rev-parse HEAD)
 
 CMD=soci-snapshotter-grpc soci
 
@@ -126,3 +127,11 @@ integration: pre-build
 	@echo "$@"
 	@echo "SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT)"
 	@GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true go test $(GO_TEST_FLAGS) -v -timeout=0 ./integration
+
+benchmarks:
+	@echo "$@"
+	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests . && sudo ../bin/PerfTests $(COMMIT) ../singleImage.csv 10
+
+benchmarks-comp:
+	@echo "$@"
+	@cd benchmark/comparisonTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/CompTests . && sudo ../bin/CompTests $(COMMIT) ../singleImage.csv 10
--- a/benchmark/benchmarkTests.go
+++ b/benchmark/benchmarkTests.go
@@ -1,0 +1,173 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package benchmark
+
+import (
+	"context"
+	"testing"
+
+	"github.com/awslabs/soci-snapshotter/benchmark/framework"
+)
+
+var (
+	outputDir         = "./output"
+	containerdAddress = "/tmp/containerd-grpc/containerd.sock"
+	containerdRoot    = "/tmp/lib/containerd"
+	containerdState   = "/tmp/containerd"
+	containerdConfig  = "../containerd-config.toml"
+	platform          = "linux/amd64"
+	sociBinary        = "../../out/soci-snapshotter-grpc"
+	sociAddress       = "/tmp/soci-snapshotter-grpc/soci-snapshotter-grpc.sock"
+	sociRoot          = "/tmp/lib/soci-snapshotter-grpc"
+	sociConfig        = "../soci_config.toml"
+	awsSecretFile     = "../aws_secret"
+)
+
+func PullImageFromECR(b *testing.B, imageRef string) {
+	ctx, cancelCtx := framework.GetTestContext()
+	containerdProcess, err := getContainerdProcess(ctx)
+	if err != nil {
+		b.Fatalf("Error Starting Containerd: %v\n", err)
+	}
+	defer containerdProcess.StopProcess(cancelCtx)
+	b.ResetTimer()
+	_, err = containerdProcess.PullImageFromECR(ctx, imageRef, platform, awsSecretFile)
+	if err != nil {
+		b.Fatalf("Error Pulling Image: %v\n", err)
+	}
+	b.StopTimer()
+	err = containerdProcess.DeleteImage(ctx, imageRef)
+	if err != nil {
+		b.Fatalf("Error Deleting Image: %v\n", err)
+	}
+}
+
+func SociRPullPullImage(
+	b *testing.B,
+	imageRef string,
+	indexDigest string) {
+	ctx, cancelCtx := framework.GetTestContext()
+	containerdProcess, err := getContainerdProcess(ctx)
+	if err != nil {
+		b.Fatalf("Failed to create containerd proc: %v\n", err)
+	}
+	defer containerdProcess.StopProcess(cancelCtx)
+	sociProcess, err := getSociProcess()
+	if err != nil {
+		b.Fatalf("Failed to create soci proc: %v\n", err)
+	}
+	defer sociProcess.StopProcess()
+	sociContainerdProc := SociContainerdProcess{containerdProcess}
+	b.ResetTimer()
+	_, err = sociContainerdProc.SociRPullImageFromECR(ctx, imageRef, indexDigest, awsSecretFile)
+	if err != nil {
+		b.Fatalf("%s", err)
+	}
+	b.StopTimer()
+}
+
+func SociFullRun(
+	b *testing.B,
+	imageRef string,
+	indexDigest string,
+	readyLine string) {
+	ctx, cancelCtx := framework.GetTestContext()
+	containerdProcess, err := getContainerdProcess(ctx)
+	if err != nil {
+		b.Fatalf("Failed to create containerd proc: %v\n", err)
+	}
+	defer containerdProcess.StopProcess(cancelCtx)
+	sociProcess, err := getSociProcess()
+	if err != nil {
+		b.Fatalf("Failed to create soci proc: %v\n", err)
+	}
+	defer sociProcess.StopProcess()
+	sociContainerdProc := SociContainerdProcess{containerdProcess}
+	b.ResetTimer()
+	image, err := sociContainerdProc.SociRPullImageFromECR(ctx, imageRef, indexDigest, awsSecretFile)
+	if err != nil {
+		b.Fatalf("%s", err)
+	}
+	container, cleanupContainer, err := sociContainerdProc.CreateSociContainer(ctx, image)
+	if err != nil {
+		b.Fatalf("%s", err)
+	}
+	defer cleanupContainer()
+	taskDetails, cleanupTask, err := sociContainerdProc.CreateTask(ctx, container)
+	if err != nil {
+		b.Fatalf("%s", err)
+	}
+	defer cleanupTask()
+	cleanupRun, err := sociContainerdProc.RunContainerTaskForReadyLine(ctx, taskDetails, readyLine)
+	if err != nil {
+		b.Fatalf("%s", err)
+	}
+	defer cleanupRun()
+	b.StopTimer()
+}
+
+func OverlayFSFullRun(
+	b *testing.B,
+	imageRef string,
+	readyLine string) {
+	ctx, cancelCtx := framework.GetTestContext()
+	containerdProcess, err := getContainerdProcess(ctx)
+	if err != nil {
+		b.Fatalf("Failed to create containerd proc: %v\n", err)
+	}
+	defer containerdProcess.StopProcess(cancelCtx)
+	b.ResetTimer()
+	image, err := containerdProcess.PullImageFromECR(ctx, imageRef, platform, awsSecretFile)
+	if err != nil {
+		b.Fatalf("%s", err)
+	}
+	container, cleanupContainer, err := containerdProcess.CreateContainer(ctx, image)
+	if err != nil {
+		b.Fatalf("%s", err)
+	}
+	defer cleanupContainer()
+	taskDetails, cleanupTask, err := containerdProcess.CreateTask(ctx, container)
+	if err != nil {
+		b.Fatalf("%s", err)
+	}
+	defer cleanupTask()
+	cleanupRun, err := containerdProcess.RunContainerTaskForReadyLine(ctx, taskDetails, readyLine)
+	if err != nil {
+		b.Fatalf("%s", err)
+	}
+	defer cleanupRun()
+	b.StopTimer()
+}
+
+func getContainerdProcess(ctx context.Context) (*framework.ContainerdProcess, error) {
+	return framework.StartContainerd(
+		containerdAddress,
+		containerdRoot,
+		containerdState,
+		containerdConfig,
+		outputDir)
+}
+
+func getSociProcess() (*SociProcess, error) {
+	return StartSoci(
+		sociBinary,
+		sociAddress,
+		sociRoot,
+		containerdAddress,
+		sociConfig,
+		outputDir)
+}
--- a/benchmark/framework/containerd_utils.go
+++ b/benchmark/framework/containerd_utils.go
@@ -1,0 +1,305 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package framework
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/cio"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/namespaces"
+	"github.com/containerd/containerd/oci"
+)
+
+var (
+	testNamespace               = "BENCHMARK_TESTING"
+	testContainerID             = "TEST_RUN_CONTAINER"
+	testEnvironment             = "TEST_RUNTIME"
+	outputFilePerm  fs.FileMode = 0777
+)
+
+type ContainerdProcess struct {
+	command *exec.Cmd
+	address string
+	root    string
+	state   string
+	stdout  *os.File
+	stderr  *os.File
+	Client  *containerd.Client
+}
+
+func StartContainerd(
+	containerdAddress string,
+	containerdRoot string,
+	containerdState string,
+	containerdConfig string,
+	containerdOutput string) (*ContainerdProcess, error) {
+	containerdCmd := exec.Command("containerd",
+		"-a", containerdAddress,
+		"--root", containerdRoot,
+		"--state", containerdState,
+		"-c", containerdConfig)
+	err := os.MkdirAll(containerdOutput, outputFilePerm)
+	if err != nil {
+		return nil, err
+	}
+	stdoutFile, err := os.Create(containerdOutput + "/containerd-stdout")
+	if err != nil {
+		return nil, err
+	}
+	containerdCmd.Stdout = stdoutFile
+	stderrFile, err := os.Create(containerdOutput + "/containerd-stderr")
+	if err != nil {
+		return nil, err
+	}
+	containerdCmd.Stderr = stderrFile
+	err = containerdCmd.Start()
+	if err != nil {
+		return nil, err
+	}
+	client, err := newClient(containerdAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	return &ContainerdProcess{
+		command: containerdCmd,
+		address: containerdAddress,
+		root:    containerdRoot,
+		stdout:  stdoutFile,
+		stderr:  stderrFile,
+		state:   containerdState,
+		Client:  client}, nil
+}
+
+func (proc *ContainerdProcess) StopProcess(cancelContext context.CancelFunc) {
+	if proc.Client != nil {
+		proc.Client.Close()
+	}
+	if proc.stdout != nil {
+		proc.stdout.Close()
+	}
+	if proc.stderr != nil {
+		proc.stderr.Close()
+	}
+	if cancelContext != nil {
+		cancelContext()
+	}
+	if proc.command != nil {
+		proc.command.Process.Kill()
+	}
+	os.RemoveAll(proc.root)
+	os.RemoveAll(proc.state)
+	os.RemoveAll(proc.address)
+}
+
+func (proc *ContainerdProcess) PullImage(
+	ctx context.Context,
+	imageRef string,
+	platform string) (containerd.Image, error) {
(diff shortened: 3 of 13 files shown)


vkuzniet:
There are some linter errors.

vkuzniet:
Would be good to think of running the benchmarking tests inside the container. You can have one container as the registry, which will contain images and SOCI indices. And the tests will be run inside the other container. This will help in cleaning up the mounts, output files, etc. You can check the folder `integration` on the approach.

vkuzniet:
Does this PR mean that #92 is not needed anymore?

Makefile
```diff
@@ -129,3 +130,7 @@ integration:
 	@echo "$@"
 	@echo "SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT)"
 	@GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true go test $(GO_TEST_FLAGS) -v -timeout=0 ./integration
+
+benchmarks:
+	@echo "$@"
+	@cd benchmark/ ; GO111MODULE=$(GO111MODULE_VALUE) go build -o bin/benchmarkTests . && sudo ./bin/benchmarkTests $(COMMIT) testSuite.csv
```
vkuzniet, line 136: Maybe make the `csv` file name as a parameter?

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
+
+func BenchmarkPullImageFromECR(b *testing.B, imageRef string) {
+	containerdProcess, err := getContainerdProcess() 
+	if err != nil {
+                fmt.Printf("Error Starting Containerd: %v\n", err)
+		b.Fatal(err)
+	}
+	defer containerdProcess.StopProcess()
+	b.ResetTimer()
+	_, err = containerdProcess.PullImageFromECR(imageRef, platform, awsSecretFile)
```
vkuzniet, line 35: `PullImageFromRegistry`? Ideally, we'd like to use other registries, not only ECR.

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
+	defer containerdProcess.StopProcess()
+	sociProcess, err := getSociProcess()	
+        if err != nil {
+                fmt.Printf("Failed to create soci proc: %v\n", err)
+		b.Fatal(err)
+	}
+	defer sociProcess.StopProcess()
+	sociContainerdProc := SociContainerdProcess{containerdProcess}
+	b.ResetTimer()
+	_, err = sociContainerdProc.SociRPullImageFromECR(imageRef, indexDigest, awsSecretFile)
```
vkuzniet, line 61: `SociRpullImage` or `SociRpullImageFromRegistry`?

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
+	return StartSoci(
+		sociBinary,
+		sociAddress,
+		sociRoot,
+		containerdAddress,
+                sociConfig,
+		outputDir)
+            }
+
+func BeforeRunContainerFromECR(
```
vkuzniet, line 199: `BeforeRunContainerFromRegistry`?

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+        image, err := proc.PullImageFromECR(imageRef, platform, awsSecretFile)
+        *imagePtr = &image
+        if err != nil {
+                fmt.Printf("Error Pulling Image: %v\n", err)
+	}
+        return 
+}
+
+func BenchmarkRunContainerFromECR(
```
vkuzniet, line 216: `BenchmarkRunContainerFromRegistry`?

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
+	b.ResetTimer()
+        cleanupRun, err := containerdProcess.RunContainerTaskForReadyLine(taskDetails, readyLine)
+	if err != nil {
+		fmt.Println(err)
+		b.Fatal(err)
+	}
+        defer cleanupRun()
+        b.StopTimer()
+}
+func BenchmarkCreateContainerFromECR(
```
vkuzniet, line 242: `BenchmarkCreateContainerFromRegistry`?

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
+        _, cleanupContainer, err := containerdProcess.CreateContainer(*image)
+	if err != nil {
+		fmt.Println(err)
+		b.Fatal(err)
+	}
+	defer cleanupContainer()
+        b.StopTimer()
+}
+
+func BenchmarkCreateTaskFromECR(
```
vkuzniet, line 256: `BenchmarkCreateTaskFromRegistry`?

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
+        _, cleanupTask, err := containerdProcess.CreateTask(container)
+	if err != nil {
+		fmt.Println(err)
+		b.Fatal(err)
+	}
+	defer cleanupTask()
+        b.StopTimer()
+}
+
+func AfterRunContainerFromECR(containerdProcess *framework.ContainerdProcess) {
```
vkuzniet, line 276: `AfterRunContainerFromRegistry`

Makefile
sbuckfelder: it's an argument of binary, did you want something different?

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
+
+	"github.com/awslabs/soci-snapshotter/benchmark/framework"
+	"github.com/containerd/containerd"
+)
+
+func BenchmarkPullImageFromECR(b *testing.B, imageRef string) {
+	containerdProcess, err := getContainerdProcess() 
+	if err != nil {
+                fmt.Printf("Error Starting Containerd: %v\n", err)
+		b.Fatal(err)
```
sparr, line 31: suggest `b.Fatalf` instead of `fmt.Printf` + `b.Fatal`, here and everywhere else

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	defer containerdProcess.StopProcess()
+	b.ResetTimer()
+	_, err = containerdProcess.PullImageFromECR(imageRef, platform, awsSecretFile)
+	if err != nil {
+                fmt.Printf("Error Pulling Image: %v\n", err)
+		b.Fatal(err)
+	}
+	b.StopTimer()
```
sparr, line 40: Is this necessary at the end of a test?

benchmark/benchmarkTests.go
sbuckfelder: This specifically uses the resolver necessary for ECR, in containerd_utils there is a function that does not use, but any registry that does not require login
```
func (proc *ContainerdProcess) PullImage(
	imageRef string,
	platform string) (containerd.Image, error)
```

benchmark/benchmarkTests.go
sbuckfelder: This is focused specifically on ECR, the alternative could be to create a more generic function that uses credentials as opposed to this ECR helper func

benchmark/benchmarkTests.go
sbuckfelder: same as above

benchmark/benchmarkTests.go
sbuckfelder: same as above

benchmark/benchmarkTests.go
sbuckfelder: same as above

benchmark/benchmarkTests.go
sbuckfelder: same as above

benchmark/benchmarkTests.go
sbuckfelder: :+1:

benchmark/benchmarkTests.go
sbuckfelder: I don't think it is necessary, but I like including it because of defers

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	defer containerdProcess.StopProcess()
+	sociProcess, err := getSociProcess()	
+        if err != nil {
+                fmt.Printf("Failed to create soci proc: %v\n", err)
+		b.Fatal(err)
+	}
+	defer sociProcess.StopProcess()
+	sociContainerdProc := SociContainerdProcess{containerdProcess}
+	b.ResetTimer()
```
sparr, line 60: function contents above here are duplicated many places. defer() prevents straightforward refactor into a reusable function. is there another way to avoid this duplication?

benchmark/framework/containerd_utils.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/containerd/containerd/cio"
+	"github.com/containerd/containerd/namespaces"
+	"github.com/containerd/containerd/oci"
+)
+
+var (
+	testNamespace               = "BENCHMARK_TESTING"
+	testContainerId             = "TEST_RUN_CONTAINER"
+	testEnvironment             = "TEST_RUNTIME"
+	outputFilePerm  fs.FileMode = 0777
```
sbuckfelder, line 41: 0666 would be more accurate as these are not executable

benchmark/soci_utils.go
```diff
@@ -0,0 +1,159 @@
...
+	"syscall"
+	"time"
+
+	"github.com/awslabs/soci-snapshotter/benchmark/framework"
+	"github.com/awslabs/soci-snapshotter/fs/source"
+	"github.com/containerd/containerd"
+)
+
+var (
+	outputFilePerm  fs.FileMode = 0777
```
sbuckfelder, line 35: 0666 would be more accurate as these are not executable

benchmark/framework/containerd_utils.go
```diff
@@ -0,0 +1,304 @@
...
+				resultChannel <- "READYLINE_STDERR"
+				continueLoop = false
+			}
+			select {
+			case _ = <-resultChannel:
+				continueLoop = false
+			default:
+				continueLoop = true
+			}
+		}
```
sparr, line 229: I am new to goroutines, so there may be no better way to do this and I wouldn't know. That said, this loop of polling *feels* like unnecessary cpu churn that I would expect to be able to do with a higher level construct of some sort.

benchmark/framework/containerd_utils.go
```diff
@@ -0,0 +1,304 @@
...
+			_ = <-exitChannel
+		}
+	}
+	return cleanupFunc, nil
+}
+
+func GetRemoteOpts(ctx context.Context, platform string) []containerd.RemoteOpt {
+	var opts []containerd.RemoteOpt
+	opts = append(opts, containerd.WithPlatform(platform))
+	opts = append(opts, containerd.WithPullUnpack)
```
sparr, line 286: can replace the two assignments with one `append(opts, containerd.WithPlatform(platform), containerd.WithPullUnpack)`
or just replace the whole thing with
`opts := [2]containerd.RemoteOpt{containerd.WithPlatform(platform), containerd.WithPullUnpack}`

benchmark/soci_utils.go
```diff
@@ -0,0 +1,159 @@
...
+	if proc.stdout != nil {
+		proc.stdout.Close()
+	}
+	if proc.stderr != nil {
+		proc.stderr.Close()
+	}
+	if proc.command != nil {
+		proc.command.Process.Kill()
+	}
+	err := os.RemoveAll(proc.address)
```
sparr, line 115: Some duplication with ContainerdProcess StopProcess, can this be deduplicated?

benchmark/framework/framework.go
```diff
@@ -0,0 +1,137 @@
...
+        "io/fs"
+	"os"
+	"testing"
+
+	"github.com/montanaflynn/stats"
+)
+
+var (
+    resultFilename = "results.json"
+    resultFilePerm fs.FileMode = 0644
```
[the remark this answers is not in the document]
sparr: Mis-copy-paste on this one?

vkuzniet:
Since `make benchmarks` code path depends on ECR, it cannot be widely used outside of AWS. It can be this PR, or probably another one, but `make benchmarks` should just be registry-agnostic.

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
```
Kern--, line 86: Discussed offline, but highlighting that the position of this reset means that we're measuring the rpull time + container create time, not just container create.

benchmark/benchmarkTests.go
```diff
@@ -0,0 +1,338 @@
...
+		fmt.Println(err)
+		b.Fatal(err)
+	}
+        container, cleanupContainer, err := sociContainerdProc.CreateSociContainer(image)
+	if err != nil {
+		fmt.Println(err)
+		b.Fatal(err)
+	}
+	defer cleanupContainer()
+	b.ResetTimer()
```
Kern--, line 164: nit: `BenchmarkSociRunForReadyLine` includes task create + run time, but `BenchmarkRunContainerFromECR` only includes run time.

vkuzniet:
When we execute in Soci mode, does the framework error out in case if there's a failover to containerd? I suppose it does not, but it would really be good to have, so we understand that the results for Soci mode are the ones, which involve the actual lazy loading.

sbuckfelder:
> When we execute in Soci mode, does the framework error out in case if there's a failover to containerd? I suppose it does not, but it would really be good to have, so we understand that the results for Soci mode are the ones, which involve the actual lazy loading.

Are you comfortable with adding this as an issue and fixing forward?

vkuzniet:
> > When we execute in Soci mode, does the framework error out in case if there's a failover to containerd? I suppose it does not, but it would really be good to have, so we understand that the results for Soci mode are the ones, which involve the actual lazy loading.
> 
> Are you comfortable with adding this as an issue and fixing forward?

Yes, I think that makes sense.

Makefile
```diff
@@ -129,3 +130,11 @@ integration:
 	@echo "$@"
 	@echo "SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT)"
 	@GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true go test $(GO_TEST_FLAGS) -v -timeout=0 ./integration
+
+benchmarks:
+	@echo "$@"
+	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests . && sudo ../bin/PerfTests $(COMMIT) ../singleImage.csv 10
```
vkuzniet, line 136: nit: `singleImage.csv` does not seem to reflect the fact that it can have multiple images inside

benchmark/framework/containerd_utils.go
```diff
@@ -0,0 +1,312 @@
...
+	task, err := container.NewTask(proc.Context, cioCreator)
+	if err != nil {
+		return nil, nil, err
+	}
+	cleanupFunc := func() {
+		stdoutPipeReader.Close()
+		stdoutPipeWriter.Close()
+		stderrPipeReader.Close()
+		stderrPipeWriter.Close()
+		processStatus, _ := task.Status(proc.Context)
```
kzys, line 186: Can you log this error?

benchmark/framework/framework.go
```diff
@@ -0,0 +1,137 @@
...
+	err = os.MkdirAll(frame.OutputDir, resultFilePerm)
+	if err != nil {
+		fmt.Printf("Failed to Create Output Dir: %v\n", err)
+	}
+	resultFileLoc := frame.OutputDir + "/" + resultFilename
+	err = os.WriteFile(resultFileLoc, json, resultFilePerm)
+	if err != nil {
+		fmt.Printf("WriteFile Error: %v\n", err)
+	}
+}
```
kzys, line 93: It may be easier to return the error and let the caller log the error. You wouldn't be able to continue the execution if you have an error from one of these.

sbuckfelder:
> > > When we execute in Soci mode, does the framework error out in case if there's a failover to containerd? I suppose it does not, but it would really be good to have, so we understand that the results for Soci mode are the ones, which involve the actual lazy loading.
> > 
> > 
> > Are you comfortable with adding this as an issue and fixing forward?
> 
> Yes, I think that makes sense.

Put some thought into this and unfortunately I think the best way to handle this would be for us to pass a setting to the soci snapshotter to fail as opposed to fallback to default.  Then for benchmarking we can use this setting,  but in production we would not.

Makefile
sbuckfelder: it's a filename, you can change it to anything you want

benchmark/framework/containerd_utils.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+	<-resultChannel
+	cleanupFunc := func() {
+		processStatus, _ := taskDetails.task.Status(proc.Context)
+		if processStatus.Status == "running" {
+			err = taskDetails.task.Kill(proc.Context, syscall.SIGKILL)
+			if err != nil {
+				fmt.Printf("Error killing task: %v\n", err)
+			}
+			exitChannel, _ := taskDetails.task.Wait(proc.Context)
```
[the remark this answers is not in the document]
sbuckfelder: If I recall I added this because I needed to wait for the process to successfully move out of the running state.  Without the wait we get these errors:
```
Tried to kill taskError deleting task: task must be stopped before deletion: running: failed precondition
Error deleting container: cannot delete running task TEST_RUN_CONTAINER-1668084161373979375: failed precondition
```

The wait ensures that the task is successfully killed before moving forward

benchmark/framework/containerd_utils.go
```diff
@@ -0,0 +1,312 @@
...
+
+type ContainerdProcess struct {
+	command       *exec.Cmd
+	address       string
+	root          string
+	state         string
+	stdout        *os.File
+	stderr        *os.File
+	Client        *containerd.Client
+	Context       context.Context
```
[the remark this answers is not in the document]
sbuckfelder: No we can pass it everywhere, I was just trying to reduce the parameters by 1 for every func.  But looks like that is not idiomatic golang.  So I'll remove it.

benchmark/framework/framework.go
sbuckfelder: I'll look into fixing this forward.  After using this for a bit I'm going to look at updating how the output is handeled.

benchmark/framework/containerd_utils.go
```diff
@@ -0,0 +1,312 @@
...
+	readyLine string) (func(), error) {
+	stdoutScanner := bufio.NewScanner(taskDetails.stdoutReader)
+	stderrScanner := bufio.NewScanner(taskDetails.stderrReader)
+
+	exitStatusC, err := taskDetails.task.Wait(proc.Context)
+	if err != nil {
+		return nil, err
+	}
+	resultChannel := make(chan string, 1)
+	go func() {
```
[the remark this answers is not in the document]
sbuckfelder: Nice - thank you for this!  Included this pattern in the new rev

benchmark/framework/containerd_utils.go
sbuckfelder: updated in new rev

vkuzniet:
> > > > When we execute in Soci mode, does the framework error out in case if there's a failover to containerd? I suppose it does not, but it would really be good to have, so we understand that the results for Soci mode are the ones, which involve the actual lazy loading.
> > > 
> > > 
> > > Are you comfortable with adding this as an issue and fixing forward?
> > 
> > 
> > Yes, I think that makes sense.
> 
> Put some thought into this and unfortunately I think the best way to handle this would be for us to pass a setting to the soci snapshotter to fail as opposed to fallback to default. Then for benchmarking we can use this setting, but in production we would not.

You should be able to identify if the container is not run in soci mode if you are not having mounts for the layers you expect. This is relatively straightforward to infer by accessing the logs. This will not require implementing that setting. Can you please file an issue for that?

benchmark/framework/containerd_utils.go
kzys: Bit confused about who would work on that :)

Also I'm fine just hardcoding these numbers on os.Mkdir() and/or os.Create().

benchmark/framework/containerd_utils.go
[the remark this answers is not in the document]
kzys: Waiting the new rev :)

benchmark/framework/framework.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (driver *BenchmarkTestDriver) calculateStats() {
+	var err error
+	driver.StdDev, err = stats.StandardDeviation(driver.TestTimes)
+	if err != nil {
+		fmt.Printf("Error Calculating Std Dev: %v\n", err)
+                driver.StdDev = -1
+	}
```
kzys, line 101: Does it happen a lot? If not, how about just returning the error instead of setting -1?

benchmark/soci_utils.go
```diff
@@ -0,0 +1,163 @@
...
+	snapshotDir := proc.root + "/snapshotter/snapshots/"
+	snapshots, err := ioutil.ReadDir(snapshotDir)
+	if err != nil {
+		fmt.Printf("Could not read dir: %s\n", snapshotDir)
+	}
+
+	for _, s := range snapshots {
+		mountpoint := snapshotDir + s.Name() + "/fs"
+		_ = syscall.Unmount(mountpoint, syscall.MNT_FORCE)
+	}
```
kzys, line 131: Can't we remove snapshots via containerd's API instead of directly unmounting them here? The framework shouldn't know SOCI snapshotter this much.

sbuckfelder: I don't believe that currently works.  In general, soci-snapshotter needs to put some work in managing it's data.  Right now we do not have any APIs to remove anything.

benchmark/framework/framework.go
sbuckfelder: This can happen if there are too few samples, so I'd rather just have the negative number.  It's useful when running just one or two runs to quickly test that it works.  We could printout the error in the std dev field, but that gets a bit tricky as it expects a number right now.  But something I would be interested in doing fixing forward.

sbuckfelder:
> > > > > When we execute in Soci mode, does the framework error out in case if there's a failover to containerd? I suppose it does not, but it would really be good to have, so we understand that the results for Soci mode are the ones, which involve the actual lazy loading.
> > > > 
> > > > 
> > > > Are you comfortable with adding this as an issue and fixing forward?
> > > 
> > > 
> > > Yes, I think that makes sense.
> > 
> > 
> > Put some thought into this and unfortunately I think the best way to handle this would be for us to pass a setting to the soci snapshotter to fail as opposed to fallback to default. Then for benchmarking we can use this setting, but in production we would not.
> 
> You should be able to identify if the container is not run in soci mode if you are not having mounts for the layers you expect. This is relatively straightforward to infer by accessing the logs. This will not require implementing that setting. Can you please file an issue for that?

https://github.com/awslabs/soci-snapshotter/issues/190

benchmark/benchmarkTests.go
kzys: - You could tie the cancellation of the context to destruction of the resources.
- Or https://pkg.go.dev/testing#B.Cleanup

benchmark/framework/framework.go
kzys: I see. I'm fine keeping that as is in this PR then.

benchmark/soci_utils.go
kzys: That's sad to know :(

kzys (approved):
I'm fine merging this one as-is. We are anyway doing some performance tuning work, and we have to have a benchmark suite in the tree.

vkuzniet:
> I'm fine merging this one as-is. We are anyway doing some performance tuning work, and we have to have a benchmark suite in the tree.

Totally agree. Let's ship and incrementally improve.

vkuzniet (dismissed):

vkuzniet (approved):

vkuzniet (changes requested):
can you please make sure that linter passes? looks like it is having some questions about `containerd_utils.go`.

sbuckfelder:
> can you please make sure that linter passes? looks like it is having some questions about `containerd_utils.go`.

Updated and passes

vkuzniet (approved):

hanyuel (approved):

PR #120, opened 2022-10-20
Remove dependency on zinfo C library

vkuzniet:

*Issue #, if available:*
#78 

*Description of changes:*
Removed linking to zinfo.a. Also moved the C files to soci.
gzip_zinfo.go also gets the unmarshaling of gzip_zinfo C struct, which is used in ztoc tests.

*Testing performed:*


soci/gzip_zinfo.go
```diff
@@ -28,6 +28,24 @@ import (
...
+	Out    int64  /* corresponding offset in uncompressed data */
+	In     int64  /* offset in input file of first full byte */
+	Bits   int8   /* number of bits (1-7) from byte at in - 1, or 0 */
+	Window []byte /* preceding 32K of uncompressed data */
+}
+
+// GzIndex corresponds to gzip_index in indexer.c
+type CZinfo struct {
+	Have     int64            /* number of list entries filled in */
+	Size     int64            /* number of list entries allocated */
```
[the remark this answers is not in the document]
vkuzniet: Yes, because we want to precisely check the contents of gzip_checkpoint. This is used in tests only.

soci/gzip_zinfo.go
```diff
@@ -139,3 +157,26 @@ func (i *GzipZinfo) ExtractData(fileName string, uncompressedSize, uncompressedO
 
 	return bytes, nil
 }
+
+func (i *GzipZinfo) UnmarshalCZinfo() (*CZinfo, error) {
```
rdpsin, line 161: Suggest renaming this to just `Unmarshal`.

soci/gzip_zinfo.go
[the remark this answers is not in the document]
rdpsin: We can technically remove one of these, since `have` will always equal `size`.

soci/gzip_zinfo.go
```diff
@@ -28,6 +28,24 @@ import (
...
+// GzipIndexPoint corresponds to gzip_index_point in indexer.c
+type GzipCheckPoint struct {
+	Out    int64  /* corresponding offset in uncompressed data */
+	In     int64  /* offset in input file of first full byte */
+	Bits   int8   /* number of bits (1-7) from byte at in - 1, or 0 */
+	Window []byte /* preceding 32K of uncompressed data */
+}
+
+// GzIndex corresponds to gzip_index in indexer.c
+type CZinfo struct {
```
rdpsin, line 42: What does the `C` stand for here?

rdpsin:
Does this change mean it is now possible to use the `soci` package as a separate module?

rdpsin (changes requested):

soci/gzip_zinfo.go
vkuzniet: This stands for go representation of C struct.

soci/gzip_zinfo.go
vkuzniet: Agree.

soci/gzip_zinfo.go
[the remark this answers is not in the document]
vkuzniet: While we have the structure defined in zinfo.h in precisely that way, I would still prefer keeping as is.

soci/gzip_zinfo.go
rdpsin: Would just keep it as `ZInfo`. `CZinfo` is just bound to confuse people.

rdpsin (changes requested):

soci/gzip_zinfo.go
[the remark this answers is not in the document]
wmesard: Can you explain the rationale for your preference? Generally speaking, test-only methods and data structures are a code smell.

soci/gzip_zinfo.go
wmesard: I think I get the blame for [coining](https://github.com/awslabs/soci-snapshotter/blob/main/soci/gzip_zinfo.go#L32) the term. But, I agree that making it public (capital "C") is leaking an implementation detail.

wmesard (changes requested):

soci/gzip_zinfo.go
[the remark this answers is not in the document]
vkuzniet: This code has no use in production. The only reason for it to exist is to be able to do `deep equal` of the gzip's Checkpoints collection. It may make more sense to move it back to testutils. 

Yes, I agree that this is a code smell, but the alternative is to directly access C struct in unit tests and do those comparisons, which doesn't seem to be worth it.

soci/gzip_zinfo.go
vkuzniet: > Does this change mean it is now possible to use the `soci` package as a separate module?

Yes.

soci/gzip_zinfo.go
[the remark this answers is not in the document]
vkuzniet: Removed the unmarshal and replaced it with returning the index of the first different byte in sequence.

soci/gzip_zinfo.go
vkuzniet: Removed the test-only code.

soci/ztoc_test.go
```diff
@@ -236,53 +233,8 @@ func TestZtocGenerationConsistency(t *testing.T) {
...
-
-					if indexPoint1.out != indexPoint2.out {
-						t.Fatalf("index1.list[%d].out=%d must be equal to index2.list[%d].out=%d", i, index1.list[i].out, i, index2.list[i].out)
-					}
-
-					if !reflect.DeepEqual(indexPoint1.window, indexPoint2.window) {
-						t.Fatalf("index1.list[%d].window must be identical to index2.list[%d].window", i, i)
-					}
-				}
+				diffIdx := getPositionOfFirstDiffInByteSlice(ztoc1.CompressionInfo.Checkpoints, ztoc2.CompressionInfo.Checkpoints)
```
rdpsin, line 236: Can just use `cmp.Diff` here.

vkuzniet: `cmp.Diff` can produce a lot of noise for checkpoints, since it produces the string difference of two slices in our case (https://github.com/google/go-cmp/blob/v0.5.9/cmp/compare.go#L113). The best would be to just return the index of the first difference and let investigate from here.

rdpsin: Since this is only run in tests, I don't see any problem with the diff output, but that's fine I guess.

rdpsin (approved):

vkuzniet:
> Does this change mean it is now possible to use the `soci` package as a separate module?

Yes, since `soci` is now self-contained. It will be done as a separate PR.

wmesard (approved):

PR #126, opened 2022-10-25
Reduce permissions for the root directory

vkuzniet:

*Issue #, if available:*

*Description of changes:*
Setting the limited permission set for snapshotter's root folder (`/var/lib/soci-snapshotter`).

*Testing performed:*
- deployed on EC2 instance, created an index and verified that the permission set applies:
```
$ sudo ls -l /var/lib
total 0
...
drwx--x--x 10 root    root    307 Oct 25 20:30 containerd
drwx--x--x  3 root    root     41 Oct 25 20:30 soci-snapshotter-grpc
...
```

### DIFF
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -17,6 +17,8 @@
 package commands
 
 import (
+	"os"
+
 	"github.com/awslabs/soci-snapshotter/fs/config"
 	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/containerd/containerd/cmd/ctr/commands"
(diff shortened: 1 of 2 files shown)


cmd/soci/commands/create.go
```diff
@@ -79,6 +81,14 @@ var CreateCommand = cli.Command{
 		}
 		spanSize := cliContext.Int64(spanSizeFlag)
 		minLayerSize := cliContext.Int64(minLayerSizeFlag)
+		// Creating the snapshotter's root path first if it does not exist, since this ensures, that
+		// it has the limited permission set as drwx--x--x.
+		// The subsequent oci.New creates a root path dir with too broad permission set.
+		if _, err := os.Stat(config.SociSnapshotterRootPath); os.IsNotExist(err) {
```
rdpsin, line 87: Do we need to check for when `err != nil && !os.IsNotExist(err)`?

vkuzniet: I don't think so. It examines the error for type `os.ErrNotExist`. The null value, will force `os.IsNotExist` to return false, which is what we need.

rdpsin: I meant, do we have to check for any errors from `os.Stat` besides `os.ErrNotExist`?

vkuzniet: Yeah, it does make sense. Updated.

cmd/soci/commands/create.go
```diff
@@ -79,6 +81,16 @@ var CreateCommand = cli.Command{
...
+			if err = os.Mkdir(config.SociSnapshotterRootPath, 0711); err != nil {
+				return err
+			}
+		} else if err != nil {
```
Kern--, line 91: Interesting. I didn't know that variable assignments in an if statement are still bound in else statements. I think this would still be clearer if we assigned the error variable in a separate line, though.

Kern-- (approved):

rdpsin (approved):

Issue #54, opened 2022-09-07
SOCI CLI integration test expansion


Kern--:
**Is your feature request related to a problem? Please describe.**
The SOCI CLI is tested for integration around the core user experience of creating/pushing/pulling SOCI indices, but it is not tested around the debug tooling for indices and ztocs. We should add integration tests to verify these debugging tools.

**Describe the solution you'd like**
The following integration tests:
- [x] `index list` #250
- [x] `index list` with image ref #250
- [x] `index info` #250
- [x] `ztoc list` #265
- [x] `ztoc list` with digest #265
- [x] `ztoc info` #334 
- [x] `ztoc get-file` #334 
- [x] `index rm` with digest #260 
- [x] `index rm` with image ref #260
- [x] `index rm` everything #260
- [ ] `create` with all platforms
- [ ] `push` with all platforms
- [ ] `image rpull` with all platforms

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

rdpsin:
Additionally, 

- `index rm`
- `index rm` with image ref
-  `index rm` everything.

djdongjin:
@rdpsin @Kern-- do we still miss tests for this three in the list

- [ ]  `create` with all platforms
- [ ] `push` with all platforms
- [ ] `image rpull` with all platforms

for 1, we have https://github.com/awslabs/soci-snapshotter/blob/main/integration/create_test.go#L168

for 2 and 3, we have https://github.com/awslabs/soci-snapshotter/blob/main/integration/push_test.go#L74-L86

not sure if they cover the `all platforms`; if not, any other platforms we want to add?

djdongjin:
Closing this one since all tests are added/exists.

We have multi platform tests linked in the comments above: https://github.com/awslabs/soci-snapshotter/issues/54#issuecomment-1362948399

@Kern-- lmk if there's something not covered.

PR #127, opened 2022-10-26
Allow multi-platform/cross-platform index build

Kern--:
*Issue #, if available:*
N/A

*Description of changes:*
This change allows users to interact with SOCI indices from multiple platforms or in a cross platform context (e.g. creating an index for ARM64 on an x86_64 machine).

CLI Changes:
* `soci create` can now take `--platform` or `--all-platforms`. By default, it will create an index for the default platform.
* `soci push` can now take `--platform` or `--all-platforms`. By default, it will push an index for the default platform.
* `soci index list` can now take `--platform`. By default, it will list all platforms.

the --platform flag uses strict matching. E.g. specifying linux/amd64 will not create an index for linux/386.

*Testing performed:*

### DIFF
--- a/cmd/soci/commands/index/list.go
+++ b/cmd/soci/commands/index/list.go
@@ -24,17 +24,38 @@ import (
 
 	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/containerd/containerd/cmd/ctr/commands"
+	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/platforms"
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/urfave/cli"
 )
+
+type filter func(ae *soci.ArtifactEntry) bool
 
 func indexFilter(ae *soci.ArtifactEntry) bool {
 	return ae.Type == soci.ArtifactEntryTypeIndex
 }
 
-func originalDigestFilter(digest string) func(*soci.ArtifactEntry) bool {
+func platformFilter(platform specs.Platform) filter {
 	return func(ae *soci.ArtifactEntry) bool {
-		return ae.Type == soci.ArtifactEntryTypeIndex && ae.OriginalDigest == digest
+		return indexFilter(ae) && ae.Platform == platforms.Format(platform)
+	}
+}
+
+func originalDigestFilter(digest string) filter {
+	return func(ae *soci.ArtifactEntry) bool {
+		return indexFilter(ae) && ae.OriginalDigest == digest
+	}
+}
+
+func anyMatch(fns []filter) filter {
+	return func(ae *soci.ArtifactEntry) bool {
+		for _, f := range fns {
+			if f(ae) {
+				return true
+			}
+		}
+		return false
 	}
 }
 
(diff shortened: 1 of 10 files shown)


cmd/soci/commands/index/list.go
```diff
@@ -24,17 +24,38 @@ import (
...
+	}
+}
+
+func originalDigestFilter(digest string) filter {
+	return func(ae *soci.ArtifactEntry) bool {
+		return indexFilter(ae) && ae.OriginalDigest == digest
+	}
+}
+
+func any(fns []filter) filter {
```
rdpsin, line 51: Maybe change this name? `any` isn't very descriptive. Also, conflicts with the `any` type.

cmd/soci/commands/push.go
```diff
@@ -45,12 +46,17 @@ If multiple soci indices exist for the given image, the most recent one will be
 After pushing the soci artifacts, they should be available in the registry. Soci artifacts will be pushed only
 if they are available in the snapshotter's local content store.
 `,
-	Flags: append(append(append(commands.RegistryFlags, commands.LabelFlag), commands.SnapshotterFlags...),
+	Flags: append(append(append(append(
```
vkuzniet, line 49: Can you please add an integration test for `push`, involving multiple platforms?

integration/create_test.go
```diff
@@ -156,12 +158,33 @@ func TestSociCreate(t *testing.T) {
 			name:           "test create for drupal:latest",
 			containerImage: "drupal:latest",
 		},
+		// The following two tests guarantee that we have tested at least
```
vkuzniet, line 161: How about making 3 test cases?
1. Create for "linux/arm64"
2. Create for "linux/amd64"
3. Create for all

rdpsin (approved):

vkuzniet (approved):

rdpsin (approved):

rdpsin (approved):

cmd/soci/commands/internal/platform.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/platforms"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/urfave/cli"
+)
+
+const (
+	PlatformFlagKey     = "platform"
+	AllPlatformsFlagKey = "all-platforms"
```
vkuzniet, line 32: would be good to have tests with create/push/pull functionality, which cover "all-platforms" flag

sbuckfelder: should we do this in a future PR and move forward?

Kern--: I'd like for this to be a fast follow. There are a number of helper methods that assume there will be exactly 1 index so I think refactoring those to allow for the all-platforms flag will propagate across the tests and it would be clearer in a separate PR.

integration/create_test.go
Kern--: See below about the all platforms tests

cmd/soci/commands/internal/platform.go
vkuzniet: I'm fine with that. Can you please create a ticket on that?

vkuzniet (approved):

cmd/soci/commands/internal/platform.go
Kern--: Updated https://github.com/awslabs/soci-snapshotter/issues/54

PR #132, opened 2022-10-27
Handle uncompressed layers nicely

kzys:
For uncompressed layers, images.DiffCompression() returns "unknown" but buildSociLayer() was only telling it wasn't gzip.

*Issue #, if available:*

*Description of changes:*

*Testing performed:*

### DIFF
--- a/soci/soci_index_test.go
+++ b/soci/soci_index_test.go
@@ -20,8 +20,10 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"testing"
 
+	"github.com/containerd/containerd/images"
 	"github.com/google/go-cmp/cmp"
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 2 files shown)


rdpsin (dismissed):

soci/soci_index_test.go
```diff
@@ -352,3 +353,15 @@ func TestNewIndexFromReader(t *testing.T) {
...
 }
+
+func TestBuildSociLayer(t *testing.T) {
+	ctx := context.Background()
+	desc := ocispec.Descriptor{
+		MediaType: images.MediaTypeDockerSchema2Layer,
+		Digest:    "layerdigest",
+	}
+	_, err := buildSociLayer(ctx, nil, desc, 0, nil, nil)
+	if err.Error() != `layer layerdigest (application/vnd.docker.image.rootfs.diff.tar) must be compressed by gzip, but got "unknown"` {
```
wmesard, line 364: This is unnecessarily brittle. Tests should protect functionality, not enshrine current behavior. When we add support for Zstandard layers, this error message will change, and the test--as currently written--will break. (I'll admit that my real complaint here is philosophical, not technical. If developers get in the habit of tweaking unit tests, they learn to ignore test failures even when a failure is trying to tell them about a real problem.)

**TL;DR:** Suggest testing only for the behavior that you want to protect. Something like:
```
        if !strings.Contains(err.Error(), "layerdigest") || !strings.Contains(err.Error(), "application/vnd.docker.image.rootfs.diff.tar") ||
                !strings.Contains(err.Error(), "compress") {
                t.Fatalf("unexpected error: %v", err)

```

soci/soci_index_test.go
```diff
@@ -352,3 +353,15 @@ func TestNewIndexFromReader(t *testing.T) {
 		})
 	}
 }
+
+func TestBuildSociLayer(t *testing.T) {
```
vkuzniet, line 357: Since this test tests for one of the error modes in `buildSociLayer`, maybe it may make sense to name it respectively, e.g. `TestBuildSociLayerErrorModes`, or something like that?

soci/soci_index_test.go
kzys: Sounds good. How about returning ErrFailedPrecondition from https://pkg.go.dev/github.com/containerd/containerd/errdefs in addition to?

kzys: Or define something like errUnsupportedCompression

rdpsin: Would prefer the latter.

rdpsin: Something like `errUnsupportedLayerFormat`.

rdpsin (approved):

vkuzniet (approved):

wmesard (approved):

PR #133, opened 2022-10-27
Run integration tests in a separate runner

kzys:
Running "make" and "make test" in the same runner may make sense since they could share Go modules.

"make integration" is much slower than others. So the benefit of sharing Go modules is marginal.

*Issue #, if available:*

*Description of changes:*

*Testing performed:*


.github/workflows/build.yml
```diff
@@ -30,4 +30,9 @@ jobs:
       - uses: actions/checkout@v3
       - run: make
       - run: make test
+  integration:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - run: make
```
rdpsin, line 37: Instead of adding `make`, maybe we can add the `pre-build` target to `make integration` in the Makefile. That'll just create the C library, and won't spend time building the binaries.

kzys: Thanks will do.

rdpsin (approved):

vkuzniet (approved):

kzys:
Can someone re-run the integration test?

rdpsin (approved):

PR #135, opened 2022-10-28
Remove build-tool-version annotation from the SOCI index manifest

wmesard:
*Description of changes:*

Remove build-tool-version annotation from the SOCI index manifest

The old zTOC format contained a separate field for the build tool version. This was deliberately removed because we want people testing against the zTOC *file format version*, not the build tool version. This PR removes the build tool version from the SOCI index manifest as well (and for the same reason). In both cases, there is still a "build tool identifier string". This is strictly informational. It is not (and should not) be parsed.

*Testing performed:*

make && make test && make check


cmd/soci/commands/create.go
```diff
@@ -30,7 +30,6 @@ import (
 
 const (
 	buildToolIdentifier = "AWS SOCI CLI"
```
kzys, line 32: If you have invalid zTOC, wouldn't you want to track down the version of this CLI?

rdpsin: Should we add the version to the identifier string?

vkuzniet: Since we have fbs representation for ztoc public, anyone can implement their own tooling to generate ztoc, so it will be hard to do so.

vkuzniet (approved):

sbuckfelder (approved):

cmd/soci/commands/create.go
wmesard: Yes, we should add the version to the string.

wmesard:
oops

Issue #128, opened 2022-10-26
[FEATURE] soci index builder as a library


dvnguyen-amzn:
**Is your feature request related to a problem? Please describe.**
Currently the index builder code is tightly coupled with Containerd. This blocks us on using the index builder module in environments without having containerd daemon running.

**Describe the solution you'd like**
We hope to move the index builder code into a separate module and the module should work without requiring on the present of the Containerd daemon. The new index builder module then can be imported by the soci-cli or other projects.

**Describe alternatives you've considered**
We need to have Containerd daemon running if we want to build soci index files.

Kern--:
This is complete.

PR #158, opened 2022-11-03
Refactor SOCI index builder code into a separate module.

dvnguyen-amzn:
*Issue #, if available:* #128

*Description of changes:*
Refactor SOCI index builder code into a separate module.
Motivation: to allow reusing the index builder code in other projects (issue #128).

This PR contains minimal change to existing soci_index.go. If this PR is accepted, we'll have another follow up PR that replaces existing index builder code with the newly created module.

*Testing performed:*
- running `make && make test && make check`
- writing an example code and use the module locally to build soci index for a pulled image

### DIFF
--- a/soci/index_builder.go
+++ b/soci/index_builder.go
@@ -1,0 +1,198 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package soci
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/platforms"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	DefaultSpanSize            = int64(1 << 22)
+	MinLayerSize               = 0
+	DefaultBuildToolIdentifier = "AWS SOCI CLI v0.1"
+)
+
+type IndexBuilder struct {
+	contentStore        content.Store
+	spanSize            int64
+	minLayerSize        int64
+	buildToolIdentifier string
+}
+
+type IndexBuilderOption func(*IndexBuilder) error
+
+func NewIndexBuilder(contentStore content.Store, opts ...IndexBuilderOption) (*IndexBuilder, error) {
+	b := &IndexBuilder{
+		contentStore:        contentStore,
+		spanSize:            DefaultSpanSize,
+		minLayerSize:        MinLayerSize,
+		buildToolIdentifier: DefaultBuildToolIdentifier,
+	}
+
+	for _, opt := range opts {
+		if err := opt(b); err != nil {
+			return nil, err
+		}
+	}
+
+	return b, nil
+}
+
+func WithSpanSizeOption(spanSize int64) IndexBuilderOption {
+	return func(b *IndexBuilder) error {
+		b.spanSize = spanSize
+		return nil
+	}
+}
+
+func WithMinLayerSizeOption(minLayerSize int64) IndexBuilderOption {
+	return func(b *IndexBuilder) error {
+		b.minLayerSize = minLayerSize
+		return nil
+	}
+}
+
+func WithBuildToolIdentifierOption(buildToolIdentifier string) IndexBuilderOption {
+	return func(b *IndexBuilder) error {
+		b.buildToolIdentifier = buildToolIdentifier
+		return nil
+	}
+}
+
+// BuildIndex builds SOCI index for an image
+// This returns SOCI index manifest and an array of io.Reader of zTOCs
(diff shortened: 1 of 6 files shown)


soci/index_builder.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/platforms"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+	"golang.org/x/sync/errgroup"
+)
+
+type IndexBuilder struct {
+	contentStore        content.Store
```
rdpsin, line 18: Do we want to use `content.Store` here? As @Kern-- mentioned, `content.Store` might be too large of an interface to use here. We can instead use ORAS's [content.Storage](https://pkg.go.dev/oras.land/oras-go/v2@v2.0.0-rc.4/content#Storage) or define our own.

soci/index_builder.go
```diff
@@ -0,0 +1,161 @@
...
+	spanSize            int64
+	minLayerSize        int64
+	buildToolIdentifier string
+}
+
+func NewIndexBuilder(contentStore content.Store) *IndexBuilder {
+	defaultSpanSize := int64(1 << 22)
```
rdpsin, line 25: We should define a const up top for defaultSpanSize

```
const (
    defaultSpanSize = 1 << 22
 )
 ```

**Comment by rdpsin** [comment]
Can you run `make add-ltag` to add headers to new files?

File: soci/index_builder.go [hunk 3]
```diff
@@ -0,0 +1,161 @@
...
+	return &IndexBuilder{
+		contentStore:        contentStore,
+		spanSize:            defaultSpanSize,
+		minLayerSize:        0,
+		buildToolIdentifier: "AWS SOCI CLI v0.1",
```
hanyuel, line 30: We probably should have the const string "AWS SOCI CLI v0.1"d defined else where and have the constructor pass in ```buildToolIdentifier```.

soci/index_builder.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (b *IndexBuilder) WithBuildToolIdentifer(buildToolIdentifier string) *IndexBuilder {
+	b.buildToolIdentifier = buildToolIdentifier
+	return b
+}
+
+// BuildIndex builds SOCI index for an image
+// This returns SOCI index manifest and an array of io.Reader of zTOCs
+func (b *IndexBuilder) BuildIndex(ctx context.Context, desc ocispec.Descriptor) (Index, []io.Reader, error) {
```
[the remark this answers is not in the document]
dvnguyen-amzn: Yes we will right after this PR

soci/index_builder.go
```diff
@@ -0,0 +1,161 @@
...
+	defaultSpanSize := int64(1 << 22)
+	return &IndexBuilder{
+		contentStore:        contentStore,
+		spanSize:            defaultSpanSize,
+		minLayerSize:        0,
+		buildToolIdentifier: "AWS SOCI CLI v0.1",
+	}
+}
+
+func (b *IndexBuilder) WithSpanSize(spanSize int64) *IndexBuilder {
```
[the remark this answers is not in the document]
dvnguyen-amzn: thanks for the link. I'll use that pattern in the next revision

soci/index_builder.go
dvnguyen-amzn: Agree that it's a big interface. Would you mind if I do a fast follow on that? The reasons are:
1. I don't know how yet
2. I don't want to introduce too much changes in a single PR.

soci/index_builder.go
dvnguyen-amzn: yes will do in the next revision

soci/index_builder.go
dvnguyen-amzn: yes will do in the next revision

soci/index_builder.go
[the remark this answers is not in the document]
dvnguyen-amzn: cc @qingshenhuang

soci/index_builder.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+type IndexBuilder struct {
+	contentStore        content.Store
+	spanSize            int64
+	minLayerSize        int64
+	buildToolIdentifier string
+}
+
+type IndexBuilderOption func(*IndexBuilder)
```
rdpsin, line 40: Maybe `IndexBuilderOption` should return an `error`, if an invalid value is ever supplied?

Then we can do


```
for _, opt := range opts {
    if err := opt(b); err != nil {
       return err
     }
}
```

dvnguyen-amzn: done

soci/index_builder.go
rdpsin: Sure.

soci/index_builder.go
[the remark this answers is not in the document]
rdpsin: Why not return `*Index` here, instead of `Index`?

soci/index_builder.go
dvnguyen-amzn: done

wmesard (approved):

soci/index_builder.go
```diff
@@ -0,0 +1,198 @@
...
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	DefaultSpanSize            = int64(1 << 22)
+	MinLayerSize               = 0
+	DefaultBuildToolIdentifier = "AWS SOCI CLI v0.1"
```
rdpsin, line 36: nit: Can probably set these to lowercase as they are not be exported from this package.

rdpsin (approved):

soci/index_builder.go
dvnguyen-amzn: makes sense. I can do it in the PR

hanyuel (approved):

Issue #99, opened 2022-10-05
"Chunk" is not a thing in the SOCI world


vkuzniet:
**Is your feature request related to a problem? Please describe.**
The Stargz and eStargz file formats have a notion of "chunks". During the conversion step, large files are split into multiple chunks, so that they can be pulled, verified and uncompressed separately. SOCI has no such notion. In the SOCI world, layers are divided into variable sized "spans" for pulling and checksumming purposes.

**Describe the solution you'd like**
Any code in SOCI that refers to `chunks` is dead code and needs to be removed.

The solution should satisfy the following criteria: 
- `cd SociSnapshotter; ag chunk | wc -l` returns 0
- If the above criteria cannot be satisfied, a comment is added to this story explaining why.

**Describe alternatives you've considered**

**Additional context**
See the following links for reference:
- https://github.com/google/crfs
- https://github.com/containerd/stargz-snapshotter/blob/main/docs/estargz.md

vkuzniet:
Closing, since it's already done.

PR #160, opened 2022-11-04
Remove cache from blob

hanyuel:
*Issue #, if available:*

*Description of changes:*
Remove the cache in ``blob`` for the following reasons: 
1) it's redundant. We already have `` span cache`` that caches contents fetched from remote. 
2) blob cache is keyed by chunk which should not be a thing in SOCI world, and we have an open issue #99 to remove chunk completely from this repository.

*Testing performed:* 
```make check && make test && make integration```

### DIFF
--- a/fs/remote/blob.go
+++ b/fs/remote/blob.go
@@ -43,17 +43,13 @@ import (
 	"fmt"
 	"io"
 	"regexp"
-	"sort"
-	"strings"
 	"sync"
 	"time"
 
-	"github.com/awslabs/soci-snapshotter/cache"
 	"github.com/awslabs/soci-snapshotter/fs/source"
 	"github.com/containerd/containerd/reference"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
-	"golang.org/x/sync/singleflight"
 )
 
 var contentRangeRegexp = regexp.MustCompile(`bytes ([0-9]+)-([0-9]+)/([0-9]+|\\*)`)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: fs/remote/blob.go [hunk 1]
```diff
@@ -63,9 +59,8 @@ type Blob interface {
 	Size() int64
 	FetchedSize() int64
 	ReadAt(p []byte, offset int64, opts ...Option) (int, error)
-	Cache(offset int64, size int64, opts ...Option) error
 	Refresh(ctx context.Context, host source.RegistryHosts, refspec reference.Spec, desc ocispec.Descriptor) error
-	Close() error
+	Close()
```
rdpsin, line 63: Why remove `error`?

rdpsin (approved):

fs/remote/blob.go
hanyuel: because there's no error to return after removing ``b.cache.Close()``.  See line 99 - line 105 of this file.

vkuzniet (approved):

fs/remote/blob.go
```diff
@@ -74,46 +69,39 @@ type blob struct {
...
-		cache:         blobCache,
 		lastCheck:     lastCheck,
 		checkInterval: checkInterval,
 		resolver:      r,
 		fetchTimeout:  fetchTimeout,
 	}
 }
 
-func (b *blob) Close() error {
+func (b *blob) Close() {
```
wmesard, line 99: `Close() error` is a standard method from the [Closer interface](https://pkg.go.dev/io#Closer). The current implementation may not return an error, but it is generally useful to have `blob` implement *Closer* in its familiar form.  a) It might want to return an error again at some point. b) Gophers [expect](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) `Close()` to return an error. So have it do so makes the code more readable. c) It costs nothing to have it do so.

wmesard (changes requested):

fs/remote/blob.go
hanyuel: Done. Please take a look at the latest revision

vkuzniet (approved):

rdpsin (approved):

Issue #149, opened 2022-11-01
Remove ORAS support from the snapshotter


rdpsin:
**Is your feature request related to a problem? Please describe.**
With oras-go migration from ORAS to OCI Artifact completed, we should look into removing ORAS support from the snapshotter. 

**Describe the solution you'd like**
ORAS support is removed.

PR #162, opened 2022-11-04
Update oras-go to support OCI Artifacts

rdpsin:
This commit updates the oras-go library to v2.0.0-rc4, which deprecates ORAS in favor of OCI Artifact. With this change, it is now possible to push and pull OCI Artifact compatible manifests.

This is a breaking change! Following this change, users will not be able to push and pull ORAS manifests anymore, and all existing manifests will be rendered unusable. Building an ORAS manifest is still possible, though.

Also, the OCI Artifact spec has renamed the `Refers` field in the manifest to `Subject` (similar to ORAS). Therefore, removed the field from SOCI Index.

*Issue #, if available:*
#149 

*Testing performed:*

Successfully pushed and pulled SOCI artifacts from an OCI Artifact compatible registry. Successfully ran multiple images using the SOCI snapshotter. 

### DIFF
--- a/integration/pull_test.go
+++ b/integration/pull_test.go
@@ -556,7 +556,7 @@ services:
     - "lazy-containerd-data:/var/lib/containerd"
     - "lazy-soci-snapshotter-grpc-data:/var/lib/soci-snapshotter-grpc"
   registry:
-    image: ghcr.io/oras-project/registry:v1.0.0-rc
+    image: ghcr.io/oci-playground/registry:v3.0.0-alpha.1
     container_name: {{.RegistryHost}}
     environment:
     - REGISTRY_AUTH=htpasswd
@@ -760,7 +760,7 @@ func buildSparseIndex(sh *shell.Shell, src imageInfo, minLayerSize int64) string
 	opts := encodeImageInfo(src)
 	indexDigest := sh.
 		X(append([]string{"ctr", "i", "pull"}, opts[0]...)...).
-		X("soci", "create", src.ref, "--min-layer-size", fmt.Sprintf("%d", minLayerSize), "--oras").
+		X("soci", "create", src.ref, "--min-layer-size", fmt.Sprintf("%d", minLayerSize)).
 		O("soci", "index", "list", "-q", "--ref", src.ref) // this will make SOCI artifact available locally
 	return strings.Trim(string(indexDigest), "\n")
 }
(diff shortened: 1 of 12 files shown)


soci/soci_index_test.go
```diff
@@ -279,18 +279,14 @@ func TestNewIndex(t *testing.T) {
...
-				if diff := cmp.Diff(index.Refers, &tc.subject); diff != "" {
-					t.Fatalf("the refers field is not equal; diff = %v", diff)
+					t.Fatalf("unexpected media type; expected = %v, got = %v", ORASManifestMediaType, mt)
 				}
+			} else if mt != OCIArtifactManifestMediaType {
+				t.Fatalf("unexpected media type; expected = %v, got = %v", OCIArtifactManifestMediaType, mt)
+			}
+
+			if diff := cmp.Diff(index.Subject, &tc.subject); diff != "" {
+				t.Fatalf("the refers field is not equal; diff = %v", diff)
```
hanyuel, line 289: Should be "the subject field"?

hanyuel:
Why is the integration test run skipped?

rdpsin:
Seems like it timed out.

integration/pull_test.go
```diff
@@ -556,7 +556,7 @@ services:
     - "lazy-containerd-data:/var/lib/containerd"
     - "lazy-soci-snapshotter-grpc-data:/var/lib/soci-snapshotter-grpc"
   registry:
-    image: ghcr.io/oras-project/registry:v1.0.0-rc
+    image: ghcr.io/oci-playground/registry:latest
```
vkuzniet, line 559: Can you use another tag, other than latest, so we have repeatability?

integration/testutil.go
```diff
@@ -251,7 +251,7 @@ services:
     - "lazy-containerd-data:/var/lib/containerd"
     - "lazy-soci-snapshotter-grpc-data:/var/lib/soci-snapshotter-grpc"
   registry:
-    image: ghcr.io/oras-project/registry:v1.0.0-rc
+    image: ghcr.io/oci-playground/registry:latest
```
vkuzniet, line 254: Can you use another tag, other than latest, so we have repeatability?

rdpsin: Done.

soci/soci_index_test.go
rdpsin: Yes, fixed.

vkuzniet (approved):

hanyuel (approved):

Issue #106, opened 2022-10-10
Re-design the background fetcher 


rdpsin:
**Is your feature request related to a problem? Please describe.**
The background fetcher is the component of the snapshotter that fetches spans in the background. In it's current state, it is not a particularly well-thought-out component, leaking abstractions and degrading snapshotter performance.

Some of the issues:

- The background fetcher acquires the lock and only releases it when it's done fetching the span from remote. This can lead to situations like #105, wherein on-demand fetches suffer due to lock contention.

- The background fetcher is very aggressive; currently, it tries to fetch all spans from all layers as fast as possible. This is undesirable (e.g., there might be cases we want to rate limit the background fetcher to avoid throttling, or prioritize certain layers' prefetch over others). 

- The background fetcher fetches a single span at a time. This is fine for when a span size is a desirable value (for e.g., 8-10MiB for S3), but no so much when it's not. For some span sizes and some repos, it may be more performant to fetch multiple spans at once.

**Describe the solution you'd like**
The background fetcher is re-architected and takes into consideration all the above points.

jamesmt-aws:
When we are fetching spans from S3-backed container registries like ECR and Dockerhub, we should be following the best practices for S3. Those include:
- Scale Connections Horizontally, so that multiple fetches are performed in parallel
- Use byte range fetches in the 8-16MiB range. If we set the span size to 8MiB we'd do this naturally, and 4MiB probably isn't too far off, but in the 4MiB span world fetching two spans at a time seems like a good idea. 
- Retry requests that are taking too long. 

All of this stuff is implemented in the AWS S3 SDK. I'm not sure if it's appropriate for SOCI to depend on that library, or if it makes more sense for us to reimplement that stuff ourselves, but I think it's important to follow best practices here. 

Here's the page on [Performance Guidelines for Amazon S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-guidelines.html) for reference.

rdpsin:
Implemented in PR #214.

PR #176, opened 2022-11-09
Temporarily turn off the background fetch until it gets re-worked

vkuzniet:

*Issue #, if available:*

*Description of changes:*
Temporarily turning off the background fetch until it gets re-worked. For more details see the issue https://github.com/awslabs/soci-snapshotter/issues/106

*Testing performed:*
- Deployed `soci-snapshotter-grpc` and `soci` to an EC2 instance, created an index for rethinkdb and ran the container workload in lazy loading mode. Observed no log record referring to background fetcher.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -175,6 +175,11 @@ func NewFilesystem(root string, cfg config.Config, opts ...Option) (_ snapshot.F
 	if ns != nil {
 		metrics.Register(ns) // Register layer metrics.
 	}
+
+	// We are turning off the background fetch until it gets re-worked.
+	// See https://github.com/awslabs/soci-snapshotter/issues/106 for more details.
+	cfg.NoBackgroundFetch = true
+
 	return &filesystem{
 		resolver:              r,
 		getSources:            getSources,


rdpsin (approved):

hanyuel (changes requested):
With this change, even when a user passes in NoBackgroundFetch=false in config parameter for running the snapshotter, the background fetch will not run.  I don’t think we should silently not run the bg fetch w/o any message when user explicitly sets the background fetch on. Sure the background fetch will be re-designed, but today it still can run. I suggest changing the behavior to not running the background fetch if no bg fetch config pass in from the user, and using the config from the user if provided.

vkuzniet:
> With this change, even when a user passes in NoBackgroundFetch=false in config parameter for running the snapshotter, the background fetch will not run. I don’t think we should silently not run the bg fetch w/o any message when user explicitly sets the background fetch on. Sure the background fetch will be re-designed, but today it still can run. I suggest changing the behavior to not running the background fetch if no bg fetch config pass in from the user, and using the config from the user if provided.

It is usually on by default. There's a config option to turn it off. The problem with intelligent handling there is that the `config.toml` is being deserialized into config struct and there's no way to know if the user set this field or it was set by default. 
If we agree that background fetcher as it is right now interferes with on-demand fetches, and it does seem so, then the easiest thing is to turn it off temporarily.

Another option would be to keep it disabled by default, e.g. name the option as `toml:background_fetch_enabled`, which will be set to false by default. In this case, the likelihood of someone enabling it would be much smaller.

sbuckfelder:
I'm comfortable with this

sbuckfelder (approved):

PR #169, opened 2022-11-08
Refactor SOCI CLI using refactored index builder modules

qingshenhuang:
*Issue #, if available:* [#128](https://github.com/awslabs/soci-snapshotter/issues/128)

*Description of changes:*

- Refactored **create** command so that it uses the newly created Index builder modules
- Consolidated soci_index.go by removing duplicate code
- Modified soci_index_test.go to test against the new index builder functions

*Testing performed:*

- Created indices using both original and new module and compare 
- Ran automated test cases and insured all test cases passed 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #179, opened 2022-11-09
Add platform option to IndexBuilder module

dvnguyen-amzn:
This change, which adopts https://github.com/awslabs/soci-snapshotter/pull/127, allows users to specify platform option when using the IndexBuilder module.

*Testing performed:*

`make && make test && make check`

### DIFF
--- a/soci/index_builder.go
+++ b/soci/index_builder.go
@@ -134,7 +134,7 @@ func (b *IndexBuilder) BuildIndex(ctx context.Context, desc ocispec.Descriptor) 
 		Size:        imgManifestDesc.Size,
 		Annotations: imgManifestDesc.Annotations,
 	}
-	return newOCIArtifactManifest(ztocsDesc, refers, annotations), ztocReaders, nil
+	return NewIndex(ztocsDesc, refers, annotations), ztocReaders, nil
 }
 
 // buildSociLayer builds the ztoc for an image layer and returns a Descriptor for the new ztoc.
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -31,9 +31,8 @@ import (
 const (
 	buildToolIdentifier = "AWS SOCI CLI v0.1"
 
-	spanSizeFlag           = "span-size"
-	minLayerSizeFlag       = "min-layer-size"
-	createORASManifestFlag = "oras"
+	spanSizeFlag     = "span-size"
+	minLayerSizeFlag = "min-layer-size"
 )
 
 // CreateCommand creates SOCI index for an image
@@ -55,10 +54,6 @@ var CreateCommand = cli.Command{
 			Name:  minLayerSizeFlag,
 			Usage: "The minimum layer size in bytes to build zTOC for. Default is 10 MiB.",
 			Value: 10 << 20,
-		},
-		cli.BoolFlag{
-			Name:  createORASManifestFlag,
-			Usage: "If set, will create an ORAS manifest instead of an OCI Artifact manifest. Default is false.",
 		},
 	),
 	Action: func(cliContext *cli.Context) error {
@@ -101,16 +96,10 @@ var CreateCommand = cli.Command{
 			return err
 		}
 
-		manifestType := soci.ManifestOCIArtifact
-		if cliContext.Bool(createORASManifestFlag) {
-			manifestType = soci.ManifestORAS
-		}
-
 		for _, plat := range ps {
 			sociIndexWithMetadata, err := soci.BuildSociIndex(ctx, cs, srcImg, spanSize, blobStore,
 				soci.WithMinLayerSize(minLayerSize),
 				soci.WithBuildToolIdentifier(buildToolIdentifier),
-				soci.WithManifestType(manifestType),
 				soci.WithPlatform(plat))
 
 			if err != nil {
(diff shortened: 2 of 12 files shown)


soci/index_builder.go
```diff
@@ -31,26 +31,29 @@ import (
 )
 
 const (
-	DefaultSpanSize            = int64(1 << 22)
-	MinLayerSize               = 0
-	DefaultBuildToolIdentifier = "AWS SOCI CLI v0.1"
+	defaultSpanSize            = int64(1 << 22)
+	minLayerSize               = 0
```
Kern--, line 35: nit: We have updated this to 10 MiB elsewhere. Can you update it here too?

soci/index_builder.go
```diff
@@ -83,17 +86,23 @@ func WithBuildToolIdentifierOption(buildToolIdentifier string) IndexBuilderOptio
 	}
 }
 
-// BuildIndex builds SOCI index for an image
+func WithPlatformOption(platform ocispec.Platform) IndexBuilderOption {
```
Kern--, line 89: nit: the community has generally fallen into either calling these functions `WithPlatform` or just `Platform`. If this is going to be our public API can we use a similar convention?

This can be a separate PR, if you agree.

soci/index_builder.go
```diff
@@ -83,17 +86,23 @@ func WithBuildToolIdentifierOption(buildToolIdentifier string) IndexBuilderOptio
...
+	return func(b *IndexBuilder) error {
+		b.platform = platform
```
Kern--, line 91: So thinking through how someone would use the builder - if they wanted to build an index for every platform the caller would have to create a new builder for each, e.g.

```
for _, platform := range platforms {
    builder := soci.NewIndexBuilder(store, WithPlatform(platform))
    index := builder.Build(ctx, desc)
}
```

This says to me that platform is a property of a particular call to `Build` and not of the `IndexBuilder` itself. It seems more natural to add a `BuildOption` concept and put the platform there.

e.g. usage like:
```
builder := soci.NewIndexBuilder(store)
for _, platform := range platforms {
    index := builder.Build(WithPlatform(platform))
}
```

Does that sound reasonable or is that almost going back to just having a `soci.BuildIndex` function?

soci/index_builder.go
```diff
@@ -83,17 +86,23 @@ func WithBuildToolIdentifierOption(buildToolIdentifier string) IndexBuilderOptio
...
+		return nil
+	}
+}
+
+// BuildIndex builds SOCI index for an image target (ocispec.Descriptor)
 // This returns SOCI index manifest and an array of io.Reader of zTOCs
-func (b *IndexBuilder) BuildIndex(ctx context.Context, desc ocispec.Descriptor) (*Index, []io.Reader, error) {
-	platform := platforms.Default()
+func (b *IndexBuilder) BuildIndex(ctx context.Context, desc ocispec.Descriptor) (*IndexWithMetadata, []io.Reader, error) {
```
Kern--, line 98: nit: `IndexBuilder.BuildIndex` feels redundant. Can this just be called `Build`?

soci/index_builder.go
```diff
@@ -134,7 +143,11 @@ func (b *IndexBuilder) BuildIndex(ctx context.Context, desc ocispec.Descriptor)
 		Size:        imgManifestDesc.Size,
 		Annotations: imgManifestDesc.Annotations,
 	}
-	return newOCIArtifactManifest(ztocsDesc, refers, annotations), ztocReaders, nil
+	index := newOCIArtifactManifest(ztocsDesc, refers, annotations)
+	return &IndexWithMetadata{
+		Index:       index,
+		ImageDigest: desc.Digest,
```
Kern--, line 149: There was another update recently that added `Platform` to `IndexWithMetadata`. Can you add that here too?

soci/index_builder.go
hanyuel: +1 to the 2nd implementation 
```
index := builder.Build(WithPlatform(platform))
```

soci/index_builder.go
dvnguyen-amzn: I'm glad that we're having this PR discussion, otherwise this could be missed. I'll update it.

soci/index_builder.go
dvnguyen-amzn: Yes completely agree. Currently we have a name conflict for other `WithXXX` functions, so that's why I added the `Option` suffix. @qingshenhuang will rename them in his PR https://github.com/awslabs/soci-snapshotter/pull/169. He's currently waiting for my PR to be merged.

soci/index_builder.go
dvnguyen-amzn: yes

soci/index_builder.go
dvnguyen-amzn: yes, will do

soci/index_builder.go
dvnguyen-amzn: That seems reasonable. I can introduce back the `BuildOption`.

dvnguyen-amzn: cc @qingshenhuang for Build function signature change

dvnguyen-amzn:
Closed in favor of https://github.com/awslabs/soci-snapshotter/pull/189

PR #180, opened 2022-11-09
Remove chunk-related code

hanyuel:
SOCI doesn't have a notion of "chunk". This commits removes the code that refers to "chunk".

*Issue #, if available:* #99

*Description of changes:*

*Testing performed:* 

```make check && make test && make integration```

### CODE REVIEW

File: cache/cache_test.go [hunk 1]
```diff
@@ -180,15 +180,15 @@ func hit(sample string) check {
 	return func(t *testing.T, c BlobCache) {
 		// test whole blob
 		key := digestFor(sample)
-		testChunk(t, c, key, 0, sample)
+		testSpan(t, c, key, 0, sample)
```
wmesard, line 183: While it is true that SOCI doesn't have a notion of "chunks", it is equally true that cache/* does not have a notion of "spans". These are unit tests of a blob cache. The things that a blob cache stores are "blobs". Or leave them as "chunks" if you prefer. But let's not confuse them with an unrelated concept from elsewhere in the code.

fs/remote/blob_test.go
```diff
@@ -58,30 +58,30 @@ import (
 const (
 	testURL            = "http://testdummy.com/v2/library/test/blobs/sha256:deadbeaf"
 	rangeHeaderPrefix  = "bytes="
-	sampleChunkSize    = 3
-	sampleMiddleOffset = sampleChunkSize / 2
+	sampleSpanSize     = 3
```
wmesard, line 61: As above, I think these are "blobs", not "spans".

wmesard (changes requested):

fs/remote/blob.go
```diff
@@ -187,28 +186,16 @@ func (b *blob) ReadAt(p []byte, offset int64, opts ...Option) (int, error) {
...
-			expectedSize = chunk.size() - upperUnread - lowerUnread
-		)
-
-		// Take it from remote registry.
-		// We get the whole chunk here.
-		allData[chunk] = newBytesWriter(p[base:base+expectedSize], lowerUnread)
-		return nil
-	})
+	// Take it from remote registry.
+	allData[reg] = newBytesWriter(p, 0)
```
vkuzniet, line 198: I don't think we need `allData` here, since there will be just a single writer now. I suggest removing it, getting writer directly and refactoring `fetchRange` to accommodate it.

vkuzniet (dismissed):

cache/cache_test.go
hanyuel: I updated this to keeping "chunk".

fs/remote/blob_test.go
hanyuel: I updated this to keeping "chunk".

fs/remote/blob.go
hanyuel: yes. I've refactored ``fetchRange`` and ``fetchRegions`` to take a single writer

fs/reader/reader.go
```diff
@@ -268,10 +266,10 @@ func WithReader(sr *io.SectionReader) CacheOption {
 	}
 }
 
-func digestVerifier(id uint32, chunkDigestStr string) (digest.Verifier, error) {
-	chunkDigest, err := digest.Parse(chunkDigestStr)
+func digestVerifier(id uint32, digestStr string) (digest.Verifier, error) {
+	digest, err := digest.Parse(digestStr)
 	if err != nil {
-		return nil, errors.Wrap(err, "invalid chunk: no digset is recorded")
+		return nil, errors.Wrap(err, "no digset is recorded")
```
vkuzniet, line 272: nit: maybe "invalid blob: no digest is recorded", but up to you

hanyuel: This func is not used. It's part of [verifier](https://github.com/awslabs/soci-snapshotter/blob/main/fs/reader/reader.go#L135) which should be removed completely but it's out of scope of this PR.

fs/remote/blob.go
```diff
@@ -263,82 +241,35 @@ func (b *blob) fetchRegions(allData map[region]io.Writer, fetched map[region]boo
...
-			reg.e = b.size - 1
-		}
-		if err := walkFn(reg); err != nil {
-			return err
-		}
-	}
-	return nil
+// fetchRange fetches content from remote blob.
+func (b *blob) fetchRange(reg region, w io.Writer, opts *options) error {
+	var fetched bool
```
vkuzniet, line 271: can you please just pass `false` here, since there's no use of `fetched` in the current context?

hanyuel: Done

vkuzniet (dismissed):

rdpsin (approved):

vkuzniet (approved):

Issue #153, opened 2022-11-02
Document all the metrics snapshotter emits


vkuzniet:
**Is your feature request related to a problem? Please describe.**
Soci snapshotter emits multiple metrics in prometheus format and through logs. They need to be documented, so anyone using snapshotter can easily understand what they are about.

**Describe the solution you'd like**
Create a doc, where you describe all the available metrics, give a brief description and describe how to read them. 
As the bare minimum, one can write the comments in `metrics.go` to document it there.

PR #185, opened 2022-11-11
Add comments with description to the metrics

vkuzniet:

*Issue #, if available:*
#153 

*Description of changes:*
This PR adds description for all metrics emitted by soci snapshotter in a form of comments in `metrics.go`.

*Testing performed:*
- `make check && make test && make integration` pass.

### DIFF
--- a/fs/metrics/common/metrics.go
+++ b/fs/metrics/common/metrics.go
@@ -62,22 +62,61 @@ const (
 
 // Lists all metric labels.
 const (
-	// prometheus metrics
-	Mount                         = "mount"
-	RemoteRegistryGet             = "remote_registry_get"
-	NodeReaddir                   = "node_readdir"
-	InitMetadataStore             = "init_metadata_store"
-	ReadOnDemand                  = "read_on_demand"
+	/**
+	Metrics emitted via prometheus endpoint. All latency metrics are measured in milliseconds,
+	unless specified otherwise.
+	*/
+	// Mount measures the duration of Mount operation for resolved layer.
+	Mount = "mount"
+
+	// RemoteRegistryGet records the roundtrip latency for remote registry GET operation.
+	RemoteRegistryGet = "remote_registry_get"
+
+	// NodeReaddir measures how long readdir operation takes (in microseconds).
+	NodeReaddir = "node_readdir"
+
+	// InitMetadataStore measures how long it takes to initialize the file metadata store.
+	InitMetadataStore = "init_metadata_store"
+
+	// ReadOnDemand measures the latency of on-demand reads (in microseconds).
+	ReadOnDemand = "read_on_demand"
+
+	// OnDemandReadAccessCount records the count of on-demand reads.
+	OnDemandReadAccessCount = "on_demand_read_access_count"
+
+	// OnDemandRemoteRegistryFetchCount records the count of the on-demand fetch requests to remote registry.
+	OnDemandRemoteRegistryFetchCount = "on_demand_remote_registry_fetch_count"
+
+	// OnDemandBytesServed counts the number of bytes served on-demand.
+	OnDemandBytesServed = "on_demand_bytes_served"
+
+	// OnDemandBytesFetched counts the number of bytes fetched on-demand.
+	OnDemandBytesFetched = "on_demand_bytes_fetched"
+
+	/**
+	Metrics logged in snapshotter's log.
+	All latencies here are measured in milliseconds, unless stated otherwise.
+	*/
+	// MountLayerToLastOnDemandFetch measures the latency of last on demand fetch,
+	// which must be invoked at the end of background fetch operation only.
+	// Since this is expected to happen only once per container launch, it is recorded as a log line,
+	// instead of being emitted as a metric.
+	// The algorithm for recording is as follows:
+	// 1. Record the mount start time.
+	// 2. Constantly record the timestamps when the on-demand fetch happening for each layer.
+	// 3. On background fetch completed, measure the difference between the last on demand fetch
+	// and mount start time and record it.
 	MountLayerToLastOnDemandFetch = "mount_layer_to_last_on_demand_fetch"
 
-	OnDemandReadAccessCount          = "on_demand_read_access_count"
-	OnDemandRemoteRegistryFetchCount = "on_demand_remote_registry_fetch_count"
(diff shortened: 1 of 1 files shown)


fs/metrics/common/metrics.go
```diff
@@ -62,22 +62,61 @@ const (
...
-	ReadOnDemand                  = "read_on_demand"
-	MountLayerToLastOnDemandFetch = "mount_layer_to_last_on_demand_fetch"
+	/**
+	Metrics emitted via prometheus endpoint. All latency metrics are measured in milliseconds,
+	unless specified otherwise.
+	*/
+	// Mount measures the duration of Mount operation for resolved layer.
+	Mount = "mount"
+
+	// RemoteRegistryGet records the roundtrip latency for remote registry GET operation .
```
hanyuel, line 72: Extra space before "."

fs/metrics/common/metrics.go
```diff
@@ -62,22 +62,61 @@ const (
...
+	RemoteRegistryGet = "remote_registry_get"
+
+	// NodeReaddir measures how long readdir operation takes (in microseconds).
+	NodeReaddir = "node_readdir"
+
+	// InitMetadataStore measures how long it takes to initialize the file metadata.
```
hanyuel, line 78: initialize the file metadata store

fs/metrics/common/metrics.go
```diff
@@ -62,22 +62,61 @@ const (
...
+	InitMetadataStore = "init_metadata_store"
+
+	// ReadOnDemand measures the latency of on-demand reads (in mocroseconds).
```
hanyuel, line 81: microsecond

fs/metrics/common/metrics.go
```diff
@@ -62,22 +62,61 @@ const (
...
+
+	// OnDemandBytesFetched counts the number of bytes fetched on-demand.
+	OnDemandBytesFetched = "on_demand_bytes_fetched"
+
+	/**
+	Metrics logged in snapshotter's log.
+	All latencies here are measured in milliseconds, unless stated otherwise.
+	*/
+	// MountLayerToLastOnDemandFetch measures the latency of last on demand fetch,
+	// which must be invoked at the end of background fetch operation only.
```
hanyuel, line 101: I'm not following here. Why does MountLayerToLastOnDemandFetch need to be invoked at the end of _background_ fetch?

fs/metrics/common/metrics.go
```diff
@@ -62,22 +62,61 @@ const (
...
+	// Since this is expected to happen only once per container launch, it is recorded as a log line,
+	// instead of being emitted as a metric.
```
hanyuel, line 103: It's not emitted as metric, so why is it included in the list of Metrics here?

fs/metrics/common/metrics.go
vkuzniet: Done.

fs/metrics/common/metrics.go
vkuzniet: Done.

fs/metrics/common/metrics.go
vkuzniet: Done.

fs/metrics/common/metrics.go
vkuzniet: Because it's written in the log. It doesn't make sense to emit a metric, for a one-time event, that is why it is written in the log.

fs/metrics/common/metrics.go
vkuzniet: This metric is inherited from stargz snapshotter. The idea is that on-demand fetches finish when background fetch is complete, since all the data for the layer is local. I'm just documenting it as it is right now. Most likely, we need to revisit the approach, given that we plan to do changes in background fetcher.

fs/metrics/common/metrics.go
vkuzniet: Everything, related to background fetch is written in the log, since it's just 1 value per layer.

fs/metrics/common/metrics.go
hanyuel: Do we log this metric now?

vkuzniet: Yes, it's logged here: https://github.com/awslabs/soci-snapshotter/blob/main/fs/fs.go#L561

vkuzniet: We inherited this metric from stargz snapshotter. Renaming and reworking that metric should be a topic of the other PR after we conclude the broader conversation on the metrics. This PR just documents what is already in there. 
On that topic, probably, it will make sense to emit it with via the prometheus endpoint, so we capture the time when last on-demand fetch for that particular layer was done.

hanyuel: ok, I can approve this PR, but I still think we should rename it in this PR because the current name is confusing and does not reflect what is being logged, and it won't take much effort to change a name.

hanyuel (approved):

fs/metrics/common/metrics.go
```diff
@@ -62,22 +62,61 @@ const (
...
+	RemoteRegistryGet = "remote_registry_get"
+
+	// NodeReaddir measures how long readdir operation takes (in microseconds).
+	NodeReaddir = "node_readdir"
+
+	// InitMetadataStore measures how long it takes to initialize the file metadata store.
+	InitMetadataStore = "init_metadata_store"
+
+	// ReadOnDemand measures the latency of on-demand reads (in microseconds).
+	ReadOnDemand = "read_on_demand"
```
wmesard, line 82: "OnDemand" is a funny term. I think we will have three kinds of reads in the fullness of time. If we can think of names that would be particularly helpful to someone doing performance analysis, that would be great. Failing that, we should use the terms that our code uses. We don't use the term "on demand" anywhere except when referring to these metrics. 

I'll suggest:
* "Synchronous reads" or maybe "Blocking reads".
* "Prefetch reads" 
* "Background reads"
I don't feel that strongly about them.

wmesard (changes requested):
Honestly, I don't think these comments are adding much value. Most of them are of the form
```
    FooBarBing measures how long it takes Foo Bar to Bing.
```
I think we should actually document them. Define precisely what they measure, why they're useful, and how someone might use them. Or just let them speak for themselves if the name is clear. (Or decide that they're actually *not* useful and remove them.)

Issue #123, opened 2022-10-21
Incapsulate Ztoc serialization logic into ZtocSerializer


vkuzniet:
**Is your feature request related to a problem? Please describe.**
Ztoc has a very complex serialization logic, which involves doing manual placement of data to compose the flatbuffer byte stream. This asks for a separate entity to do that, instead of relying on `ztocToFlatbuf` and `flatbufToZtoc`.

**Describe the solution you'd like**
Implement ZtocSerializer and replace `ztocToFlatbuf` and `flatbufToZtoc` by using the respective `Serialize/Deserialize` methods.

**Describe alternatives you've considered**
Keep it as is.

**Additional context**
N/A

vkuzniet:
Done.

PR #188, opened 2022-11-11
Add ZtocSerializer

vkuzniet:

*Issue #, if available:*
#123 

*Description of changes:*
Refactored ztoc serialization logic to include it into a single unit - `ZtocSerializer`.

*Testing performed:*

### DIFF
--- a/soci/ztoc_serializer.go
+++ b/soci/ztoc_serializer.go
@@ -1,0 +1,244 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package soci
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"sort"
+	"time"
+
+	ztoc_flatbuffers "github.com/awslabs/soci-snapshotter/soci/fbs/ztoc"
+	flatbuffers "github.com/google/flatbuffers/go"
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// ZtocMarshaler incapsulates the serialization logic for Ztoc.
+// Currently, it implements serialization/deserialization to/from flatbuffers.
+type ZtocMarshaler struct {
+}
+
+// Marshal serializes Ztoc to its flatbuffers schema and returns a reader along with the descriptor (digest and size only).
+// If not successful, it will return an error.
+func (zm ZtocMarshaler) Marshal(ztoc *Ztoc) (io.Reader, ocispec.Descriptor, error) {
+	flatbuf, err := zm.ztocToFlatbuffer(ztoc)
+	if err != nil {
+		return nil, ocispec.Descriptor{}, err
+	}
+
+	buf := bytes.NewReader(flatbuf)
+	dgst := digest.FromBytes(flatbuf)
+	size := len(flatbuf)
+	return buf, ocispec.Descriptor{
+		Digest: dgst,
+		Size:   int64(size),
+	}, nil
+}
+
+// Unmarshal takes the reader with flatbuffers byte stream and deserializes it ztoc.
+// In case if there's any error situation during deserialization from flatbuffers, there will be an error returned.
+func (zm ZtocMarshaler) Unmarshal(serializedZtoc io.Reader) (*Ztoc, error) {
+	flatbuf, err := io.ReadAll(serializedZtoc)
+	if err != nil {
+		return nil, err
+	}
+
+	return zm.flatbufToZtoc(flatbuf)
+}
+
+func (zm ZtocMarshaler) flatbufToZtoc(flatbuffer []byte) (z *Ztoc, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			z = nil
+			err = fmt.Errorf("cannot unmarshal ztoc: %v", r)
+		}
+	}()
+
+	ztoc := new(Ztoc)
+	ztocFlatbuf := ztoc_flatbuffers.GetRootAsZtoc(flatbuffer, 0)
+	ztoc.Version = string(ztocFlatbuf.Version())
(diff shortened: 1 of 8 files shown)


rdpsin:
Can you explain the reasoning behind this change? AFAICT, this doesn't change anything except needing to build a ZtocSerializer object first?

vkuzniet:
> Can you explain the reasoning behind this change? AFAICT, this doesn't change anything except needing to build a ZtocSerializer object first?

Yes, this doesn't change anything, but implements the separation of concerns for Ztoc's serialization. ZtocSerializer will be the only one who knows how to serialize/deserialize Ztoc.

soci/ztoc_serializer.go
```diff
@@ -0,0 +1,227 @@
...
+
+// ZtocSerializer incapsulates the serialization logic for Ztoc.
+// Currently, it implements serialization/deserialization to/from flatbuffers.
+type ZtocSerializer struct {
+}
+
+// Serialize serializes Ztoc to its flatbuffers schema and returns a reader along with the descriptor (digest and size only).
+// In case if there's any error situation during serialization to flatbuffers, there will be a panic issued.
+// This is due to the nature of flatbuffers go library, which issues panic for any error.
+func (zs *ZtocSerializer) Serialize(ztoc *Ztoc) (io.Reader, ocispec.Descriptor, error) {
```
rdpsin, line 39: np: This can be `func (zs ZtocSerializer)`.

hanyuel (approved):

soci/ztoc_serializer.go
```diff
@@ -0,0 +1,227 @@
...
+	ztoc_flatbuffers "github.com/awslabs/soci-snapshotter/soci/fbs/ztoc"
+	flatbuffers "github.com/google/flatbuffers/go"
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// ZtocSerializer incapsulates the serialization logic for Ztoc.
+// Currently, it implements serialization/deserialization to/from flatbuffers.
+type ZtocSerializer struct {
+}
```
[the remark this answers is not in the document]
vkuzniet: I left it as a struct, since it has two methods: Serialize and Deserialize. Potentially the ZtocSerializer may get some state, but it's not needed as of today.

soci/ztoc_serializer.go
kzys: +1

vkuzniet: Done.

kzys (dismissed):

soci/ztoc_serializer.go
```diff
@@ -0,0 +1,227 @@
...
+	dgst := digest.FromBytes(flatbuf)
+	size := len(flatbuf)
+	return buf, ocispec.Descriptor{
+		Digest: dgst,
+		Size:   int64(size),
+	}, nil
+}
+
+// Deserialize takes the reader with flatbuffers byte stream and deserializes it ztoc.
+// In case if there's any error situation during deserialization from flatbuffers, there will be a panic issued.
```
rdpsin, line 51: Can we do a `recover()` and issue an error? I can imagine someone would be able to pass in an arbitrary blob as a Ztoc. If that happens, it would crash the snapshotter and all workloads would be impacted.

soci/ztoc_serializer.go
```diff
@@ -0,0 +1,227 @@
...
+
+// ZtocSerializer incapsulates the serialization logic for Ztoc.
+// Currently, it implements serialization/deserialization to/from flatbuffers.
+type ZtocSerializer struct {
+}
+
+// Serialize serializes Ztoc to its flatbuffers schema and returns a reader along with the descriptor (digest and size only).
+// In case if there's any error situation during serialization to flatbuffers, there will be a panic issued.
+// This is due to the nature of flatbuffers go library, which issues panic for any error.
+func (zs ZtocSerializer) Serialize(ztoc *Ztoc) (io.Reader, ocispec.Descriptor, error) {
```
rdpsin, line 39: Maybe change from `Serialize/Deserialie` to `Marshal/Unmarshal` or `Encode/Decode` (they seem to be more idiomatic Go notation)?

soci/ztoc_serializer.go
vkuzniet: Done. I think, that makes sense.

soci/ztoc_serializer.go
vkuzniet: Done.

soci/ztoc_test.go
```diff
@@ -607,6 +609,28 @@ func TestWriteZtoc(t *testing.T) {
...
+			serializedZtoc: genRandomByteData(50000),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			zs := ZtocSerializer{}
+			r := bytes.NewReader(tc.serializedZtoc)
+			if _, err := zs.Unmarshal(r); err == nil {
+				t.Fatalf("")
```
rdpsin, line 628: Can we be more descriptive here? 

"Expected error but got nil"

vkuzniet: Good catch. Done.

soci/ztoc_serializer.go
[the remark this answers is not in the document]
Kern--: We've talked in the past about having a separate `ztoc` package. Wouldn't it make sense to just have `ztoc.Serialize` and `ztoc.Deserialize` functions rather than these types?

soci/ztoc_serializer.go
Kern--: If we go with this change, we should probably also rename the struct to `ZTOCMarshaler`.

vkuzniet: Done.

soci/ztoc_serializer.go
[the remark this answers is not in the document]
vkuzniet: In the world as it is today, it makes sense. I wanted to keep it as type for the case if it gets some state in the future. This way, we don't need to change the invocations throughout the code.

hanyuel (approved):

soci/ztoc_serializer.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	return zm.flatbufToZtoc(flatbuf)
+}
+
+func (zm ZtocMarshaler) flatbufToZtoc(flatbuffer []byte) (z *Ztoc, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			z = nil
+			err = fmt.Errorf("cannot unmarshal Ztoc")
```
rdpsin, line 69: Can you add more context to the error message?

```
err = fmt.Errorf("cannot unmarshal ztoc: %v", r)
```

vkuzniet: Done.

hanyuel (approved):

rdpsin (approved):

Issue #184, opened 2022-11-10
[BENCHMARKING] Add file access pattern information


sbuckfelder:
To better analyze our benchmarking results we need to log out which files are being requested and when they are being requested.  

**Possible Solution:**
We could do this as part of debug logging and then parse the logs to get the file access pattern

vkuzniet:
Since #196 is merged, path forward is:
1. Execute the container workload
2. Retrieve the log file
3. Execute something like `cat logs | grep -w "FUSE operation" | cut -d ',' -f3-`, which will produce the output in the way outlined below
```
"operation":"Getattr","path":"","time":"2022-11-22T22:07:07.516131503Z"}
"operation":"Getattr","path":"","time":"2022-11-22T22:07:07.532138857Z"}
"operation":"Getattr","path":"","time":"2022-11-22T22:07:07.546298630Z"}
"operation":"Getattr","path":"","time":"2022-11-22T22:07:07.561092639Z"}
"operation":"Getattr","path":"","time":"2022-11-22T22:07:07.578447067Z"}
"operation":"Getattr","path":"","time":"2022-11-22T22:07:21.476433825Z"}
"operation":"Getxattr","path":"etc","time":"2022-11-22T22:07:21.488078862Z"}
"operation":"Getxattr","path":"etc","time":"2022-11-22T22:07:21.488161976Z"}
"operation":"Getxattr","path":"etc","time":"2022-11-22T22:07:21.488297596Z"}
"operation":"Getxattr","path":"etc","time":"2022-11-22T22:07:21.488394330Z"}
"operation":"Getxattr","path":"etc","time":"2022-11-22T22:07:21.488527780Z"}
"operation":"Getxattr","path":"etc","time":"2022-11-22T22:07:21.488623926Z"}
"operation":"Getxattr","path":"etc/group","time":"2022-11-22T22:07:21.489083397Z"}
"operation":"Open","path":"etc/group","time":"2022-11-22T22:07:21.489172398Z"}
"operation":"Read","path":"etc/group","time":"2022-11-22T22:07:21.489263811Z"}
"operation":"Getxattr","path":"proc","time":"2022-11-22T22:07:21.751545884Z"}
"operation":"Getxattr","path":"dev","time":"2022-11-22T22:07:21.752071272Z"}
"operation":"Getxattr","path":"sys","time":"2022-11-22T22:07:21.752770652Z"}
[...]

PR #196, opened 2022-11-21
Add file access logging

vkuzniet:

*Issue #, if available:*
#184 

*Description of changes:*
To be able to collect file access patterns via fuse, this change adds the debug logging in node.go for FUSE operations as well as config option to enable that logging. This capability should never be used in production. It is turned off by default.

*Testing performed:*
- Deployed snapshotter to ec2 instance, updated `/etc/soci-snapshotter-grpc/config.toml` to set `log_fuse_operations` to true. Ran rethinkdb workload in soci mode and observed the following lines in the log (the workload without setting `log_fuse_operations` to true does not produce the log lines with fuse operations information): 

```
...
{"level":"debug","msg":"[FUSE] operation=Getxattr on path=etc/resolv.conf","time":"2022-11-21T22:17:49.035050285Z"}
{"level":"debug","msg":"[FUSE] operation=Getxattr on path=data","time":"2022-11-21T22:17:49.036099425Z"}
{"level":"debug","msg":"[FUSE] operation=Getxattr on path=data","time":"2022-11-21T22:17:49.036198736Z"}
{"level":"debug","msg":"[FUSE] operation=Getattr on path=data","time":"2022-11-21T22:17:49.037002695Z"}
{"level":"debug","msg":"[FUSE] operation=Getxattr on path=etc/passwd","time":"2022-11-21T22:17:49.037338965Z"}
{"level":"debug","msg":"[FUSE] operation=Open on path=etc/passwd","time":"2022-11-21T22:17:49.037430002Z"}
{"level":"debug","msg":"[FUSE] operation=Getxattr on path=etc/group","time":"2022-11-21T22:17:49.037588983Z"}
{"level":"debug","msg":"[FUSE] operation=Open on path=etc/group","time":"2022-11-21T22:17:49.037667408Z"}
{"level":"debug","msg":"[FUSE] operation=Read on path=etc/passwd","time":"2022-11-21T22:17:49.037766639Z"}
{"level":"debug","msg":"[FUSE] operation=Getxattr on path=usr","time":"2022-11-21T22:17:49.038198795Z"}
...
```

### DIFF
--- a/fs/layer/node.go
+++ b/fs/layer/node.go
@@ -88,7 +88,7 @@ var opaqueXattrs = map[OverlayOpaqueType][]string{
 	OverlayOpaqueUser:    {"user.overlay.opaque"},
 }
 
-func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseInode uint32, opaque OverlayOpaqueType) (fusefs.InodeEmbedder, error) {
+func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseInode uint32, opaque OverlayOpaqueType, logFSOperations bool) (fusefs.InodeEmbedder, error) {
 	rootID := r.Metadata().RootID()
 	rootAttr, err := r.Metadata().GetAttr(rootID)
 	if err != nil {
@@ -99,11 +99,12 @@ func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseIno
 		return nil, fmt.Errorf("unknown overlay opaque type")
 	}
 	ffs := &fs{
-		r:            r,
-		layerDigest:  layerDgst,
-		baseInode:    baseInode,
-		rootID:       rootID,
-		opaqueXattrs: opq,
+		r:               r,
+		layerDigest:     layerDgst,
+		baseInode:       baseInode,
+		rootID:          rootID,
+		opaqueXattrs:    opq,
+		logFSOperations: logFSOperations,
 	}
 	ffs.s = ffs.newState(layerDgst, blob)
 	return &node{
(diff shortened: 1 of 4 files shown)


rdpsin:
I think we should enable FS logging when `log-level=debug` is enabled, instead of having another option in the config.

fs/layer/node.go
```diff
@@ -149,6 +151,12 @@ type node struct {
 	entsCached bool
 }
 
+func (n *node) logOperation(ctx context.Context, operationName string) {
+	if n.fs.logFSOperations {
+		log.G(context.Background()).Debugf("[FUSE] operation=%s on path=%s", operationName, n.Path(nil))
```
rdpsin, line 156: We should use `WithFields` here.

```
log.G(ctx).WithFields(logrus.Fields{
          "operation": operationName,
          "path": n.Path(nil),
}).Debugf("FUSE operation")
```

rdpsin:
https://github.com/awslabs/soci-snapshotter/issues/184 mentions logging the access time as well. This PR does not do that. Will there be a follow-up PR?

fs/layer/node.go
sbuckfelder: +1

sbuckfelder:
I do think we should use logrus fields as @rdpsin pointed out, but other then that I'm happy with this

wmesard:
> 



>

wmesard:
> I think we should enable FS logging when `log-level=debug` is enabled, instead of having another option in the config

A separate config option for this is probably a good idea. This has the potential to fill up the local disk with log files, consume lots of computes and iops, and generally wreak havoc. It needs to be used with caution.

vkuzniet:
> ~#184 mentions logging the access time as well. This PR does not do that. Will there be a follow-up PR?~
> 
> Nevermind. Just saw the output.
> 
> Is there a way to relate the time requested to when the container process started, i.e. log the time relative to container start?

I think, you can easily do so within benchmarking framework, where you just record the container start time and then start parsing logs accounting for that start time.

fs/layer/node.go
vkuzniet: Done.

rdpsin (approved):

fs/layer/node.go
```diff
@@ -165,6 +176,7 @@ var _ = (fusefs.InodeEmbedder)((*node)(nil))
 var _ = (fusefs.NodeReaddirer)((*node)(nil))
 
 func (n *node) Readdir(ctx context.Context) (fusefs.DirStream, syscall.Errno) {
+	n.logOperation(context.Background(), "Readdir")
```
kzys, line 179: Why doesn't it take the incoming context (`ctx`)?

vkuzniet: Updated to take the incoming context.

rdpsin (approved):

PR #197, opened 2022-11-22
Add Stargz for Benchmarking Boilerplate

sbuckfelder:

This add all of the boilerplate for stargz, but none of the logic.  It does add a Makefile target `make benchmarks-stargz` that will run and return very quickly - as it does not do anything.  These empty shells can be filled in similar to how soci was implemented.

One important thing to note is that we will need to build new images for stargz and have a file that references them.

### DIFF
--- a/benchmark/benchmarkTests.go
+++ b/benchmark/benchmarkTests.go
@@ -153,6 +153,25 @@ func OverlayFSFullRun(
 	b.StopTimer()
 }
 
+func StargzFullRun(
+	b *testing.B,
+	imageRef string,
+	readyLine string) {
+	ctx, cancelCtx := framework.GetTestContext()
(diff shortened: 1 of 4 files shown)


benchmark/benchmarkTests.go
```diff
@@ -171,3 +185,8 @@ func getSociProcess() (*SociProcess, error) {
 		sociConfig,
 		outputDir)
 }
+
+func getStargzProcess() (*StargzProcess, error) {
```
vkuzniet, line 189: Can you please address the concerns from linter?

sbuckfelder: Done

hanyuel (approved):

PR #202, opened 2022-11-22
Make "make benchmarks" work in the fresh clone

kzys:
It still results zeros though.

*Issue #, if available:*

*Description of changes:*

*Testing performed:*

### DIFF
--- a/benchmark/framework/aws_utils.go
+++ b/benchmark/framework/aws_utils.go
@@ -32,7 +32,12 @@ func (proc *ContainerdProcess) PullImageFromECR(
 	platform string,
 	awsSecretFile string) (containerd.Image, error) {
 	opts := GetRemoteOpts(ctx, platform)
-	opts = append(opts, containerd.WithResolver(GetECRResolver(ctx, awsSecretFile)))
+
+	resolver, err := GetECRResolver(ctx, awsSecretFile)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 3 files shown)


benchmark/singleImage.csv
```diff
@@ -0,0 +1 @@
+Debian,docker.io/library/debian:latest,"",""
```
kzys, line 1: Do we have a list of public images we should test against?

benchmark/framework/aws_utils.go
```diff
@@ -32,20 +32,25 @@ func (proc *ContainerdProcess) PullImageFromECR(
 	platform string,
 	awsSecretFile string) (containerd.Image, error) {
 	opts := GetRemoteOpts(ctx, platform)
-	opts = append(opts, containerd.WithResolver(GetECRResolver(ctx, awsSecretFile)))
+
+	resolver, err := GetECRResolver(ctx, awsSecretFile)
```
sbuckfelder, line 36: I'm happy with this change - but has nothing to do with your PR as per your comment.

benchmark/singleImage.csv
sbuckfelder: Great question, we originally had not included one because the only registry we had was an AWS internal preprod.  If I recall Docker hub now support referrers API.  If so we can move there, it would be great to be more public.

hanyuel:
Could you provide some context for why this change is needed / what problem we are seeing with ``make benchmarks``?

Issue #183, opened 2022-11-10
[BENCHMARKING] Add stargz for benchmark comparisons


sbuckfelder:
Soci-snapshotter started as a fork of stargz, another lazy loading snapshotter.  Currently the benchmarking framework supports only soci-snapshotter and overlayfs.  Let's add stargz support so we can sanity check that soci is providing comparable benefits as stargz.

**Possible Solution** : 
To run with stargz we will need to ensure that the stargz daemon is available on the machine and running.  To do this we can get the binaries from the stargz github releases (https://github.com/containerd/stargz-snapshotter/releases) and then run it in a similar fashion to what is currently being done in the _soci_utils_ file.

PR #206, opened 2022-11-29
Start and stop stargz process for benchmarking

hanyuel:
Add the ability of starting and stopping the stargz snapshotter process in running benchmarks.

To run the stargz benchmark, run:
``make benchmarks-stargz``
It uses the default location of the stargz binary installed by the stargz snapshotter's makefile.

Or, we can optionally pass in the path to the stargz binary to the command:
``make benchmarks-stargz STARGZ_BINARY=<path-to-binary>``

Also moved ``ImageDescriptor`` and ``GetImageListFromCsv`` to a new ``utils.go`` file as they can be shared by both soci and stargz tests.

*Issue #, if available:* [#183
](https://github.com/awslabs/soci-snapshotter/issues/183)
*Description of changes:*

*Testing performed:* 
``make benchmarks-stargz`` runs successfully.
``make benchmarks-stargz STARGZ_BINARY=<path-to-binary>`` runs successfully
``make benchmarks-comp`` pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ SOCI_SNAPSHOTTER_PROJECT_ROOT ?= $(shell pwd)
 LTAG_TEMPLATE_FLAG=-t ./.headers
 FBS_FILE_PATH=$(CURDIR)/soci/fbs/ztoc.fbs
 COMMIT=$(shell git rev-parse HEAD)
+STARGZ_BINARY?=/usr/local/bin/containerd-stargz-grpc
 
 CMD=soci-snapshotter-grpc soci
 
@@ -138,4 +139,4 @@ benchmarks-comp:
 
 benchmarks-stargz:
 	@echo "$@"
-	@cd benchmark/stargzTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/StargzTests . && sudo ../bin/StargzTests $(COMMIT) ../singleImage.csv 10
+	@cd benchmark/stargzTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/StargzTests . && sudo ../bin/StargzTests $(COMMIT) ../singleImage.csv 10 $(STARGZ_BINARY)
--- a/benchmark/benchmarkTests.go
+++ b/benchmark/benchmarkTests.go
@@ -36,6 +36,9 @@ var (
 	sociRoot          = "/tmp/lib/soci-snapshotter-grpc"
 	sociConfig        = "../soci_config.toml"
 	awsSecretFile     = "../aws_secret"
+	stargzAddress     = "/tmp/containerd-stargz-grpc/containerd-stargz-grpc.sock"
+	stargzConfig      = "../stargz_config.toml"
+	stargzRoot        = "/tmp/lib/containerd-stargz-grpc"
 )
 
 func PullImageFromECR(ctx context.Context, b *testing.B, imageRef string) {
@@ -161,13 +164,14 @@ func StargzFullRun(
 	ctx context.Context,
 	b *testing.B,
 	imageRef string,
-	readyLine string) {
+	readyLine string,
+	stargzBinary string) {
 	containerdProcess, err := getContainerdProcess(ctx)
 	if err != nil {
 		b.Fatalf("Failed to create containerd proc: %v\n", err)
 	}
 	defer containerdProcess.StopProcess()
-	stargzProcess, err := getStargzProcess()
+	stargzProcess, err := getStargzProcess(stargzBinary)
 	if err != nil {
 		b.Fatalf("Failed to create stargz proc: %v\n", err)
 	}
@@ -195,7 +199,11 @@ func getSociProcess() (*SociProcess, error) {
 		outputDir)
 }
 
-func getStargzProcess() (*StargzProcess, error) {
-	//TODO
-	return nil, nil
+func getStargzProcess(stargzBinary string) (*StargzProcess, error) {
+	return StartStargz(
+		stargzBinary,
+		stargzAddress,
+		stargzRoot,
+		stargzConfig,
+		outputDir)
 }
--- a/benchmark/stargz_utils.go
+++ b/benchmark/stargz_utils.go
@@ -16,13 +16,103 @@
 
 package benchmark
 
+import (
+	"errors"
+	"fmt"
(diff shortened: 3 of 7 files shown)


benchmark/utils.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	csv, err := csv.NewReader(csvFile).ReadAll()
+	if err != nil {
+		return nil, err
+	}
+	var images []ImageDescriptor
+	for _, image := range csv {
+		images = append(images, ImageDescriptor{
+			ShortName: image[0],
+			ImageRef:  image[1],
```
sbuckfelder, line 44: won't this break for SociIndexManifestRef?  That was why these were in the main.go files.  Another option if you prefer this method is to have the SociIndexManifestRef be the last one.

benchmark/stargz_utils.go
```diff
@@ -16,13 +16,104 @@
...
+	"os"
+	"os/exec"
+	"syscall"
+	"time"
+)
+
+var (
+	stargzAddress = "/tmp/containerd-stargz-grpc/containerd-stargz-grpc.sock"
+	stargzRoot    = "/tmp/lib/containerd-stargz-grpc"
+	stargzConfig  = "../snapshotter_config.toml"
```
sbuckfelder, line 31: the snapshotter config reads
```
[cri_keychain]
  enable_keychain = true 
  image_service_path = "/tmp/containerd-grpc/containerd.sock"
```
I'm guessing we can use it for both?

benchmark/stargz_utils.go
```diff
@@ -16,13 +16,104 @@
...
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"syscall"
+	"time"
+)
+
+var (
+	stargzAddress = "/tmp/containerd-stargz-grpc/containerd-stargz-grpc.sock"
```
sbuckfelder, line 29: noticed this is here as opposed to in `benchmark/benchmark_tests.go` any reason why you chose a different pattern?

hanyuel: It is stargz specific, so I put it in ``stargz_utils.go``.

benchmark/stargz_utils.go
hanyuel: yes

sbuckfelder:
This PR does lots more refactoring then it does **Start and stop stargz process for benchmarking**.  I'm very open to the changes, but could you explain in the description what you are doing and why you made those decisions?

sbuckfelder (dismissed):
I believe these changes will break the soci snapshotter tests, could you run the `benchmarks-comp`?  Thanks!

hanyuel:
@sbuckfelder the refactoring is because there's common code that should be shared by soci and stargz tests.

benchmark/benchmarkTests.go
```diff
@@ -34,8 +34,10 @@ var (
 	sociBinary        = "../../out/soci-snapshotter-grpc"
 	sociAddress       = "/tmp/soci-snapshotter-grpc/soci-snapshotter-grpc.sock"
 	sociRoot          = "/tmp/lib/soci-snapshotter-grpc"
-	sociConfig        = "../soci_config.toml"
+	snapshotterConfig = "../snapshotter_config.toml"
```
vkuzniet, line 37: how do you distinguish between soci and stargz snapshotter configs?

benchmark/utils.go
hanyuel: Moved ``SociIndexManifestRef`` to the the last one in ``ImageDescriptor``

sbuckfelder (dismissed):

benchmark/benchmarkTests.go
hanyuel: i think you meant [this config](https://github.com/awslabs/soci-snapshotter/blob/main/benchmark/containerd-config.toml). It is used in [getContainerdProcess](https://github.com/awslabs/soci-snapshotter/blob/main/benchmark/benchmarkTests.go#L184) which is used for soci pulling images. So a similar config for stargz will be added in a future PR of adding the functionality of stargz pulling image.

vkuzniet: I'm still concerned with the former [soci_config.toml](https://github.com/awslabs/soci-snapshotter/blob/main/benchmark/soci_config.toml). With the proposed setup, it will be shared between stargz and soci snapshotters when running different benchmarks. The problem is that these configs are not 1:1 anymore, since they've diverged already. I'm afraid that one of the snapshotters might get misconfigured if we go this way.

Makefile
```diff
@@ -138,4 +138,4 @@ benchmarks-comp:
 
 benchmarks-stargz:
 	@echo "$@"
-	@cd benchmark/stargzTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/StargzTests . && sudo ../bin/StargzTests $(COMMIT) ../singleImage.csv 10
+	@cd benchmark/stargzTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/StargzTests . && sudo ../bin/StargzTests $(COMMIT) ../singleImage.csv 10 $(stargzBinary)
```
Kern--, line 141: Can we default the variable to the default install location? (`/usr/local/bin/containerd-stargz-grpc`)

hanyuel: Where does that "default" come from? I don't have it. The field is to pass in the ``containerd-stargz-grpc`` binary built by the stargz package.

Kern--: The default is the location that the stargz makefile will install the stargz snapshotter. It would be a nice convenience feature to be able to just run `make benchmarks-stargz` and have it run against the system installed stargz snapshotter. 

If we set the variable at the start of the makefile like this, we should be able to get both that default and the ability to pass it explicitly:
```
stargzBinary ?= /usr/local/bin/containerd-stargz-grpc
```

Side nit:
Is it idiomatic Make to name the variable with screaming snake case? (i.e. STARGZ_BINARY)

hanyuel: Requested change is added in the latest revision.

benchmark/benchmarkTests.go
hanyuel: Separated soci_config and stargz_config in the latest revision.

Kern-- (approved):

vkuzniet (approved):

PR #207, opened 2022-11-30
Run "make benchmarks" on GitHub Actions

kzys:
While we might not be able to trust the numbers, this will help us to make sure that the target doesn't have any build errors.

*Issue #, if available:*

*Description of changes:*

*Testing performed:*

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,3 +44,11 @@ jobs:
         with:
           go-version: '>=1.18.0'
       - run: make integration
+  benchmarks:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '>=1.18.0'
+      - run: make benchmarks


kzys:
We actually have build errors right now.

@sbuckfelder or @hanyuel - Would it be fixed by one of your PRs?

kzys:
> Also, I think to do this we need to investigate if we can use a public repo (I think Docker might support referrers now?)

It will be, according to https://www.docker.com/blog/announcing-docker-hub-oci-artifacts-support/

> But, our mission doesn’t end here! We’re continually working to bolster our OCI support. While the OCI Artifact Specification [is considered a release candidate](https://github.com/opencontainers/image-spec/commit/19a74bcb54ba211005a68d85c6b359c2947721ce), full Docker Hub support for OCI Reference Types and the accompanying [Referrers API](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#api) is on the horizon. Stay tuned for upcoming enhancements, improved repo organization, and more.

We also could consider https://github.com/opencontainers/distribution-spec/blob/main/spec.md#backwards-compatibility. oras-go is supporting that apparently (see https://github.com/oras-project/oras-go/issues/354).

sbuckfelder:
So just to confirm we are agreed that this PR is on hold until we can get images and soci artifacts in a public repo?

vkuzniet:
#239

PR #208, opened 2022-11-30
Fixes broken build

sbuckfelder:

*Issue #, if available:*

*Description of changes:*
This fixes the broken build.  The build was broken for two reasons: 
1) the new stargz test did not implement the new logs (thanks @kzys  for finding this) 
2) if an output dir was not present it would error out (thanks @hanyuel for finding this)

*Testing performed:*
- make
- make check
- make benchmarks
- make benchmarks-stargz

Made sure to also check if this works with or without the output dir added

### DIFF
--- a/benchmark/comparisonTest/main.go
+++ b/benchmark/comparisonTest/main.go
@@ -52,6 +52,11 @@ func main() {
 		panic(errMsg)
 	}
 
+	err = os.Mkdir(outputDir, 0755)
+	if err != nil && !os.IsExist(err) {
+		panic(err)
+	}
+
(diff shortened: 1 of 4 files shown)


benchmark/comparisonTest/main.go
```diff
@@ -52,6 +52,11 @@ func main() {
 		panic(errMsg)
 	}
 
+	err = os.Mkdir(outputDir, 0755)
```
sbuckfelder, line 55: 755 is the default permissions if you would just `mkdir`

hanyuel (approved):

PR #212, opened 2022-11-30
Remove unnecessary buffer in fs.reader

hanyuel:
*Issue #, if available:*

*Description of changes:*
Remove unnecessary temporary buffer ``content`` in fs.reader when reading file contents.

*Testing performed:*
``make check`` && ``make test`` && ``make integration``

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -75,9 +75,8 @@ import (
 )
 
 const (
-	defaultFuseTimeout    = time.Second
-	defaultMaxConcurrency = 2
-	fusermountBin         = "fusermount"
+	defaultFuseTimeout = time.Second
+	fusermountBin      = "fusermount"
 )
 
 var (
--- a/fs/reader/reader.go
+++ b/fs/reader/reader.go
@@ -209,6 +209,9 @@ type file struct {
 
 // ReadAt reads the file when the file is requested by the container
 func (sf *file) ReadAt(p []byte, offset int64) (int, error) {
+	if len(p) == 0 {
+		return 0, nil
+	}
 	uncompFileSize := sf.fr.GetUncompressedFileSize()
 	if soci.FileSize(offset) >= uncompFileSize {
 		return 0, io.EOF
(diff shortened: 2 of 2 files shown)


fs/fs.go
```diff
@@ -75,9 +75,8 @@ import (
 )
 
 const (
-	defaultFuseTimeout    = time.Second
-	defaultMaxConcurrency = 2
```
hanyuel, line 79: This change is because ``defaultMaxConcurrency`` is unused so it fails the lint style check.

vkuzniet:
Can you please fix the linter error?

fs/reader/reader.go
```diff
@@ -227,11 +227,10 @@ func (sf *file) ReadAt(p []byte, offset int64) (int, error) {
 	commonmetrics.IncOperationCount(commonmetrics.OnDemandRemoteRegistryFetchCount, sf.gr.layerSha) // increment the number of on demand file fetches from remote registry
 	sf.gr.setLastReadTime(time.Now())
 
-	contents, err := io.ReadAll(r)
+	n, err := io.ReadFull(r, p[0:expectedSize])
 	if err != nil && err != io.EOF {
 		return 0, err
 	}
-	n := copy(p, contents)
 	if soci.FileSize(n) != expectedSize {
```
Kern--, line 234: Can we combine this with the previous error case?

The possible returns from `io.ReadFull` are (let `buflen == len(p[0:expectedSize]`):
1) n == buflen, err == nil
2) n == 0 err == `io.EOF`
3) 0 < n < buflen, err == `io.ErrUnexpectedEOF`
4) n == ?, err == other read error

Can we just combine these into:
```
n, err := io.ReadFull(r, p[0:expectedSize])
if err != nil {
        return 0, fmt.Errorf("unexpected copied data size for on-demand fetch. read = %d, expected = %d: %w", n, expectedSize, err)
}
```

Kern--: Fair point - could we guard for that with a precondition?
```
if len(p) == 0 {
    return 0, nil
}
```

rdpsin (dismissed):

fs/reader/reader.go
hanyuel: Done

rdpsin (approved):

Kern-- (approved):

sbuckfelder (approved):

PR #214, opened 2022-12-02
Implement background fetcher

rdpsin:
This PR implements a global, rate-limited background fetcher. 

Features:

- `backgroundFetcher.silencePeriod`: Whenever a new layer comes in, all background fetches are suspended for `silencePeriod` (configurable)

- `backgroundFetcher.fetchPeriod`: At most 1 background fetch will be initiated every `fetchPeriod` (configurable)

### DIFF
--- a/fs/backgroundfetcher/background_fetcher.go
+++ b/fs/backgroundfetcher/background_fetcher.go
@@ -1,0 +1,170 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package backgroundfetcher
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/containerd/containerd/log"
+	"golang.org/x/time/rate"
+)
+
+type Option func(*BackgroundFetcher) error
+
+func WithSilencePeriod(period time.Duration) Option {
+	return func(bf *BackgroundFetcher) error {
+		bf.silencePeriod = period
+		return nil
+	}
+}
+
+func WithFetchPeriod(period time.Duration) Option {
+	return func(bf *BackgroundFetcher) error {
+		bf.fetchPeriod = period
+		return nil
+	}
+}
+
+func WithMaxQueueSize(size int) Option {
+	return func(bf *BackgroundFetcher) error {
+		bf.maxQueueSize = size
+		return nil
+	}
+}
+
+// An interface for a type to "pause" the background fetcher.
+// Useful for mocking in unit tests.
+type pauser interface {
+	pause(time.Duration)
+}
+
+type defaultPauser struct{}
+
+func (p defaultPauser) pause(d time.Duration) {
+	time.Sleep(d)
+}
+
+// A backgroundFetcher is responsible for fetching spans from layers
+// in the background.
+type BackgroundFetcher struct {
+	silencePeriod time.Duration
+	fetchPeriod   time.Duration
+	maxQueueSize  int
+
+	rateLimiter *rate.Limiter
+
+	bfPauser pauser
+
+	// All span managers are added to the channel and picked up in Run().
+	// If a span manager is still able to fetch, it is reinserted into the chanel.
+	workQueue chan Resolver
+	closeChan chan struct{}
+	pauseChan chan struct{}
+}
+
+func NewBackgroundFetcher(opts ...Option) (*BackgroundFetcher, error) {
+	bf := new(BackgroundFetcher)
+	for _, o := range opts {
+		if err := o(bf); err != nil {
+			return nil, err
+		}
+	}
+	// Create a rate-limiter that will fetch every bf.fetchPeriod
+	// with a burst capacity of 1 (i.e., it will never invoke more than 1 bg-fetch
+	// within bf.fetchPeriod)
+	bf.rateLimiter = rate.NewLimiter(rate.Every(bf.fetchPeriod), 1)
+	bf.workQueue = make(chan Resolver, bf.maxQueueSize)
+	bf.closeChan = make(chan struct{})
+	bf.pauseChan = make(chan struct{})
+
+	if bf.bfPauser == nil {
+		bf.bfPauser = defaultPauser{}
+	}
+
+	return bf, nil
+}
+
+// Add a new Resolver to be background fetched from.
+// Sends the resolver through the channel, which will be received in the Run() method.
+func (bf *BackgroundFetcher) Add(resolver Resolver) {
+	bf.workQueue <- resolver
+}
+
+func (bf *BackgroundFetcher) Close() error {
+	bf.closeChan <- struct{}{}
+	return nil
+}
+
+// Sends a signal to pause the background fetcher for silencePeriod on the next iteration.
+func (bf *BackgroundFetcher) Pause() {
+	bf.pauseChan <- struct{}{}
+}
+
+func (bf *BackgroundFetcher) pause(ctx context.Context) {
+	needPause := false
+loop:
+	for {
+		select {
+		// A new image has been mounted. Need to pause the background fetcher
+		case <-bf.pauseChan:
+			needPause = true
+		default:
+			break loop
+		}
+	}
+	if needPause {
+		log.G(ctx).WithField("silencePeriod", bf.silencePeriod).Debug("new image mounted, pausing the background fetcher for silence period")
+		bf.bfPauser.pause(bf.silencePeriod)
+	}
+}
+
+func (bf *BackgroundFetcher) Run(ctx context.Context) error {
+	for {
+		// Pause the background fetcher if necessary.
+		bf.pause(ctx)
+
+		select {
+		case <-bf.closeChan:
+			return nil
+		case <-ctx.Done():
+			return nil
+		default:
+		}
+
+		select {
+		case lr := <-bf.workQueue:
+			if lr.Closed() {
+				continue
+			}
+			go func() {
+				more, err := lr.Resolve(ctx)
+				if more {
+					bf.workQueue <- lr
+				} else if err != nil {
+					log.G(ctx).WithError(err).Warn("error trying to resolve layer, removing it from the queue")
+				}
+			}()
+		default:
+		}
+
+		if err := bf.rateLimiter.Wait(ctx); err != nil {
+			return fmt.Errorf("background fetch: error while waiting for rate limiter: %w", err)
+		}
+	}
+}
--- a/fs/backgroundfetcher/background_fetcher_test.go
+++ b/fs/backgroundfetcher/background_fetcher_test.go
@@ -1,0 +1,194 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package backgroundfetcher
+
+import (
+	"compress/gzip"
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/awslabs/soci-snapshotter/cache"
+	spanmanager "github.com/awslabs/soci-snapshotter/fs/span-manager"
+	"github.com/awslabs/soci-snapshotter/util/testutil"
+	"github.com/awslabs/soci-snapshotter/ztoc"
+	"github.com/opencontainers/go-digest"
+)
+
+func withPauser(p pauser) Option {
+	return func(bf *BackgroundFetcher) error {
+		bf.bfPauser = p
+		return nil
+	}
+}
+
+type countingPauser struct {
+	mu    sync.Mutex
+	count int
+}
+
+func (c *countingPauser) pause(time.Duration) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.count++
+}
+
+func TestBackgroundFetcherPause(t *testing.T) {
+	p := &countingPauser{}
+	bf, err := NewBackgroundFetcher(WithSilencePeriod(0), withPauser(p))
+	if err != nil {
+		t.Fatal(err)
+	}
+	go bf.Run(context.Background())
+	defer bf.Close()
+	bf.Pause()
+
+	time.Sleep(10 * time.Millisecond)
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	if p.count != 1 {
+		t.Fatalf("unexpected pause count; expected 1, got %v", p.count)
+	}
+}
+
+func TestBackgroundFetcherRun(t *testing.T) {
+	testCases := []struct {
+		name     string
+		waitTime time.Duration
+		entries  [][]testutil.TarEntry
+	}{
+		{
+			name:     "background fetcher fetches all data for single span manager",
+			waitTime: 1 * time.Second,
+			entries: [][]testutil.TarEntry{
+				{
+					testutil.File("test", string(genRandomByteData(10000000))),
+				},
+			},
+		},
+		{
+			name:     "background fetcher fetches all data for multiple span managers",
+			waitTime: 1 * time.Second,
+			entries: [][]testutil.TarEntry{
+				{
+					testutil.File("test1", string(genRandomByteData(10000000))),
+				},
+				{
+					testutil.File("test2", string(genRandomByteData(20000000))),
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+
+			type testInfo struct {
+				sm    *spanmanager.SpanManager
+				cache *countingCache
+				ztoc  *ztoc.Ztoc
+			}
+
+			var infos []testInfo
+			for _, entries := range tc.entries {
+				ztoc, sr, err := ztoc.BuildZtocReader(entries, gzip.DefaultCompression, 1000000)
+				if err != nil {
+					t.Fatalf("error building span manager and section reader: %v", err)
+				}
+				cache := &countingCache{}
+				sm := spanmanager.New(ztoc, sr, cache)
+				infos = append(infos, testInfo{sm, cache, ztoc})
+			}
+
+			bf, err := NewBackgroundFetcher(WithFetchPeriod(0))
+			if err != nil {
+				t.Fatalf("unable to construct background fetcher: %v", err)
+			}
+
+			go bf.Run(context.Background())
+			defer bf.Close()
+
+			for _, info := range infos {
+				bf.Add(NewSequentialResolver(digest.FromString("test"), info.sm))
+			}
+
+			time.Sleep(tc.waitTime)
+
+			for _, info := range infos {
+				if info.cache.addCount != int(info.ztoc.CompressionInfo.MaxSpanID)+1 {
+					t.Fatalf("unexpected number of adds to cache; expected %d, got %d", info.ztoc.CompressionInfo.MaxSpanID+1, info.cache.addCount)
+				}
+
+				// The first 10 bytes of a compressed gzip archive is the gzip header.
+				// We don't fetch it when lazy-loading; therefore, subtracting 10 from the total compressed archive size.
+				compressedSize := info.ztoc.CompressedArchiveSize - 10
+				if info.cache.addBytes != int64(compressedSize) {
+					t.Fatalf("unexpected number of bytes added to cache; expected %d, got %d", compressedSize, info.cache.addBytes)
+				}
+			}
+		})
+	}
+}
+
+// countingCache is an implementation of cache.BlobCache
+// which counts the number of times `cache.Add` was invoked
+// and the number of bytes added to the cache.
+// All writes to the cache succeed.
+type countingCache struct {
+	addCount int
+	addBytes int64
+}
+
+var _ cache.BlobCache = &countingCache{}
+
+func (c *countingCache) Add(key string, opts ...cache.Option) (cache.Writer, error) {
+	return &countingWriter{c}, nil
+}
+
+func (c *countingCache) Get(key string, opts ...cache.Option) (cache.Reader, error) {
+	return nil, nil
+}
+
+func (c *countingCache) Close() error {
+	return nil
+}
+
+type countingWriter struct {
+	cache *countingCache
+}
+
+var _ cache.Writer = &countingWriter{}
+
+func (c *countingWriter) Write(p []byte) (int, error) {
+	c.cache.addBytes += int64(len(p))
+	c.cache.addCount++
+	return len(p), nil
+}
+
+func (c *countingWriter) Close() error {
+	return nil
+}
+
+func (c *countingWriter) Commit() error {
+	return nil
+}
+
+func (c *countingWriter) Abort() error {
+	return nil
+}
--- a/fs/backgroundfetcher/resolver.go
+++ b/fs/backgroundfetcher/resolver.go
@@ -1,0 +1,96 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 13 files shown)


fs/background-fetcher/background_fetcher.go
```diff
@@ -0,0 +1,162 @@
+/*
```
kzys, line 1: How about moving that to fs/backgroundfetcher/background_fetcher.go or fs/bakground_fetcher.go?

- Go package names are generally the same as the file hierarchy even that have multiple words (see https://github.com/golang/go/tree/master/src/index/suffixarray)
- https://dave.cheney.net/practical-go/presentations/qcon-china.html#_consider_fewer_larger_packages

fs/background-fetcher/background_fetcher.go
```diff
@@ -0,0 +1,162 @@
...
+
+type Option func(*BackgroundFetcher) error
+
+const (
+	// Amount of time the background fetcher will wait once a new layer comes in
+	// before (re)starting fetches.
+	defaultSilencePeriod = 5 * time.Second
+
+	// The minimum size of data to be fetched in a single fetch.
+	defaultMinFetchSize = 1 << 22 // 4 MiB
```
kzys, line 45: How about defining mib as a constant? I cannot recall 22 bits shift is 2 MiB :)

fs/background-fetcher/background_fetcher.go
```diff
@@ -0,0 +1,162 @@
...
+	return &BackgroundFetcher{
+		minFetchSize:  defaultMinFetchSize,
+		silencePeriod: defaultSilencePeriod,
+		fetchPeriod:   defaultFetchPeriod,
+		waitQueue:     make(chan *layerResolver),
+		closeChan:     make(chan struct{}),
+	}
+}
+
+// Adds a new span manager to be background fetched from.
```
kzys, line 117: Adds -> Add

Let's follow https://go.dev/doc/comment

fs/background-fetcher/background_fetcher.go
```diff
@@ -0,0 +1,162 @@
...
+		// If any new span managers have come in, wait for silencePeriod.
+		if bf.timer != nil {
+			<-bf.timer.C
+		}
+
+		if err := bf.rateLimiter.Wait(ctx); err != nil {
+			return fmt.Errorf("background fetch: error while waiting for rate limiter: %w", err)
+		}
+
+		select {
```
kzys, line 146: You should check ctx.Done() as well to cancel this when the context is cancelled.

fs/background-fetcher/background_fetcher.go
hanyuel: I think it makes sense to have  a ``backgroundfetcher`` package as we expect to see the fetcher implementation to be extended in the future.

fs/background-fetcher/background_fetcher.go
```diff
@@ -0,0 +1,162 @@
...
+	layerDigest     digest.Digest
+	nextSpanFetchID soci.SpanID
+}
+
+type Option func(*BackgroundFetcher) error
+
+const (
+	// Amount of time the background fetcher will wait once a new layer comes in
+	// before (re)starting fetches.
+	defaultSilencePeriod = 5 * time.Second
```
hanyuel, line 42: With the silence period being set to 5 seconds, now I understand more why the background fetcher is more about availability than performance.

fs/background-fetcher/background_fetcher.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func NewBackgroundFetcher(opts ...Option) (*BackgroundFetcher, error) {
+	bf := newWithDefaults()
+	for _, o := range opts {
+		if err := o(bf); err != nil {
+			return nil, err
+		}
+	}
+	// Create a rate-limited that will fetch every bf.fetchPeriod
```
hanyuel, line 100: ``Create a rate-limited`` __?

fs/background-fetcher/background_fetcher.go
```diff
@@ -0,0 +1,162 @@
...
+	// with a burst capacity of 1 (i.e., it will never invoke more than 1 bg-fetch
```
hanyuel, line 101: We probably want ``burst capacity`` to be configurable for the bg fetcher as well.

fs/background-fetcher/background_fetcher.go
```diff
@@ -0,0 +1,162 @@
...
+func WithFetchPeriod(period time.Duration) Option {
+	return func(bf *BackgroundFetcher) error {
+		bf.fetchPeriod = period
+		return nil
+	}
+}
+
+// A backgroundFetcher is responsible for fetching spans from layers
+// in the background.
+type BackgroundFetcher struct {
```
Kern--, line 75: Is this the right abstraction? The BackgroundFetcher handles both work queueing, but also the state of the work to be done. Could we separate it into a WorkQueue/JobQueue and a job that actually knows how to do a single unit of background fetching? Is there an existing implementation of a queue that we could use?

fs/background-fetcher/background_fetcher.go
rdpsin: Oops. Friday morning blues.

fs/background-fetcher/background_fetcher.go
rdpsin: FYI, that default is completely arbitrary.

fs/background-fetcher/background_fetcher.go
rdpsin: Ack.

fs/background-fetcher/background_fetcher.go
rdpsin: Ack.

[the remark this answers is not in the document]
rdpsin: Yes; the idea is that maybe can plumb this value down the fetch stack in the future to fetch a minimum specified size. I can remove it for now if we think it's a red herring.

fs/span-manager/span_manager.go
```diff
@@ -162,6 +163,49 @@ func (m *SpanManager) buildAllSpans() {
...
+	}
+
+	m.addSpanToCache(fmt.Sprintf("%d", spanID), compressedBuf)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	return s.setState(fetched)
+}
+
 func (m *SpanManager) ResolveSpan(spanID soci.SpanID, r *io.SectionReader) error {
```
hanyuel, line 209: This function can be removed.

fs/span-manager/span_manager.go
```diff
@@ -162,6 +163,49 @@ func (m *SpanManager) buildAllSpans() {
...
+
+	if n != len(compressedBuf) {
+		return fmt.Errorf("unexpected data size for reading compressed span. read = %d, expected = %d", n, len(compressedBuf))
+	}
+
+	if err := m.verifySpanContents(compressedBuf, spanID); err != nil {
+		return err
+	}
+
+	m.addSpanToCache(fmt.Sprintf("%d", spanID), compressedBuf)
```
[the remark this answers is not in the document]
rdpsin: Hmm, I think you're right. I was under the impression that `BlobCache` internally uses a lock. Seems I was wrong. Will fix it.

fs/background-fetcher/background_fetcher.go
rdpsin: Hmm, does it make sense to do something like 


```
type Resolver interface {
    Resolve(ctx context.Context) error 
}

func (lr *layerResolve) Resolve() error {
    // fetch from data
}

...

type BackgroundFetcher struct {
     ...
     workQueue chan Resolver
}


func (bf *BackgroundFetcher) Run() error {
   ...
   select {
        ...
        case lr := <- workQueue:
                  err := lr.ResolveNext()
                  if err == nil {
                      workQueue <- lr
                   }
       }
 }
```

Kern--: I think that makes sense. It could make it easier for us to, for example, add an optional LOD for a layer to the bgfetcher later.

fs/span-manager/span_manager.go
rdpsin: There are a bunch of tests that rely on that method. I don't want to remove it just yet, will remove in a future PR.

fs/background-fetcher/background_fetcher.go
rdpsin: Done.

fs/background-fetcher/background_fetcher.go
rdpsin: Renamed package to `backgroundfetcher`. We also have a `spanmanager` package in a `span-manager` directory that we should update in a future PR.

fs/span-manager/span_manager.go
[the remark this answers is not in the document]
rdpsin: Fixed.

fs/background-fetcher/background_fetcher.go
rdpsin: Done.

fs/backgroundfetcher/background_fetcher.go
```diff
@@ -0,0 +1,152 @@
...
+	bf := newWithDefaults()
+	for _, o := range opts {
+		if err := o(bf); err != nil {
+			return nil, err
+		}
+	}
+	// Create a rate-limiter that will fetch every bf.fetchPeriod
+	// with a burst capacity of 1 (i.e., it will never invoke more than 1 bg-fetch
+	// within bf.fetchPeriod)
+	bf.rateLimiter = rate.NewLimiter(rate.Every(bf.fetchPeriod), 1)
```
kzys, line 93: How about setting bf.timer here as well? Then you don't have to have `timer != nil` checks below.

fs/backgroundfetcher/background_fetcher.go
```diff
@@ -0,0 +1,152 @@
...
+	return nil
+}
+
+func (bf *BackgroundFetcher) Run(ctx context.Context) error {
+	for {
+
+		// If any new span managers have come in, wait for silencePeriod.
+		if bf.timer != nil {
+			<-bf.timer.C
+		}
```
kzys, line 130: Is it possible to put this channel read in the select below? Ideally speaking, the passed context's cancellation should be honored while waiting this timer.

fs/backgroundfetcher/background_fetcher.go
```diff
@@ -0,0 +1,152 @@
...
+
+		select {
+		case <-bf.closeChan:
+			return nil
+		case <-ctx.Done():
+			return nil
+		case lr := <-bf.waitQueue:
+			go func() {
+				more, err := lr.Resolve(ctx)
+				// if error, append it to the waitQueue to retry
```
kzys, line 144: Would it retry that forever?

fs/span-manager/span_manager.go
```diff
@@ -162,6 +163,49 @@ func (m *SpanManager) buildAllSpans() {
 	}
 }
 
+// FetchSpanBackground invokes the reader to fetch the span in the background and cache it.
+// It is invoked by the backgroundFetcher.
+func (m *SpanManager) FetchSpanInBackground(spanID soci.SpanID) error {
```
kzys, line 168: Doesn't it block until the span is fetched? We don't have to mention "Background" if so.

kzys (changes requested):

fs/backgroundfetcher/background_fetcher.go
rdpsin: Tbh, I wasn't sure what to do in this case? Retrying forever seems wasteful. Should we define a max error count?  And retry until it succeeds or error count reaches max error count? Something like this:

```
go func() {
   for i := 0; i < maxErrorCount; i++ {
       more, err := lr.Resolve(ctx)
       if err != nil {
           continue
        } else {
            break
       }
    }
}
```

kzys: Looks good to me. We ideally should honor some HTTP status codes (e.g. don't retry in some cases) and headers (e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After), but we can start from just limiting the # of retries.

rdpsin: We already use [retryablehttp](https://github.com/hashicorp/go-retryablehttp). https://github.com/awslabs/soci-snapshotter/blob/main/fs/remote/resolver.go#L184-L201

fs/span-manager/span_manager.go
rdpsin: Renamed.

fs/backgroundfetcher/background_fetcher.go
rdpsin: Done.

fs/backgroundfetcher/background_fetcher.go
rdpsin: Done.

fs/backgroundfetcher/background_fetcher.go
```diff
@@ -0,0 +1,165 @@
...
+			go func() {
+				for i := 0; i < bf.maxRetryCount; i++ {
+					more, err := lr.Resolve(ctx)
+					// if error, append it to the waitQueue to retry
+					if more {
+						bf.waitQueue <- lr
+						break
+					} else if err != nil {
+						continue
+					}
```
djdongjin, line 154: The `else continue` part seems redundant.

rdpsin: There can 3 case:

- true, nil: add to the queue again for next span fetch and break out
- false, nil: don't add to the queue and break out
- false, err: retry

fs/backgroundfetcher/resolver.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (lr *sequentialLayerResolver) Resolve(ctx context.Context) (bool, error) {
+	log.G(ctx).WithTime(time.Now()).WithFields(logrus.Fields{
+		"layer":  lr.layerDigest,
+		"spanId": lr.nextSpanFetchID,
+	}).Debug("fetching span in background")
+
+	err := lr.FetchSingleSpan(lr.nextSpanFetchID)
+	if err == nil || errors.Is(err, sm.ErrSpanStateNotUnrequested) {
```
hanyuel, line 64: So why do we have ``FetchSingleSpan`` return  an error``ErrSpanStateNotUnrequested``? We're not handling it as an error here

fs/fs.go
```diff
@@ -181,6 +188,7 @@ func NewFilesystem(root string, cfg config.Config, opts ...Option) (_ snapshot.F
 		entryTimeout:        entryTimeout,
 		negativeTimeout:     negativeTimeout,
 		orasStore:           store,
+		bgFetcher:           bgFetcher,
```
hanyuel, line 191: ``bgFetcher`` is part of ``layer.Resolver``. Why do we need ``bgFetcher`` as part of ``fileSystem``? I don't see ``filesystem.bgFetcher`` is invoked anywhere

fs/span-manager/span_manager.go
```diff
@@ -162,6 +163,49 @@ func (m *SpanManager) buildAllSpans() {
...
+	if n != len(compressedBuf) {
+		return fmt.Errorf("unexpected data size for reading compressed span. read = %d, expected = %d", n, len(compressedBuf))
+	}
+
+	if err := m.verifySpanContents(compressedBuf, spanID); err != nil {
+		return err
+	}
+
+	s.mu.Lock()
+	m.addSpanToCache(fmt.Sprintf("%d", spanID), compressedBuf)
```
Kern--, line 203: We should verify that state is still `requested` before doing this or else we have a race where we could replace previously uncompressed data with compressed data.

fs/layer/layer.go
```diff
@@ -325,6 +330,8 @@ func (r *Resolver) Resolve(ctx context.Context, hosts source.RegistryHosts, refs
 	log.G(ctx).Debugf("[Resolver.Resolve]Initialized metadata store for layer sha=%v", desc.Digest)
 
 	spanManager := spanmanager.New(ztoc, sr, spanCache, cache.Direct())
+	bgLayerResolver := backgroundfetcher.NewSequentialResolver(desc.Digest, spanManager)
+	r.bgFetcher.Add(bgLayerResolver)
```
Kern--, line 334: There should probably be someway to unregister a `layerResolver` (e.g. because the container exited). We already call `layer.Done()` when the layer is unmounted so we should probably hook the background layer resolver into that.

fs/backgroundfetcher/background_fetcher.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (bf *BackgroundFetcher) Run(ctx context.Context) error {
+	for {
+		select {
+		case <-bf.closeChan:
+			return nil
+		case <-ctx.Done():
+			return nil
+		case <-bf.timer.C:
```
Kern--, line 143: Does this do the right thing? From my understanding of `select`, this is a case with an empty body, so it's effectively ignoring the silence period and will just select from the `waitQueue` whenever it's available.

fs/backgroundfetcher/background_fetcher.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// All span managers are added to the channel and picked up in Run().
+	// If a span manager is still able to fetch, it is reinserted into the chanel.
+	waitQueue chan Resolver
+
+	closeChan chan struct{}
+
+	// The timer that the bg-fetcher will wait on until it is ready to (re)start
+	// background fetches.
+	timer *time.Timer
```
Kern--, line 91: nit: can we call this something more descriptive like `silenceTimer`?

fs/backgroundfetcher/background_fetcher.go
rdpsin: What should the behavior be then? `retryablehttp` with retry a fixed number of times. After 'n' failed retries, should we just remove the layer from being background fetched?

kzys: I'm inclined to do so. The availability gain by adding this retry wouldn't outweigh the risk of overloading a container registry. We should start from the specific scenarios that retryablehttp couldn't cover.

fs/fs.go
rdpsin: Not yet, but maybe in the future (e.g., dynamically adjusting rate limit).

fs/layer/layer.go
rdpsin: Makes sense. I don't think `Done()` is the right place to do so, because it's called whenever a ref count in the lrucache is decreased. This means that layer can still be mounted, but we won't do background fetch anymore. Example: https://github.com/awslabs/soci-snapshotter/blob/192f026b6267a08241cfe7e9e9f1346b81a49ccc/fs/fs.go#L384-L386

Instead, decided to close the resolver in `layer.close()`.

fs/backgroundfetcher/background_fetcher_test.go
```diff
@@ -0,0 +1,194 @@
...
+	c.count++
+}
+
+func TestBackgroundFetcherPause(t *testing.T) {
+	p := &countingPauser{}
+	bf, err := NewBackgroundFetcher(WithSilencePeriod(0), withPauser(p))
+	if err != nil {
+		t.Fatal(err)
+	}
+	go bf.Run(context.Background())
```
kzys, line 57: How about passing a cancel-able context and cancel that at the end of the function? I'd like to make sure the goroutine is killed at the end of this test.

fs/backgroundfetcher/background_fetcher_test.go
```diff
@@ -0,0 +1,194 @@
...
+				sm := spanmanager.New(ztoc, sr, cache)
+				infos = append(infos, testInfo{sm, cache, ztoc})
+			}
+
+			bf, err := NewBackgroundFetcher(WithFetchPeriod(0))
+			if err != nil {
+				t.Fatalf("unable to construct background fetcher: %v", err)
+			}
+
+			go bf.Run(context.Background())
```
kzys, line 124: Same as above. Let's make sure all goroutines are cancelled at the end.

fs/backgroundfetcher/resolver.go
```diff
@@ -0,0 +1,100 @@
...
+		return true, nil
+	}
+	if errors.Is(err, sm.ErrExceedMaxSpan) {
+		return false, nil
+	}
+
+	log.G(ctx).WithError(err).WithFields(logrus.Fields{
+		"layer":  lr.layerDigest,
+		"spanId": lr.nextSpanFetchID,
+	}).Warn("error trying to background fetch span")
```
kzys, line 97: Wouldn't it be double-logged? How about including this information to the returning error?

fs/span-manager/span_manager.go
```diff
@@ -162,6 +162,54 @@ func (m *SpanManager) buildAllSpans() {
 	}
 }
 
+// FetchSingleSpan invokes the reader to fetch the span in the background and cache it.
+// It is invoked by the BackgroundFetcher.
+func (m *SpanManager) FetchSingleSpan(spanID soci.SpanID) error {
```
kzys, line 167: Wouldn't it block until the span is fetched?

fs/backgroundfetcher/background_fetcher.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+type Option func(*BackgroundFetcher) error
+
+const (
+
+	// Specifies the maximum size of the work queue i.e., the maximum number
+	// of span managers that can be queued. In case of overflow, the `Add` call
+	// will block until a span manager is removed from the workqueue.
+	defaultWorkQueueSize = 100
```
kzys, line 35: Does it have metrics/logs to tune the work queue size? Can we expose the queue length somewhere?

fs/span-manager/span_manager.go
rdpsin: Yes, but it's called in a separate goroutine. The bg-fetcher won't be blocked on this call.

fs/backgroundfetcher/background_fetcher.go
rdpsin: Sure, I can make this configurable.

fs/backgroundfetcher/background_fetcher_test.go
rdpsin: There is a `defer bf.Close()` call. Is there a benefit to having a cancellable context?

fs/fs.go
```diff
@@ -149,12 +159,36 @@ func NewFilesystem(root string, cfg config.Config, opts ...Option) (_ snapshot.F
...
+		log.G(context.Background()).WithFields(logrus.Fields{
+			"fetchPeriod":   bgFetchPeriod,
+			"silencePeriod": bgSilencePeriod,
+		}).Info("constructing background fetcher")
+
+		bgFetcher, err = bf.NewBackgroundFetcher(bf.WithFetchPeriod(bgFetchPeriod), bf.WithSilencePeriod(bgSilencePeriod))
+		if err != nil {
+			return nil, fmt.Errorf("cannot create background fetcher: %w", err)
+		}
+		go bgFetcher.Run(context.Background())
```
djdongjin, line 188: Where is `bgFetcher` closed when necessary?

fs/backgroundfetcher/background_fetcher_test.go
kzys: Oh I wasn't aware about that.

fs/backgroundfetcher/background_fetcher.go
```diff
@@ -0,0 +1,167 @@
...
+				} else if err != nil {
+					log.G(ctx).WithError(err).Warn("error trying to resolve layer, removing it from the queue")
+				}
+			}()
+		default:
+		}
+
+		if err := bf.rateLimiter.Wait(ctx); err != nil {
+			log.G(ctx).WithError(err).Warn("background fetcher: error while waiting for rate limiter")
+			return fmt.Errorf("background fetch: error while waiting for rate limiter: %w", err)
```
kzys, line 164: Generally speaking, it is better to do either logging (because the caller cannot catch the error) or returning an error (bacause the caller can catch the error), not both.

fs/backgroundfetcher/resolver.go
```diff
@@ -0,0 +1,102 @@
...
+	return &sequentialLayerResolver{
+		base: &base{
+			SpanManager: spanManager,
+			layerDigest: layerDigest,
+		},
+	}
+}
+
+func (lr *sequentialLayerResolver) Resolve(ctx context.Context) (bool, error) {
+	log.G(ctx).WithTime(time.Now()).WithFields(logrus.Fields{
```
kzys, line 81: The logger will log the time by default. Do you need WithTime()?

fs/backgroundfetcher/resolver.go
```diff
@@ -0,0 +1,102 @@
...
+		"layer":  lr.layerDigest,
+		"spanId": lr.nextSpanFetchID,
+	}).Debug("fetching span in background")
```
kzys, line 84: I'm inclined to not mention "background" in this. This function itself shouldn't aware whether this is used in background or not.

fs/fs.go
```diff
@@ -149,12 +164,45 @@ func NewFilesystem(root string, cfg config.Config, opts ...Option) (_ snapshot.F
...
+	}
+
+	bgSilencePeriod := time.Duration(cfg.BackgroundFetchConfig.SilencePeriodMSec) * time.Millisecond
+	if bgSilencePeriod == 0 {
+		bgSilencePeriod = defaultBgSilencePeriod
+	}
+
+	bgMaxQueueSize := cfg.BackgroundFetchConfig.MaxQueueSize
+	if bgMaxQueueSize == 0 {
+		bgMaxQueueSize = defaultBgMaxQueueSize
```
kzys, line 179: How do users figure out the appropriate queue size? Do we have enough logs/metrics to tell users that the size is good?

kzys (dismissed):
Non-blocking though.

fs/backgroundfetcher/background_fetcher.go
rdpsin: Yes, will submit a new revision that will address this.

fs/fs.go
rdpsin: Making it configurable.

vkuzniet (approved):

kzys (approved):

hanyuel (approved):

PR #216, opened 2022-12-02
Fixed image descriptor returned from csv. 

turan18:
Soci benchmarks on ecr pull were failing since the image descriptors returned after reading the test csv mismatched the ReadyLine and SociIndexManifest properties.

*Issue #, if available:*

*Description of changes:*

*Testing performed:*

### DIFF
--- a/benchmark/utils.go
+++ b/benchmark/utils.go
@@ -40,18 +40,15 @@ func GetImageListFromCsv(csvLoc string) ([]ImageDescriptor, error) {
 	}
 	var images []ImageDescriptor
 	for _, image := range csv {
-		if len(image) < 3 {
+		if len(image) < 4 {
 			return nil, errors.New("image input is not sufficient")
-		}
-		var sociIndexManifestRef string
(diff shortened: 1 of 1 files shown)


benchmark/utils.go
```diff
@@ -40,18 +40,15 @@ func GetImageListFromCsv(csvLoc string) ([]ImageDescriptor, error) {
 	}
 	var images []ImageDescriptor
 	for _, image := range csv {
-		if len(image) < 3 {
+		if len(image) < 4 {
 			return nil, errors.New("image input is not sufficient")
 		}
-		var sociIndexManifestRef string
```
hanyuel, line 46: This change was made specifically for stargz test which does _not_ have soci index

hanyuel (changes requested):
This PR will break ``make benchmarks-stargz``.

benchmark/utils.go
turan18: Seems like I was using an outdated test image set. Apologies and feel free to close.

PR #218, opened 2022-12-05
[BENCHMARKS] Added a parser to the benchmark framework to collect fuse file access patterns

turan18:
Added a parser to the benchmark framework to collect fuse file access patterns.

This can be enabled by first adding `log_fuse_operations = true` to the soci_config.toml found within the benchmark framework to enable file access info collection. You can then add the optional flag `-parseFileAccessPatterns` or `-parseFileAccessPatterns=true` to the start of the `make benchmarks`  target and this will enable the parsing and logging of access patterns. This will run once for each image in the test set and output the results into a directory `.../performanceTest/output/file_access_logs` wherein there will be a parsed log file for each image. Each log will be named `<image name>_access_patterns`.

The parsed file access logs for an image will look like:

```
{
 "ImageName": "Rabbitmq",
 "containerStartTime": "2022-12-08T18:01:47.723749763Z",
 "TotalOperationCounts": {
  "Getattr": 6036,
  "Getxattr": 12671,
  "Listxattr": 4,
  "Open": 1943,
  "Read": 1443,
  "ReadLink": 428,
  "Readdir": 4872
 },
 "operations": [
     ...
     {
      "operation": "Getattr",
      "path": "usr/lib/x86_64-linux-gnu/libc-2.31.so",
      "firstAccessTimeAfterStart": "317.842289ms",
      "count": 8
    },
  ...
  ]
}
```

The parser will collect every *unique* fuse operation (a specific operation on a specific path) and note the amount of times that the fuse operation was called. It will also log the the amount of time after the the container start event in which the operation was invoked. The log file will be sorted by `firstAccessTimeAfterStart` with the first fuse operation at the top of log. The total count for each fuse operation will also be logged.

*Note*: make sure the soci grpc server was started with `--log-level debug` set. 

*Note:* It is worth noting that there will be negative values for `firstAccessTimeAfterStart` for the first few fuse operations. This is because there are still fuse operations happening before the invocation of `task.Start`. 

*Issue #, if available:*

*Description of changes:*

*Testing performed:*

### DIFF
--- a/benchmark/framework/framework.go
+++ b/benchmark/framework/framework.go
@@ -46,7 +46,7 @@ type BenchmarkTestDriver struct {
 	NumberOfTests  int              `json:"numberOfTests"`
 	BeforeFunction func()           `json:"-"`
 	TestFunction   func(*testing.B) `json:"-"`
-	AfterFunction  func()           `json:"-"`
+	AfterFunction  func() error     `json:"-"`
 	TestsRun       int              `json:"-"`
 	TestTimes      []float64        `json:"testTimes"`
 	StdDev         float64          `json:"stdDev"`
@@ -77,7 +77,11 @@ func (frame *BenchmarkFramework) Run(ctx context.Context) {
 		}
 		testDriver.calculateStats()
 		if testDriver.AfterFunction != nil {
-			testDriver.AfterFunction()
+			err := testDriver.AfterFunction()
+			if err != nil {
+				fmt.Printf("After function error: %v\n", err)
+			}
+
 		}
 	}
 
--- a/benchmark/framework/parser/file_access.go
+++ b/benchmark/framework/parser/file_access.go
@@ -1,0 +1,177 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package bparser
+
+import (
+	"encoding/json"
+	"io"
+	"os"
+	"sort"
+	"strings"
+	"time"
+)
+
+var (
+	outputDir           = "./output/"
+	FileAccessDir       = outputDir + "file_access_logs/"
+	sociLogs            = outputDir + "soci-snapshotter-stderr"
+	containerdLogs      = outputDir + "containerd-stderr"
+	sociLogsFuseMessage = "FUSE operation"
+)
+
+type FileAccessPatterns struct {
+	ImageName           string         `json:"ImageName"`
+	ContainerStartTime  time.Time      `json:"containerStartTime"`
+	TotalOperationCount map[string]int `json:"TotalOperationCounts"`
+	Operations          []Operation    `json:"operations"`
+}
+
+type SociLog struct {
+	Msg string `json:"msg"`
+}
+type BaseOperation struct {
+	Level     string    `json:"level"`
+	Msg       string    `json:"msg"`
+	Operation string    `json:"operation"`
+	Path      string    `json:"path"`
+	Time      time.Time `json:"time"`
+}
+
+type Operation struct {
+	Operation                 string `json:"operation"`
+	Path                      string `json:"path"`
+	FirstAccessTimeAfterStart string `json:"firstAccessTimeAfterStart"`
+	Count                     int    `json:"count"`
+}
+
+func ParseFileAccesses(imageName string) error {
+	startTime, err := getTaskStartTime()
+	if err != nil {
+		return err
+	}
+	totalCounts := make(map[string]int)
+	fa := FileAccessPatterns{
+		ImageName:           imageName,
+		ContainerStartTime:  *startTime,
+		TotalOperationCount: totalCounts,
+	}
+	file, err := os.Open(sociLogs)
+	if err != nil {
+		return err
+	}
+
+	defer file.Close()
+
+	b, err := io.ReadAll(file)
+	if err != nil {
+		return err
+	}
+
+	logs := strings.Split(string(b), "\n")
+
+	m := make(map[string]Operation)
+	for _, log := range logs {
+		sociLog := SociLog{}
+		if log != "" {
+			err := json.Unmarshal([]byte(log), &sociLog)
+			if err != nil {
+				return err
+			}
+
+		}
+		if sociLog.Msg == sociLogsFuseMessage {
+			var tempOperation BaseOperation
+			err := json.Unmarshal([]byte(log), &tempOperation)
+			if err != nil {
+				return err
+			}
+			op := tempOperation.Operation + tempOperation.Path
+
+			val, ok := m[op]
+
+			if ok {
+				val.Count = val.Count + 1
+				m[op] = val
+
+			} else {
+				m[op] = Operation{
+					Operation:                 tempOperation.Operation,
+					Path:                      tempOperation.Path,
+					FirstAccessTimeAfterStart: tempOperation.Time.Sub(fa.ContainerStartTime).String(),
+					Count:                     1,
+				}
+			}
+			count, ok := fa.TotalOperationCount[tempOperation.Operation]
+			if ok {
+				fa.TotalOperationCount[tempOperation.Operation] = count + 1
+			} else {
+				fa.TotalOperationCount[tempOperation.Operation] = 1
+			}
+		}
+	}
+	keys := make([]string, 0, len(m))
+	for key := range m {
+		keys = append(keys, key)
+	}
+	sort.SliceStable(keys, func(i, j int) bool {
+		t1, _ := time.ParseDuration(m[keys[i]].FirstAccessTimeAfterStart)
+		t2, _ := time.ParseDuration(m[keys[j]].FirstAccessTimeAfterStart)
+		return t1 < t2
+	})
+
+	for _, key := range keys {
+		fa.Operations = append(fa.Operations, m[key])
+	}
+	json, err := json.MarshalIndent(fa, "", " ")
+	if err != nil {
+		return err
+	}
+	imageFileAccessLogPath := FileAccessDir + imageName + "_access_patterns"
+	err = os.WriteFile(imageFileAccessLogPath, json, 0644)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getTaskStartTime() (*time.Time, error) {
+	var taskStartTime time.Time
+	file, err := os.Open(containerdLogs)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	data, err := io.ReadAll(file)
+	if err != nil {
+		return nil, err
+	}
+	logs := strings.Split(string(data), "\n")
+	for i := len(logs) - 1; i >= 0; i-- {
+		log := logs[i]
+		if strings.Contains(log, "/tasks/start") {
+			l := strings.Split(log, " ")
+			temp := strings.ReplaceAll(l[0], "time=", "")
+			taskStartTime, err = time.Parse(time.RFC3339, temp[1:len(temp)-1])
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+	return &taskStartTime, nil
+}
--- a/benchmark/performanceTest/main.go
+++ b/benchmark/performanceTest/main.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"os"
 	"strconv"
@@ -24,6 +25,7 @@ import (
 
 	"github.com/awslabs/soci-snapshotter/benchmark"
 	"github.com/awslabs/soci-snapshotter/benchmark/framework"
+	bparser "github.com/awslabs/soci-snapshotter/benchmark/framework/parser"
 )
 
 var (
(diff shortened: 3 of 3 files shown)


benchmark/framework/containerd_utils.go
```diff
@@ -258,6 +261,17 @@ func (proc *ContainerdProcess) RunContainerTaskForReadyLine(
 	if err := taskDetails.task.Start(ctx); err != nil {
 		return nil, err
 	}
+	if parseFileAccessPatterns {
+		data := bparser.FileAccessPatterns{
+			ContainerStartTime: time.Now(),
```
vkuzniet, line 266: Is container task started yet?

benchmark/framework/parser/file_access.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err = file.Close(); err != nil {
+			return
+		}
+	}()
+	b, err := ioutil.ReadAll(file)
```
vkuzniet, line 55: [ioutil](https://pkg.go.dev/io/ioutil) is deprecated as of Go 1.16. Can you go with io's ReadAll?

benchmark/framework/parser/file_access.go
```diff
@@ -0,0 +1,134 @@
...
+	"os"
+	"sort"
+	"strings"
+	"time"
+)
+
+var (
+	outputDir      = "./output/"
+	sociLogs       = outputDir + "soci-snapshotter-stderr"
+	FileAccessLogs = outputDir + "fileAccessPatterns"
```
vkuzniet, line 15: This does not tell any details about container image. Probably, this information can be embedded somewhere, e.g. output file name, or be the part of fileAccessPatterns file.

benchmark/framework/containerd_utils.go
turan18: At this point the container process should be running. It is not tracking the time of ready line. Should it?

benchmark/framework/parser/file_access.go
turan18: Sure.

benchmark/performanceTest/main.go
```diff
@@ -38,6 +39,10 @@ func main() {
 		errMsg := fmt.Sprintf("Failed to parse number of test %s with error:%v\n", os.Args[3], err)
 		panic(errMsg)
 	}
+	var parseFileAccessPatterns bool
+	if numberOfTests == 1 {
```
vkuzniet, line 43: This doesn't seem right to me. Can you add a separate parameter, which comes as an input to performanceTest?

benchmark/framework/containerd_utils.go
vkuzniet: Are we measuring the ContainerStartTime right? This should be the time, container task starts.

vkuzniet:
Can you run `make add-ltag` to add header to the new file you are introducing?

hanyuel:
Can you add an example of ``fileAccessLogs`` in the PR description?

Makefile
```diff
@@ -140,3 +140,7 @@ benchmarks-comp:
 benchmarks-stargz:
 	@echo "$@"
 	@cd benchmark/stargzTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/StargzTests . && sudo ../bin/StargzTests $(COMMIT) ../singleImage.csv 10 $(STARGZ_BINARY)
+
+benchmark-collect-file-access-patterns:
```
hanyuel, line 144: Why made a new target instead of making it a command line option of``make benchmark``?

benchmark/performanceTest/main.go
hanyuel: +1 I suggest further adding a separate parameter ``parseFileAccess`` to ``make benchmarks`` and have it parsed here

benchmark/framework/parser/file_access.go
```diff
@@ -0,0 +1,134 @@
...
+			err := json.Unmarshal([]byte(log), &tempOperation)
+			if err != nil {
+				return err
+			}
+			op := tempOperation.Operation + tempOperation.Path
+
+			val, ok := m[op]
+
+			if ok {
+				copyOp := val
```
Kern--, line 76: nit: `m` is a map of structs (as opposed to pointers) so `val` is already a copy and this line is unnecessary.

benchmark/framework/containerd_utils.go
```diff
@@ -258,6 +261,17 @@ func (proc *ContainerdProcess) RunContainerTaskForReadyLine(
 	if err := taskDetails.task.Start(ctx); err != nil {
 		return nil, err
 	}
+	if parseFileAccessPatterns {
```
Kern--, line 264: Should this be after the container exits? Right now we'll only get partial data based on however far into the benchmark run the container is when the parser runs, right?

benchmark/framework/containerd_utils.go
```diff
@@ -205,7 +207,8 @@ func (proc *ContainerdProcess) CreateTask(
 func (proc *ContainerdProcess) RunContainerTaskForReadyLine(
 	ctx context.Context,
 	taskDetails *TaskDetails,
-	readyLine string) (func(), error) {
+	readyLine string,
+	parseFileAccessPatterns bool) (func(), error) {
```
Kern--, line 211: Maybe this was in another PR that I missed, but are we always emitting these logs but only parsing them if this flag is enabled? Do we need to do something to ensure that the logs are enabled when we try to parse them?

benchmark/framework/parser/file_access.go
```diff
@@ -0,0 +1,134 @@
...
+		return err
+	}
+
+	file, err := os.Open(sociLogs)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err = file.Close(); err != nil {
```
Kern--, line 51: I think this is effectively a no-op right now. What was the goal of this block?

benchmark/framework/containerd_utils.go
vkuzniet: We are not always emitting these logs. You need to run soci snapshotter in debug mode and also `log_fuse_operations` should be set to true in config.toml.

sbuckfelder:
At the highest level I would have interest in this tool not being so tightly coupled to the benchmarker, this tool can have value as a standalone piece.  The use case I imagine would be a customer turning on the file-access-pattern logs and then using the parser as a standalone.

benchmark/framework/containerd_utils.go
sbuckfelder: This looks like the correct place to get container start time - but I'm not sure why we want to inject the other logs here

sbuckfelder: Is this something we need?

vkuzniet:
> At the highest level I would have interest in this tool not being so tightly coupled to the benchmarker, this tool can have value as a standalone piece. The use case I imagine would be a customer turning on the file-access-pattern logs and then using the parser as a standalone.

Since we need to collect the file access pattern data and the tool seem to do the job, can we agree on making file_access.go an independent parametrized unit within the benchmarking framework for now and do the job collecting the file access pattern information? 
As the second iteration, the parser can be taken out of benchmarking framework and we can either write separate tool if needed.

benchmark/framework/containerd_utils.go
turan18: The container start time is now being pulled from  containerd stderr at the time in which the `containerd.events.TaskStart` event is emitted, instead of logging the time after `task.Start` returns.

benchmark/benchmarkTests.go
```diff
@@ -89,7 +89,8 @@ func SociFullRun(
 	imageRef string,
 	indexDigest string,
 	readyLine string,
-	testName string) {
+	testName string,
+	parseFileAccessPatterns bool) {
```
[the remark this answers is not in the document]
turan18: Since the collection of `containerStartTime` happens within the parser itself (see : https://github.com/awslabs/soci-snapshotter/pull/218#discussion_r1041388504) this optional parameter is no longer needed. The running of the parser happens as a part of the `benchmark.Run()` call as the parser is now assigned to the `AfterFunction` attribute of the test driver which will execute a given function after each image is fully ran.

benchmark/framework/containerd_utils.go
turan18: The parser will read the logs from both `containerd-stderr` and `soci-snapshotter-stderr` after the tests for an image are run.  This was initially there to record the start time of the container. ( No longer getting start time here)

benchmark/benchmarkTests.go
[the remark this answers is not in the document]
turan18: Something like:


```
opts = []bparser.ParseOpts{bparser.WithParseFileAccessPaterns,bparser.WithParseLifecycleOperations}
...
driver := framework.BenchmarkTestDriver{
...
AfterOpts: opts
...
}
```

benchmark/framework/parser/file_access.go
```diff
@@ -0,0 +1,153 @@
...
+	return err
+}
+
+func getTaskStartTime() (*time.Time, error) {
+	var taskStartTime time.Time
+	file, err := os.Open(containerdLogs)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
```
vkuzniet, line 136: Why not just `defer file.Close()`?

benchmark/framework/parser/file_access.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+type Operation struct {
+	Operation                 string `json:"operation"`
+	Path                      string `json:"path"`
+	FirstAccessTimeAfterStart string `json:"firstAccessTimeAfterStart"`
+	Count                     int    `json:"count"`
+}
+
+func ParseFileAccessPatterns(imageName string) error {
```
vkuzniet, line 55: Maybe name it `ParseFileAccesses`?

benchmark/framework/parser/file_access.go
turan18: To catch the error if one is produced by `file.Close()`. Although it will most likely be not needed

benchmark/framework/parser/file_access.go
turan18: Sounds good

benchmark/framework/framework.go
```diff
@@ -77,7 +77,11 @@ func (frame *BenchmarkFramework) Run(ctx context.Context) {
 		}
 		testDriver.calculateStats()
 		if testDriver.AfterFunction != nil {
-			testDriver.AfterFunction()
+			err := testDriver.AfterFunction()
+			if err != nil {
+				fmt.Printf("File Access Parsing Error: %v\n", err)
```
kzys, line 82: AfterFunction could be anything. Don't mention "File Access Parsing Error".

benchmark/framework/parser/file_access.go
kzys: Does it return an error if the file is open for reading?

benchmark/framework/parser/file_access.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return err
+	}
+
+	logs := strings.Split(string(b), "\n")
+
+	m := make(map[string]Operation)
+
+	for _, log := range logs {
+		if strings.Contains(log, "FUSE") {
```
kzys, line 82: Does it work if the file path itself is containing the word `FUSE`?

turan18: Ahhh... This would fail in that case. I could potentially look for "FUSE operation" perhaps. Or rather unmarshal to a struct since I believe every log contains at least `msg:` field and then filter by `msg` value.

benchmark/framework/framework.go
turan18: Got it will remove it.

benchmark/performanceTest/main.go
```diff
@@ -38,6 +39,20 @@ func main() {
 		errMsg := fmt.Sprintf("Failed to parse number of test %s with error:%v\n", os.Args[3], err)
 		panic(errMsg)
 	}
+	var parseFileAccessPatterns bool
+	if len(os.Args) == 5 && os.Args[4] == "parseFileAccessPatterns" {
+		fileAccessDir := outputDir + "/file_access_logs"
+		parseFileAccessPatterns = true
+		err := os.RemoveAll(fileAccessDir)
```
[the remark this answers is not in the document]
turan18: This is to remove older image logs that may be present. All logs including soci and containerd stdio persist each run

benchmark/performanceTest/main.go
```diff
@@ -38,6 +39,20 @@ func main() {
 		errMsg := fmt.Sprintf("Failed to parse number of test %s with error:%v\n", os.Args[3], err)
 		panic(errMsg)
 	}
+	var parseFileAccessPatterns bool
```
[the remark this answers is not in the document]
turan18: Sure

benchmark/framework/parser/file_access.go
turan18: It doesnt

benchmark/framework/parser/file_access.go
turan18: I have also added the ability to log the total count for each fuse operation. This may be useful for determining  possible correlations between any specific fuse operation and the total benchmark time.

benchmark/performanceTest/main.go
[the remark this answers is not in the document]
kzys: It is scary to see the command is removing ./output no matter where that is. What if I have something there?

Doesn't it work if you don't remove the directory? Unix's open syscall will just overwrite existing files. Note that MkdirAll is no-op if the directory is already there.

https://pkg.go.dev/os#MkdirAll

vkuzniet (approved):

vkuzniet:
LGTM.

kzys (approved):

benchmark/framework/parser/file_access.go
```diff
@@ -0,0 +1,177 @@
...
+		TotalOperationCount: totalCounts,
+	}
+	file, err := os.Open(sociLogs)
+	if err != nil {
+		return err
+	}
+
+	defer file.Close()
+
+	b, err := io.ReadAll(file)
```
Kern--, line 79: nit: a more idiomatic way to do this is to use a `bufio.Scanner` (https://pkg.go.dev/bufio#Scanner)

```
scanner := bufio.NewScanner(file)
for scanner.Scan() {
    log := scanner.Text()
}
```

benchmark/framework/parser/file_access.go
```diff
@@ -0,0 +1,177 @@
...
+
+			} else {
+				m[op] = Operation{
+					Operation:                 tempOperation.Operation,
+					Path:                      tempOperation.Path,
+					FirstAccessTimeAfterStart: tempOperation.Time.Sub(fa.ContainerStartTime).String(),
+					Count:                     1,
+				}
+			}
+			count, ok := fa.TotalOperationCount[tempOperation.Operation]
```
Kern--, line 118: nit: maps in go return the value type's zero value if not present which is 0 for integers. It should be safe to skip these checks and just do:
```
fa.TotalOperationCount[tempOperation.Operation]++
```

If it's not there it will add 1 to 0 and store it which is what we want.

Kern-- (approved):

Kern--:
Just a couple non-blocking nits. I'm fine to merge as is and we can always address those in a follow up.

Issue #229, opened 2022-12-12
Remove HTTP round trip for determining blob size and get blob sizes from layer descriptors


hanyuel:
**Is your feature request related to a problem? Please describe.**
When the snapshotter resolves a blob/layer, it needs to know the blob size. Currently the snapshotter does an [HTTP round trip call just to get the blob size](https://github.com/awslabs/soci-snapshotter/blob/main/fs/remote/resolver.go#L262-L263). This HTTP round trip is not needed as we can/should get the blob size from the image manifest.

**Describe the solution you'd like**
1. Remove the http round trip for getting blob size: https://github.com/awslabs/soci-snapshotter/blob/main/fs/remote/resolver.go#L262-L263
2. Get the blob size from manifest instead

**Describe alternatives you've considered**
None.

hanyuel:
The change is merged. Resolving.

PR #234, opened 2022-12-14
Get blob size from layer descriptors and remove redundant HTTP round trip

hanyuel:
*Issue #, if available:* #229

*Description of changes:*
Removed the HTTP round trip call that is just for getting blob size when resolving blobs.

Instead, we get blob sizes from layer descriptors. Create a new label "targetImageLayersSizeLabel" to add blob sizes to layers' labels/annotations. The label is parsed during layer mount to get blob sizes.

*Testing performed:*
``make test``
``make integration``

### DIFF
--- a/fs/remote/resolver.go
+++ b/fs/remote/resolver.go
@@ -154,14 +154,14 @@ func (r *Resolver) resolveFetcher(ctx context.Context, hosts source.RegistryHost
 	}
 
 	log.G(ctx).WithError(handlersErr).WithField("ref", refspec.String()).WithField("digest", desc.Digest).Debugf("using default handler")
-	hf, size, err := newHTTPFetcher(ctx, fc)
+	hf, err := newHTTPFetcher(ctx, fc)
 	if err != nil {
 		return nil, 0, err
 	}
 	if blobConfig.ForceSingleRangeMode {
 		hf.singleRangeMode()
 	}
-	return hf, size, err
+	return hf, desc.Size, err
 }
 
 type fetcherConfig struct {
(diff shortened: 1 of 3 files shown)


fs/remote/resolver.go
```diff
@@ -154,14 +154,14 @@ func (r *Resolver) resolveFetcher(ctx context.Context, hosts source.RegistryHost
...
-	hf, size, err := newHTTPFetcher(ctx, fc)
+	hf, err := newHTTPFetcher(ctx, fc)
 	if err != nil {
 		return nil, 0, err
 	}
 	if blobConfig.ForceSingleRangeMode {
 		hf.singleRangeMode()
 	}
-	return hf, size, err
+	return hf, desc.Size, err
```
vkuzniet, line 164: just a quick question. have you verified, that `desc.Size` contains non-zero value?

vkuzniet: please don't mind, I think I found the answer in the source.go

vkuzniet (approved):

kzys (approved):

Kern-- (approved):

rdpsin:
I'm confused why we're doing all this when `c.Annotations[targetSizeLabel]` already contains the image size. All we need to do is pass that value from `fs.Mount` to `makeBlob`.

hanyuel:
@rdpsin we need to know the sizes of all layers of the image to resolve neighboring layers..

sparr:
#536 requests conditionally getting the layer size from the registry if the label/annotation is missing.

PR #235, opened 2022-12-14
Avoid `Resolve` when no ztoc for layers

djdongjin:

*Issue #, if available:*

*Description of changes:*

In `fs.Mount`, if a layer doesn't have `ztoc` its `Resolve` call will fail. We should avoid these unnecessary `Resolve` calls.

*Testing performed:*

`make test`
`make integration`

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -349,9 +349,10 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	go func() {
 		rErr := fmt.Errorf("failed to resolve target")
 		for _, s := range src {
-			sociDesc := ocispec.Descriptor{}
-			if desc, ok := c.imageLayerToSociDesc[s.Target.Digest.String()]; ok {
-				sociDesc = desc
+			sociDesc, ok := c.imageLayerToSociDesc[s.Target.Digest.String()]
+			if !ok {
+				rErr = errors.Wrapf(rErr, "unable to resolve layer %q from %q, mounting entire layer using overlayfs: %s", s.Target.Digest, s.Name, "no ztoc for the layer")
+				continue
 			}
 
 			l, err := fs.resolver.Resolve(ctx, s.Hosts, s.Name, s.Target, sociDesc)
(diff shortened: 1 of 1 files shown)


fs/fs.go
```diff
@@ -349,17 +349,16 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	go func() {
 		rErr := fmt.Errorf("failed to resolve target")
 		for _, s := range src {
-			sociDesc := ocispec.Descriptor{}
-			if desc, ok := c.imageLayerToSociDesc[s.Target.Digest.String()]; ok {
-				sociDesc = desc
+			if sociDesc, ok := c.imageLayerToSociDesc[s.Target.Digest.String()]; ok {
```
Kern--, line 352: nit: can we invert this to reduce some nesting?
```
sociDesc, ok := ...
if !ok {
    rErr = error.Wrapf(...)
    continue
}
l, err := fs.resolver.Resovle(...)
...
```

Kern-- (approved):

fs/fs.go
kzys: +1

kzys (commented):
What would be a good way to test this change? I want to have some automated tests since we couldn't make it right, but Mount is really complicated. So I'm unsure we can write unit tests easily...

fs/fs.go
```diff
@@ -349,9 +349,10 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	go func() {
 		rErr := fmt.Errorf("failed to resolve target")
 		for _, s := range src {
-			sociDesc := ocispec.Descriptor{}
-			if desc, ok := c.imageLayerToSociDesc[s.Target.Digest.String()]; ok {
-				sociDesc = desc
+			sociDesc, ok := c.imageLayerToSociDesc[s.Target.Digest.String()]
+			if !ok {
+				rErr = errors.Wrapf(rErr, "failed to resolve layer %q from %q: %v", s.Target.Digest, s.Name, "no ztoc for the layer")
```
kzys, line 354: We know some layers don't have corresponding zTOC because of the configuration below;

https://github.com/awslabs/soci-snapshotter/blob/a6759ef18fb5028daddbe704494eb76e7653fc9e/cmd/soci/commands/create.go#L53-L58

So is it worth to report?

rdpsin:
@kzys We have integration tests running with sparse indices (layers with no ztocs).

fs/fs.go
rdpsin, line 354: How about instead of `failed to resolve layer`, we use something like `unable to resolve layer, mounting locally`?

fs/fs.go
```diff
@@ -371,16 +372,17 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 		go func() {
 			// Avoids to get canceled by client.
 			ctx := log.WithLogger(context.Background(), log.G(ctx).WithField("mountpoint", mountpoint))
-			sociDesc := ocispec.Descriptor{}
-			if descriptor, ok := c.imageLayerToSociDesc[desc.Digest.String()]; ok {
-				sociDesc = descriptor
+			sociDesc, ok := c.imageLayerToSociDesc[desc.Digest.String()]
+			if !ok {
+				log.G(ctx).Debug("failed to pre-resolve: no ztoc for the layer")
```
rdpsin, line 377: Same as above.

fs/fs.go
hanyuel: And in such case, we shouldn't report an error

djdongjin: I reword the error/log message as suggested by @rdpsin below. I think it's worth to report.

We know some layers doesn't have zTOC layer, but suppose if there is a bug causing a large layer should have zTOC but it's not in `c.imageLayerToSociDesc`, we'll know it's due to this (ztoc missing) instead of `Resolve` failure.

djdongjin:
@hanyuel updated. thanks

rdpsin (approved):

kzys (approved):

vkuzniet (approved):

PR #243, opened 2022-12-19
WIP Test running multiple containers at the same time

sparr:
*Description of changes:*
Add two integration tests, which each spin up an integration environment via Docker Compose. Both tests pull and build indices for images. The first test runs two containers from the same image. The second test runs[1] containers from different images. Both tests then perform a task within the running containers.

[1] The second test as currently submitted runs just one container; this will be changed in short order.

There are also a few other TODO/FIXME in the PR which I intend to discuss and address.

There is also one description mistake in an unrelated test fixed here, to avoid having to separately PR a single line fix.

*Testing performed:*
Ran the new tests repeatedly.


integration/pull_test.go
```diff
@@ -108,7 +108,7 @@ func TestOptimizeConsistentSociArtifact(t *testing.T) {
 			containerImage: "nginx:latest",
 		},
 		{
-			name:           "soci artifact is consistently built for drupal",
+			name:           "soci artifact is consistently built for alpine",
 			containerImage: "alpine:latest",
```
djdongjin, line 112: Do we want to change `name` to alpine or change the `containerImage` to drupal?

integration/run_test.go
```diff
@@ -0,0 +1,356 @@
...
+	"time"
+
+	shell "github.com/awslabs/soci-snapshotter/util/dockershell"
+	"github.com/awslabs/soci-snapshotter/util/dockershell/compose"
+	"github.com/awslabs/soci-snapshotter/util/testutil"
+)
+
+const (
+	drupalImage = "drupal:latest"
+)
```
djdongjin, line 46: nit: maybe put the const in the `testutil` pkg and export (`DrupalImage`).

integration/run_test.go
```diff
@@ -0,0 +1,356 @@
...
+
+[plugins."io.containerd.snapshotter.v1.soci"]
+root_path = "/var/lib/soci-snapshotter-grpc/"
+
+[plugins."io.containerd.snapshotter.v1.soci".blob]
+check_always = true
+
+{{.AdditionalConfig}}
+`, struct {
+		RegistryHost     string
```
djdongjin: `RegistryHost` seems not used by `containerdConfigYaml`. do we need it?

integration/run_test.go
```diff
@@ -0,0 +1,356 @@
...
+	"github.com/awslabs/soci-snapshotter/util/dockershell/compose"
+	"github.com/awslabs/soci-snapshotter/util/testutil"
+)
+
+const (
+	drupalImage = "drupal:latest"
+)
+
+// TestRunMultipleContainersSame tests if running the same container twice works
+func TestRunMultipleContainersSame(t *testing.T) {
```
djdongjin, line 49: nit: seems they are not "same container" but "same image". Suggest to change to `TestRunMutipleContainersWithSameImage`

integration/run_test.go
```diff
@@ -0,0 +1,356 @@
...
+	image := regConfig.mirror(imageName).ref
+	sh.X("soci", "run", "-d", "--rm", "--snapshotter=soci", image, "test1")
+	sh.X("soci", "run", "-d", "--rm", "--snapshotter=soci", image, "test2")
+	// FIXME ...Different test demonstrates that running `curl` too soon can produce "connection refused"
+	sh.X("ctr", "task", "exec", "--exec-id", "test1-curl", "test1", "curl", "http://127.0.0.1")
+	sh.X("ctr", "task", "exec", "--exec-id", "test2-curl", "test2", "curl", "http://127.0.0.1")
+}
+
+// TestRunMultipleContainersDifferent tests if running two different containers at the same time works
+func TestRunMultipleContainersDifferent(t *testing.T) {
```
djdongjin, line 54: similarly, `TestRunMultipleContainersWithDifferentImages`

integration/run_test.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+	if err := testutil.WriteFileContents(sh, filepath.Join(caCertDir, "domain.crt"), crt, 0600); err != nil {
+		t.Fatalf("failed to write %v: %v", caCertDir, err)
+	}
+	sh.
+		X("update-ca-certificates").
+		Retry(100, "nerdctl", "login", "-u", regConfig.user, "-p", regConfig.pass, regConfig.host)
+
+	// imageNames := []string{nginxImage, drupalImage}
+	imageNames := []string{nginxImage}
```
djdongjin: seems `imageNames := []string{nginxImage, drupalImage}` should be used?

integration/run_test.go
```diff
@@ -0,0 +1,356 @@
...
+	// Run the containers
+	for _, imageName := range imageNames {
+		image := regConfig.mirror(imageName).ref
+		containerName := "test_" + regexp.MustCompile(`[^a-zA-Z0-9 ]+`).ReplaceAllString(imageName, "_")
+		sh.X("soci", "run", "-d", "--rm", "--snapshotter=soci", image, containerName)
+	}
+
+	time.Sleep(5 * time.Second) // FIXME find better solution for "connection refused" from `curl` below too soon
+
+	// Do something in each container
```
djdongjin, line 117: If the `// Do something in each container` is the same in the two tests, maybe you can have a single function that does the testing logic (create images, create containers, check, etc). And probably for the two testcases, each can just pass in the function something like "which container uses which image".

integration/pull_test.go
sparr: If we're going to change the actual container being tested, I think that would deserve its own PR and discussion. This version of the change has no practical impact, which is why I felt comfortable including it here.

integration/run_test.go
sparr: I hadn't originally intended both containers to be http servers. I'd probably want to move the "do something" out into a function that can be passed in, possibly on a per container basis?

integration/run_test.go
sparr: Moved this and the three other related constants from pull_test, all to testutil.

Why export?

djdongjin: just noticed there are two `testutil`. in the 2nd case export is not needed since they are in the same package :)

https://github.com/awslabs/soci-snapshotter/blob/078ae664ed40b06679fce302f792812c97f7adef/util/testutil/util.go

https://github.com/awslabs/soci-snapshotter/blob/078ae664ed40b06679fce302f792812c97f7adef/integration/testutil.go

integration/run_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+// TestRunMultipleContainersWithSameImage runs two containers with the same image at the same time
+func TestRunMultipleContainersWithSameImage(t *testing.T) {
+	RunMultipleWebContainers(t, []string{nginxImage, nginxImage})
+}
+
+// TestRunMultipleContainersWithDifferentImages runs two containers with different images at the same time
+func TestRunMultipleContainersWithDifferentImages(t *testing.T) {
+	RunMultipleWebContainers(t, []string{nginxImage, drupalImage})
+}
```
rdpsin, line 52: We should use table driven tests here. You can take a look at how we've implemented other integration tests.

integration/run_test.go
```diff
@@ -0,0 +1,128 @@
...
+	for _, imageName := range deduplicate(imageNames) {
+		// Mirror image
+		copyImage(sh, dockerhub(imageName), regConfig.mirror(imageName))
+		indexDigest := optimizeImage(sh, regConfig.mirror(imageName))
+
+		// Pull image
+		sh.X("ctr", "i", "pull", "--user", regConfig.creds(), regConfig.mirror(imageName).ref)
+
+		// Create SOCI metadata
+		sh.X("soci", "create", regConfig.mirror(imageName).ref)
```
rdpsin, line 98: The `optimizeImage` function already creates a soci index.

integration/run_test.go
```diff
@@ -0,0 +1,128 @@
...
+		sh.X("soci", "image", "rpull", "--user", regConfig.creds(), "--soci-index-digest", indexDigest, regConfig.mirror(imageName).ref)
```
rdpsin, line 99: You already pulled the image in line 95. So this `rpull` will just skip the lazy-loading since the image already exists in the local storage.

integration/run_test.go
```diff
@@ -0,0 +1,128 @@
...
+		containerName := "test_" + fmt.Sprint(index) + "_" + regexp.MustCompile(`[^a-zA-Z0-9 ]+`).ReplaceAllString(imageName, "_")
+		sh.X("soci", "run", "-d", "--rm", "--snapshotter=soci", image, containerName)
+	}
+
+	time.Sleep(5 * time.Second) // FIXME find better solution for "connection refused" from `curl` below too soon
+
+	// Do something in each container
+	for index, imageName := range imageNames {
+		containerName := "test_" + fmt.Sprint(index) + "_" + regexp.MustCompile(`[^a-zA-Z0-9 ]+`).ReplaceAllString(imageName, "_")
+		sh.X("ctr", "task", "exec", "--exec-id", "test-curl", containerName, "curl", "http://127.0.0.1")
```
rdpsin, line 114: Normally, we don't run a process in the snapshotter. We compare the contents of a directory (`/usr`) with the corresponding contents created by the overlayfs snapshotter. Can we do something similar here?

rdpsin (dismissed):

integration/run_test.go
sparr: I've refactored into a single function for testing, the two Test functions are just one line each now.

integration/run_test.go
sparr: Similar to https://github.com/awslabs/soci-snapshotter/blob/main/integration/pull_test.go#L91-L114 ?

integration/run_test.go
sparr: Good catch, thanks. Can you confirm that https://github.com/awslabs/soci-snapshotter/blob/main/integration/pull_test.go#L786-L794 is also redundant?

integration/run_test.go
sparr: Reaching out about this elsewhere.

integration/run_test.go
sparr: I was under the impression that these tests were intended to be about the functioning of the containers, not just about whether they start up. I tried to aim in the middle here, but I expect we could go so far as a test that would perform some operation then check the results within the container.
Nailing this down probably requires a more specific definition of what exactly we're trying to test.

integration/run_test.go
sparr: Do we have any table driven tests that do things in parallel?

integration/run_test.go
rdpsin: Hmm, looks like it. Should probably reach out to @vkuzniet about this; he will have the most context.

integration/run_test.go
rdpsin: No, I don't think so.

integration/run_test.go
vkuzniet: Yep, it is redundant. Pull and create happen during `optimizeImage`.

integration/run_test.go
vkuzniet: Not exactly. `containerd` and `soci-snapshotter` do not share the content store. What will happen, is that for each layer in container image, there will be a respective call to `Prepare()`, which will go over mounting the layer depending on Soci index configuration.

integration/run_test.go
```diff
@@ -0,0 +1,128 @@
...
+func TestRunMultipleContainersWithSameImage(t *testing.T) {
+	RunMultipleWebContainers(t, []string{nginxImage, nginxImage})
+}
+
+// TestRunMultipleContainersWithDifferentImages runs two containers with different images at the same time
+func TestRunMultipleContainersWithDifferentImages(t *testing.T) {
+	RunMultipleWebContainers(t, []string{nginxImage, drupalImage})
+}
+
+func RunMultipleWebContainers(t *testing.T, imageNames []string) {
```
kzys, line 54: You don't have to export this function.

integration/run_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+		// Create SOCI metadata
+		sh.X("soci", "create", regConfig.mirror(imageName).ref)
+		sh.X("soci", "image", "rpull", "--user", regConfig.creds(), "--soci-index-digest", indexDigest, regConfig.mirror(imageName).ref)
+	}
+
+	// Run the containers
+	for index, imageName := range imageNames {
+		image := regConfig.mirror(imageName).ref
+		containerName := "test_" + fmt.Sprint(index) + "_" + regexp.MustCompile(`[^a-zA-Z0-9 ]+`).ReplaceAllString(imageName, "_")
```
kzys, line 105: Can you extract this regex + ReplaceAllString to a function?

integration/testutil.go
```diff
@@ -67,6 +67,13 @@ const (
 	dockerLibrary                = "public.ecr.aws/docker/library/"
 )
 
+const (
+	alpineImage = "alpine:latest"
+	nginxImage  = "nginx:latest"
+	ubuntuImage = "ubuntu:latest"
+	drupalImage = "drupal:latest"
```
kzys, line 74: Is it possible to pin versions?

integration/run_test.go
sparr: Are you suggesting that the two tests I've implemented could be a single test with two sub-tests, each driven from a table, or that each of my tests should have its list of containers to run as a table? If the latter, I can't see how to map the existing examples for table driven subtests to this situation where the containers need to run in parallel and make up a single logical test with no sub units.

integration/run_test.go
sparr: I'm fixing this in my PR. I'll probably open another PR later to clean up the redundancy elsewhere.

integration/run_test.go
sparr: @vkuzniet I think that's the behavior we want here. Can you clarify if you're also suggesting some change?

integration/run_test.go
sparr: Yes

integration/testutil.go
sparr: It is possible, but other existing tests rely on most of these values. I think that should be a separate PR.

integration/run_test.go
rdpsin: They do not share the content store for artifacts, but share it for images.

integration/run_test.go
Kern--: The goal of these tests is to make sure we don't accidentally share state between containers in a way that breaks things.

E.g. a previous version of the snapshotter only lazily loaded the first rpulled image (https://github.com/awslabs/soci-snapshotter/pull/77) and we couldn't detect that from our existing tests which only ran 1 task in a clean environment.

I think verifying that the container workload comes up will tell us that the snapshotter works in this case, but we may want additional verifications. At the very least verifying no fallback to overlayfs. We do that in the single task tests (e.g. https://github.com/awslabs/soci-snapshotter/blob/main/integration/pull_test.go#L234-L241) but it might be easier to do it here after we have a like a metric or something on fallback.

integration/run_test.go
kzys: I think, TestRunMultipleContainersWithSameImage and TestRunMultipleContainersWithDifferentImages could be merged. Both of them are just calling runMultipleWebContainers.

integration/testutil.go
Kern--: Can you create an issue for that please?

sparr: https://github.com/awslabs/soci-snapshotter/issues/248

integration/run_test.go
sparr: @Kern-- does that clear this to merge, or do you want this PR to add the mount-counting check, or for this PR to wait for the fallback metric functionality?

integration/run_test.go
vkuzniet: > @vkuzniet I think that's the behavior we want here. Can you clarify if you're also suggesting some change?

I see. If the intention of the test is to lazily pull the container image twice, then I'm fine with that.

vkuzniet: > They do not share the content store for artifacts, but share it for images.

For download/decompress part, snapshotter explicitly uses the same content store as for artifacts. See https://github.com/awslabs/soci-snapshotter/blob/main/fs/unpacker.go#L76 for more details. Snapshotter does not have an access to containerd's content addressable store and we didn't have an intention to make this happen in the first place. Anyway, if the call to rpull lands on Prepare(), snapshotter will take over and will attempt to lazily load the layer to the best of its ability (more details here: https://github.com/awslabs/soci-snapshotter/blob/main/snapshot/snapshot.go#L263).

kzys (commented):
The change looks good to me. Can you tidy up your commit history? I think, even just having one commit is fine for me.

integration/run_test.go
Kern--: Yes, we can merge without the mount-checking. Can you create a follow up ticket?

sbuckfelder (dismissed):

integration/run_test.go
sparr: Per standup and post-scrum conversations, here is a manual recreation of the critical steps of this test, outside of a containerized environment. I believe this shows that the `ctr i pull` then `soci image rpull` produces the desired behavior with two of the layers being served by soci through the fuse mounted filesystems.

```
% docker run -d -p 5000:5000 --restart=always --name registry ghcr.io/oci-playground/registry:v3.0.0-alpha.1

12ce889755d9b883e941223c84e041f92a2ab5e32ab5ccbdbd0230ff2615cc0d

% curl http://localhost:5000/v2/_catalog

{"repositories":[]}

% sudo ctr i pull docker.io/library/nginx:latest                      
docker.io/library/nginx:latest:                                                   resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:0047b729188a15da49380d9506d65959cce6d40291ccfb4e039f5dc7efd33286:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:9a821cadb1b13cb782ec66445325045b2213459008a41c72d8d87cde94b33c8c: done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:5f63362a3fa390a685ae42e1936feeca3e4fba185bdc46fb66cf184036611f7d:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:50c68654b16f458108a537c9842c609f647a022fbc5a9b6bde1ffb60b77c2349:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:88d3ab68332da2aa6cc8d83c9dfe95905dc899d9b8fb302ebae2bf9a6b167c40:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:40b838968eeab5abc9fb941a8e3ee1377660bb02672153cada52bc9d4e0595b7:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:3f4ca61aafcd4fc07267a105067db35c0f0ac630e1970f3cd0c7bf552780e985:    done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:1403e55ab369cd1c8039c34e6b4d47ca40bbde39c371254c7cba14756f472f52:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:3ed295c083ec7246873f1b98bbc7b634899c99f6d2d901e2f9f5220d871830dd:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 0.7 s                                                                    total:  52.3 M (74.6 MiB/s)                                      
unpacking linux/amd64 sha256:0047b729188a15da49380d9506d65959cce6d40291ccfb4e039f5dc7efd33286...
done: 1.266077552s

% sudo ctr i tag docker.io/library/nginx:latest localhost:5000/nginx:latest
localhost:5000/nginx:latest

% sudo ctr i push --platform linux/amd64 --plain-http localhost:5000/nginx:latest   
manifest-sha256:9a821cadb1b13cb782ec66445325045b2213459008a41c72d8d87cde94b33c8c: done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:1403e55ab369cd1c8039c34e6b4d47ca40bbde39c371254c7cba14756f472f52:   done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 0.4 s                                                                    total:  9.0 Ki (22.5 KiB/s)                                      

% curl http://localhost:5000/v2/_catalog

{"repositories":["nginx"]}

% sudo soci create localhost:5000/nginx:latest
layer sha256:5f63362a3fa390a685ae42e1936feeca3e4fba185bdc46fb66cf184036611f7d -> ztoc skipped
layer sha256:3ed295c083ec7246873f1b98bbc7b634899c99f6d2d901e2f9f5220d871830dd -> ztoc skipped
layer sha256:40b838968eeab5abc9fb941a8e3ee1377660bb02672153cada52bc9d4e0595b7 -> ztoc skipped
layer sha256:88d3ab68332da2aa6cc8d83c9dfe95905dc899d9b8fb302ebae2bf9a6b167c40 -> ztoc skipped
layer sha256:50c68654b16f458108a537c9842c609f647a022fbc5a9b6bde1ffb60b77c2349 -> ztoc sha256:678ea3b10a0dfab41fd7dfdc39596485f0e213e067043d43aa9f1d74458516ff
layer sha256:3f4ca61aafcd4fc07267a105067db35c0f0ac630e1970f3cd0c7bf552780e985 -> ztoc sha256:1c5fe514fa61890c0f1684fdbdf8a14f61d3a61bbed1c503cee5a7887f25e0ca

% sudo soci index list
DIGEST                                                                     SIZE    IMAGE REF                         PLATFORM
sha256:cff81469d225a23cd8aefa7641279029def23325a02b6406a6a1f54c9b642ecd    1083    docker.io/library/nginx:latest    linux/amd64    
sha256:cff81469d225a23cd8aefa7641279029def23325a02b6406a6a1f54c9b642ecd    1083    localhost:5000/nginx:latest       linux/amd64    

% sudo soci push --plain-http localhost:5000/nginx:latest
pushing soci index with digest: sha256:cff81469d225a23cd8aefa7641279029def23325a02b6406a6a1f54c9b642ecd
pushing artifact with digest: sha256:678ea3b10a0dfab41fd7dfdc39596485f0e213e067043d43aa9f1d74458516ff
pushing artifact with digest: sha256:1c5fe514fa61890c0f1684fdbdf8a14f61d3a61bbed1c503cee5a7887f25e0ca
skipped artifact with digest: sha256:9a821cadb1b13cb782ec66445325045b2213459008a41c72d8d87cde94b33c8c
successfully pushed artifact with digest: sha256:1c5fe514fa61890c0f1684fdbdf8a14f61d3a61bbed1c503cee5a7887f25e0ca
successfully pushed artifact with digest: sha256:678ea3b10a0dfab41fd7dfdc39596485f0e213e067043d43aa9f1d74458516ff
pushing artifact with digest: sha256:cff81469d225a23cd8aefa7641279029def23325a02b6406a6a1f54c9b642ecd
successfully pushed artifact with digest: sha256:cff81469d225a23cd8aefa7641279029def23325a02b6406a6a1f54c9b642ecd

% sudo soci-snapshotter-grpc&> ~/soci-snapshotter-logs &
[1] 27651

% sudo soci image rpull --plain-http --soci-index-digest sha256:cff81469d225a23cd8aefa7641279029def23325a02b6406a6a1f54c9b642ecd localho
st:5000/nginx:latest
fetching sha256:9a821cad... application/vnd.docker.distribution.manifest.v2+json
fetching sha256:1403e55a... application/vnd.docker.container.image.v1+json
fetching sha256:3ed295c0... application/vnd.docker.image.rootfs.diff.tar.gzip
fetching sha256:88d3ab68... application/vnd.docker.image.rootfs.diff.tar.gzip
fetching sha256:40b83896... application/vnd.docker.image.rootfs.diff.tar.gzip
fetching sha256:5f63362a... application/vnd.docker.image.rootfs.diff.tar.gzip

% mount   
[boilerplate output omitted]
fusectl on /sys/fs/fuse/connections type fusectl (rw,relatime)
overlay on /local/docker/overlay2/e903856086088109af564197ad1752045447f628fa49b31aef83b09736b8ee6d/merged type overlay (rw,relatime,lowerdir=/local/docker/overlay2/l/DL3KS6KPL25RMZ3PVT56XFK7ER:/local/docker/overlay2/l/C4OAMNIB7AF56EUYPNS2BOMDBR:/local/docker/overlay2/l/SXOW7OH6WBV6RG5LXPF5TCRBCS:/local/docker/overlay2/l/DEMTSB4QMCQCRCRSSSNDAAAAI2:/local/docker/overlay2/l/TZSZV6EQADHJUIAYDQ7UDCBM4V,upperdir=/local/docker/overlay2/e903856086088109af564197ad1752045447f628fa49b31aef83b09736b8ee6d/diff,workdir=/local/docker/overlay2/e903856086088109af564197ad1752045447f628fa49b31aef83b09736b8ee6d/work)
nsfs on /run/docker/netns/b773e06fa9e8 type nsfs (rw)
/local/home/clrnc/src/soci-snapshotter/soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/1/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
/local/home/clrnc/src/soci-snapshotter/soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/2/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)

% sudo ctr run -d --rm --snapshotter soci --net-host localhost:5000/nginx:latest nginxExample

% sudo ctr c ls
CONTAINER       IMAGE                          RUNTIME                  
nginxExample    localhost:5000/nginx:latest    io.containerd.runc.v2    

% sudo ctr t ls
TASK            PID      STATUS    
nginxExample    29651    RUNNING

% mount
[boilerplate output omitted]
fusectl on /sys/fs/fuse/connections type fusectl (rw,relatime)
overlay on /local/docker/overlay2/e903856086088109af564197ad1752045447f628fa49b31aef83b09736b8ee6d/merged type overlay (rw,relatime,lowerdir=/local/docker/overlay2/l/DL3KS6KPL25RMZ3PVT56XFK7ER:/local/docker/overlay2/l/C4OAMNIB7AF56EUYPNS2BOMDBR:/local/docker/overlay2/l/SXOW7OH6WBV6RG5LXPF5TCRBCS:/local/docker/overlay2/l/DEMTSB4QMCQCRCRSSSNDAAAAI2:/local/docker/overlay2/l/TZSZV6EQADHJUIAYDQ7UDCBM4V,upperdir=/local/docker/overlay2/e903856086088109af564197ad1752045447f628fa49b31aef83b09736b8ee6d/diff,workdir=/local/docker/overlay2/e903856086088109af564197ad1752045447f628fa49b31aef83b09736b8ee6d/work)
nsfs on /run/docker/netns/b773e06fa9e8 type nsfs (rw)
/local/home/clrnc/src/soci-snapshotter/soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/1/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
/local/home/clrnc/src/soci-snapshotter/soci on /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/2/fs type fuse.rawBridge (rw,nodev,relatime,user_id=0,group_id=0,allow_other)
overlay on /run/containerd/io.containerd.runtime.v2.task/default/nginxExample/rootfs type overlay (rw,relatime,lowerdir=/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/6/fs:/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/5/fs:/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/4/fs:/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/3/fs:/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/2/fs:/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/1/fs,upperdir=/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/7/fs,workdir=/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/7/work)
```

Kern-- (approved):

integration/run_test.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+
+			// Run the containers
+			for index, container := range tt.containers {
+				image := regConfig.mirror(container.containerImage).ref
+				containerName := "test_" + fmt.Sprint(index) + "_" + makeImageNameValid(container.containerImage)
+				sh.X("soci", "run", "-d", "--rm", "--snapshotter=soci", image, containerName)
+			}
+
+			time.Sleep(2 * time.Second) // FIXME find better solution for "connection refused" from `curl` below too soon
```
Kern--, line 131: It looks like this might not be quite long enough, the most recent run failed with:
```
curl: (7) Failed to connect to 127.0.0.1 port 80: Connection refused
```
https://github.com/awslabs/soci-snapshotter/actions/runs/3760427063/jobs/6391131866#step:4:14365


To get this merged, I think we can just increase this timeout for now. As a follow up, it might be worth switching from testing that we make a single request and it succeeds to retrying the request and making sure it succeeds at least once within a timeout.

sparr: Switched to `curl --retry --retry-connrefused` with what felt like sensible settings

Kern-- (approved):

rdpsin (approved):

Issue #247, opened 2022-12-21
De-duplicate integration test code


rdpsin:
**Is your feature request related to a problem? Please describe.**
There is quite a bit of code duplication going on in the integration and unit tests. 

Some examples:

-  [ ] dedup `genRandomByteData`
```
% ag -Q "func genRandomByteData" .
ztoc/ztoc_test.go
624:func genRandomByteData(size int) []byte {

fs/span-manager/span_manager_test.go
344:func genRandomByteData(size compression.Offset) []byte {

fs/backgroundfetcher/resolver_test.go
82:func genRandomByteData(size int64) []byte {

```

- [ ] `optimizeImage` and `optimizeImageWithOpts` functions in the integration tests.
- [x] identical snapshotter configs are defined separately in every single integration test. (#294)

We should de-duplicate all this code.

djdongjin:
Meanwhile, probably we should also move those reusable helper functions to a common place such as `testutil.go` or something, instead of creating a helper function in `integration/aaa_test.go` and reusing in `integration/bbb_test.go`.

PR #250, opened 2022-12-21
Add integration tests for `soci index ls` and `soci index info`

djdongjin:

*Issue #, if available:*
Part of https://github.com/awslabs/soci-snapshotter/issues/54

*Description of changes:*

This PR adds integration tests for `soci index ls` and `soci index info`.

*Testing performed:*

### DIFF
--- a/integration/create_test.go
+++ b/integration/create_test.go
@@ -73,22 +73,6 @@ func TestSociCreateSparseIndex(t *testing.T) {
 				t.Fatalf("cannot get index data: %v", err)
 			}
 
-			if index.MediaType != ocispec.MediaTypeArtifactManifest {
-				t.Fatalf("unexpected index media type; expected = %v, got = %v", ocispec.MediaTypeArtifactManifest, index.MediaType)
-			}
-
-			if index.ArtifactType != soci.SociIndexArtifactType {
-				t.Fatalf("unexpected index artifact type; expected = %v, got = %v", soci.SociIndexArtifactType, index.ArtifactType)
-			}
-
-			expectedAnnotations := map[string]string{
-				soci.IndexAnnotationBuildToolIdentifier: "AWS SOCI CLI v0.1",
-			}
-
-			if diff := cmp.Diff(index.Annotations, expectedAnnotations); diff != "" {
-				t.Fatalf("unexpected index annotations; diff = %v", diff)
-			}
-
 			imageManifestJSON := fetchContentByDigest(sh, manifestDigest)
 			imageManifest := new(ocispec.Manifest)
 			if err := json.Unmarshal(imageManifestJSON, imageManifest); err != nil {
@@ -102,29 +86,7 @@ func TestSociCreateSparseIndex(t *testing.T) {
 				}
 			}
 
-			blobs := index.Blobs
-			if len(blobs) != len(includedLayers) {
-				t.Fatalf("unexpected blob count; expected=%v, got=%v", len(includedLayers), len(blobs))
-			}
-
-			for _, blob := range blobs {
-				blobContent := fetchContentFromPath(sh, blobStorePath+"/"+trimSha256Prefix(blob.Digest.String()))
-				blobSize := int64(len(blobContent))
-				blobDigest := digest.FromBytes(blobContent)
-				layerDigest := blob.Annotations[soci.IndexAnnotationImageLayerDigest]
-
-				if _, ok := includedLayers[layerDigest]; !ok {
-					t.Fatalf("found ztoc for layer %v in index but should not have built ztoc for it", layerDigest)
-				}
-
-				if blobSize != blob.Size {
-					t.Fatalf("unexpected blob size; expected = %v, got = %v", blob.Size, blobSize)
-				}
-
-				if blobDigest != blob.Digest {
-					t.Fatalf("unexpected blob digest; expected = %v, got = %v", blob.Digest, blobDigest)
-				}
-			}
+			validateSociIndex(t, sh, *index, manifestDigest, includedLayers)
 		})
 	}
 }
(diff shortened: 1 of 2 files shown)


integration/index_test.go
```diff
@@ -0,0 +1,166 @@
...
+			existHandler: existHandlerQuiet,
+		},
+		{
+			name:         "`soci index ls --ref imgRef` should only list soci indices for the image",
+			command:      []string{"soci", "index", "list", "--ref", testImages[0].imgRef},
+			filter:       func(img testImageIndex) bool { return img.imgRef == testImages[0].imgRef },
+			existHandler: existHandlerFull,
+		},
+		{
+			name:         "`soci index ls --platform arm64` should only list soci indices for arm64 platform",
```
Kern--, line 145: nit: `linux/arm64`

integration/index_test.go
```diff
@@ -0,0 +1,166 @@
...
+			command:      []string{"soci", "index", "list", "-q"},
+			filter:       func(img testImageIndex) bool { return true },
+			existHandler: existHandlerQuiet,
+		},
+		{
+			name:         "`soci index ls --ref imgRef` should only list soci indices for the image",
+			command:      []string{"soci", "index", "list", "--ref", testImages[0].imgRef},
+			filter:       func(img testImageIndex) bool { return img.imgRef == testImages[0].imgRef },
+			existHandler: existHandlerFull,
+		},
```
Kern--, line 143: Can we add an explicit test here that if you do `soci index ls --ref imgRef -q` you get exactly the image ref and nothing else?

e.g. the `existHandler` should be 
```
func(output string, img testImageIndex) bool {
    return strings.Trim(output) == img.sociIndexDigest
}
```

djdongjin: fixed

integration/index_test.go
djdongjin: fixed

Kern-- (approved):

integration/index_test.go
```diff
@@ -0,0 +1,178 @@
...
+				t.Fatalf("invalid soci index media type: expected: %s, actual: %s", soci.OCIArtifactManifestMediaType, sociIndex.MediaType)
+			}
+			if sociIndex.ArtifactType != soci.SociIndexArtifactType {
+				t.Fatalf("invalid soci index artifact type: expected: %s, actual: %s", soci.SociIndexArtifactType, sociIndex.ArtifactType)
+			}
+		})
+	}
+}
+
+func TestSociIndexLs(t *testing.T) {
```
rdpsin, line 100: np: 'TestSociIndexList'

integration/index_test.go
```diff
@@ -0,0 +1,178 @@
...
+	testImages := prepareSociIndices(t, sh)
+
+	existHandlerFull := func(output string, img testImageIndex) bool {
+		// full output should have both img ref and soci index digest
+		return strings.Contains(output, img.imgRef) && strings.Contains(output, img.sociIndexDigest)
+	}
+
+	existHandlerQuiet := func(output string, img testImageIndex) bool {
+		// quiet output should only have "soci_index_digest\n"
+		return strings.Contains(output, img.sociIndexDigest+"\n")
```
rdpsin, line 115: np: Can we do 'strings.Trim(output, "\n"), img.sociIndexDigest` here?

integration/index_test.go
```diff
@@ -0,0 +1,178 @@
...
+			rawJSON := sh.O("soci", "index", "info", img.sociIndexDigest)
+			if err := json.Unmarshal(rawJSON, &sociIndex); err != nil {
+				t.Fatalf("invalid soci index from digest %s: %v", img.sociIndexDigest, rawJSON)
+			}
+			if sociIndex.MediaType != soci.OCIArtifactManifestMediaType {
+				t.Fatalf("invalid soci index media type: expected: %s, actual: %s", soci.OCIArtifactManifestMediaType, sociIndex.MediaType)
+			}
+			if sociIndex.ArtifactType != soci.SociIndexArtifactType {
+				t.Fatalf("invalid soci index artifact type: expected: %s, actual: %s", soci.SociIndexArtifactType, sociIndex.ArtifactType)
+			}
```
rdpsin, line 95: I think we probably want to test for `subject`, `annotations` and `blobs` fields.

integration/index_test.go
djdongjin: The output might have multiple index digests (digest1\ndigest2\n...\n). Trim makes the last one as `digest` but the rest as `digest\n`. Maybe we can use `strings.HasPrefix(output, img.SociIndexDigest+"\n")` which matches the whole line and also works in multi index output?

integration/index_test.go
djdongjin: fixed

integration/index_test.go
djdongjin: fixed

integration/index_test.go
djdongjin: made some refactor and reuse some code from `create` integration test. It checks all now.

djdongjin:
PTAL @Kern-- @rdpsin

Made some refactor to reduce duplication on soci index validation. Thanks

integration/index_test.go
```diff
@@ -0,0 +1,179 @@
...
+	existHandlerFull := func(output string, img testImageIndex) bool {
+		// full output should have both img ref and soci index digest
+		return strings.Contains(output, img.imgInfo.ref) && strings.Contains(output, img.sociIndexDigest)
+	}
+
+	existHandlerQuiet := func(output string, img testImageIndex) bool {
+		// a given soci index should match exactly one line in the quiet output
+		// for the first index, it should have prefix of digest+\n
+		// for the rest, it should have `\n` before and after its digest
+		return strings.HasPrefix(output, img.sociIndexDigest+"\n") || strings.Contains(output, "\n"+img.sociIndexDigest+"\n")
```
rdpsin, line 116: All this string manipulation seems flaky to me. What if we decide to change the output? I think it make sense to have `--json` output. Doesn't have to be part of this PR, but something to think about.

rdpsin (approved):

integration/create_test.go
```diff
@@ -190,46 +152,62 @@ func TestSociCreate(t *testing.T) {
...
-			}
+			validateSociIndex(t, sh, *sociIndex, m, nil)
+		})
+	}
+}
 
-			expectedAnnotations := map[string]string{
-				soci.IndexAnnotationBuildToolIdentifier: "AWS SOCI CLI v0.1",
-			}
+func validateSociIndex(t *testing.T, sh *shell.Shell, sociIndex soci.Index, imgManifestDigest string, includedLayers map[string]struct{}) {
```
Kern--, line 165: nit: can we move this to testutil.go or something?

Kern-- (approved):

integration/create_test.go
djdongjin: Yeah I think it's good to move these functional helpers to a common place. Maybe we can track this in https://github.com/awslabs/soci-snapshotter/issues/247, and do it two steps (first make changes so it's reusable (in this PR), then we move those helpers to a better place without change). 

I found doing so is easier to review (easier to visualize what changes are made; otherwise it will be deleting a func in one file and adding a slightly changed func in another file).

PR #257, opened 2022-12-28
Add integration test for arbitrary ztoc blobs

rdpsin:
This commit adds an integration test to verify snapshotter functionality when an arbitrary blob is passed as a ztoc to the snapshotter. In this case, instead of mounting the layer as a FUSE mount for lazy-loading, the snapshotter will download the entire layer and mount it as a normal overlayfs mount.

*Issue #, if available:*

*Description of changes:*

*Testing performed:*

### DIFF
--- a/integration/pull_test.go
+++ b/integration/pull_test.go
@@ -41,10 +41,12 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/awslabs/soci-snapshotter/soci"
 	shell "github.com/awslabs/soci-snapshotter/util/dockershell"
 	"github.com/awslabs/soci-snapshotter/util/dockershell/compose"
 	"github.com/awslabs/soci-snapshotter/util/testutil"
 	"github.com/containerd/containerd/platforms"
+	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/rs/xid"
 )
@@ -112,7 +114,6 @@ func TestOptimizeConsistentSociArtifact(t *testing.T) {
 			containerImage: "alpine:latest",
 		},
 	}
-	const blobStorePath = "/var/lib/soci-snapshotter-grpc/content/blobs/sha256"
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			rebootContainerd(t, sh, "", "")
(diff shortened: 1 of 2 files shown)


integration/pull_test.go
```diff
@@ -501,6 +504,148 @@ level = "debug"
 	}
 }
 
+// TestPullWithArbitraryBlobAsZtoc tests the snapshotter behavior if an arbitrary blob is passed
```
djdongjin, line 507: should we highlight "an arbitrary blob _with invalid ztoc format_"? Since in the future we may add another test for "an arbitrary blob _with valid ztoc format but invalid data_".

integration/pull_test.go
```diff
@@ -501,6 +504,148 @@ level = "debug"
...
+// as a Ztoc. In this case, the flatbuffer deserialization will fail, which will lead
+// to the snapshotter mounting the layer as a normal overlayfs mount.
+func TestPullWithArbitraryBlobAsZtoc(t *testing.T) {
+	regConfig := newRegistryConfig()
+	getContainerdConfigYaml := func(disableVerification bool) []byte {
```
djdongjin, line 512: Can we move `getContainerdConfigYaml` and `getSnapshotterConfigYaml` to a common place like `testutil`? Both are duplicated multiple times. (2 in `push_test.go`, 4 in `pull_test.go`)

https://github.com/search?q=repo%3Aawslabs%2Fsoci-snapshotter+%22getContainerdConfigYaml+%3A%3D%22&type=code

rdpsin: I think it should be done as part of #247. Removing it will touch multiple tests, and that seems out of scope of this PR.

djdongjin: SGTM

Kern-- (approved):

integration/pull_test.go
```diff
@@ -112,7 +114,6 @@ func TestOptimizeConsistentSociArtifact(t *testing.T) {
 			containerImage: "alpine:latest",
 		},
 	}
-	const blobStorePath = "/var/lib/soci-snapshotter-grpc/content/blobs/sha256"
```
sbuckfelder, line 115: Where does this come from now?

sbuckfelder (approved):

integration/pull_test.go
rdpsin: https://github.com/awslabs/soci-snapshotter/blob/main/integration/create_test.go#L32

hanyuel (dismissed):

Kern-- (approved):

hanyuel (approved):

Issue #252, opened 2022-12-21
[BUG] `soci index rm` doesn't handle failure case proplerly


djdongjin:
**Describe the bug**

`soci index rm indexDigest` doesn't return an error code and/or print a warning/error message when the indexDigest doesn't exist.

**Steps To Reproduce**
```shell
➜  soci-snapshotter git:(main) sudo soci index ls
DIGEST                                                                     SIZE    IMAGE REF                          PLATFORM       CREATED
sha256:b21cba14e88c476d703f6ec6a99d75bbba40cb6193e292da007b90014fcd9218    369     docker.io/library/alpine:latest    linux/amd64    n/a
sha256:c066d1313c553f44765af6104338ae62ecb0cd5f804f09b6541a887497a24d34    730     docker.io/library/alpine:latest    linux/amd64    n/a
➜  soci-snapshotter git:(main) sudo soci index rm
➜  soci-snapshotter git:(main) echo $?
0
➜  soci-snapshotter git:(main) sudo soci index rm fsadfsdafsdafsadfds
➜  soci-snapshotter git:(main) echo $?
0
```

**Expected behavior**

The command should print an error message and return a non-zero exit code if the rm fails, similar to `soci index info`:

```shell
➜  soci-snapshotter git:(main) sudo soci index info
soci: invalid checksum digest format
➜  soci-snapshotter git:(main) sudo soci index info fdsafasdfsdafd
soci: invalid checksum digest format
➜  soci-snapshotter git:(main) echo $?
1
➜  soci-snapshotter git:(main) sudo soci index info sha256:b21cba14e88c476d703f6ec6a99d75bbba40cb6193e292da007b90014fcd9fff
soci: sha256:b21cba14e88c476d703f6ec6a99d75bbba40cb6193e292da007b90014fcd9fff: : not found
➜  soci-snapshotter git:(main) echo $?
1
```

**Configuration (please complete the following information):**
 - OS: Amazon Linux 2
 - Snapshotter Version: main(2cf2a1607c016406946f79d01130dbfc304ed8e5)
 - Containerd Version: containerd github.com/containerd/containerd v1.7.0-beta.0-294-g36641eade 36641eade608deb2159ca4ac9973ea059f17570a

**Additional context**

Issue #251, opened 2022-12-21
[BUG] `soci index rm` doesn't really delete the index data in storage


djdongjin:
**Describe the bug**

After `soci index rm indexDigest`, `soci index ls` doesn't show the index, but `soci index info` can still get the index.

**Steps To Reproduce**
```bash
$ sudo soci index list
DIGEST                                                                     SIZE    IMAGE REF                          PLATFORM       CREATED
sha256:1f223fc2899404ec0e4ce34f0ea0b975f3355c9232d2d6ff0f05978b3d094db1    1083    docker.io/library/nginx:latest     linux/arm64    n/a
sha256:b21cba14e88c476d703f6ec6a99d75bbba40cb6193e292da007b90014fcd9218    369     docker.io/library/alpine:latest    linux/amd64    n/a
sha256:c066d1313c553f44765af6104338ae62ecb0cd5f804f09b6541a887497a24d34    730     docker.io/library/alpine:latest    linux/amd64    n/a

$ sudo soci index rm sha256:1f223fc2899404ec0e4ce34f0ea0b975f3355c9232d2d6ff0f05978b3d094db1

$ sudo soci index list
DIGEST                                                                     SIZE    IMAGE REF                          PLATFORM       CREATED
sha256:b21cba14e88c476d703f6ec6a99d75bbba40cb6193e292da007b90014fcd9218    369     docker.io/library/alpine:latest    linux/amd64    n/a
sha256:c066d1313c553f44765af6104338ae62ecb0cd5f804f09b6541a887497a24d34    730     docker.io/library/alpine:latest    linux/amd64    n/a

$ sudo soci index info sha256:1f223fc2899404ec0e4ce34f0ea0b975f3355c9232d2d6ff0f05978b3d094db1
[...]

PR #260, opened 2022-12-28
Add integration tests for `soci index rm` command

djdongjin:

*Issue #, if available:*

Part of #54 

*Description of changes:*

Add 3 integration tests for `soci index rm`.

*Testing performed:*

make test && make check && make integration

### DIFF
--- a/integration/index_test.go
+++ b/integration/index_test.go
@@ -177,3 +177,46 @@ func TestSociIndexList(t *testing.T) {
 		})
 	}
 }
+
+func TestSociIndexRemove(t *testing.T) {
+	sh, done := newSnapshotterBaseShell(t)
+	defer done()
+	rebootContainerd(t, sh, "", "")
+
+	t.Run("soci index rm indexDigest removes an index", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


hanyuel:
Should the integ tests be designed to fail with the bug #251 still open?

djdongjin:
> Should the integ tests be designed to fail with the bug #251 still open?

Ideally yes. 

The integration tests in this PR are only based on the current behavior. E.g., 1/ `soci index rm` only deletes metadata (so not shown in `soci index list`) but not delete in storage (so `soci index info` still work); 2/ `soci index rm invalidString` doesn't retrun error or print error message.

When we fix #251, we should add extra checks that check the removed index is also removed from storage (e.g., `soci index info indexDigest` returns error, or check the blob storage).

Similarly, when we fix #252, we should add an integration test that check `soci index rm randomStringOrDigest` should return an error.

These extra checks / new integration tests will be addition to the integration tests in this PR.

Kern-- (approved):

hanyuel (approved):

PR #261, opened 2022-12-28
Allow metrics to listen on a unix socket

### LINKED ISSUE #256: Allow metrics server to listen on a UNIX socket
Currently, the `metrics_address` config value is expected to be a network address. Since the bound port must be known in order to scrape the metrics, it is usually fixed. This means that another process in the same network namespace cannot bind 0.0.0.0 on the same port as the snapshotter's metrics server.

It would be convenient to be able to bind the metrics server to a unix socket instead. 

N/A

**Additional context**
N/A

Kern--:
*Issue #, if available:*
closes #256 

*Description of changes:*
This change allows the snapshotter to be configured to listen on a unix socket for serving metrics. This can be useful when the snapshotter runs in an environment where we don't want to consume a port (e.g. because we expect another program to bind 0.0.0.0 to the same port).

*Testing performed:*

### DIFF
--- a/cmd/soci-snapshotter-grpc/main.go
+++ b/cmd/soci-snapshotter-grpc/main.go
@@ -83,6 +83,7 @@ const (
 	defaultLogLevel            = logrus.InfoLevel
 	defaultRootDir             = "/var/lib/soci-snapshotter-grpc"
 	defaultImageServiceAddress = "/run/containerd/containerd.sock"
+	defaultMetricsNetwork      = "tcp"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


hanyuel (approved):

cmd/soci-snapshotter-grpc/main.go
```diff
@@ -232,12 +236,23 @@ func serve(ctx context.Context, rpc *grpc.Server, addr string, rs snapshots.Snap
 
 	errCh := make(chan error, 1)
 
+	var cleanupFns []func() error
+	defer func() {
```
sbuckfelder, line 240: I like this

PR #270, opened 2023-01-03
Add fuse package and separate grpc errors and logs in getting started doc

sparr:
*Description of changes:*
1. mention optional dependency on `fuse` package
2. separate stdout and stderr when running soci-snapshotter-grpc so that logs remain valid json

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -29,6 +29,9 @@ https://github.com/awslabs/soci-snapshotter.git`
 Often fixed with `sudo yum install zlib-devel`
 - having the proper go version installed, as of this document please use 1.18+
 
+**fuse** - confirm FUSE is installed for mounting userspace filesystems with `fusermount -V`
+(optional, enables mounting without root access)
+
 ## Start a local ORAS registry
 Most OCI registries do not currently support the "referrers" feature.
 SOCI relies on this feature to link its indices and manifests.  The [ORAS
@@ -231,6 +234,12 @@ In the example command we arbitrarily are pushing the logs to a file named
 
 ```
 sudo ./soci-snapshotter-grpc&> ~/soci-snapshotter-logs &
+```
+
+Alternately, you can split up stdout (json logs) and stderr (plain text errors):
+
+```
+sudo ./soci-snapshotter-grpc 2> ~/soci-snapshotter-errors 1> ~/soci-snapshotter-logs &
 ```
 
 Once the snapshotter is running we can call the rpull command from the SOCI CLI.


docs/GETTING_STARTED.md
```diff
@@ -29,6 +29,8 @@ https://github.com/awslabs/soci-snapshotter.git`
 Often fixed with `sudo yum install zlib-devel`
 - having the proper go version installed, as of this document please use 1.18+
 
+**fuse** - confirm FUSE is installed for mounting userspace filesystems with `fusermount -V` (optional)
```
wmesard, line 32: This change raises more questions than it answers. "Optional prerequisite" sounds like a bit of an oxymoron. A prerequisite--that is, all the items in this section--should be required to use SOCI and/or get through the rest of this Getting Started guide. If we're going to list this as optional, we should describe *why* it's optional; i.e., what won't work as a result of not having it.

wmesard (changes requested):

docs/GETTING_STARTED.md
sparr: "optional prerequisite" doesn't sound any more oxymoronic to me than "optional dependency", which is what this is.

re what won't work as a result of not having it, I think the answer to that is fuse mounts without root access. I could add something about that here?

wmesard: > noun: pre-requisite  
>    a thing that is required as a prior condition for something else to happen or exist.

> I could add something about that here

Sounds good.

Kern-- (dismissed):

djdongjin (approved):

docs/GETTING_STARTED.md
```diff
@@ -29,6 +29,9 @@ https://github.com/awslabs/soci-snapshotter.git`
 Often fixed with `sudo yum install zlib-devel`
 - having the proper go version installed, as of this document please use 1.18+
 
+**fuse** - confirm FUSE is installed for mounting userspace filesystems with `fusermount -V`
+(optional, enables mounting without root access)
```
wmesard, line 33: I guess. This does need to be documented. And I guess until we have a user guide, this is a place to do it. But a *Getting Started* guide should be low-friction. Every added step, every added bit of information, every chance for something to go wrong translates to a non-zero number of users who will walk away before successfully using SOCI. Since the examples below all use `sudo`, I'm still left wondering why I'm being told about this now.

When this *is* added to the [non-existent] user guide, it should be completely documented, since the code is [a bit opaque](https://github.com/awslabs/soci-snapshotter/blob/106fdbb1ea903aa2535cb0395f8764d72892e796/fs/fs.go#L509) about the implications. When I look at the effect that this has in the [fuse client code](https://github.com/hanwen/go-fuse/blob/master/fuse/mount_linux.go#L137-L138), it makes me want to *uninstall* the fuse package. I trust *mount()*. I don't trust "magic".

wmesard (approved):

docs/GETTING_STARTED.md
```diff
@@ -230,7 +233,7 @@ In the example command we arbitrarily are pushing the logs to a file named
 `~/soci-snapshotter-logs`, but please feel free to choose a different location:
 
 ```
-sudo ./soci-snapshotter-grpc&> ~/soci-snapshotter-logs &
+sudo ./soci-snapshotter-grpc 2> ~/soci-snapshotter-errors 1> ~/soci-snapshotter-logs &
```
vkuzniet, line 236: What about keeping both options (everything in one place and stdout is separate from stderr) with clear definitions of which workflow is more feasible for which? 

And with that, you can just write that it is recommended to use one of the options as the default.

docs/GETTING_STARTED.md
kzys: Is this because fusermount is typically setuid'd?

https://github.com/libfuse/libfuse#security-implications

> The fusermount3 program is installed setuid root. This is done to allow normal users to mount their own filesystem implementations.

As libfuse's README said,

> FUSE (Filesystem in Userspace) is an interface for userspace programs to export a filesystem to the Linux kernel. The FUSE project consists of two components: the fuse kernel module (maintained in the regular kernel repositories) and the libfuse userspace library (maintained in this repository). libfuse provides the reference implementation for communicating with the FUSE kernel module.

We should separate the kernel module and the libfuse userspace library along with userspace binaries. The former is must, the latter is optional.

Then regarding the optional dependency, can you confirm that the snapshotter works without root permissions if there is fusermount? If it works end-to-end, mentioning that is valuable for customers, if not, we don't have to ask customers to have fusermount.

djdongjin (approved):

docs/GETTING_STARTED.md
sparr: I was really hoping with this PR to just make the existing documentation and reported behavior more clear. I can open an issue to do further research if necessary.

sbuckfelder (approved):

Kern-- (approved):

docs/GETTING_STARTED.md
sparr: https://github.com/awslabs/soci-snapshotter/issues/389

PR #277, opened 2023-01-03
Add metrics for overlayfs fallback

### LINKED ISSUE #263: Add metric for OverlayFS Fallback
Currently, if something goes wrong loading a ztoc, we log a message. We should also emit a metric in this case.

We should emit a prometheus metric when we can't load a ztoc for some reason. It should not be emitted when we fallback to overlayfs because there is no ztoc for the layer in the index.

N/A

**Additional context**
N/A

rdpsin:
This commit adds a metric `overlay_mount_fallback_count` for when the snapshotter falls back to using overlayfs instead of FUSE mount for mounting the layer.

Note that a layer not having a corresponding ztoc is NOT classified as an error, even though `fs.Mount` returns an error in that case and the layer is mounted as a normal overlayfs mount.

*Issue #, if available:*
Resolves #263 

*Testing performed:*

1) Pulling an image with some layers not having corresponding ztoc does not increase metric.

```
% sudo ./soci create docker.io/library/rabbitmq:latest
layer sha256:cae1c228689d2f294439828746d67d06536e5cf331393d0529e18ce4c8f222f8 -> ztoc skipped
layer sha256:2fa407913eb6838a8d49fffcfb8a03cf3c0baa23f3744a8e40947031171c97a6 -> ztoc skipped
layer sha256:92e035229b114a43bd976252b210632f191d41ae606b4eafaf029f0d06c0f454 -> ztoc skipped
layer sha256:c17f6444939027393a0458ce0a87ed947fc83abeb8f29f24ee93a418d33b141c -> ztoc skipped
layer sha256:34bc881cb21a868fef333c95673b993c83a9c9ce4f62d320e73191cb453ef582 -> ztoc skipped
layer sha256:dd3bdcc0ac44e0e8e87eb1f7f7703302579ca7016343eb675cfd96dffb00ff01 -> ztoc skipped
layer sha256:03dea6b5e70e630bb5bf621528463788e46cf2571f98f3f591e10633f1b650d9 -> ztoc skipped
layer sha256:c07a0c5fb44ca573ef37ad94dc3c0456f08c49b4e7591a9fde8a04dd2e29c143 -> ztoc sha256:4311ab5a6df31034221275c229463cdd07102971a2ffd9873f8fed7486e12233
layer sha256:846c0b181fff0c667d9444f8378e8fcfa13116da8d308bf21673f7e4bea8d580 -> ztoc sha256:38902e5b8587d142e2dbc2f16bbcb045e02a8de9ad38f7c91427fbe62a937f1c
layer sha256:3ddf939b446499c0937604d712f7e13340db9687e710d61c8b0b278700ba9e38 -> ztoc sha256:1ab89fcafc5d2353b92cfb73d3ffde91ed3e5a272309e132dde828f0bd5cca15

% sudo ./soci index ls
DIGEST                                                                     SIZE    IMAGE REF                            PLATFORM       CREATED    
sha256:f117044f3ed73c5fb9780ca4a2ba50dc2f1bfa1aaf320738bb007cb73982df6a    1436    docker.io/library/rabbitmq:latest    linux/amd64    25s ago    

sudo ./soci image rpull --soci-index-digest sha256:f117044f3ed73c5fb9780ca4a2ba50dc2f1bfa1aaf320738bb007cb73982df6a docker.io/library/rabbitmq:latest
fetching sha256:90c62dd5... application/vnd.docker.distribution.manifest.list.v2+json
fetching sha256:603be6b7... application/vnd.docker.distribution.manifest.v2+json
fetching sha256:3ddcc140... application/vnd.docker.container.image.v1+json

% curl localhost:6060/metrics | grep fallback_count
 // Empty output, equivalent to 0
 ```
 
 2) Pulling an image with invalid SOCI digest results in non-zero metrics count.
 
 ```
 % sudo ./soci image rpull --soci-index-digest fafafafafafafa docker.io/library/rabbitmq:latest
fetching sha256:90c62dd5... application/vnd.docker.distribution.manifest.list.v2+json
fetching sha256:603be6b7... application/vnd.docker.distribution.manifest.v2+json
fetching sha256:3ddcc140... application/vnd.docker.container.image.v1+json


% curl localhost:6060/metrics | grep fallback_count 
soci_fs_operation_count{layer="",operation_type="overlay_mount_fallback_count"} 10 // number of layers pulled synchronously
```

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -245,6 +245,8 @@ func NewFilesystem(root string, cfg config.Config, opts ...Option) (_ snapshot.F
 }
 
 type sociContext struct {
+	cachedErr            error
+	cachedErrMu          sync.RWMutex
 	bgFetchPauseOnce     sync.Once
 	fetchOnce            sync.Once
 	sociIndex            *soci.Index
@@ -254,6 +256,14 @@ type sociContext struct {
 func (c *sociContext) Init(ctx context.Context, imageRef, indexDigest, imageManifestDigest string, store orascontent.Storage) error {
 	var retErr error
 	c.fetchOnce.Do(func() {
+		defer func() {
+			if retErr != nil {
+				c.cachedErrMu.Lock()
+				c.cachedErr = retErr
+				c.cachedErrMu.Unlock()
+			}
+		}()
+
 		refspec, err := reference.Parse(imageRef)
 		if err != nil {
 			retErr = err
@@ -296,6 +306,9 @@ func (c *sociContext) Init(ctx context.Context, imageRef, indexDigest, imageMani
 		c.sociIndex = index
 		c.populateImageLayerToSociMapping(index)
 	})
+	c.cachedErrMu.RLock()
+	retErr = c.cachedErr
+	c.cachedErrMu.RUnlock()
 	return retErr
 }
 
@@ -376,7 +389,6 @@ func (fs *filesystem) getSociContext(ctx context.Context, imageRef, indexDigest,
 func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[string]string) (retErr error) {
 	// Setting the start time to measure the Mount operation duration.
 	start := time.Now()
-
 	ctx = log.WithLogger(ctx, log.G(ctx).WithField("mountpoint", mountpoint))
 
 	sociIndexDigest, ok := labels[source.TargetSociIndexDigestLabel]
@@ -411,11 +423,11 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 		errChan    = make(chan error)
 	)
 	go func() {
-		rErr := fmt.Errorf("failed to resolve target")
+		var rErr error
 		for _, s := range src {
 			sociDesc, ok := c.imageLayerToSociDesc[s.Target.Digest.String()]
 			if !ok {
-				rErr = errors.Wrapf(rErr, "unable to resolve layer %q from %q, mounting entire layer using overlayfs: %s", s.Target.Digest, s.Name, "no ztoc for the layer")
+				rErr = fmt.Errorf("unable to resolve layer %q from %q, mounting entire layer using overlayfs: %w", s.Target.Digest, s.Name, snapshot.ErrNoZtoc)
 				continue
 			}
 
@@ -424,7 +436,7 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 				resultChan <- l
 				return
 			}
-			rErr = errors.Wrapf(rErr, "failed to resolve layer %q from %q: %v", s.Target.Digest, s.Name, err)
+			rErr = fmt.Errorf("failed to resolve layer %q from %q: %w", s.Target.Digest, s.Name, err)
 		}
 		errChan <- rErr
 	}()
@@ -438,7 +450,7 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 			ctx := log.WithLogger(context.Background(), log.G(ctx).WithField("mountpoint", mountpoint))
 			sociDesc, ok := c.imageLayerToSociDesc[desc.Digest.String()]
 			if !ok {
-				log.G(ctx).Debugf("unable to resolve layer %q, mounting entire layer using overlayfs: %s", desc.Digest, "no ztoc for the layer")
+				log.G(ctx).Debugf("unable to resolve layer %q, mounting entire layer using overlayfs: %v", desc.Digest, snapshot.ErrNoZtoc)
 				return
 			}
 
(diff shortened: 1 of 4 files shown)


fs/metrics/common/metrics.go
```diff
@@ -73,6 +73,10 @@ const (
 
 	// TODO this metric is not available now. This needs to go down to BlobReader where the actuall http call is issued
 	SynchronousBytesFetched = "synchronous_bytes_fetched"
+
+	// Number of times the snapshotter falls back to use a normal overlay mount instead of mounting the layer as a FUSE mount.
+	// Note that a layer not having a ztoc is NOT classified as an error, even though `fs.Mount` returns an error in that case.
+	OverlayMountFallbackCount = "overlay_mount_fallback_count"
```
rdpsin, line 79: Not sure about the name. Maybe something like `synchronous_layer_fetch` fits better?

Kern--: `synchronous_layer_fetch_count` applies equally to the "no ztoc" case. 

What we really want to communicate with an operator here is that lazy loading for this layer failed. Our naming here and in logs seems like it's also trying to communicate that this isn't fatal for the container launch, but I wonder if this is downplaying the severity. Operators should be monitoring this and alarming if it spikes because it represents a significant degradation in the experience.

With that in mind, what about something like `lazy_mount_failure_count`? Something still feels slightly off about that to me, but I can't quite place what it is.

Kern-- (changes requested):

fs/metrics/common/metrics.go
hanyuel: I'm fine with ``overlay_mount_fallback_count``

hanyuel (approved):

fs/metrics/common/metrics.go
hanyuel: I don't think I've ever heard the term "lazy mount". ``fuse_mount_faillure_count`` sounds more straightforward to me if we report it as an error metric.

fs/fs.go
```diff
@@ -428,7 +440,7 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 			ctx := log.WithLogger(context.Background(), log.G(ctx).WithField("mountpoint", mountpoint))
 			sociDesc, ok := c.imageLayerToSociDesc[desc.Digest.String()]
 			if !ok {
-				log.G(ctx).Debugf("unable to resolve layer %q, mounting entire layer using overlayfs: %s", desc.Digest, "no ztoc for the layer")
+				log.G(ctx).Debugf("unable to resolve layer %q, mounting entire layer using overlayfs: %v", desc.Digest, snapshot.ErrNoZtoc)
```
djdongjin, line 443: line 420 uses `%w` for `ErrNoZtoc` and here is using `%v`.

fs/metrics/common/metrics.go
rdpsin: Changed to `asynchronous_mount_failure_count`.

fs/fs.go
rdpsin: Unlike line 420, we're just logging the error, not returning it.

fs/metrics/common/metrics.go
Kern--: Can we go with @hanyuel's suggestion instead? I think it's the most clear out of the ideas here.

Kern-- (approved):

fs/fs.go
```diff
@@ -287,6 +297,9 @@ func (c *sociContext) Init(ctx context.Context, imageRef, indexDigest, imageMani
 		c.sociIndex = index
 		c.populateImageLayerToSociMapping(index)
 	})
+	c.cachedErrMu.Lock()
+	retErr = c.cachedErr
+	c.cachedErrMu.Unlock()
```
sparr, line 302: This could be an Rlock on a RWMutex to allow simultaneous reads. This is not a blocking review, feel free to resolve and merge if nothing else is in the way and you aren't pushing any further changes otherwise.

djdongjin (approved):

integration/metrics_test.go
```diff
@@ -64,3 +72,123 @@ func TestMetrics(t *testing.T) {
...
+			indexDigestFn: func(sh *shell.Shell, image imageInfo) string {
+				return optimizeImageWithOpts(sh, image, 1<<22, 0)
+			},
+			expectedFallbackCount: 0,
+		},
+		{
+			name:  "image with some layers not having ztoc and no fs.Mount results in 0 overlay fallback",
+			image: "rabbitmq:latest",
+			indexDigestFn: func(s1 *shell.Shell, image imageInfo) string {
+				return optimizeImageWithOpts(sh, image, 1<<22, 10<<20)
```
djdongjin, line 141: typo: `s1 *shell.Shell` -> `sh *shell.Shell`

hanyuel (approved):

integration/metrics_test.go
```diff
@@ -64,3 +72,123 @@ func TestMetrics(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			rebootContainerd(t, sh, "", "")
+
+			imgInfo := dockerhub(tc.image)
+
+			sh.X("ctr", "i", "pull", imgInfo.ref)
+			indexDigest := tc.indexDigestFn(sh, imgInfo)
```
djdongjin, line 184: Not sure if it has impact, but `ctr i pull` is called twice for some testcases: one here, one in `optimizeImageWithOpts`

https://github.com/awslabs/soci-snapshotter/blob/2cf2a1607c016406946f79d01130dbfc304ed8e5/integration/push_test.go#L211

rdpsin: Hmm, I don't think it has any impact.

Kern-- (approved):

Issue #249, opened 2022-12-21
Make layer mount timeout configurable


rdpsin:
**Is your feature request related to a problem? Please describe.**
Currently, the layer mount timeout is set to 30s. 

https://github.com/awslabs/soci-snapshotter/blob/2cf2a1607c016406946f79d01130dbfc304ed8e5/fs/fs.go#L453-L455

We should make it configurable. 

**Describe the solution you'd like**
A field in the snapshotter's config enables end users to configure the Mount timeout as needed. 

**Describe alternatives you've considered**
Keep it as 30s.

PR #279, opened 2023-01-03
Configurable layer mount timeout

hanyuel:
*Issue #, if available:* #249

*Description of changes:*
Make layer mount timeout configurable, defaulting to 30 seconds. It can be set in the config toml file of the snapshotter.

*Testing performed:*
- ``make test``, ``make integration``. 
- Set ``mount_timeout`` in the snapshotter's config file, made the layer resolving pending, and verified ``Mount`` timed out after ``mount_timeout`` seconds.

### DIFF
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -63,6 +63,7 @@ type Config struct {
 	MaxConcurrency                   int64  `toml:"max_concurrency"`
 	NoPrometheus                     bool   `toml:"no_prometheus"`
 	PrioritizedTaskSilencePeriodMSec int    `toml:"prioritized_task_silence_period_msec"`
+	MountTimeoutSec                  int    `toml:"mount_timeout_sec"`
 
 	// BlobConfig is config for layer blob management.
 	BlobConfig `toml:"blob"`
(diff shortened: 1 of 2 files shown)


fs/config/config.go
```diff
@@ -63,6 +63,7 @@ type Config struct {
 	MaxConcurrency                   int64  `toml:"max_concurrency"`
 	NoPrometheus                     bool   `toml:"no_prometheus"`
 	PrioritizedTaskSilencePeriodMSec int    `toml:"prioritized_task_silence_period_msec"`
+	MountTimeout                     int    `toml:"mount_timeout"`
```
Kern--, line 66: We're kind of inconsistent about this, but we sometimes include the unit as a suffix for durations.

e.g. CacheConfig uses `_sec`
https://github.com/awslabs/soci-snapshotter/blob/main/fs/config/config.go#L81

but FuseCondig doesn't
https://github.com/awslabs/soci-snapshotter/blob/main/fs/config/config.go#L97

I think it makes sense to use the same convention everywhere and my preference would be to use the suffixes. What do you think?

hanyuel: ``MountTimeoutSec``? Sure

turan18 (approved):

djdongjin (approved):

Kern-- (approved):

PR #280, opened 2023-01-04
Add fuse operation failure metrics

### LINKED ISSUE #274: [FEATURE] Add FUSE operation failure metrics
Currently, we use a state file in the FUSE filesystem to report errors that occur in that filesystem. This is good for logging errors, but doesn't provide metrics.

A prometheus metric for number of failures for each FUSE operation.

We could alternatively parse the state file and create metrics from that data, however this seems like an extra step seeing as the snapshotter already emits metrics itself.

**Additional context**

djdongjin:

*Issue #, if available:*

Fix #274

*Description of changes:*

1. Add failure count metric for fuse operations
2. Format op name strings to const

*Testing performed:*

### DIFF
--- a/fs/layer/node.go
+++ b/fs/layer/node.go
@@ -88,6 +88,38 @@ var opaqueXattrs = map[OverlayOpaqueType][]string{
 	OverlayOpaqueUser:    {"user.overlay.opaque"},
 }
 
+const (
+	// fuse operations
+	fuseOpGetattr         = "node.Getattr"
+	fuseOpGetxattr        = "node.Getxattr"
+	fuseOpListxattr       = "node.Listxattr"
+	fuseOpLookup          = "node.Lookup"
+	fuseOpOpen            = "node.Open"
+	fuseOpReaddir         = "node.Readdir"
+	fuseOpFileRead        = "file.Read"
+	fuseOpFileGetattr     = "file.Getattr"
+	fuseOpWhiteoutGetattr = "whiteout.Getattr"
+)
+
+var fuseOpFailureMetrics = map[string]string{
+	fuseOpGetattr:         commonmetrics.FuseNodeGetattrFailureCount,
+	fuseOpListxattr:       commonmetrics.FuseNodeListxattrFailureCount,
+	fuseOpLookup:          commonmetrics.FuseNodeLookupFailureCount,
+	fuseOpOpen:            commonmetrics.FuseNodeOpenFailureCount,
+	fuseOpReaddir:         commonmetrics.FuseNodeReaddirFailureCount,
+	fuseOpFileRead:        commonmetrics.FuseFileReadFailureCount,
+	fuseOpFileGetattr:     commonmetrics.FuseFileGetattrFailureCount,
+	fuseOpWhiteoutGetattr: commonmetrics.FuseWhiteoutGetattrFailureCount,
+}
+
+func incFuseOpFailureMetric(operationName string, layer digest.Digest) {
+	metric, ok := fuseOpFailureMetrics[operationName]
+	if !ok {
+		metric = commonmetrics.FuseUnknownFailureCount
+	}
+	commonmetrics.IncOperationCount(metric, layer)
+}
+
 func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseInode uint32, opaque OverlayOpaqueType, logFSOperations bool) (fusefs.InodeEmbedder, error) {
 	rootID := r.Metadata().RootID()
 	rootAttr, err := r.Metadata().GetAttr(rootID)
@@ -176,7 +208,7 @@ var _ = (fusefs.InodeEmbedder)((*node)(nil))
 var _ = (fusefs.NodeReaddirer)((*node)(nil))
 
 func (n *node) Readdir(ctx context.Context) (fusefs.DirStream, syscall.Errno) {
-	n.logOperation(ctx, "Readdir")
+	n.logOperation(ctx, fuseOpReaddir)
 	ents, errno := n.readdir()
 	if errno != 0 {
 		return nil, errno
@@ -223,7 +255,8 @@ func (n *node) readdir() ([]fuse.DirEntry, syscall.Errno) {
 		})
 		return true
 	}); err != nil || lastErr != nil {
-		n.fs.s.report(fmt.Errorf("node.Readdir: err = %v; lastErr = %v", err, lastErr))
+		incFuseOpFailureMetric(fuseOpReaddir, n.fs.layerDigest)
+		n.fs.s.report(fmt.Errorf("%s: err = %v; lastErr = %v", fuseOpReaddir, err, lastErr))
 		return nil, syscall.EIO
 	}
 
@@ -232,7 +265,8 @@ func (n *node) readdir() ([]fuse.DirEntry, syscall.Errno) {
 		if !normalEnts[w[len(whiteoutPrefix):]] {
 			ino, err := n.fs.inodeOfID(id)
 			if err != nil {
-				n.fs.s.report(fmt.Errorf("node.Readdir: err = %v; lastErr = %v", err, lastErr))
+				incFuseOpFailureMetric(fuseOpReaddir, n.fs.layerDigest)
+				n.fs.s.report(fmt.Errorf("%s: err = %v; lastErr = %v", fuseOpReaddir, err, lastErr))
 				return nil, syscall.EIO
 			}
 			ents = append(ents, fuse.DirEntry{
@@ -256,6 +290,7 @@ func (n *node) readdir() ([]fuse.DirEntry, syscall.Errno) {
 var _ = (fusefs.NodeLookuper)((*node)(nil))
 
 func (n *node) Lookup(ctx context.Context, name string, out *fuse.EntryOut) (*fusefs.Inode, syscall.Errno) {
+	n.logOperation(ctx, fuseOpLookup)
 
 	isRoot := n.isRootNode()
 
@@ -275,19 +310,22 @@ func (n *node) Lookup(ctx context.Context, name string, out *fuse.EntryOut) (*fu
 		case *node:
 			ino, err := n.fs.inodeOfID(tn.id)
 			if err != nil {
-				n.fs.s.report(fmt.Errorf("node.Lookup: %v", err))
+				incFuseOpFailureMetric(fuseOpLookup, n.fs.layerDigest)
+				n.fs.s.report(fmt.Errorf("%s: %v", fuseOpLookup, err))
 				return nil, syscall.EIO
 			}
 			entryToAttr(ino, tn.attr, &out.Attr)
 		case *whiteout:
 			ino, err := n.fs.inodeOfID(tn.id)
 			if err != nil {
-				n.fs.s.report(fmt.Errorf("node.Lookup: %v", err))
+				incFuseOpFailureMetric(fuseOpLookup, n.fs.layerDigest)
+				n.fs.s.report(fmt.Errorf("%s: %v", fuseOpLookup, err))
 				return nil, syscall.EIO
 			}
 			entryToAttr(ino, tn.attr, &out.Attr)
 		default:
-			n.fs.s.report(fmt.Errorf("node.Lookup: uknown node type detected"))
+			incFuseOpFailureMetric(fuseOpLookup, n.fs.layerDigest)
+			n.fs.s.report(fmt.Errorf("%s: uknown node type detected", fuseOpLookup))
 			return nil, syscall.EIO
 		}
 		return cn, 0
@@ -312,7 +350,8 @@ func (n *node) Lookup(ctx context.Context, name string, out *fuse.EntryOut) (*fu
 		if whID, wh, err := n.fs.r.Metadata().GetChild(n.id, fmt.Sprintf("%s%s", whiteoutPrefix, name)); err == nil {
 			ino, err := n.fs.inodeOfID(whID)
 			if err != nil {
-				n.fs.s.report(fmt.Errorf("node.Lookup: %v", err))
+				incFuseOpFailureMetric(fuseOpLookup, n.fs.layerDigest)
+				n.fs.s.report(fmt.Errorf("%s: %v", fuseOpLookup, err))
 				return nil, syscall.EIO
 			}
 			return n.NewInode(ctx, &whiteout{
@@ -327,7 +366,8 @@ func (n *node) Lookup(ctx context.Context, name string, out *fuse.EntryOut) (*fu
 
 	ino, err := n.fs.inodeOfID(id)
 	if err != nil {
-		n.fs.s.report(fmt.Errorf("node.Lookup: %v", err))
+		incFuseOpFailureMetric(fuseOpLookup, n.fs.layerDigest)
+		n.fs.s.report(fmt.Errorf("%s: %v", fuseOpLookup, err))
 		return nil, syscall.EIO
 	}
 	return n.NewInode(ctx, &node{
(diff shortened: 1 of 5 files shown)


fs/layer/node.go
```diff
@@ -88,7 +88,34 @@ var opaqueXattrs = map[OverlayOpaqueType][]string{
 	OverlayOpaqueUser:    {"user.overlay.opaque"},
 }
 
-func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseInode uint32, opaque OverlayOpaqueType, logFSOperations bool) (fusefs.InodeEmbedder, error) {
+const (
+	// fuse operations
+	fuseOpGetattr     = "Getattr"
```
djdongjin, line 93: If we can rename these node operation as `node.opName`, then we can reuse them in state file reports instead of using strings. (e.g., `n.fs.s.report(fmt.Errorf("node.Open: %v", err)`).

fs/layer/node.go
```diff
@@ -413,15 +459,18 @@ func (n *node) Listxattr(ctx context.Context, dest []byte) (uint32, syscall.Errn
 
 var _ = (fusefs.NodeReadlinker)((*node)(nil))
 
-func (n *node) Readlink(ctx context.Context) ([]byte, syscall.Errno) {
-	n.logOperation(ctx, "ReadLink")
+func (n *node) Readlink(ctx context.Context) (bt []byte, errno syscall.Errno) {
+	n.logOperation(ctx, fuseOpReadlink)
+	defer n.incOperationFailureMetric(ctx, fuseOpReadlink, errno)
```
djdongjin, line 464: I added log and metrics for this one although it always return 0. I think it's worthwhile unless it's called very frequently. (same for `Statfs`)

Just in case later on if we change the implementation to returnning non-zero on error we don't need to add log/metric again.

fs/layer/node.go
```diff
@@ -160,6 +189,16 @@ func (n *node) logOperation(ctx context.Context, operationName string) {
 	}
 }
 
+func (n *node) incOperationFailureMetric(ctx context.Context, operationName string, errno syscall.Errno) {
+	if errno != 0 && !n.fs.noPrometheus {
```
rdpsin, line 193: Is `!n.fs.noPrometheus` necessary?

djdongjin: `logFSOperations bool` gives me an impression that metrics should only be released when `!noPrometheus` but it seems not the case :) I'll remove it.

djdongjin:
> Can you post an example of the metrics in the `Testing` section?

Actually I haven't thought about a good test case to test fuse operation failure. Only checked the metrics can be captured if the `errno != 0` restriction is moved. Will spend sometime looking into the part of code.

Thanks for any suggestion on how to integration test fuse operation failure :)

rdpsin:
A good exercise would be to construct ztoc's containing invalid data (invalid dictionaries and/or invalid metadata). Create an index from those ztocs, write them to the local content store, and do a normal `rpull`. Something like https://github.com/awslabs/soci-snapshotter/blob/4f78a38da307bf53dda3030aaa5ea860767697c4/integration/pull_test.go#L570-L613 but the ztocs are still de-serializable.

fs/layer/node.go
djdongjin: fixed

fs/metrics/common/metrics.go
```diff
@@ -71,6 +71,21 @@ const (
...
+	// fuse operation failure metrics
+	// - node level
+	FuseGetattrFailureCount   = "fuse_getattr_failure_count"
+	FuseGetxattrFailureCount  = "fuse_getxattr_failure_count"
+	FuseListxattrFailureCount = "fuse_listxattr_failure_count"
+	FuseLookupFailureCount    = "fuse_lookup_failure_count"
+	FuseOpenFailureCount      = "fuse_open_failure_count"
+	FuseReadlinkFailureCount  = "fuse_readlink_failure_count"
+	FuseReaddirFailureCount   = "fuse_readdir_failure_count"
+	FuseStatfs                = "fuse_statfs_failure_count"
```
hanyuel, line 83: Why does ``FuseStatfs`` not follow the name``*FailureCount`` ?

fs/layer/node.go
hanyuel: We shouldn't log an error metric if there's no error. It would only cause confusions.

hanyuel (dismissed):

fs/layer/node.go
```diff
@@ -413,15 +457,18 @@ func (n *node) Listxattr(ctx context.Context, dest []byte) (uint32, syscall.Errn
...
 	ent := n.attr
 	return []byte(ent.LinkName), 0
 }
 
 var _ = (fusefs.NodeStatfser)((*node)(nil))
 
-func (n *node) Statfs(ctx context.Context, out *fuse.StatfsOut) syscall.Errno {
+func (n *node) Statfs(ctx context.Context, out *fuse.StatfsOut) (errno syscall.Errno) {
+	n.logOperation(ctx, fuseOpStatfs)
+	defer n.incOperationFailureMetric(ctx, fuseOpStatfs, errno)
```
hanyuel, line 471: Same as my comment above. No errors here

fs/metrics/common/metrics.go
hanyuel: I see it's logged for every ``Statfs`` operation. I suggest removing this metric as this PR is about fuse failure metrics.

fs/layer/node.go
djdongjin: SGTM. I'll remove this one and below.

integration/metrics_test.go
```diff
@@ -192,3 +201,189 @@ disable_verification = %v
...
+	for _, blob := range index.Blobs {
+		ztocDigest := blob.Digest.String()
+		blobContent := fetchContentFromPath(sh, blobStorePath+"/"+trimSha256Prefix(ztocDigest))
+		zt, err := ztoc.Unmarshal(bytes.NewReader(blobContent))
+		if err != nil {
+			t.Fatalf("invalid ztoc %s from soci index %s: %v", ztocDigest, indexDigest, err)
+		}
+
+		// manipulate the ztoc
+		for i, md := range zt.TOC.Metadata {
```
rdpsin, line 341: I think we can have another test case where fs metadata is correct but dictionaries are corrupted.

integration/metrics_test.go
```diff
@@ -192,3 +201,189 @@ disable_verification = %v
...
+			if !strings.Contains(line, "fuse") || strings.Contains(line, commonmetrics.FuseMountFailureCount) {
+				continue
+			}
+
+			parts := strings.Split(line, " ")
+			if metricCount, err := strconv.Atoi(parts[len(parts)-1]); err == nil && metricCount != 0 {
+				if !expectFuseOperationFailure {
+					t.Fatalf("unexpected fuse operation failure: %s", line)
+				}
+				t.Logf("fuse operation failure metric: %s", line)
```
rdpsin, line 292: Do we need this log?

djdongjin: I think we need. if we are not expecting fuse errors but there are errors from metrics output, we want to know what the  fuse error is

integration/metrics_test.go
djdongjin: do you mean this `CompressionInfo`? what's the expected fuse operation failures by manipulating these 3 fields?

https://github.com/awslabs/soci-snapshotter/blob/e83d75262b736cff39f44d7e66ab7422f2837792/ztoc/ztoc.go#L60-L62

rdpsin: Yes. Incorrect `SpanDigests` would mean that compressed data will not be available locally. Incorrect `Checkpoints` would mean that decompression will fail.

integration/ztoc_test.go
```diff
@@ -100,3 +93,11 @@ func TestSociZtocList(t *testing.T) {
 		}
 	})
 }
+
+func sociIndexFromDigest(t *testing.T, sh *shell.Shell, indexDigest string) (index soci.Index) {
```
rdpsin, line 97: Maybe return an error here and do a `t.Fatal` in the test? I don't know how I feel about non-test functions being able to fail the test.

integration/ztoc_test.go
```diff
@@ -100,3 +93,11 @@ func TestSociZtocList(t *testing.T) {
...
+	rawSociIndexJSON := sh.O("soci", "index", "info", indexDigest)
+	if err := json.Unmarshal(rawSociIndexJSON, &index); err != nil {
+		t.Fatalf("invalid soci index from digest %s: %v", indexDigest, rawSociIndexJSON)
```
rdpsin, line 100: Can you add `err` here?

rdpsin (dismissed):

integration/metrics_test.go
djdongjin: I think we can add it in a follow up PR. Here I refactored the `buildIndexByManipulatingZtocData` to accept a `manipulator` func, so it's easy to try different manipulations.

I found even valid index/ztoc produces some fuse failures, so we need to first figure out _what ztoc manipulation leads to which fuse failure_, and then check the specific fuse failure metric.

integration/ztoc_test.go
djdongjin: changed both `sociIndexFromDigest` and `buildIndexByManipulatingZtocData` to not require a `t *tesing.T` but return an error.

djdongjin: also move `sociIndexFromDigest` to `index_test.go`, since it's more related to `index` and used by multiple test files. (probably we should have some `testutil_index.go, testutil_ztoc.go` etc, but before we have that it might be good to collocate these utils with most relevant tests)

integration/index_test.go
```diff
@@ -220,3 +222,11 @@ func TestSociIndexRemove(t *testing.T) {
...
 	})
 }
+
+func sociIndexFromDigest(sh *shell.Shell, indexDigest string) (index soci.Index, err error) {
+	rawSociIndexJSON := sh.O("soci", "index", "info", indexDigest)
+	if err = json.Unmarshal(rawSociIndexJSON, &index); err != nil {
+		err = fmt.Errorf("invalid soci index from digest %s: %s", indexDigest, err)
+	}
+	return
+}
```
sparr, line 232: Could you move this out to testutil and refactor TestSociIndexInfo to use this? Will save work in the future redundancy reducing refactor.

djdongjin: Will refactor `TestSociIndexInfo` to use this. But I would prefer to keep it in `index_test.go` since it's more relevant. Right now `testutil.go` is more for environment setup (shell, containerd, snapshotter, etc) (and already really long :)).

Personally I think we should have some `testutil_ztoc.go`/`testutil_index.go` to host ztoc/index related test logics. That should be a separate refactor and before that we can keep the relevant helpers in more relevant `*_test.go` for easy discovery

djdongjin: fixed

integration/metrics_test.go
```diff
@@ -192,3 +201,204 @@ disable_verification = %v
...
+				return indexDigest
+			},
+			metricToCheck:              commonmetrics.FuseFileReadFailureCount,
+			expectFuseOperationFailure: true,
+		},
+	}
+
+	// checkFuseOperationFailureMetrics checks if output from metrics endpoint includes
+	// a specific fuse operation failure metrics (or any fuse op failure if empty string is given)
+	checkFuseOperationFailureMetrics := func(t *testing.T, output string, metricToCheck string, expectFailure bool) {
```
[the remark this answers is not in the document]
djdongjin: fixed.

integration/metrics_test.go
```diff
@@ -192,3 +201,204 @@ disable_verification = %v
...
+	}{
+		{
+			name:  "image with valid ztocs and index doesn't cause fuse file.read failures",
+			image: rabbitmqImage,
+			indexDigestFn: func(t *testing.T, sh *shell.Shell, image imageInfo) string {
+				return optimizeImageWithOpts(sh, image, 1<<22, 0)
+			},
+			// even a valid index/ztoc produces some fuse failure such as
+			// fuse_lookup_failure_count, fuse_getxattr_failure_count, so we only
+			// check a specific fuse failure metric.
```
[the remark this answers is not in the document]
djdongjin: All fuse failures from a container with valid index (at least for the test) are `Lookup` and `Getxattr`. And some logs follow a pattern of `Lookup*n, Getxattr*n, Open, file.Read`. So I guess these `Lookup` and `Getxattr` failures are expected?  as the snapshotter goes down layers to find a file on a specific layer.

```
# logs from a valid container+index
{"level":"debug","msg":"FUSE operation","operation":"Lookup","path":"usr/lib","time":"2023-01-10T20:48:49.925292563Z"}
# look up `usr/lib/x86_64-linux-gnu` twice
{"level":"debug","msg":"FUSE operation","operation":"Lookup","path":"usr/lib/x86_64-linux-gnu","time":"2023-01-10T20:48:49.925420845Z"}
{"level":"debug","msg":"FUSE operation","operation":"Lookup","path":"usr/lib/x86_64-linux-gnu","time":"2023-01-10T20:48:49.925546599Z"}
{"level":"debug","msg":"FUSE operation","operation":"Getxattr","path":"usr/lib/x86_64-linux-gnu/ld-2.31.so","time":"2023-01-10T20:48:49.925726346Z"}
{"level":"debug","msg":"FUSE operation","operation":"Getxattr","path":"usr/lib/x86_64-linux-gnu/ld-2.31.so","time":"2023-01-10T20:48:49.925849631Z"}
{"level":"debug","msg":"FUSE operation","operation":"Open","path":"usr/lib/x86_64-linux-gnu/ld-2.31.so","time":"2023-01-10T20:48:49.925953246Z"}
{"level":"debug","msg":"FUSE operation","operation":"file.Read","path":"usr/lib/x86_64-linux-gnu/ld-2.31.so","time":"2023-01-10T20:48:49.926048155Z"}
```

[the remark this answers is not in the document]
kzys: I wonder what we can do with these metrics if we constantly have some failures. Looking the trend might be helpful though.

Is it possible to differentiate cases where an image doesn't have the requested file (which is mostly fine) and other cases where it actually indicates some issues?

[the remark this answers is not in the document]
djdongjin: Looking into how we use the state file to record fuse failures:

For `Lookup`, it only records the failure on specific errors (e.g., not all non-0 errno): [1](https://github.com/awslabs/soci-snapshotter/blob/7827c4f0d219714c035173214528f8a4448c3d01/fs/layer/node.go#L314), [2](https://github.com/awslabs/soci-snapshotter/blob/7827c4f0d219714c035173214528f8a4448c3d01/fs/layer/node.go#L321), [3](https://github.com/awslabs/soci-snapshotter/blob/7827c4f0d219714c035173214528f8a4448c3d01/fs/layer/node.go#L326).
For `Getxattr`, it doesn't record failures at all: [link](https://github.com/awslabs/soci-snapshotter/blob/7827c4f0d219714c035173214528f8a4448c3d01/fs/layer/node.go#L408).

Maybe we can do the same for metrics as well? (to be consistent when we record failure in state file, instead of for all non-0 errno).

integration/metrics_test.go
```diff
@@ -192,3 +201,204 @@ disable_verification = %v
...
+	newIndexPath := fmt.Sprintf("%s/%s", blobStorePath, trimSha256Prefix(newIndexDigest.String()))
+	if err := testutil.WriteFileContents(sh, newIndexPath, b, 0600); err != nil {
+		return "", fmt.Errorf("cannot write index %s to path %s: %s", newIndexDigest, newIndexPath, err)
+	}
+	return strings.Trim(newIndexDigest.String(), "\n"), nil
+}
+
+// checkFuseOperationFailureMetrics checks if output from metrics endpoint includes
+// a specific fuse operation failure metrics (or any fuse op failure if an empty string is given)
+func checkFuseOperationFailureMetrics(t *testing.T, output string, metricToCheck string, expectOpFailure bool) {
```
rdpsin, line 381: Can this also return an error instead of doing `t.Fatal`?

fs/metrics/common/metrics.go
```diff
@@ -71,6 +71,19 @@ const (
 	SynchronousReadRegistryFetchCount = "synchronous_read_remote_registry_fetch_count" // TODO revisit (wrong place)
 	SynchronousBytesServed            = "synchronous_bytes_served"
 
+	// fuse operation failure metrics
+	// - node level
+	FuseGetattrFailureCount   = "fuse_getattr_failure_count"
```
rdpsin, line 76: Maybe these should be `fuse_node_*`, similar to `fuse_file_read_failure_count`?

integration/metrics_test.go
djdongjin: yeah I thought about this too, but `t` is used to log fuse op failure lines. That'll be helpful to debug if we have unexpected fuse failures in the tests.

integration/metrics_test.go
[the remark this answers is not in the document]
djdongjin: updated to make the fuse failure metrics consistent with when/how we record fuse failure in state file. It avoids confusing metrics (high failure metric count but no/less failure record in state file) and helps fine-tune fuse failure in both state file and metrics.

kzys (approved):

PR #285, opened 2023-01-05
Add test to confirm image removal results in fuse unmounting

### LINKED ISSUE #175: Add more production-like integration tests
The current set of integration tests verify that the basic lazy loading features offered by the SOCI snapshotter are working. There are, however, some gaps between the tests and they way the snapshotter will be used in a production system. It would be helpful to bridge some of those gaps to give us confidence that the current state as well as the state of each PR is production ready.

- [x] A set of integration test that tests running multiple lazy-loaded containers simultaneously (with various compositions, e.g. 5 of the same container, 5 different containers, etc)
- [ ] An integration test that verifies that mounts are cleaned up appropriately - we can launch a container with `--rm` and verify that there are no FUSE mounts after the container exits.
- [x] An integration test that verifies that 2 of the same container running simultaneously can't write to each other

There is room for additional refactoring and unit tests to give us confidence in production readiness. More at metrics and logs to improve debuggability of the snapshotter would help as well.

**Additional context**
N/A

sparr:
*Issue #, if available:*
closes #175 

*Description of changes:*
Add an integration test that pulls and rpulls an image, confirms the presence of fuse mounts, then removes the image and confirms the fuse mounts have been unmounted.

*Testing performed:*
Ran the test with additional debug logging to check status of image and mounts between test steps.

### DIFF
--- a/integration/pull_test.go
+++ b/integration/pull_test.go
@@ -1086,3 +1086,65 @@ func addConfig(t *testing.T, sh *shell.Shell, conf string, cmds ...string) []str
 	}
 	return append(cmds, "--config", configPath)
 }
+
+// TestRpullImageThenRemove pulls and rpulls an image then removes it to confirm fuse mounts are unmounted
+func TestRpullImageThenRemove(t *testing.T) {
+	regConfig := newRegistryConfig()
+	sh, done := newShellWithRegistry(t, regConfig)
+	defer done()
+
+	// Initialize config files for containerd and snapshotter
+	additionalConfig := ""
+	if !isTestingBuiltinSnapshotter() {
+		additionalConfig = proxySnapshotterConfig
+	}
+	containerdConfigYaml := testutil.ApplyTextTemplate(t, `
+version = 2
+
+[plugins."io.containerd.snapshotter.v1.soci"]
+root_path = "/var/lib/soci-snapshotter-grpc/"
+
+[plugins."io.containerd.snapshotter.v1.soci".blob]
+check_always = true
+
+{{.AdditionalConfig}}
+`, struct {
+		AdditionalConfig string
+	}{
+		AdditionalConfig: additionalConfig,
+	})
+
+	// Setup environment
+	if err := testutil.WriteFileContents(sh, defaultContainerdConfigPath, []byte(containerdConfigYaml), 0600); err != nil {
+		t.Fatalf("failed to write %v: %v", defaultContainerdConfigPath, err)
+	}
+	sh.
+		X("update-ca-certificates").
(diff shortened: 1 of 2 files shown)


integration/pull_test.go
```diff
@@ -1086,3 +1086,54 @@ func addConfig(t *testing.T, sh *shell.Shell, conf string, cmds ...string) []str
...
+
+	rebootContainerd(t, sh, "", "")
+
+	containerImage := nginxImage
+
+	copyImage(sh, dockerhub(containerImage), regConfig.mirror(containerImage))
+	indexDigest := optimizeImage(sh, regConfig.mirror(containerImage))
+	sh.X("soci", "image", "rpull", "--user", regConfig.creds(), "--soci-index-digest", indexDigest, regConfig.mirror(containerImage).ref)
+
+	checkFuseMounts(t, sh, 6)
```
sparr, line 1133: Magic number could be calculated from the manifest for the downloaded image. Is that desired here?

vkuzniet: In the case w/o sparse index, yes. However, the current state of snapshotter has the `min-layer-size` set to non-zero value, which implies having a sparse index. This magic number can still be calculated from manifest, but you will need to specify `min-layer-size` explicitly when building soci index for that test.

rdpsin:
We should also test the scenario where sending a SIGTERM to the snapshotter should clean up the FUSE mounts. Should that be part of this PR or a separate PR?

sparr:
@rdpsin I'd prefer one new test (possibly with sub-tests) per commit, and so far the feedback I've gotten suggests one squashed commit per PR. Good call on that case, though. Open an Issue?

integration/pull_test.go
sparr: Could you please elaborate on this?

> the current state of snapshotter has the min-layer-size set to non-zero value

Kern--: We skip creating ztocs for layers that are smaller than `min-layer-size` (currently 10 MiB, I believe) so we might be wrong if we use the image manifest to figure out how many fuse mounts to expect.

Since the index iteself is the source of truth, I think we should get the number of ztocs from the index and verify that there are that many FUSE mounts.

Sort of like this:
https://github.com/awslabs/soci-snapshotter/blob/main/integration/index_test.go#L83-L87

We should probably also verify that there is at least 1 ztoc to make sure we don't accidentally raise the default min-layer-size so high that this test skips all layers.

sparr: 10<<20 bytes is the default, but optimizeImage specifies a min layer size of 0 when calling buildSparseIndex (which should produce a non-sparse index?)

Kern-- (approved):

djdongjin (approved):

PR #287, opened 2023-01-05
Add nil/zero slice check to gzip_zinfo

djdongjin:

*Issue #, if available:*

*Description of changes:*

This change checks if slices are zero-length and return error if so. Otherwise the snapshotter will `panic` if a zero/nil slice is used when taking pointer of the slice. (for example, a ztoc with valid format but malicious data within it)

*Testing performed:*

### DIFF
--- a/compression/gzip_zinfo.go
+++ b/compression/gzip_zinfo.go
@@ -34,6 +34,9 @@ type GzipZinfo struct {
 
 // NewGzipZinfo creates a new instance of `GzipZinfo` from cZinfo byte blob on zTOC.
 func NewGzipZinfo(checkpoints []byte) (*GzipZinfo, error) {
+	if len(checkpoints) == 0 {
+		return nil, fmt.Errorf("empty checkpoints")
+	}
 	cZinfo := C.blob_to_zinfo(unsafe.Pointer(&checkpoints[0]))
 	if cZinfo == nil {
 		return nil, fmt.Errorf("cannot convert blob to gzip_zinfo")
@@ -70,7 +73,7 @@ func (i *GzipZinfo) Close() {
 func (i *GzipZinfo) Bytes() ([]byte, error) {
 	blobSize := C.get_blob_size(i.cZinfo)
 	bytes := make([]byte, uint64(blobSize))
-	if bytes == nil {
+	if len(bytes) == 0 {
 		return nil, fmt.Errorf("could not allocate byte array of size %d", blobSize)
 	}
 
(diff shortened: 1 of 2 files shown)


sparr (approved):
LGTM

compression/gzip_zinfo.go
```diff
@@ -109,6 +112,12 @@ func (i *GzipZinfo) SpanIDToUncompressedOffset(spanID SpanID) Offset {
 // ExtractDataFromBuffer wraps the call to `C.extract_data_from_buffer`, which takes in the compressed bytes
 // and returns the decompressed bytes.
 func (i *GzipZinfo) ExtractDataFromBuffer(compressedBuf []byte, uncompressedSize, uncompressedOffset Offset, spanID SpanID) ([]byte, error) {
+	if len(compressedBuf) == 0 {
+		return nil, fmt.Errorf("empty compressed buffer")
+	}
+	if uncompressedSize <= 0 {
```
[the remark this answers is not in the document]
djdongjin: thanks for the reminder. changed to `error if <0, return []byte{} if == 0`.

compression/gzip_zinfo.go
```diff
@@ -70,7 +73,7 @@ func (i *GzipZinfo) Close() {
 func (i *GzipZinfo) Bytes() ([]byte, error) {
 	blobSize := C.get_blob_size(i.cZinfo)
 	bytes := make([]byte, uint64(blobSize))
-	if bytes == nil {
+	if len(bytes) == 0 {
```
rdpsin, line 76: Is this necessary? I think the Go runtime would panic if it can't allocate enough memory.

djdongjin: if the passed `blobSize` is 0,`make` will return an _empty (not nil)_ slice. So `bytes` will pass the if check and panic later when `&bytes[0]` is called. Both nil/empty slices have 0-length.

Kern-- (dismissed):

rdpsin:
LGTM. Can you add unit tests for this too?

djdongjin:
> LGTM. Can you add unit tests for this too?

How about creating an issue for this first? Currently `gzip_zinfo` doesn't have any unit test and I think we should add tests to it more thoroughly in a separate PR.

Also this change blocks #280. After adding integration test for #280, the error case triggers panic in gzip_zinfo instead of fuse operation failure.

vkuzniet:
> > LGTM. Can you add unit tests for this too?
> 
> How about creating an issue for this first? Currently `gzip_zinfo` doesn't have any unit test and I think we should add tests to it more thoroughly in a separate PR.
> 
> Also this change blocks #280. After adding integration test for #280, the error case triggers panic in gzip_zinfo instead of fuse operation failure.

You can at least add unit tests that test your changes in particular and create an issue to provide more extensive coverage. My guess, that writing unit tests for `gzip_zinfo` (within the new ticket) will not be high priority for a while and constantly be postponed, so having at least something for newly added code should definitely help.

djdongjin:
> > > LGTM. Can you add unit tests for this too?
> > 
> > 
> > How about creating an issue for this first? Currently `gzip_zinfo` doesn't have any unit test and I think we should add tests to it more thoroughly in a separate PR.
> > Also this change blocks #280. After adding integration test for #280, the error case triggers panic in gzip_zinfo instead of fuse operation failure.
> 
> You can at least add unit tests that test your changes in particular and create an issue to provide more extensive coverage. My guess, that writing unit tests for `gzip_zinfo` (within the new ticket) will not be high priority for a while and constantly be postponed, so having at least something for newly added code should definitely help.

SGTM. Will add some tests.

compression/gzip_zinfo_test.go
```diff
@@ -0,0 +1,141 @@
...
+			compressedBuf:    []byte("foobar"),
+			uncompressedSize: -1,
+			expectError:      true,
+		},
+		{
+			name:             "zero uncompressedSize should return empty byte slice",
+			gzipZinfo:        GzipZinfo{},
+			compressedBuf:    []byte("foobar"),
+			uncompressedSize: 0,
+			expectError:      false,
```
rdpsin, line 87: np: Don't really need this line.

rdpsin (approved):

Kern-- (approved):

PR #289, opened 2023-01-06
Improve fs.Mount logging

### LINKED ISSUE #272: Improve logging around OverlayFS fallback

When we create a snapshot for an image layer, we have a 3 step fall back process:
1) Attempt to mount as a FUSE filesystem using a ztoc
2) Attempt to mount as an overlayfs filesystem from the snapshotter itself
3) Defer to the containerd client to do a normal ahead of time pull

The logs we have to indicate when and why we do these fallback is confusing. If any step fails, we get an error and then we log a warning with both that error and an additional message. Below are the logs and what they mean:

---
__error__:
> failed to resolve layer: unable to resolve layer "$SHA" from "$SRC", mounting entire layer using overlayfs: no ztoc for the layer: failed to resolve target

__msg__:
> failed to prepare remote snapshot

__meaning__:
The SOCI index has no zTOC for this layer. This is a very common situation because lazy loading small layers causes more overhead than just downloading it ahead of time. 

---

__error__:
>  failed to resolve layer: failed to resolve layer "$SHA" from "$SRC": cannot get ztoc; download and unpack this layer in container runtime for now: cannot unmarshal ztoc: runtime error: slice bounds out of range [786502718:1000000]: failed to resolve target

__msg__:
> failed to prepare remote snapshot

__meaning__:
The SOCI index does have a zTOC, but we failed to use it to set up a FUSE filesystem for some reason. This is likely a real error that people care about - it shouldn't affect the container launch, but it will skip lazy loading. 

It's also worth noting that "download and unpack this layer in container runtime for now" is not correct. We will attempt to download and unpack the layer from within the snapshotter before we defer to the runtime client. 

---

__error__:
> cannot unpack the layer: cannot fetch layer: unable to fetch descriptor ($SHA) from remote store: GET "$SRC": response status code 403: denied: Your authorization token has expired. Reauthenticate and try again."

__msg__:
> failed to prepare snapshot; deferring to container runtime

__meaning__:
We didn't lazily load the layer (either because there was no zTOC or we failed to setup the FUSE filesystem - that will be in a previous log) and we also failed to download the layer in the snapshotter. We will defer to the runtime client to do a normal ahead-of-time pull.

---
1) The logs should clearly indicate that "no ztoc for layer" is not an error by making it an info log and changing "failed to resolve" to "skipping".
2) We currently have 3 variants of "unable to resolve layer" per real error message when preparing a remote snapshot - we should collapse that to 1.
3) Right now our terminology is "remote snapshot" and "snapshot" - we should change "snapshot" to "ahead-of-time snapshot" or similar to reduce ambiguity.

N/A

**Additional context**
N/A

rdpsin:

*Issue #, if available:*
Resolves #272

*Description of changes:*

*Testing performed:*

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -427,8 +427,12 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 		for _, s := range src {
 			sociDesc, ok := c.imageLayerToSociDesc[s.Target.Digest.String()]
 			if !ok {
-				rErr = fmt.Errorf("unable to resolve layer %q from %q, mounting entire layer using overlayfs: %w", s.Target.Digest, s.Name, snapshot.ErrNoZtoc)
-				continue
+				log.G(ctx).WithFields(logrus.Fields{
+					"layerDigest": s.Target.Digest.String(),
+					"image":       s.Name.String(),
+				}).Infof("skipping mounting layer as FUSE mount: %v", snapshot.ErrNoZtoc)
+				rErr = fmt.Errorf("skipping mounting layer %s as FUSE mount: %w", s.Target.Digest.String(), snapshot.ErrNoZtoc)
+				break
 			}
 
 			l, err := fs.resolver.Resolve(ctx, s.Hosts, s.Name, s.Target, sociDesc)
@@ -450,7 +454,7 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 			ctx := log.WithLogger(context.Background(), log.G(ctx).WithField("mountpoint", mountpoint))
 			sociDesc, ok := c.imageLayerToSociDesc[desc.Digest.String()]
 			if !ok {
-				log.G(ctx).Debugf("unable to resolve layer %q, mounting entire layer using overlayfs: %v", desc.Digest, snapshot.ErrNoZtoc)
+				log.G(ctx).WithError(snapshot.ErrNoZtoc).WithField("layerDigest", desc.Digest.String()).Debug("skipping layer pre-resolve")
 				return
 			}
 
@@ -470,12 +474,14 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	select {
 	case l = <-resultChan:
 	case err := <-errChan:
-		log.G(ctx).WithError(err).Debug("failed to resolve layer")
-		retErr = fmt.Errorf("failed to resolve layer: %w", err)
+		retErr = err
 		return
 	case <-time.After(fs.mountTimeout):
-		log.G(ctx).Debug("failed to resolve layer (timeout)")
-		retErr = fmt.Errorf("failed to resolve layer (timeout)")
+		log.G(ctx).WithFields(logrus.Fields{
+			"timeout":     fs.mountTimeout.String(),
+			"layerDigest": labels[source.TargetDigestLabel],
+		}).Info("timeout waiting for layer to resolve")
+		retErr = fmt.Errorf("timeout waiting for layer %s to resolve", labels[source.TargetDigestLabel])
 		return
 	}
 	defer func() {
@@ -494,7 +500,7 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	node, err := l.RootNode(0)
 	if err != nil {
 		log.G(ctx).WithError(err).Warnf("Failed to get root node")
-		retErr = errors.Wrapf(err, "failed to get root node")
+		retErr = fmt.Errorf("failed to get root node: %w", err)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


fs/fs.go
```diff
@@ -427,7 +427,11 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 		for _, s := range src {
 			sociDesc, ok := c.imageLayerToSociDesc[s.Target.Digest.String()]
 			if !ok {
-				rErr = fmt.Errorf("unable to resolve layer %q from %q, mounting entire layer using overlayfs: %w", s.Target.Digest, s.Name, snapshot.ErrNoZtoc)
+				log.G(ctx).WithFields(logrus.Fields{
+					"layerDigest": s.Target.Digest.String(),
+					"image":       s.Name.String(),
+				}).Info("skipping mounting layer as FUSE mount since it has no ztoc")
```
djdongjin, line 433: `Info(....)` -> `Infof("skipping mounting layer as FUSE mount: %v", snapshot.ErrNoZtoc)`

fs/fs.go
```diff
@@ -470,12 +474,12 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
...
-		log.G(ctx).WithError(err).Debug("failed to resolve layer")
-		retErr = fmt.Errorf("failed to resolve layer: %w", err)
+		log.G(ctx).WithError(err).Debug("unable to resolve layer for lazy loading")
+		retErr = err
 		return
 	case <-time.After(fs.mountTimeout):
-		log.G(ctx).Debug("failed to resolve layer (timeout)")
-		retErr = fmt.Errorf("failed to resolve layer (timeout)")
+		log.G(ctx).WithField("timeout", fs.mountTimeout.String()).Info("timed out waiting for layer to resolve")
+		retErr = fmt.Errorf("timed out waiting for layer to resolve")
```
djdongjin, line 482: will this show which layer cause timeout?

djdongjin: nit: `timed out` -> `timeout`

rdpsin: It will now. :)

djdongjin (dismissed):

vkuzniet (dismissed):

turan18 (approved):

fs/fs.go
```diff
@@ -427,7 +427,11 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 		for _, s := range src {
 			sociDesc, ok := c.imageLayerToSociDesc[s.Target.Digest.String()]
 			if !ok {
-				rErr = fmt.Errorf("unable to resolve layer %q from %q, mounting entire layer using overlayfs: %w", s.Target.Digest, s.Name, snapshot.ErrNoZtoc)
+				log.G(ctx).WithFields(logrus.Fields{
+					"layerDigest": s.Target.Digest.String(),
+					"image":       s.Name.String(),
+				}).Infof("skipping mounting layer as FUSE mount: %v", snapshot.ErrNoZtoc)
+				rErr = fmt.Errorf("skipping mounting layer %s as FUSE mount: %w", s.Target.Digest.String(), snapshot.ErrNoZtoc)
 				continue
```
Kern--, line 435: nit: if there is more than 1 source, this message will be logged multiple times, right? Should we just break here?

fs/fs.go
```diff
@@ -470,12 +474,15 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	select {
 	case l = <-resultChan:
 	case err := <-errChan:
-		log.G(ctx).WithError(err).Debug("failed to resolve layer")
-		retErr = fmt.Errorf("failed to resolve layer: %w", err)
+		log.G(ctx).WithError(err).Debug("unable to resolve layer for lazy loading")
```
Kern--, line 477: Is this double logged since we logged it in `fs.Mount`?

fs/fs.go
```diff
@@ -427,8 +427,12 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
...
 			sociDesc, ok := c.imageLayerToSociDesc[s.Target.Digest.String()]
 			if !ok {
-				rErr = fmt.Errorf("unable to resolve layer %q from %q, mounting entire layer using overlayfs: %w", s.Target.Digest, s.Name, snapshot.ErrNoZtoc)
-				continue
+				log.G(ctx).WithFields(logrus.Fields{
+					"layerDigest": s.Target.Digest.String(),
+					"image":       s.Name.String(),
+				}).Infof("skipping mounting layer as FUSE mount: %v", snapshot.ErrNoZtoc)
+				rErr = fmt.Errorf("skipping mounting layer %s as FUSE mount: %w", s.Target.Digest.String(), snapshot.ErrNoZtoc)
+				break
```
hanyuel, line 435: Why changed to ``break``?

rdpsin: If a layer does not have a corresponding ztoc, it does not make sense to continue looking at other sources. So breaking out and returning early.

djdongjin (approved):

hanyuel (approved):

Issue #258, opened 2022-12-28
[BUG] `soci index info` outputs binary when given a ztoc digest


sparr:
```
% soci ztoc list
DIGEST                                                                     SIZE       LAYER DIGEST                                                               
sha256:1ab89fcafc5d2353b92cfb73d3ffde91ed3e5a272309e132dde828f0bd5cca15    1103000    sha256:3ddf939b446499c0937604d712f7e13340db9687e710d61c8b0b278700ba9e38    
sha256:38902e5b8587d142e2dbc2f16bbcb045e02a8de9ad38f7c91427fbe62a937f1c    1119144    sha256:846c0b181fff0c667d9444f8378e8fcfa13116da8d308bf21673f7e4bea8d580    
sha256:4311ab5a6df31034221275c229463cdd07102971a2ffd9873f8fed7486e12233    1055280    sha256:c07a0c5fb44ca573ef37ad94dc3c0456f08c49b4e7591a9fde8a04dd2e29c143 
```

```
% soci index info sha256:1ab89fcafc5d2353b92cfb73d3ffde91ed3e5a272309e132dde828f0bd5cca15
($01��
      0
       @�T�


             ��4��D��T�d�t$��4��D��TGsha256:528e6a801c3f5a830ab24173ac345483fec64a6ba13701861449ea496d54b2c8Gsha256:55898e4139874fb4b397b3ff7fa690602f92f1e0aa536c6e79e26a6d7569e253Gsha256:42b7331978563d48b2f3ec495
[output truncated]
```

This should probably print an error message, ideally pointing the user to `soci ztoc info`

Kern--:
The PR for this should include an integration test to verify the expected behavior.

djdongjin:
place to add integration test for this: https://github.com/awslabs/soci-snapshotter/blob/4bd409abdc9394459caae5de193fa323783c262f/integration/index_test.go#L73

PR #293, opened 2023-01-06
Improve "soci index info" command w/ ztoc digest as input

hanyuel:
*Issue #, if available:* #258

*Description of changes:*
When given a ztoc digest as input, "soci index info" command returns an error message pointing users to use "soci ztoc info" instead.

*Testing performed:*
make test, make integration

### DIFF
--- a/cmd/soci/commands/index/info.go
+++ b/cmd/soci/commands/index/info.go
@@ -18,10 +18,13 @@ package index
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 
 	"github.com/awslabs/soci-snapshotter/fs/config"
+	"github.com/awslabs/soci-snapshotter/soci"
+
 	"github.com/opencontainers/go-digest"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/urfave/cli"
@@ -38,6 +41,17 @@ var infoCommand = cli.Command{
 		if err != nil {
 			return err
 		}
+		db, err := soci.NewDB()
+		if err != nil {
+			return err
+		}
+		artifactType, err := db.GetArtifactType(digest.String())
+		if err != nil {
+			return err
+		}
+		if artifactType == soci.ArtifactEntryTypeLayer {
+			return fmt.Errorf("the provided digest is of ztoc not SOCI index. Use \"soci ztoc info\" command to get detailed info of ztoc")
+		}
 		storage, err := oci.New(config.SociContentStorePath)
 		if err != nil {
 			return err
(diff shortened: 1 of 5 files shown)


cmd/soci/commands/index/info.go
```diff
@@ -52,6 +62,20 @@ var infoCommand = cli.Command{
...
+	if err != nil {
+		return false, err
+	}
+	ae, err := db.GetArtifactEntry(digest)
+	if err != nil {
+		return false, err
+	}
+	if ae.Type == soci.ArtifactEntryTypeLayer {
+		return true, nil
+	}
```
djdongjin, line 79: why not check if `ae.Type == soci.ArtifactEntryTypeIndex` and only proceed if it's a soci index? Or just let it return `ArtifactEntryType` so it can be reused by `ztoc` (e.g., if we only want to get ztoc not soci index)

djdongjin (changes requested):

cmd/soci/commands/index/info.go
hanyuel: Because it would be nice to redirect users to ``soci ztoc`` command to get detailed info of ztoc, if the input is ztoc digest, which is what is proposed in the description of the ticket #258

djdongjin: We should show the error message to redirect users to `soci ztoc`, and also make this function resuable. Otherwise we'll have a almost identical function for `isIndexDigest`.`soci ztoc info` has a similar issue:

```shell
$ soci index list
DIGEST                                                                     SIZE    IMAGE REF                          PLATFORM       CREATED
sha256:e0ad85cffcf93aba3e1b75cb7297e1c8995b4714d966354be56b8fbbef98af6c    731     docker.io/library/ubuntu:latest    linux/amd64    95h36m35s ago
$ soci ztoc info sha256:e0ad85cffcf93aba3e1b75cb7297e1c8995b4714d966354be56b8fbbef98af6c
soci: cannot unmarshal ztoc: runtime error: slice bounds out of range [1701651067:731]
```

Suggest to name it `checkSociDigestType` and return the type, or `validateSociDigestType` and pass a digest and expected type.

hanyuel: We _are_ showing the error msg of redirecting users to use ``soci ztoc info``. Please check line 49.

djdongjin: Yes, I mean "I agree we should show the error message to redirect users to soci ztoc (your line 49), **but also make this function resuable**" :)

hanyuel: Added ``GetArtifactType`` method in ``soci/artifacts.go``

turan18 (approved):

djdongjin (approved):

PR #294, opened 2023-01-09
Refactor/deduplicate intergration test code

djdongjin:
*Issue #, if available:*

Part of #247

*Description of changes:*

It has 3 commits, each for an independent integration test refactor:

1. Use const for image names instead of plain strings.
2. Refactor containerd and snapshotter config creation: after the refactor, we should only use the helper function and pass the specific `additional config`. As an example: [`containerdMirrorConfig`](https://github.com/awslabs/soci-snapshotter/blob/ac0838de498d0aa9b26ba249a19265f17b7d0167/integration/pull_test.go#L709-L722), [`snapshotterMetricsConfig`](https://github.com/awslabs/soci-snapshotter/blob/ac0838de498d0aa9b26ba249a19265f17b7d0167/integration/metrics_test.go#L35-L42).
3. Move rebootContainerd to testutil: it's not related to `test_pull.go` and used almost by every integraion test.

*Testing performed:*

make test && make check && make integration

### DIFF
--- a/integration/testutil.go
+++ b/integration/testutil.go
@@ -65,6 +65,7 @@ const (
 	builtinSnapshotterFlagEnv    = "BUILTIN_SNAPSHOTTER"
 	buildArgsEnv                 = "DOCKER_BUILD_ARGS"
 	dockerLibrary                = "public.ecr.aws/docker/library/"
+	blobStorePath                = "/var/lib/soci-snapshotter-grpc/content/blobs/sha256"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


rdpsin (dismissed):

integration/testutil.go
```diff
@@ -506,3 +506,64 @@ func genRandomByteData(size int64) []byte {
...
+	m = testutil.NewRemoteSnapshotMonitor(testutil.NewTestingReporter(t), outR, errR)
+
+	// make sure containerd and soci-snapshotter-grpc are up-and-running
+	sh.Retry(100, "ctr", "snapshots", "--snapshotter", "soci",
+		"prepare", "connectiontest-dummy-"+xid.New().String(), "")
+
+	return m
+}
+
+func removeDirContents(sh *shell.Shell, dir string) {
```
kzys, line 559: Can't you `rm -fr {dir}` instead of using `find`?

djdongjin: This is a direct move of https://github.com/awslabs/soci-snapshotter/blob/cc303960c39d34d53066f8908495a116b046a25e/integration/pull_test.go#L1029-L1088, since it's more a `testutil` than some logic for `pull_test.go`. I didn't change any code so it's easier to review (e.g., only relocation with no slight change in a long `+++`.)

I agree `rm -rf dir` is more straightforward than original `find` way. Will update this function.

djdongjin: It seems `rm -rf dir` directly will cause some [device busy error](https://github.com/awslabs/soci-snapshotter/actions/runs/3886387536/jobs/6631429030#step:4:928). I think we can keep this part of code unchanged here since this PR is a refactor and other PRs can utilize the containerd/snapshotter config dedup from this PR early.

I'll make a separate PR to fix this (that will also easier to review). WDYT @kzys

kzys: Interesting. How about commenting the reasons to use `find` then? I'm fine not using `rm -fr`.

djdongjin: added comments. thanks

kzys (approved):

rdpsin (approved):

Issue #271, opened 2023-01-03
[FEATURE] Produce log output when container creation succeeds


sparr:
**Is your feature request related to a problem? Please describe.**
When containerd asks soci for a mount command, various failure modes will produce log output. However, success produces silence in the logs, making it hard to immediately tell whether containerd reached soci at all or fell back to a different snapshotter.

**Describe the solution you'd like**
One line of log output, probably at info level, declaring success.

**Describe alternatives you've considered**
Checking other logs and more directly monitoring activity while testing functionality for which success is uncertain.

turan18:
@sparr Do you mean when a FUSE mount is successful (IE per layer)? Or the final mount call when a container is being created/run

Kern--:
We already have logs that we're starting to work on a layer, unsuccessfully mounted as FUSE, and unsuccessfully mounted as overlay. I think it would make sense to add a per-layer log when either of those succeeds as well.

turan18:
re: It seems that per-layer both local and remote snapshot success is logged already.
https://github.com/awslabs/soci-snapshotter/blob/main/snapshot/snapshot.go#L288
https://github.com/awslabs/soci-snapshotter/blob/main/snapshot/snapshot.go#L311

The current log-level is set to debug so we can potentially set it to info, as that is the default level we set.

PR #301, opened 2023-01-09
Change local and remote snapshot success log level to Info 

turan18:
~~Added log statement to `Mounts()` function that will print if mount returns are successful on an active snapshot when a container is created/run.~~

Changed the log-level of success logs for remote or local snapshots to `Info` from `Debug` as `Info` is used by default.

*Issue #, if available:*
Fixes : https://github.com/awslabs/soci-snapshotter/issues/271

*Testing performed:*

`make test` && `make integration`

Log:
```
{
  "key": "default/1/extract-503501741-XIir sha256:0002c93bdb3704dd9e36ce5153ef637f84de253015f3ee330468dccdeacad60b",
  "level": "info",
  "msg": "Remote snapshot successfully prepared.",
  "parent": "",
  "remote-snapshot-prepared": "true",
  "time": "2023-01-10T21:26:05.821705631Z"
}

```

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -285,7 +285,7 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 			err := o.commit(ctx, true, target, key, append(opts, snapshots.WithLabels(base.Labels))...)
 			if err == nil || errdefs.IsAlreadyExists(err) {
 				// count also AlreadyExists as "success"
-				log.G(lCtx).WithField(remoteSnapshotLogKey, prepareSucceeded).Debug("prepared remote snapshot")
+				log.G(lCtx).WithField(remoteSnapshotLogKey, prepareSucceeded).Info("Remote snapshot successfully prepared.")
 				return nil, errors.Wrapf(errdefs.ErrAlreadyExists, "target snapshot %q", target)
 			}
 			log.G(lCtx).WithField(remoteSnapshotLogKey, prepareFailed).
@@ -308,7 +308,7 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 				// count also AlreadyExists as "success"
 				// there's no need to provide any details on []mount.Mount because mounting is already taken care of
 				// by snapshotter
-				log.G(lCtx).WithField(remoteSnapshotLogKey, prepareSucceeded).Debug("prepared local snapshot")
+				log.G(lCtx).WithField(remoteSnapshotLogKey, prepareSucceeded).Info("Local snapshot successfully prepared")
 				return nil, errors.Wrapf(errdefs.ErrAlreadyExists, "target snapshot %q", target)
 			}
 			log.G(lCtx).WithField(remoteSnapshotLogKey, prepareFailed).


snapshot/snapshot.go
```diff
@@ -343,7 +343,11 @@ func (o *snapshotter) Mounts(ctx context.Context, key string) ([]mount.Mount, er
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get active mount")
 	}
-	return o.mounts(ctx, s, key)
+	mounts, err := o.mounts(ctx, s, key)
+	if err == nil {
+		log.G(ctx).Info("SOCI mount successfully returned")
```
djdongjin, line 348: should we print which layer is mounted successfully? (`s.ID`)

turan18: As in the mount command? This will simply return the mount command with the upperdir and workdir set to the active snapshot.

turan18: Or did you mean the active snapshot digest?

djdongjin: I mean in the underlying `o.mounts` we log layer digest in the error message. should we also log the layer digest in this info line?

https://github.com/awslabs/soci-snapshotter/blob/eabacb9e7fd8c27fa27895aefaeaaaa9826ee831/snapshot/snapshot.go#LL627C6-L627C6

turan18: I think that makes sense for more verbosity.

snapshot/snapshot.go
```diff
@@ -343,7 +343,11 @@ func (o *snapshotter) Mounts(ctx context.Context, key string) ([]mount.Mount, er
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get active mount")
 	}
-	return o.mounts(ctx, s, key)
+	mounts, err := o.mounts(ctx, s, key)
+	if err == nil {
```
hanyuel, line 347: We log  "[prepared remote snapshot](https://github.com/awslabs/soci-snapshotter/blob/cc303960c39d34d53066f8908495a116b046a25e/snapshot/snapshot.go#L288)" when Mount is successful. I don't think this new line of log is necessary. I suggest amending the existing log message to ``SOCI mount successfully returned. Prepared remote snapshot``.

turan18: I may have misinterpreted what the issue might of have been getting at. Since currently successful remote snapshot mounts are logged, I thought that we may have needed to log if the snapshotter returns a successful set of mounts when running/creating a container.

hanyuel: Personally I don't think this PR is really needed. However I think it would be nice to amend the existing log msg to clearly state that mount is successful like I suggested above.

kzys (commented):
- There are Prepare and View that return mountpoints. Why do we log from Mount? Does containerd call Mount after calling Prepare and/or View?
- Let's say we have 2 large layers with zTOC and 2 small layers without zTOC, would we have 2 "SOCI mount ..." messages? Does it log something for the layers without zTOCs?

turan18:
> * There are Prepare and View that return mountpoints. Why do we log from Mount? Does containerd call Mount after calling Prepare and/or View?
> 
>     * Let's say we have 2 large layers with zTOC and 2 small layers without zTOC, would we have 2 "SOCI mount ..." messages? Does it log something for the layers without zTOCs?



I was under the impression that we wanted to log a success message when we are running a container, since on pull we already have logs that emit on successful remote or local snapshot creation.

https://github.com/awslabs/soci-snapshotter/blob/main/snapshot/snapshot.go#L288
https://github.com/awslabs/soci-snapshotter/blob/main/snapshot/snapshot.go#L311 


This log should be outputted only once when we attempt to run a container(either in read only or read/write) and containerd reaches out to the snapshotter to return the mounts. Before this during container creation, I believe Prepare would be called once to return the mounts with the active snapshot. 

https://github.com/containerd/containerd/blob/main/container.go#L243

turan18:
> > * There are Prepare and View that return mountpoints. Why do we log from Mount? Does containerd call Mount after calling Prepare and/or View?
> >   
> >   * Let's say we have 2 large layers with zTOC and 2 small layers without zTOC, would we have 2 "SOCI mount ..." messages? Does it log something for the layers without zTOCs?
> 
> I was under the impression that we wanted to log a success message when we are running a container, since on pull we already have logs that emit on successful remote or local snapshot creation. Layers without ztocs would be prepared locally.
> 
> https://github.com/awslabs/soci-snapshotter/blob/main/snapshot/snapshot.go#L288 https://github.com/awslabs/soci-snapshotter/blob/main/snapshot/snapshot.go#L311
> 
> This log should be outputted only once when we attempt to run a container(either in read only or read/write) and containerd reaches out to the snapshotter to return the mounts. Before this during container creation, I believe Prepare would be called once to return the mounts with the active snapshot.
> 
> https://github.com/containerd/containerd/blob/main/container.go#L243

No longer logging with Mounts. Context: https://github.com/awslabs/soci-snapshotter/issues/271#issuecomment-1377740994

kzys (approved):

rdpsin (approved):

hanyuel (approved):

Kern-- (approved):

PR #304, opened 2023-01-11
Add retries in case of span verification failure

### LINKED ISSUE #286: Gracefully handle span verification failures
Currently, at ztoc build time, we compute each span's sha256 digest and save it in the ztoc. At runtime, when we pull the span from the registry, we compute its digest and compare it to the one in the ztoc to make sure we've fetched the correct bytes. If span verification fails for some reason, we return an error, which translates to a read error for the container workload. However, it is entire possible that the incorrect digest was due to some kind of transient error, and retrying will fetch the correct bytes. 

1) Log the failure, and retry the fetch up to 'x' times. (configurable?)
2) If verification still fails after 'x' times, 
     - if it's a synchronous read, kill the container
     - if it's a background fetch, do nothing, because it is possible that the span may never be read. If a future sync read fetches the span, it will fail and kill the container. 
  

N/A

**Additional context**
During on-demand failure, why not return an error and let the application handle it? Because this is a bizarre failure mode, that most apps probably aren't coded to handle. It's like a single-sector failure on an HDD. Maybe we'll add a config option to do this at some point. But for now, we have to be conservative and kill the container.

rdpsin:
This commit adds retries to the span manager in case of span verification failure. The span manager will retry spanmanager.retries amount of times before returning an error. Added a config option span_manager.max_retries to set the retry value.

*Issue #, if available:*
Resolves #286 

*Description of changes:*

*Testing performed:*

### DIFF
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -82,6 +82,10 @@ type BlobConfig struct {
 	MaxRetries           int   `toml:"max_retries"`
 	MinWaitMSec          int   `toml:"min_wait_msec"`
 	MaxWaitMSec          int   `toml:"max_wait_msec"`
+
+	// MaxSpanVerificationRetries defines the number of times fetch will be invoked in case of
+	// span verification failure.
+	MaxSpanVerificationRetries int `toml:"max_span_verification_retries"`
 }
 
 type DirectoryCacheConfig struct {
--- a/fs/span-manager/span_manager.go
+++ b/fs/span-manager/span_manager.go
@@ -47,6 +47,11 @@ const (
 	uncompressed
 )
 
+const (
+	// Default number of tries fetching data from remote and verifying the digest.
+	defaultSpanVerificationFailureRetries = 3
+)
+
 // map of valid span transtions. Key is the current state and value is valid new states.
 var stateTransitionMap = map[spanState][]spanState{
 	unrequested:  {unrequested, requested},
@@ -92,12 +97,13 @@ func (s *span) validateStateTransition(newState spanState) error {
 }
 
 type SpanManager struct {
-	cache    cache.BlobCache
-	cacheOpt []cache.Option
-	index    *compression.GzipZinfo
-	r        *io.SectionReader // reader for contents of the spans managed by SpanManager
-	spans    []*span
-	ztoc     *ztoc.Ztoc
+	cache                             cache.BlobCache
+	cacheOpt                          []cache.Option
+	index                             *compression.GzipZinfo
+	r                                 *io.SectionReader // reader for contents of the spans managed by SpanManager
+	spans                             []*span
+	ztoc                              *ztoc.Ztoc
+	maxSpanVerificationFailureRetries int
 }
 
 type spanInfo struct {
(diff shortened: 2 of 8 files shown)


fs/span-manager/span_manager.go
```diff
@@ -98,6 +100,7 @@ type SpanManager struct {
 	r        *io.SectionReader // reader for contents of the spans managed by SpanManager
 	spans    []*span
 	ztoc     *ztoc.Ztoc
+	retries  int
```
sparr, line 103: Could we get a comment here? Is this max retries? Number of retries so far? Something else? I can figure it out from the code, but that's a lot of effort to figure out the meaning of the struct.

fs/span-manager/span_manager.go
```diff
@@ -126,6 +129,10 @@ func New(ztoc *ztoc.Ztoc, r *io.SectionReader, cache cache.BlobCache, cacheOpt .
 		r:        r,
 		spans:    spans,
 		ztoc:     ztoc,
+		retries:  retries,
+	}
+	if m.retries == 0 {
+		m.retries = defaultRetries
```
sparr, line 135: 0 seems like a dangerous sentinel value here. Someone might specify 0 when they actually want 0. -1 may be safer. This might also feel weird to golang devs who expect defaults to be 0 and 0s to be defaults, though. Fixing this is not a blocker, but a response/explanation before merge would be appreciated.

fs/config/config.go
```diff
@@ -74,6 +76,13 @@ type Config struct {
 	BackgroundFetchConfig `toml:"background_fetch"`
 }
 
+type SpanManagerConfig struct {
```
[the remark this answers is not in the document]
hanyuel: We do have [BlobConfig](https://github.com/awslabs/soci-snapshotter/blob/792a8996dc02311d9ca85c23fd18ae5563212f28/fs/config/config.go#L77) though.

fs/span-manager/span_manager.go
```diff
@@ -163,6 +170,28 @@ func (m *SpanManager) buildAllSpans() {
...
 
+// fetchWithRetries fetches the requested data and verifies that the span digest matches the one in the ztoc m.retries times.
+// It does not retry when there is an error fetching the data, because retries already happen lower in the stack in httpFetcher.
+// If there is an error fetching data from remote, it is not an transient error.
+func (m *SpanManager) fetchWithRetries(spanID compression.SpanID, buf []byte, offset compression.Offset) error {
+	var (
+		err error
+		n   int
+	)
+	for i := 0; i < m.retries; i++ {
```
hanyuel, line 181: If ``m.retries`` is set to 0, which is no retry, we do not fetch at all? That doesn't sound right.

hanyuel (changes requested):

fs/span-manager/span_manager.go
hanyuel: My comment relates to a comment from @sparr above about when user actually specifically wants to set retry to 0 and yet we convert 0 to the default value.

fs/span-manager/span_manager.go
rdpsin: Fair enough. I've renamed it to maxSpanVerificationTries, which now defaults to 1 + (number of retries requested). It will only default to `defaultRetries` if it's still less than or equal to 0.

fs/span-manager/span_manager.go
rdpsin: Fair enough. I've renamed it to maxSpanVerificationTries, which now defaults to 1 + (number of retries requested). It will only default to `defaultRetries` if it's still less than or equal to 0.

fs/span-manager/span_manager.go
```diff
@@ -45,6 +45,8 @@ const (
 	fetched
 	// A span is in Uncompressed state when it's uncompressed and its uncompressed content is cached.
 	uncompressed
+	// Default number of tries fetching data from remote and verifying the digest.
+	defaultRetries = 3
```
djdongjin, line 49: this will conflict with the `iota`. the `uncompressed`'s value is 3 based on its order.

suggest to move it out of this iota const block.

fs/span-manager/span_manager.go
```diff
@@ -92,12 +94,13 @@ func (s *span) validateStateTransition(newState spanState) error {
...
-	r        *io.SectionReader // reader for contents of the spans managed by SpanManager
-	spans    []*span
-	ztoc     *ztoc.Ztoc
+	cache                    cache.BlobCache
+	cacheOpt                 []cache.Option
+	index                    *compression.GzipZinfo
+	r                        *io.SectionReader // reader for contents of the spans managed by SpanManager
+	spans                    []*span
+	ztoc                     *ztoc.Ztoc
+	maxSpanVerificationTries int
```
djdongjin, line 103: we should keep the name consistent, in config it's `MaxSpanVerificationRetries`, in default it's `defaultRetries`

fs/span-manager/span_manager.go
```diff
@@ -113,19 +116,23 @@ type spanInfo struct {
...
+		cache:                    cache,
+		cacheOpt:                 cacheOpt,
+		index:                    index,
+		r:                        r,
+		spans:                    spans,
+		ztoc:                     ztoc,
+		maxSpanVerificationTries: retries + 1,
+	}
+	if m.maxSpanVerificationTries <= 0 {
+		m.maxSpanVerificationTries = defaultRetries
```
djdongjin, line 135: I feel this is a little confusing to me: try sometimes = retry+1 and sometimes try = retry

```
try = retry + 1
if try <= 0 (retry <= -1)
then try = default retry
```
which means we're using `retry` differently: for 1st, retry is the extra tries; for 2nd, retry is the total tries.

I suggest we only use `retry`: if retry >= 0, keep the value; if retry <0, use the default.
when fetch, we use `i < retry + 1` (minimum 1 fetch, plus x retry)

djdongjin (approved):

djdongjin (approved):

PR #308, opened 2023-01-12
Parameterize artifacts db

dvnguyen-amzn:
Currently artifacts db path is hard-coded at "/var/lib/soci-snapshotter-grpc/artifacts.db". It's problematic in Lambda environment as "/var/lib" is read-only in Lambda.

This commit adds an option to specify artifacts db the index builder initializer.

*Issue #, if available:* https://github.com/awslabs/soci-snapshotter/issues/307

*Description of changes:*
- Add a `path` parameter to the NewDB() initializer
- Add artifacts db path option to the soci index builder build option
 
*Testing performed:*
- Build and execute the `soci` cli locally
- Integrate soci_index module to a Lambda and test the Lambda

Signed-off-by: Dave Nguyen <dvnguyen@amazon.com>
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #309, opened 2023-01-12
Refactor test code

### LINKED ISSUE #247: De-duplicate integration test code

djdongjin:
*Issue #, if available:*

Fix #247. (This PR fixes remaining tasks in the issue, and I haven't seen new noticable refactors for test code)

*Description of changes:*

1. commit1: dedup genRandomByteData and move to `testutil` pkg, rename to `RandomByteData` to be consistent with existing helper like `RandomUInt64`.
2. commit2: simplify `optimizeImage`, `optimizeImageWithOpts`, `buildSparseIndex`, which are duplicated and confusing (e.g., what does `optimize` mean, which is basically build a soci index).
    1. Combine and rename them two 2 functions: `buildIndex` and `buildSparseIndex`.
    2. `buildIndex` builds an index with `min-layer-size=0` and `min-span-size=default`.
    3. `buildSparseIndex` builds an index by passing the two parameters.

*Testing performed:*

make test && make integration

### DIFF
--- a/integration/testutil_soci.go
+++ b/integration/testutil_soci.go
@@ -1,0 +1,72 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package integration
+
+import (
+	"fmt"
+	"strings"
+
+	shell "github.com/awslabs/soci-snapshotter/util/dockershell"
+	"github.com/containerd/containerd/platforms"
+	"github.com/opencontainers/go-digest"
+)
+
+const (
+	// copied from `soci/soci_index.go` for convenience so we don't always need to
+	// import the `soci` pkg only to use the default values.
+	// default span size (4MiB)
+	defaultSpanSize = int64(1 << 22)
+	// min layer size (10MiB)
+	defaultMinLayerSize = 10 << 20
+)
+
+// buildIndex builds a soci index with default span size and 0 min-layer-size, so every layer
+// will have a corresponding ztoc.
+func buildIndex(sh *shell.Shell, src imageInfo) string {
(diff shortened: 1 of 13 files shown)


integration/testutil_soci.go
```diff
@@ -0,0 +1,72 @@
...
+	shell "github.com/awslabs/soci-snapshotter/util/dockershell"
+	"github.com/containerd/containerd/platforms"
+	"github.com/opencontainers/go-digest"
+)
+
+const (
+	// copied from `soci/soci_index.go` for convenience so we don't always need to
+	// import the `soci` pkg only to use the default values.
+	// default span size (4MiB)
+	defaultSpanSize = int64(1 << 22)
```
hanyuel, line 48: Can we just use https://github.com/awslabs/soci-snapshotter/blob/792a8996dc02311d9ca85c23fd18ae5563212f28/soci/soci_index.go#L56-L58 ?

djdongjin: we can, but that means we need to import the soci pkg whenever it's used and use `soci.defaultSpanSize` (instead of `defaultSpanSize`). So I copied here for convenience.

Kern--:
On the topic of naming, we had a previous discussion that was tabled that we should probably rename all our "testutil" type files to end in `_test` so that they can't accidentally be imported into non-test packges:

https://github.com/awslabs/soci-snapshotter/pull/107#pullrequestreview-1141590789

We can probably defer that again for now, but we should create an issue this time and fix it.

integration/testutil_soci.go
```diff
@@ -0,0 +1,72 @@
...
+	// min layer size (10MiB)
+	defaultMinLayerSize = 10 << 20
+)
+
+// buildIndex builds a soci index with default span size and 0 min-layer-size, so every layer
+// will have a corresponding ztoc.
+func buildIndex(sh *shell.Shell, src imageInfo) string {
```
Kern--, line 55: I think functional parameters are a more idiomatic way to do this:

```
func buildIndex(sh *shell.Shell, src imageInfo, opts ...buildIndexOpts) string
```

The benefits being:
1) We have 1 name in our codebase
2) We only have to specify the options we care about (and not have a magic number when we want to skip one of min layer size and default span size)
3) We don't have to update all the call sites if we add additional parameters.

djdongjin: That makes sense to me, but I feel it worths a separate PR given that it requires more change and here we are just dedup some functions (and below):

This will be almost a direct copy of [soci.buildConfig](https://github.com/awslabs/soci-snapshotter/blob/792a8996dc02311d9ca85c23fd18ae5563212f28/soci/soci_index.go#L132-L167). Maybe we should just export `buildConfig` and use it in, e.g., `buildIndex`, instead of duplicating these opt functions.

WDYT @Kern--

rdpsin (approved):

integration/testutil_soci.go
Kern--: A separate PR is ok, but I think I'd prefer to do the duplication because this models CLI parameters which may be different than the library options. E.g. https://github.com/awslabs/soci-snapshotter/pull/308 adds an `ArtifactDB` to `buildConfig` which doesn't make sense in this context.

Kern-- (approved):

integration/testutil_soci.go
djdongjin: SGTM.

Issue #253, opened 2022-12-22
[BUG] `soci ztoc list --digest soc_index_digest` doesn't work as expected


djdongjin:
**Describe the bug**

The `--digest` flag and description is confusing (ztoc digest, soci index digest, or layer digest?) and has some bugs.

`soci index digest` seems more reasonable: list `ztoc`s belonging to that index. But now it refers to a ztoc digest, which doesn't quite useful (list `ztoc`s whose digest == given `ztoc` digest).

Also it doesn't validate the given digest.

**Steps To Reproduce**
```bash
# create an index for every layer
$ soci index ls
DIGEST    SIZE    IMAGE REF    PLATFORM    CREATED
$ soci create docker.io/library/ubuntu:latest --min-layer-size 0
layer sha256:6e3729cf69e0ce2de9e779575a1fec8b7fb5efdfa822829290ab6d5d1bc3e797 -> ztoc sha256:a50472c6492422c5839a6e601b9bf992a6652ea64f4ce7d58acfafe4367c5116
# soci index created
$ soci index ls --ref  docker.io/library/ubuntu:latest
DIGEST                                                                     SIZE    IMAGE REF                          PLATFORM       CREATED
sha256:e0ad85cffcf93aba3e1b75cb7297e1c8995b4714d966354be56b8fbbef98af6c    731     docker.io/library/ubuntu:latest    linux/amd64    54s ago
# given a soci index, return empty.
$ soci ztoc list --digest sha256:e0ad85cffcf93aba3e1b75cb7297e1c8995b4714d966354be56b8fbbef98af6c
DIGEST    SIZE    LAYER DIGEST
# given a layer digest, return empty.
$ soci ztoc list --digest sha256:6e3729cf69e0ce2de9e779575a1fec8b7fb5efdfa822829290ab6d5d1bc3e797
DIGEST    SIZE    LAYER DIGEST
# given a ztoc digest, return this ztoc.
$ soci ztoc list --digest sha256:a50472c6492422c5839a6e601b9bf992a6652ea64f4ce7d58acfafe4367c5116
DIGEST                                                                     SIZE       LAYER DIGEST
sha256:a50472c6492422c5839a6e601b9bf992a6652ea64f4ce7d58acfafe4367c5116    1196624    sha256:6e3729cf69e0ce2de9e779575a1fec8b7fb5efdfa822829290ab6d5d1bc3e797
# given an invalide digest, return empty and 0 exit code.
[...]

PR #314, opened 2023-01-13
`soci ztoc list` takes --ztoc-digest and --image-ref

hanyuel:
*Issue #, if available:* #253

*Description of changes:*
1)Add an optional parameter ``--image-ref`` that filters ztoc to those of a specific image
2)rename ``--digest`` parameter to ``--ztoc-digest``
3)Add ``--verbose`` flag that enables displaying extra debugging messages
Now we use the command like this
``soci ztoc list --ztoc-digest <ztoc-digest> --image-ref <ref> --v``

If both ``--image-ref`` and ``--ztoc-digest`` are provided in the command, we
will validate if the specified ztoc is with the specified image. If the
ztoc is not with the image, the command will return an error.

The command is designed to take image reference not SOCI index digest for the following reason:
To filter ztocs by a SOCI index, we need to know the layer digests associated with the SOCI index, because
In the artifacts DB, the ztoc entries have only [layer digests](https://github.com/awslabs/soci-snapshotter/blob/792a8996dc02311d9ca85c23fd18ae5563212f28/soci/soci_index.go#L343) but no index/image digests. And we are not able to get layer digests for SOCI index from the DB, because index artifact entries only have [image manifest and image digests](https://github.com/awslabs/soci-snapshotter/blob/792a8996dc02311d9ca85c23fd18ae5563212f28/soci/soci_index.go#L417-L418), unless we add layer digests to  index artifact db entries That would require changing the structure of the artifact entries, which I'm not inclined to do for a debugging command line tool.

*Testing performed:*
Added new integration test for the command ``soci ztoc list --image-ref <ref>``

### DIFF
--- a/cmd/soci/commands/ztoc/list.go
+++ b/cmd/soci/commands/ztoc/list.go
@@ -22,6 +22,10 @@ import (
 	"text/tabwriter"
 
 	"github.com/awslabs/soci-snapshotter/soci"
+	"github.com/containerd/containerd/cmd/ctr/commands"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/platforms"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/urfave/cli"
 )
 
@@ -30,8 +34,16 @@ var listCommand = cli.Command{
 	Description: "list ztocs",
 	Flags: []cli.Flag{
 		cli.StringFlag{
-			Name:  "digest",
+			Name:  "ztoc-digest",
 			Usage: "filter ztocs by digest",
+		},
+		cli.StringFlag{
+			Name:  "image-ref",
+			Usage: "filter ztocs to those that are associated with a specific image",
+		},
+		cli.BoolFlag{
+			Name:  "verbose, v",
+			Usage: "display extra debugging messages",
 		},
 	},
 	Action: func(cliContext *cli.Context) error {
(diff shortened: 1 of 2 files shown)


cmd/soci/commands/ztoc/list.go
```diff
@@ -30,25 +34,69 @@ var listCommand = cli.Command{
...
-		digest := cliContext.String("digest")
+		ztocDgst := cliContext.String("ztoc-digest")
+		imgRef := cliContext.String("image-ref")
 
 		var artifacts []*soci.ArtifactEntry
-		db.Walk(func(ae *soci.ArtifactEntry) error {
-			if ae.Type == soci.ArtifactEntryTypeLayer &&
-				(digest == "" || ae.Digest == digest) {
-				artifacts = append(artifacts, ae)
+		if imgRef == "" {
```
sbuckfelder, line 54: What is the expected behavior when both are present?  Does it make sense to split the two into private functions?

sparr (approved):
LGTM

turan18 (dismissed):

cmd/soci/commands/ztoc/list.go
```diff
@@ -30,25 +34,69 @@ var listCommand = cli.Command{
...
+				return err
 			}
-			return nil
-		})
+			var layers []ocispec.Descriptor
+			for _, p := range platform {
+				manifest, err := images.Manifest(ctx, client.ContentStore(), img.Target, platforms.OnlyStrict(p))
+				if err == nil {
+					layers = append(layers, manifest.Layers...)
+					break
```
Kern--, line 82: Why do we break here? What if the user has created indices for multiple platforms?

cmd/soci/commands/ztoc/list.go
hanyuel: Updated the PR to include the case where both --image-ref and --ztoc-digest are present. We will validate if the specified ztoc is with the specified image. If the ztoc is not with the image, the command will return an error.

cmd/soci/commands/ztoc/list.go
hanyuel: Here, we only need to find the image layers.

Kern--: Image layers depend on the platform. If you create an index for 2 platforms, this will only show the ztocs for 1 platform.

Kern--: > That is true only if the image have different layers for different platforms.

This is likely to be true for any layer that contains binaries.

> If we don't break here, the next `images.Manifest` call will error out as it walks through the image again trying to read all manifests again and that read will fail.

I'm not sure I understand this. I would expect `images.Manifest` to succeed for every platform that I have pulled.

hanyuel: Just found out some image can have multiple platforms in their image index, but manifests for some platforms don't exist, which will result in a failed ``images.Manifest`` call for those platforms.  In this case, we will print a warning message and continue looking for manifests other other platforms. I've updated this in the latest revision.

cmd/soci/commands/ztoc/list.go
```diff
@@ -30,25 +34,85 @@ var listCommand = cli.Command{
...
 			}
-			return nil
-		})
+			var layers []ocispec.Descriptor
+			for _, p := range platform {
+				manifest, err := images.Manifest(ctx, client.ContentStore(), img.Target, platforms.OnlyStrict(p))
+				if err != nil {
+					// print a warning message if a manifest can't be resolved
+					// continue looking for manifests of other platforms
+					fmt.Printf("no image manifest for platform %s/%s. err: %v\n", p.Architecture, p.OS, err)
```
Kern--, line 83: Possibly for a follow up, but can we hide this behind a `--verbose` flag or something? I would expect that in most cases a user is working with a single platform and this just adds unnecessary clutter. It could be useful for debugging if you created indices for all platforms and then deleted the image, for example. The ztocs would still be there, but the filtering wouldn't work.

Kern-- (dismissed):

turan18 (dismissed):

rdpsin (approved):

cmd/soci/commands/ztoc/list.go
hanyuel: Honestly we can just not print anything in this case, and only add the layers from the available manifests. The purpose of ``soci ztoc list`` command is to filter out all _available_ ztocs. What platforms the image is not on is not really its concern.

hanyuel: Added ``--verbose`` flag

turan18 (approved):

cmd/soci/commands/ztoc/list.go
```diff
@@ -30,25 +34,90 @@ var listCommand = cli.Command{
...
-			Name:  "digest",
+			Name:  "ztoc-digest",
 			Usage: "filter ztocs by digest",
 		},
+		cli.StringFlag{
+			Name:  "image-ref",
+			Usage: "filter ztocs to those that are associated with a specific image",
+		},
+		cli.BoolFlag{
+			Name:  "verbose, v",
```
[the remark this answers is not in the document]
hanyuel: In this case, we need the default to be not showing output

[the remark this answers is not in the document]
Kern--: The convention we're setting up here is the normal output is human readable, `--quiet` is scriptable (i.e. only the minimum information so you can pipe it into xargs or something), and `--verbose` adds debug info so you can see what it's doing.

PR #315, opened 2023-01-13
Check overlayfs fallback in mult. container tests

### LINKED ISSUE #254: Improve integration tests for multiple container workloads by checking for overlayfs fallback
Per https://github.com/awslabs/soci-snapshotter/pull/243#discussion_r1054961714, once we have a metric for overlays fallback occurrences, we should revisit the multiple container tests and have them also check to confirm that the containers are using soci, in addition to the current (pending merge) behavior of checking the functionality of the workload.

rdpsin:

*Issue #, if available:*
Resolves #254 

*Description of changes:*

*Testing performed:*

### DIFF
--- a/integration/run_test.go
+++ b/integration/run_test.go
@@ -87,6 +87,11 @@ func TestRunMultipleContainers(t *testing.T) {
 			if err := testutil.WriteFileContents(sh, defaultContainerdConfigPath, getContainerdConfigYaml(t, false), 0600); err != nil {
 				t.Fatalf("failed to write %v: %v", defaultContainerdConfigPath, err)
 			}
+
+			if err := testutil.WriteFileContents(sh, defaultSnapshotterConfigPath, getSnapshotterConfigYaml(t, false, tcpMetricsConfig), 0600); err != nil {
+				t.Fatalf("failed to write %v: %v", defaultSnapshotterConfigPath, err)
+			}
+
 			sh.
 				X("update-ca-certificates").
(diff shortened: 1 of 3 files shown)


integration/run_test.go
```diff
@@ -77,6 +79,28 @@ func TestRunMultipleContainers(t *testing.T) {
 		},
 	}
 
+	checkOverlayFallbackCount := func(output string, expected int) error {
```
djdongjin, line 82: This seems a duplicate of https://github.com/awslabs/soci-snapshotter/blob/792a8996dc02311d9ca85c23fd18ae5563212f28/integration/metrics_test.go#L121 can you extract it to a separate func and reuse here?

hanyuel (dismissed):

integration/run_test.go
```diff
@@ -110,6 +115,12 @@ func TestRunMultipleContainers(t *testing.T) {
 				sh.X("soci", "run", "-d", "--rm", "--snapshotter=soci", image, getTestContainerName(index, container))
 			}
 
+			// Verify that no mounts fallback to overlayfs
+			curlOutput := string(sh.O("curl", tcpMetricsAddress+metricsPath))
```
sparr, line 119: Why shell out to curl instead of using net/http or a higher level request library?

rdpsin: "localhost" is in the context of the testing container. Using net/http would still require us to build some kind of binary inside the container and run it.

djdongjin (approved):

hanyuel (approved):

Issue #333, opened 2023-01-19
[BUG] [INTEG-TEST] prepareSociIndices returns incorrect ztocs for images


turan18:
**Describe the bug**
Images returned by `prepareSociIndices` do not have the correct ztocs as each images `ztocDigests` field is populated with the list of all ztocs rather the ztocs that only correspond to their layers. The `getZtocDigests()` just returns  the list of all generated ztocs.  
https://github.com/awslabs/soci-snapshotter/blob/235146bfc76f92a39a40e54b382fa88b1f9224cb/integration/index_test.go#L72-L97

**Expected behavior**

Either add a separate function named `getZtocDigestsForImage` or rename the current function. This function should only return a list of ztocs corresponding to a specific image. This should be resolved quickly once https://github.com/awslabs/soci-snapshotter/pull/314 is merged as you can simply query for a list of ztocs corresponding to an image by passing in the `--image-ref` parameter.

**Configuration (please complete the following information):**
 - OS: 
 - Snapshotter Version:
 - Containerd Version: 

**Additional context**

djdongjin:
I think we should change `getZtocDigests` to get the index in `imageInfo` and then get ztocs from the index's blobs.

https://github.com/awslabs/soci-snapshotter/blob/235146bfc76f92a39a40e54b382fa88b1f9224cb/integration/index_test.go#L83-L97

turan18:
> I think we can change `getZtocDigests` to get the soci index in `imageInfo` and then get ztocs from the index's blobs.
> 
> https://github.com/awslabs/soci-snapshotter/blob/235146bfc76f92a39a40e54b382fa88b1f9224cb/integration/index_test.go#L83-L97
> 
> Something like this:
> 
> https://github.com/awslabs/soci-snapshotter/blob/235146bfc76f92a39a40e54b382fa88b1f9224cb/integration/ztoc_test.go#L61-L71

Yea I think that'll work as well, I just though that the `--image-ref` solution would be a little cleaner, although I dont really mind either or

djdongjin:
> Yea I think that'll work as well, I just though that the --image-ref solution would be a little cleaner, although I dont really mind either or

Yeah both work for me :)

PR #334, opened 2023-01-19
Add `soci ztoc info` and `soci ztoc get-file` integration tests

turan18:

Added integration tests for `ztoc info` and `get-file`

*Issue #, if available:*

#54

*Testing performed:*

~~At this point both integration tests fail due to #333~~
 

### DIFF
--- a/integration/ztoc_test.go
+++ b/integration/ztoc_test.go
@@ -17,13 +17,47 @@
 package integration
 
 import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"encoding/json"
+	"fmt"
+	"io"
+	"math/rand"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
+	"github.com/awslabs/soci-snapshotter/compression"
 	"github.com/awslabs/soci-snapshotter/soci"
+	"github.com/awslabs/soci-snapshotter/util/dockershell"
+	"github.com/awslabs/soci-snapshotter/util/testutil"
+	"github.com/awslabs/soci-snapshotter/ztoc"
+	"github.com/google/go-cmp/cmp"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
+
+type Info struct {
+	Version           string             `json:"version"`
+	BuildTool         string             `json:"build_tool"`
+	Size              int64              `json:"size"`
+	SpanSize          compression.Offset `json:"span_size"`
+	NumSpans          compression.SpanID `json:"num_spans"`
+	NumFiles          int                `json:"num_files"`
+	NumMultiSpanFiles int                `json:"num_multi_span_files"`
+	Files             []FileInfo         `json:"files"`
+}
+
+type FileInfo struct {
+	Filename  string             `json:"filename"`
+	Offset    int64              `json:"offset"`
+	Size      int64              `json:"size"`
+	Type      string             `json:"type"`
+	StartSpan compression.SpanID `json:"start_span"`
+	EndSpan   compression.SpanID `json:"end_span"`
+}
 
 func TestSociZtocList(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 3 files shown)


integration/ztoc_test.go
```diff
@@ -17,14 +17,33 @@
...
+	BuildTool string     `json:"build_tool"`
+	Files     []FileInfo `json:"files"`
+}
+
+type FileInfo struct {
+	Filename string `json:"filename"`
+	Offset   int64  `json:"offset"`
+	Size     int64  `json:"size"`
+}
+
```
rdpsin, line 46: Can we not use them from ztoc/info.go?

util/testutil/util.go
```diff
@@ -117,3 +118,11 @@ func RandomByteData(size int64) []byte {
 	rand.Read(b)
 	return b
 }
+func GenerateRandomDigest() string {
+	s := "test string"
+	h := sha256.New()
+	h.Write([]byte(s))
+	hash := h.Sum(nil)
+
+	return fmt.Sprintf("sha256:%x", string(hash))
```
rdpsin, line 127: Can just do 

```
dgst := digest.FromString(s) // or dgst.FromBytes
return dgst.String()
```

util/testutil/util.go
```diff
@@ -117,3 +118,11 @@ func RandomByteData(size int64) []byte {
 	rand.Read(b)
 	return b
 }
+func GenerateRandomDigest() string {
+	s := "test string"
```
rdpsin, line 122: Not really random. 😅

rdpsin (approved):

util/testutil/util.go
turan18: Oh yeah, alot cleaner will do

util/testutil/util.go
turan18: Oof forgot about that part haha

integration/ztoc_test.go
turan18: We can but that would introduce an extra dependency to the root go.mod.

djdongjin: If they're in `cmd` pkg, then we can't import it in integration test. Otherwise we'll have circular dependency. (`./cmd` depends on `./`)

turan18: Yeah, I'm not sure where we should put extra metadata like this

integration/ztoc_test.go
```diff
@@ -97,3 +116,214 @@ func TestSociZtocList(t *testing.T) {
...
+			},
+		}
+		for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+				var zinfo Info
+				output, err := sh.OLog("soci", "ztoc", "info", tt.ztocDigest)
+				if !tt.expectErr {
+					err := json.Unmarshal(output, &zinfo)
+					if err != nil {
+						t.Fatalf("Expected Info type got %s", output)
```
djdongjin, line 203: nit: `Expected` -> `expected`

integration/ztoc_test.go
```diff
@@ -97,3 +116,214 @@ func TestSociZtocList(t *testing.T) {
...
+					for _, blob := range sociIndex.Blobs {
+						if blob.Digest.String() == ztocDigest {
+							layerDigest = blob.Annotations[soci.IndexAnnotationImageLayerDigest]
+							break
+						}
+					}
+
+					layerPath := filepath.Join(containerdBlobStorePath, trimSha256Prefix(layerDigest))
+					contents := sh.X("mkdir", "-p", tempLayerPath).
+						X("tar", "-xzf", layerPath, "-C", tempLayerPath).
```
rdpsin, line 317: Instead of doing this, can we just use `gzip.NewReader` with `tar.NewReader`? 

For example, something like:

```
layerContents := sh.O("cat", layerPath)
gzipReader := gzip.NewReader(bytes.NewReader(layerContents))
tarReader := tar.NewReader(gzipReader)

// for each file in tarReader, if regular file, compare contents
```

integration/ztoc_test.go
```diff
@@ -145,3 +166,227 @@ func ztocExistChecker(t *testing.T, listOutputLines []string, img testImageIndex
...
+	verifyOutputStream := func(t *testing.T, sh *dockershell.Shell, contents, output []byte) {
+		if len(output) > 0 {
+			output = output[:len(output)-1]
+			res := bytes.Compare(contents, output)
+			if res != 0 {
+				t.Fatalf("output: %s expected: %s", output, contents)
+			}
+
+		} else {
+			o := sh.O("cat", tempOutputStream)
```
rdpsin, line 307: Not sure why we need this?

integration/ztoc_test.go
```diff
@@ -145,3 +166,227 @@ func ztocExistChecker(t *testing.T, listOutputLines []string, img testImageIndex
...
+			}
+		}
+		return path
+	}
+	verifyOutputStream := func(t *testing.T, sh *dockershell.Shell, contents, output []byte) {
+		if len(output) > 0 {
+			output = output[:len(output)-1]
+			res := bytes.Compare(contents, output)
+			if res != 0 {
+				t.Fatalf("output: %s expected: %s", output, contents)
```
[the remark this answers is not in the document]
turan18: Sure

integration/ztoc_test.go
```diff
@@ -145,3 +166,227 @@ func ztocExistChecker(t *testing.T, listOutputLines []string, img testImageIndex
...
+				t.Fatalf("output: %s expected: %s", o, contents)
+			}
+
+		}
+	}
+
+	for _, img := range testImages {
+		img := img
+		ztocDigest := img.ztocDigests[0]
+		fileName := getFilePathWithinZtoc(t, ztocDigest)
```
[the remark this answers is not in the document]
turan18: This would return the path for the first non-zero sized regular file. Sure

integration/ztoc_test.go
turan18: This was there to check whether the output was to stdout or to a `--output` file. Since I was only testing filepaths with a non-zero size, if the output was empty that would implicitly mean that it was redirected to an output file

turan18: I can move this inside the test function, to make it cleaner

integration/ztoc_test.go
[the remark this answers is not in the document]
turan18: @rdpsin This will be really IO intensive / take alot of time, should this only be for the stdout test and not the output test. I just ran this and it takes around 5 mins, just for stdout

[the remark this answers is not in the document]
rdpsin: Can we do something like select a random subset of files (like a maximum of 10 files)? I'm afraid just testing a single file at the start/end can lead to false positives. Depends on whether we want the tests to be deterministic all the time.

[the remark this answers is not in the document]
turan18: Yeah , I think that makes sense. And I do think the tests at least in this case shouldn't be deterministic.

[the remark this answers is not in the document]
rdpsin: I do think we want them to be somewhat determinstic (or atleast catch regressions when and if they happen). FWIW, I don't think 5 mins is really all that time. :)

integration/ztoc_test.go
```diff
@@ -145,3 +179,248 @@ func ztocExistChecker(t *testing.T, listOutputLines []string, img testImageIndex
...
+	testImages := prepareSociIndices(t, sh)
+
+	getFullZtoc := func(sh *dockershell.Shell, ztocPath string) (*ztoc.Ztoc, error) {
+		output := sh.O("cat", ztocPath)
+		reader := bytes.NewReader(output)
+		z, err := ztoc.Unmarshal(reader)
+		return z, err
+	}
+
+	verifyInfoOutput := func(t *testing.T, zinfo Info, ztoc *ztoc.Ztoc) error {
```
djdongjin, line 197: `t *testing.T` is not used in the `func`, can you remove it? Also can you move `verifyInfoOutput` to a separate func (put it in this file is fine) since it's long and not use other variables outside of it.

integration/ztoc_test.go
```diff
@@ -145,3 +179,248 @@ func ztocExistChecker(t *testing.T, listOutputLines []string, img testImageIndex
...
+
+	testImages := prepareSociIndices(t, sh)
+
+	var (
+		tempOutputStream = "test.txt"
+		randomFile       = string(testutil.RandomByteData(10))
+		randomZtocDigest = testutil.RandomDigest()
+	)
+
+	getRandomFilePathsWithinZtoc := func(t *testing.T, ztocDigest string, numFiles int) []string {
```
djdongjin, line 308: can you remove `t *testing.T` here

integration/ztoc_test.go
```diff
@@ -145,3 +179,248 @@ func ztocExistChecker(t *testing.T, listOutputLines []string, img testImageIndex
...
+				regPaths = append(regPaths, file.Filename)
+			}
+		}
+		for i := 0; i < numFiles; i++ {
+			randPaths = append(randPaths, regPaths[rand.Intn(len(regPaths))])
+		}
+		return randPaths
+	}
+
+	verifyOutputStream := func(t *testing.T, contents, output []byte) error {
```
djdongjin, line 327: can you remove `t *testing.T` here  as well

integration/ztoc_test.go
```diff
@@ -145,3 +179,248 @@ func ztocExistChecker(t *testing.T, listOutputLines []string, img testImageIndex
...
+			return nil
+		}
+		return fmt.Errorf("expected %s got %s", contents, output)
+	}
+
+	for _, img := range testImages {
+		img := img
+		ztocDigest := img.ztocDigests[0]
+		var layerDigest string
+		sociIndex, _ := sociIndexFromDigest(sh, img.sociIndexDigest)
```
djdongjin, line 339: We should check the returned `err` here and `t.Fatal(err)` if not nil.

integration/ztoc_test.go
```diff
@@ -145,3 +179,248 @@ func ztocExistChecker(t *testing.T, listOutputLines []string, img testImageIndex
...
+		}
+		for _, tt := range testCases {
+			t.Run(tt.name, func(t *testing.T) {
+				if !tt.expectedErr {
+					for _, f := range files {
+						cmd := append(tt.cmd, f)
+						output, err := sh.OLog(cmd...)
+						if err != nil {
+							t.Fatalf("failed to return file contents")
+						}
```
djdongjin, line 389: we should print the err if not nil: `t.Fatalf("failed to return file contents: %v", err)`

integration/ztoc_test.go
```diff
@@ -145,3 +179,248 @@ func ztocExistChecker(t *testing.T, listOutputLines []string, img testImageIndex
...
+						var contents []byte
+						for {
+							h, err := tarReader.Next()
+							if err == io.EOF {
+								break
+							}
+							if h.Name == f {
+								contents, err = io.ReadAll(tarReader)
+								if err != nil {
+									t.Fatalf("failed getting original file content")
```
djdongjin, line 402: same here

integration/ztoc_test.go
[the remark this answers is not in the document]
hanyuel: 5 mins does sound a little too long for running the test...A random subset of files should work.

integration/ztoc_test.go
```diff
@@ -145,3 +394,40 @@ func ztocExistChecker(t *testing.T, listOutputLines []string, img testImageIndex
 	t.Fatalf("invalid ztoc from index %s for image %s:\n expected ztoc: digest: %s, size: %s, layer digest: %s\n actual output lines: %s",
 		img.sociIndexDigest, img.imgInfo.ref, ztocDigest, size, layerDigest, listOutputLines)
 }
+
+func verifyInfoOutput(t *testing.T, zinfo Info, ztoc *ztoc.Ztoc) error {
```
rdpsin, line 398: Do we need to pass `testing.T` here?

integration/ztoc_test.go
```diff
@@ -130,6 +164,221 @@ func TestSociZtocList(t *testing.T) {
...
+			},
+		}
+		for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+				var zinfo Info
+				output, err := sh.OLog("soci", "ztoc", "info", tt.ztocDigest)
+				if !tt.expectErr {
+					err := json.Unmarshal(output, &zinfo)
+					if err != nil {
+						t.Fatalf("expected Info type got %s", output)
```
djdongjin, line 222: can you print the `err` as well? (`t.Fatalf("expected Info type got %s: %v", output, err)`)

integration/ztoc_test.go
```diff
@@ -130,6 +164,221 @@ func TestSociZtocList(t *testing.T) {
...
+					}
+					ztocPath := filepath.Join(blobStorePath, trimSha256Prefix(tt.ztocDigest))
+					ztoc, err := getFullZtoc(sh, ztocPath)
+					if err != nil {
+						t.Fatalf("failed getting original ztoc")
```
djdongjin, line 227: can you print the `err` as well?

integration/ztoc_test.go
```diff
@@ -130,6 +164,221 @@ func TestSociZtocList(t *testing.T) {
...
+					}
+					err = verifyInfoOutput(t, zinfo, ztoc)
+					if err != nil {
+						t.Fatal(err)
+					}
+				} else if err == nil {
+					t.Fatal("failed to return error")
```
djdongjin, line 234: can you print the `err` as well?

turan18: The error is being printed.

integration/ztoc_test.go
```diff
@@ -130,6 +164,221 @@ func TestSociZtocList(t *testing.T) {
...
+		}
+		return randPaths
+	}
+
+	verifyOutputStream := func(contents, output []byte) error {
+		d := cmp.Diff(contents, output)
+		if d == "" {
+			return nil
+		}
+		return fmt.Errorf("expected %s got %s", contents, output)
```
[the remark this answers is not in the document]
turan18: Sure

rdpsin (approved):

djdongjin (approved):

integration/ztoc_test.go
```diff
@@ -130,6 +164,221 @@ func TestSociZtocList(t *testing.T) {
...
+
+	getFullZtoc := func(sh *dockershell.Shell, ztocPath string) (*ztoc.Ztoc, error) {
+		output := sh.O("cat", ztocPath)
+		reader := bytes.NewReader(output)
+		z, err := ztoc.Unmarshal(reader)
+		return z, err
+	}
+
+	for _, img := range testImages {
+		img := img
```
sparr, line 183: Why?

Kern--: This is something @turan18 and I discussed. Go's loops have a somewhat surprising behavior where the loop variable `img` is the same **instance** across all iterations. If you create a closure over it, you're closing over effectively a pointer which will be updated on the next loop. If you introduce `img := img`, you create a local variable that shadows the loop variable and is independent between executions.

The classic example is the following:

```
for _, n := range []int{1,2,3,4,5} {
    go func() {
        fmt.Println(n)
    }()
}

output:
5
5
5
5
5
```

There's a detailed explanation here:  https://go.dev/doc/faq#closures_and_goroutines


It's not a huge issue for us now because `t.Run` is synchronous by default. But if we tried to make the subtests run in parallel, we might only test the last test case in the list.

turan18: When go routines are spawned from within a loop, they share the same reference to the thing being iterated, in this case the `img` value. When `t.Run` is called within a loop a go routine is spawned and each routine has a reference to the same value of `img`. To fix this you can reassign `img` to itself to make sure each routine has its own copy of img. Although, with these tests its not really needed as `img` is never directly referenced within t.Run, it is still good to have in cases where it may be referenced.

A better explanation here: https://appliedgo.com/blog/goroutine-closures

turan18: Oops didn't see @Kern-- commented 😅

sparr (approved):

PR #336, opened 2023-01-19
Fixed image ztocs returned by prepareSociIndices()

turan18:

*Issue #, if available:*

Fixes: #333 

*Description of changes:*

Fixed `prepareSociIndices` to return the correct set of ztocs for each image.

*Testing performed:*

### DIFF
--- a/integration/index_test.go
+++ b/integration/index_test.go
@@ -69,7 +69,7 @@ func prepareSociIndices(t *testing.T, sh *dockershell.Shell) []testImageIndex {
 		}
 		img.imgInfo = dockerhub(img.imgName, withPlatform(platform))
 		img.sociIndexDigest = buildIndex(sh, img.imgInfo)
-		ztocDigests, err := getZtocDigests(sh, img.imgInfo)
+		ztocDigests, err := getZtocDigestsForImage(sh, img.imgInfo)
 		if err != nil {
 			t.Fatalf("could not get ztoc digests: %v", err)
 		}
@@ -80,8 +80,8 @@ func prepareSociIndices(t *testing.T, sh *dockershell.Shell) []testImageIndex {
 	return testImages
 }
 
-func getZtocDigests(sh *dockershell.Shell, img imageInfo) ([]string, error) {
-	ztocInfoBytes := sh.O("soci", "ztoc", "list")
+func getZtocDigestsForImage(sh *dockershell.Shell, img imageInfo) ([]string, error) {
+	ztocInfoBytes := sh.O("soci", "ztoc", "list", "--image-ref", img.ref)
 	scanner := bufio.NewScanner(bytes.NewReader(ztocInfoBytes))
 	scanner.Split(bufio.ScanLines)
 	var lines []string


integration/index_test.go
```diff
@@ -80,8 +80,8 @@ func prepareSociIndices(t *testing.T, sh *dockershell.Shell) []testImageIndex {
 	return testImages
 }
 
-func getZtocDigests(sh *dockershell.Shell, img imageInfo) ([]string, error) {
-	ztocInfoBytes := sh.O("soci", "ztoc", "list")
+func getZtocDigestsForImage(sh *dockershell.Shell, img imageInfo) ([]string, error) {
+	ztocInfoBytes := sh.O("soci", "ztoc", "list", "--image-ref", img.ref)
 	scanner := bufio.NewScanner(bytes.NewReader(ztocInfoBytes))
```
rdpsin, line 85: Maybe we should add a `quiet` flag to avoid this.

rdpsin (approved):

integration/index_test.go
turan18: Agreed. I suppose that can be a separate PR

turan18: We could also add it to `soci index list` too

integration/index_test.go
```diff
@@ -80,8 +80,8 @@ func prepareSociIndices(t *testing.T, sh *dockershell.Shell) []testImageIndex {
 	return testImages
 }
 
-func getZtocDigests(sh *dockershell.Shell, img imageInfo) ([]string, error) {
-	ztocInfoBytes := sh.O("soci", "ztoc", "list")
+func getZtocDigestsForImage(sh *dockershell.Shell, img imageInfo) ([]string, error) {
```
djdongjin, line 83: what if the image has multiple soci indices generated? Will that break any tests?

integration/index_test.go
rdpsin: We already have it. :) 

https://github.com/awslabs/soci-snapshotter/blob/bff4b07408aa3df86d5932a036880c5bf2ff6f4a/cmd/soci/commands/index/list.go#L73-L76

turan18: Nice didnt catch that

integration/index_test.go
turan18: We are only using one platform for each image, so there should only be one index I think? Regardless, It shouldn't as I am only testing one ztoc digest anyway.

hanyuel (approved):

integration/index_test.go
hanyuel: >  what if the image has multiple soci indices generated

It does not. For the integ test, we pull image for only 1 platform 
https://github.com/awslabs/soci-snapshotter/blob/bff4b07408aa3df86d5932a036880c5bf2ff6f4a/integration/testutil_soci.go#L63

Kern--: It doesn't except where we test multi-platform create
https://github.com/awslabs/soci-snapshotter/blob/main/integration/create_test.go#L122-L131

Maybe there's value in adding a `platform` flag to `ztoc list` at some point.

Kern-- (approved):

PR #343, opened 2023-01-23
Add runtime configuration for minimum layer size for remote mount

### LINKED ISSUE #172: A runtime override for min-layer-size

`soci create --min-layer-size` let's users say "don't bother creating a zTOC for small layers, in which case the snapshotter will just pull the layer at launch time like the default snapshotter does. This is great for performance tuning a particular image, but it means that multiple SOCI indices have to be generated and pushed when running benchmarks to find the right size. 

A more convenient solution would be `min_layer_size` option in  _snapshotterConfig_. If set, it would mean, "even if the SOCI index supplied a zTOC for the layer, if the layer is small, then ignore the zTOC and do the MountLocal() thing instead.

Another use case for this feature would be to allow administrators to override what's in the SOCI index when running on systems with a very slow or unreliable network connection to the remote registry. (To do this today, they would have to create and managed multiple  SOCI indices.)

sparr:
*Issue #, if available:*
closes #172 

*Description of changes:*
Adds an option to `soci-snapshotter-grpc/config.toml` to set a minimum size for remote mounting. Smaller layers will always be locally mounted.

*Testing performed:*
Ran automated test, and manual similar tests with other images and thresholds, with and without new config.

### DIFF
--- a/integration/pull_test.go
+++ b/integration/pull_test.go
@@ -792,3 +792,40 @@ func TestRpullImageThenRemove(t *testing.T) {
 
 	checkFuseMounts(t, sh, 0)
 }
+
+// TestRpullImageWithMinLayerSize pulls and rpulls an image with a runtime min_layer_size to confirm small layers are mounted locally
+func TestRpullImageWithMinLayerSize(t *testing.T) {
+	const minLayerSize = "25000000"
+	const minLayerSizeConfig = `
+min_layer_size=` + minLayerSize + `
+`
+
+	regConfig := newRegistryConfig()
+	sh, done := newShellWithRegistry(t, regConfig)
+	defer done()
+
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, false, minLayerSizeConfig))
+
+	containerImage := rabbitmqImage
+
+	copyImage(sh, dockerhub(containerImage), regConfig.mirror(containerImage))
+	// withMinLayerSize(0) creates ztoc for all 10 layers instead of the 3 that are >= 10MB
+	indexDigest := buildIndex(sh, regConfig.mirror(containerImage), withMinLayerSize(0))
+
+	rawJSON := sh.O("soci", "index", "info", indexDigest)
+	var sociIndex soci.Index
+	if err := json.Unmarshal(rawJSON, &sociIndex); err != nil {
+		t.Fatalf("invalid soci index from digest %s: %v", indexDigest, rawJSON)
+	}
+
+	if len(sociIndex.Blobs) == 0 {
+		t.Fatalf("soci index %s contains 0 blobs, invalidating this test", indexDigest)
+	}
+
+	sh.X("soci", "image", "rpull", "--user", regConfig.creds(), "--soci-index-digest", indexDigest, regConfig.mirror(containerImage).ref)
+
+	// default index creation min layer size of 10MB would produce 3 fuse mounts
+	// specified index creation min layer size of 0 would produce 10 fuse mounts
+	// specified runtime min layer size of 25MB will produce 2 fuse mounts
+	checkFuseMounts(t, sh, 2)
+}
--- a/service/config.go
+++ b/service/config.go
@@ -48,6 +48,9 @@ type Config struct {
 
 	// ResolverConfig is config for resolving registries.
 	ResolverConfig `toml:"resolver"`
+
+	// MinLayerSize skips remote mounting of smaller layers
+	MinLayerSize int `toml:"min_layer_size"`
 }
 
 // KubeconfigKeychainConfig is config for kubeconfig-based keychain.
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -38,6 +38,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -102,18 +103,29 @@ type FileSystem interface {
 // SnapshotterConfig is used to configure the remote snapshotter instance
 type SnapshotterConfig struct {
 	asyncRemove bool
+
+	// minLayerSize skips remote mounting of smaller layers
+	minLayerSize int
 }
 
 // Opt is an option to configure the remote snapshotter
 type Opt func(config *SnapshotterConfig) error
 
-// AsynchronousRemove defers removal of filesystem content until
+// WithAsynchronousRemove defers removal of filesystem content until
 // the Cleanup method is called. Removals will make the snapshot
 // referred to by the key unavailable and make the key immediately
 // available for re-use.
-func AsynchronousRemove(config *SnapshotterConfig) error {
+func WithAsynchronousRemove(config *SnapshotterConfig) error {
 	config.asyncRemove = true
 	return nil
+}
+
+// WithMinLayerSize sets the smallest layer that will be mounted remotely.
+func WithMinLayerSize(minLayerSize int) Opt {
+	return func(config *SnapshotterConfig) error {
+		config.minLayerSize = minLayerSize
+		return nil
+	}
 }
 
 type snapshotter struct {
@@ -124,6 +136,8 @@ type snapshotter struct {
 	// fs is a filesystem that this snapshotter recognizes.
 	fs        FileSystem
 	userxattr bool // whether to enable "userxattr" mount option
+
+	minLayerSize int // minimum layer size for remote mounting
 }
 
 // NewSnapshotter returns a Snapshotter which can use unpacked remote layers
@@ -167,11 +181,12 @@ func NewSnapshotter(ctx context.Context, root string, targetFs FileSystem, opts 
 	}
 
 	o := &snapshotter{
-		root:        root,
-		ms:          ms,
-		asyncRemove: config.asyncRemove,
-		fs:          targetFs,
-		userxattr:   userxattr,
+		root:         root,
+		ms:           ms,
+		asyncRemove:  config.asyncRemove,
+		fs:           targetFs,
+		userxattr:    userxattr,
+		minLayerSize: config.minLayerSize,
 	}
 
 	if err := o.restoreRemoteSnapshot(ctx); err != nil {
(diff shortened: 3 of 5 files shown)


snapshot/snapshot.go
```diff
@@ -116,6 +120,12 @@ func AsynchronousRemove(config *SnapshotterConfig) error {
 	return nil
 }
 
+// MinLayerSize sets the smallest layer that will be mounted remotely.
+func MinLayerSize(config *SnapshotterConfig, minLayerSize int) error {
```
rdpsin, line 124: This should really be a function that returns `Opt`.

```
func WithMinLayerSize(minLayerSize int) Opt {
      return func(c *SnapshotterConfig) error {
               c.minLayerSize = minLayerSize
               return nil
      }
}
```

service/config.go
```diff
@@ -48,6 +48,9 @@ type Config struct {
 
 	// ResolverConfig is config for resolving registries.
 	ResolverConfig `toml:"resolver"`
+
+	// MinLayerSize skips remote mounting of smaller layers
+	MinLayerSize int `toml:"min_layer_size"`
```
rdpsin, line 53: #345 adds `SnapshotterConfig` to `fs/config`. This field should be part of that.

np: How about renaming it to something like `MinLayerSizeOverride`, to distinguish it from the `MinLayerSize` in the CLI?

snapshot/snapshot.go
```diff
@@ -268,37 +281,54 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
...
-			log.G(lCtx).WithField(remoteSnapshotLogKey, prepareFailed).
-				WithError(err).Warn("failed to prepare remote snapshot")
-			if !errors.Is(err, ErrNoZtoc) {
-				commonmetrics.IncOperationCount(commonmetrics.FuseMountFailureCount, digest.Digest(""))
+
+		localOverride := false
+		if strVal, ok := base.Labels["containerd.io/snapshot/remote/image.layers.size"]; ok {
+			if intVal, err := strconv.Atoi(strVal); err == nil {
+				if intVal <= o.minLayerSize {
+					localOverride = true
```
rdpsin, line 297: Should we emit a metric here?

service/config.go
sbuckfelder: Heck might as well call it `MinLayerSizeConfigOverride` no reason not to be explicit

sparr: SnapshotterConfig is used for different purposes than snapshotterConfig. I think it would need to be in both. I'd be moving it out of the snapshotter struct, not out of the config for the main app.

snapshot/snapshot.go
sparr: I discussed this with Kern. This function follows the model of AsynchronousRemove, which directly manipulates the config instead of returning a function that manipulates the config. This model is distinct from the WithFoo model used elsewhere in the project for options.
Making this change would require also changing AsynchronousRemove, and the handler that calls them.

snapshot/snapshot.go
sparr: And, if so, which sort of metric?

service/config.go
sparr: The CLI min layer size is for creating, not preparing/mouinting. This new option is not an override of that option, it applies at a different point in the workflow. If you use both, the larger one overrides the smaller, in either order.

snapshot/snapshot.go
```diff
@@ -268,37 +281,54 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
...
 		lCtx := log.WithLogger(ctx, log.G(ctx).WithField("key", key).WithField("parent", parent))
-		if err := o.prepareRemoteSnapshot(lCtx, key, base.Labels); err != nil {
-			log.G(lCtx).WithField(remoteSnapshotLogKey, prepareFailed).
-				WithError(err).Warn("failed to prepare remote snapshot")
-			if !errors.Is(err, ErrNoZtoc) {
-				commonmetrics.IncOperationCount(commonmetrics.FuseMountFailureCount, digest.Digest(""))
+
+		localOverride := false
+		if strVal, ok := base.Labels["containerd.io/snapshot/remote/image.layers.size"]; ok {
+			if intVal, err := strconv.Atoi(strVal); err == nil {
```
Kern--, line 295: We should log an error if we can't convert to an int.

snapshot/snapshot.go
```diff
@@ -268,37 +281,54 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
...
 		//       log is used by tests in this project.
 		lCtx := log.WithLogger(ctx, log.G(ctx).WithField("key", key).WithField("parent", parent))
-		if err := o.prepareRemoteSnapshot(lCtx, key, base.Labels); err != nil {
-			log.G(lCtx).WithField(remoteSnapshotLogKey, prepareFailed).
-				WithError(err).Warn("failed to prepare remote snapshot")
-			if !errors.Is(err, ErrNoZtoc) {
-				commonmetrics.IncOperationCount(commonmetrics.FuseMountFailureCount, digest.Digest(""))
+
+		localOverride := false
+		if strVal, ok := base.Labels["containerd.io/snapshot/remote/image.layers.size"]; ok {
```
Kern--, line 294: nit: can we export this field and use that instead? 

https://github.com/awslabs/soci-snapshotter/blob/main/fs/source/source.go#L95

snapshot/snapshot.go
```diff
@@ -268,37 +281,54 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
...
+			if intVal, err := strconv.Atoi(strVal); err == nil {
+				if intVal <= o.minLayerSize {
```
Kern--, line 296: nit: I think this should be strictly less than. If I say "Min layer size is 10MiB" I would expect a file that is exactly 10MiB to be remotely mounted. TBF, I wouldn't be surprised if we never hit this case, but it seems esrier to parse to me.

snapshot/snapshot.go
```diff
@@ -682,10 +712,12 @@ func (o *snapshotter) mounts(ctx context.Context, s storage.Snapshot, checkKey s
 
 }
 
+// Produces a file path like "{snapshotter.root}/snapshots/{id}/fs"
```
Kern--, line 715: nit: I think we should follow the normal godoc format and start with the name of the thing we're documenting. It makes it easier to export the function later.

snapshot/snapshot.go
Kern--: `AsynchronousRemove` is already an `Opt`. Personally, I think `AsynchronousRemove` should be renamed to `WithAsynchronousRemove` and we should use @rdpsin's suggestion here.

Kern-- (dismissed):

Kern--:
LGTM pending tests.

rdpsin:
LGTM. Can we have an integration test for this?

integration/pull_test.go
```diff
@@ -49,6 +49,14 @@ import (
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
+const (
+	minLayerSize = "20000000"
```
Kern--, line 53: nit for follow up: the constant for `minLayerSize` and expected FUSE mounts means that this test could fail unexpectedly if the number or size of layers in the latest image change. We can partially solve this with pinning the version we test  against, but it would still be annoying if tests failed when upgrading the pinned version.

Instead of making these constants, it would be nice if we can choose a size that is exactly the median size of layers in the image and then calculate how many fuse mounts we expect. That way we can be sure it won't start failing after an update and we can test the edge case where min layer size is exactly the size of a layer. I think we can do all this by calling `rebootContainerd` after creating the SOCI index to remove the local state of containerd/the snapshotter.

Can you create an issue for that follow up?

Kern-- (approved):

integration/pull_test.go
sparr: https://github.com/awslabs/soci-snapshotter/issues/373

integration/pull_test.go
```diff
@@ -845,3 +853,42 @@ func TestRpullImageThenRemove(t *testing.T) {
...
+		t.Fatalf("invalid soci index from digest %s: %v", indexDigest, rawJSON)
+	}
+
+	if len(sociIndex.Blobs) == 0 {
+		t.Fatalf("soci index %s contains 0 blobs, invalidating this test", indexDigest)
+	}
+
+	sh.X("soci", "image", "rpull", "--user", regConfig.creds(), "--soci-index-digest", indexDigest, regConfig.mirror(containerImage).ref)
+
+	checkFuseMounts(t, sh, 3) // default behavior produces 10 Fuse mounts
```
rdpsin, line 893: Is this right? The default behavior should produce 3 mounts, since the defaultMinLayerSize is set to 10MiB.

integration/pull_test.go
```diff
@@ -845,3 +853,42 @@ func TestRpullImageThenRemove(t *testing.T) {
...
+	sh.
+		X("update-ca-certificates").
+		Retry(100, "nerdctl", "login", "-u", regConfig.user, "-p", regConfig.pass, regConfig.host)
+
+	rebootContainerd(t, sh, "", "")
+
+	containerImage := rabbitmqImage
+
+	copyImage(sh, dockerhub(containerImage), regConfig.mirror(containerImage))
+	indexDigest := buildIndex(sh, regConfig.mirror(containerImage))
```
rdpsin, line 879: We should probably pass `WithMinLayerSize(0)` to ensure that ztocs for all layers are being built.

integration/pull_test.go
sparr: I did run it both ways, and it produced 10 without a runtime min layer size set. I'll see if the creation-time min layer size is getting overridden in some way, or what other explanation there is for my results.

integration/pull_test.go
sparr: That feels like a different case that might deserve its own test. What I'd like to be doing here is comparing to the default behavior.

Kern--: https://github.com/awslabs/soci-snapshotter/pull/363 just changed the default behavior for tests from 0 to `defaultMinLayerSize`. `WithMinLayerSize(0)` is probably equivalent to what you actually tested against.

integration/pull_test.go
sparr: See other thread below.

integration/pull_test.go
sparr: Good catch! That would explain the discrepancy.

sparr: I believe that I've addressed this.

rdpsin (approved):

Kern-- (approved):

Issue #273, opened 2023-01-03
[BUG] Integration tests leave behind docker compose artifacts


sparr:
**Describe the bug**
`make integration` leaves docker images behind from the compose operations

**Steps To Reproduce**
1. `make integration`
2. `make integration`
3. `make integration`
4. `docker image ls`

**Expected behavior**
Artifacts are removed after tests are complete. It is possible that non-image artifacts will be found during investigation of this issue, which should probably also be removed.

sparr:
`docker image prune` may naively address this, but would also affect unrelated orphan images the user has on their system and wouldn't want to have deleted.

PR #344, opened 2023-01-23
Added --rmi flag to cleanup function to remove image artficats after integration tests

turan18:
Added --rmi flag to compose down clean up func to remove the image defined/built by our compose testing service.

*Issue #, if available:*

Fixes: #273

*Description of changes:*

*Testing performed:*

```
$ make integration
....

$ docker image ls 

REPOSITORY   TAG       IMAGE ID   CREATED   SIZE

```


util/testutil/util.go
```diff
@@ -75,7 +78,7 @@ func StreamTestingLogToFile(destPath string) (func() error, error) {
 }
 
 // GetProjectRoot returns the path to the directory where the source code of this project reside.
-func GetProjectRoot(t *testing.T) string {
+func GetProjectRoot(t TestingFatal) string {
```
djdongjin, line 81: Instead of having a new interface only so that error can be `Fatal`ed directly by `testing.T` and `log.Logger`, I'd prefer removing the arg and return (string, err) and let the caller errors out, regardless `testing.T` or `log.Logger`.

djdongjin: same for `ApplyTextTemplate`

util/dockershell/compose/compose.go
```diff
@@ -91,7 +91,118 @@ func WithStdio(stdout, stderr io.Writer) Option {
...
+	}
+	if err := cmd.Run(); err != nil {
+		return nil, err
+	}
+	return cleanups, nil
+}
+
+// Starts all services defined in the docker-compose yaml and returned the execution environment for each service
+
+func Up(dockerComposeYaml string, opts ...Option) (*Compose, error) {
```
djdongjin, line 136: nit: remove empty line and start comment with func name (`// Up starts ....`)

util/dockershell/compose/compose.go
```diff
@@ -91,7 +91,118 @@ func WithStdio(stdout, stderr io.Writer) Option {
 	}
 }
 
-// New creates a new Compose of the specified docker-compose yaml data.
+// Builds all services defined in the docker-compose yaml
+
+func Build(dockerComposeYaml string, opts ...Option) ([]func() error, error) {
```
djdongjin, line 96: remove empty line and start comment with func name

integration/testutil.go
```diff
@@ -612,3 +636,33 @@ func checkOverlayFallbackCount(output string, expected int) error {
...
+		compose.WithStdio(testutil.TestingLogDest()),
+	}
+
+	return compose.Build(composeYaml, cOpts...)
+
+}
+
+// This function should run after integration tests.
+
+func tearDown(c []func() error) {
```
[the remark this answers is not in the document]
djdongjin: also suggest use a more descriptive arg name for `c`

djdongjin (changes requested):

util/testutil/util.go
turan18: Yea I think that makes things a little cleaner. I'll change it

util/dockershell/compose/compose.go
```diff
@@ -91,7 +91,118 @@ func WithStdio(stdout, stderr io.Writer) Option {
...
+	if err != nil {
+		return nil, err
+	}
+	confFile := filepath.Join(tmpContext, "docker-compose.yml")
+	if err := os.WriteFile(confFile, []byte(dockerComposeYaml), 0600); err != nil {
+		return nil, err
+	}
+
+	var cleanups []func() error
+	for i := 0; i < 3; i++ {
```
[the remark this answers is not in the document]
turan18: The first and second calls to compose down with `--rmi` flag set do not remove the images but rather it untags them. Only on the third call are images removed.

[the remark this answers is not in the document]
djdongjin: curious why is this 🤔, since the command already has `--rmi all` and why both 1st and 2nd only untag them.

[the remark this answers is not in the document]
turan18: The `all` flag will remove all images regardless if they have a custom tag or not. The `local` flag will remove everything that doesn't have a custom tag. 

> why both 1st and 2nd only untag them

Yeah I'm not sure why that is, maybe as a safety net or just to keep images for cache in case you wanted to restart services.

integration/testutil.go
```diff
@@ -101,37 +101,148 @@ level = "debug"
...
 		AdditionalConfig    string
 	}{
 		DisableVerification: disableVerification,
 		AdditionalConfig:    strings.Join(additionalConfigs, "\n"),
-	}))
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	return []byte(s)
```
sparr, line 231: Merge conflict with #378 if that lands first. This is not a blocker, just a heads up; I can fix on my end if this lands first.

sparr (approved):

djdongjin (approved):

PR #346, opened 2023-01-24
Decouple ztoc creation and compression algorithm

### LINKED ISSUE #233: Refactor the Ztoc creation
Per #103, the snapshotter's Ztoc creation needs to be adjusted to allow for easy addition of new types of compression algorithms. For more details, please refer to the design doc in #103.
This story requires implementing the following:
1. TocBuilder (please ensure that this type will be a common one for any compression algorithm)
2. Interface ZinfoBuilder with specific implementation for gzip (aka GzipZinfoBuilder)
3. ztoc.Builder, which combines 1 and 2. It will be shared for any compression algorithm. For more details see the design doc from #103.
4. soci.buildSociIndex must be using ztoc.Builder for building Ztoc.

The steps to implement are outlined in the above section.

Keep everything as is.

**Additional context**
#103

djdongjin:

*Issue #, if available:* Fix #233

*Description of changes:*

Decouple ztoc creation and compression by adding a `ztoc.Builder` which has:

1. A single `ztoc.TocBuilder` that builds `TOC` and works with both gzip and zstd.
2. One `ztoc.ZinfoBuilder` per compression algorithm (currently, only `gzip`).

`ztoc.Builder` has one export func `BuildZtoc` which checks compression algorithm and builds `Ztoc`(`TOC` via `TocBuilder`, `Zinfo` via `ZinfoBuilder`).

`ztoc.TocBuilder` has one export func `TocFromFile` which creates `TOC` given compress algorithm and tar filename. It supports different compression by creating different tar readers through compression libraries (e.g., `gzip.NewReader`, `zstd.NewReader`). 

`ztoc.ZinfoBuilder` interface has one export func `ZinfoFromFile` which creates `Zinfo` given compress tar filename and span size. Currently it has one implementation for `gzip` (`gzipZinfoBuilder`).

*Testing performed:*

make test && make integration

### DIFF
--- a/soci/soci_index.go
+++ b/soci/soci_index.go
@@ -270,6 +270,7 @@ type IndexBuilder struct {
 	contentStore content.Store
 	blobStore    orascontent.Storage
 	config       *buildConfig
+	ztocBuilder  *ztoc.Builder
 }
 
 // NewIndexBuilder returns an `IndexBuilder` that is used to create soci indices.
@@ -292,6 +293,7 @@ func NewIndexBuilder(contentStore content.Store, blobStore orascontent.Storage, 
 		contentStore: contentStore,
 		blobStore:    blobStore,
 		config:       config,
+		ztocBuilder:  ztoc.NewBuilder(config.buildToolIdentifier),
 	}, nil
 }
 
@@ -316,7 +318,7 @@ func (b *IndexBuilder) Build(ctx context.Context, img images.Image) (*IndexWithM
 		eg.Go(func() error {
 			desc, err := b.buildSociLayer(ctx, l)
 			if err != nil {
-				return fmt.Errorf("could not build zTOC for %s: %w", l.Digest.String(), err)
+				return fmt.Errorf("could not build zTOC for layer %s: %w", l.Digest.String(), err)
 			}
 			sociLayersDesc[i] = desc
 			return nil
@@ -352,7 +354,8 @@ func (b *IndexBuilder) Build(ctx context.Context, img images.Image) (*IndexWithM
 	}, nil
 }
 
-// buildSociLayer builds the ztoc for an image layer and returns a Descriptor for the new ztoc.
+// buildSociLayer builds a ztoc for an image layer (`desc`) and returns ztoc descriptor.
+// It may skip building ztoc (e.g., if layer size < `minLayerSize`) and return nil.
 func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descriptor) (*ocispec.Descriptor, error) {
 	if !images.IsLayerType(desc.MediaType) {
 		return nil, errNotLayerType
@@ -362,11 +365,12 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 		fmt.Printf("layer %s -> ztoc skipped\n", desc.Digest)
 		return nil, nil
 	}
+
 	compression, err := images.DiffCompression(ctx, desc.MediaType)
 	if err != nil {
 		return nil, fmt.Errorf("could not determine layer compression: %w", err)
 	}
-	if compression != "gzip" {
+	if compression != ztoc.CompressionGzip {
 		return nil, fmt.Errorf("layer %s (%s) must be compressed by gzip, but got %q: %w",
 			desc.Digest, desc.MediaType, compression, errUnsupportedLayerFormat)
 	}
@@ -391,7 +395,7 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 		return nil, errors.New("the size of the temp file doesn't match that of the layer")
 	}
 
-	toc, err := ztoc.BuildZtoc(tmpFile.Name(), b.config.spanSize, b.config.buildToolIdentifier)
+	toc, err := b.ztocBuilder.BuildZtoc(tmpFile.Name(), b.config.spanSize, ztoc.WithCompression(compression))
 	if err != nil {
 		return nil, err
 	}
@@ -426,7 +430,7 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 
 	ztocDesc.MediaType = SociLayerMediaType
 	ztocDesc.Annotations = map[string]string{
-		IndexAnnotationImageLayerMediaType: ocispec.MediaTypeImageLayerGzip,
+		IndexAnnotationImageLayerMediaType: desc.MediaType,
 		IndexAnnotationImageLayerDigest:    desc.Digest.String(),
 	}
 	return &ztocDesc, err
@@ -461,10 +465,7 @@ func skipBuildingZtoc(desc ocispec.Descriptor, cfg *buildConfig) bool {
 		return false
 	}
 	// avoid the file access if the layer size is below threshold
-	if desc.Size < cfg.minLayerSize {
-		return true
-	}
-	return false
+	return desc.Size < cfg.minLayerSize
 }
 
 // GetImageManifestDescriptor gets the descriptor of image manifest
--- a/util/testutil/tar.go
+++ b/util/testutil/tar.go
@@ -43,6 +43,8 @@ import (
 	"os"
 	"strings"
 	"time"
+
+	"github.com/klauspost/compress/zstd"
 )
 
 // TarEntry is an entry of tar.
@@ -118,6 +120,39 @@ func BuildTarGz(ents []TarEntry, compressionLevel int, opts ...BuildTarOption) i
 			return
 		}
 		if err := gw.Close(); err != nil {
+			pw.CloseWithError(err)
+			return
+		}
+		pw.Close()
+	}()
+	return pr
+}
+
+// BuildTarZstd builds a tar blob with zstd compression.
+func BuildTarZstd(ents []TarEntry, compressionLevel int, opts ...BuildTarOption) io.Reader {
+	var bo BuildTarOptions
+	for _, o := range opts {
+		o(&bo)
+	}
+	pr, pw := io.Pipe()
+	go func() {
+		zw, err := zstd.NewWriter(pw, zstd.WithEncoderLevel(zstd.EncoderLevelFromZstd(compressionLevel)))
+		if err != nil {
+			pw.CloseWithError(err)
+			return
+		}
+		tw := tar.NewWriter(zw)
+		for _, ent := range ents {
+			if err := ent.AppendTar(tw, bo); err != nil {
+				pw.CloseWithError(err)
+				return
+			}
+		}
+		if err := tw.Close(); err != nil {
+			pw.CloseWithError(err)
+			return
+		}
+		if err := zw.Close(); err != nil {
 			pw.CloseWithError(err)
 			return
 		}
--- a/ztoc/toc_builder.go
+++ b/ztoc/toc_builder.go
@@ -1,0 +1,203 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package ztoc
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/awslabs/soci-snapshotter/compression"
+	"github.com/klauspost/compress/zstd"
+)
+
+// TarProvider creates a tar reader from a compressed file reader (e.g., a gzip file reader),
+// which can be used by `TocBuilder` to create `TOC` from it.
+type TarProvider func(file *os.File) (io.Reader, error)
+
+// TarProviderGzip creates a tar reader from gzip reader.
+func TarProviderGzip(compressedReader *os.File) (io.Reader, error) {
+	return gzip.NewReader(compressedReader)
+}
+
+// TarProviderZstd creates a tar reader from zstd reader.
+func TarProviderZstd(compressedReader *os.File) (io.Reader, error) {
+	return zstd.NewReader(compressedReader)
+}
+
+// TocBuilder builds the `TOC` part of a ztoc and works with different
+// compression algorithms (e.g., gzip, zstd) with a registered `TarProvider`.
+type TocBuilder struct {
+	tarProviders map[string]TarProvider
+}
+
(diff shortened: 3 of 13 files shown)


soci/soci_index.go
```diff
@@ -40,23 +40,23 @@ import (
...
+	// IndexAnnotationImageLayerDigest is the index annotation for image layer digest
 	IndexAnnotationImageLayerDigest = "com.amazon.soci.image-layer-digest"
-	// index annotation for build tool identifier
+	// IndexAnnotationBuildToolIdentifier is the index annotation for build tool identifier
 	IndexAnnotationBuildToolIdentifier = "com.amazon.soci.build-tool-identifier"
-	// media type for OCI Artifact manifest
+	// OCIArtifactManifestMediaType is the media type for OCI Artifact manifest
 	OCIArtifactManifestMediaType = "application/vnd.oci.artifact.manifest.v1+json"
-	// default span size (4MiB)
+	// defaultSpanSize (4MiB)
```
hanyuel, line 55: This documentation doesn't look clearer than the before.

soci/soci_index.go
```diff
@@ -40,23 +40,23 @@ import (
...
 	defaultSpanSize = int64(1 << 22)
-	// min layer size (10MiB)
+	// minLayerSize to build a ztoc for a layer (10MiB)
```
hanyuel, line 57: We should have better explanation of what "min layer" means.

soci/soci_index.go
djdongjin: This more follows [go doc comment](https://go.dev/doc/comment) that comments start with name. I'm okay to change to back since it's not exported.

soci/soci_index.go
djdongjin: A better place to explain this is the cli flag [`Usage`](https://github.com/awslabs/soci-snapshotter/blob/1e87baff6ad9fec046c39e3ed7c4245b40efea79/cmd/soci/commands/create.go#L55) which is user-facing and [`WithMinLayerSize `](https://github.com/awslabs/soci-snapshotter/blob/1e87baff6ad9fec046c39e3ed7c4245b40efea79/soci/soci_index.go#L148) which is exported.

Do you think the cli usage is clear enough? (`The minimum layer size in bytes to build zTOC for. Default is 10 MiB.`) Or we can change to `The minimum layer size in bytes to build zTOC for. Default is 10 MiB. Layers smaller than this won't have zTOC (and not lazy pulled).`

ztoc/zinfo_builder.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/awslabs/soci-snapshotter/compression"
+	"github.com/opencontainers/go-digest"
+)
+
+// ZinfoBuilder builds the `zinfo` part of a ztoc. This interface should be
+// implemented for each compression algorithm we support.
+type ZinfoBuilder interface {
+	// ZinfoFromFile builds zinfo given a compressed tar filename and span size, and calculate the size of the file.
+	ZinfoFromFile(filename string, spanSize int64) (zinfo CompressionInfo, fs compression.Offset, err error)
```
hanyuel, line 32: Can we please rename [CompressionInfo](https://github.com/awslabs/soci-snapshotter/blob/1e87baff6ad9fec046c39e3ed7c4245b40efea79/ztoc/ztoc.go#L59) to ``zInfo``? This naming inconsistency is confusing.

djdongjin: Agree, especially we're calling them `zinfo` in most cases (e.g. docs). Will open an issue to track it separately since it requires more changes (cli, fs, ztoc and flatbuffer).

soci/soci_index.go
```diff
@@ -355,13 +341,46 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
...
+func skipBuildingZtoc(desc ocispec.Descriptor, cfg *buildConfig) bool {
+	if cfg == nil {
+		return false
+	}
+	// avoid the file access if the layer size is below threshold
+	if desc.Size < cfg.minLayerSize {
+		return true
+	}
+	return false
+}
```
rdpsin, line 381: np: Can just do `return desc.Size < cfg.minLayerSize`.

util/testutil/tar.go
```diff
@@ -122,6 +126,58 @@ func BuildTarGz(ents []TarEntry, compressionLevel int, opts ...BuildTarOption) i
 	return pr
 }
 
+// BuildTarZstd builds a tar blob with zstd compression.
```
rdpsin, line 129: Do we need this? At least not until we have zstd decompression right?

ztoc/toc_builder.go
```diff
@@ -0,0 +1,207 @@
...
+type TocBuilder struct{}
+
+// NewTocBuilder return a `TocBuilder` struct.
+func NewTocBuilder() TocBuilder {
+	return TocBuilder{}
+}
+
+// TocFromFile creates a `TOC` given a layer blob filename and the compression
+// algorithm used by the layer.
+func (tb TocBuilder) TocFromFile(algorithm, filename string) (TOC, compression.Offset, error) {
```
rdpsin, line 41: How about passing an `io.Reader` here instead of `filename`?

ztoc/ztoc_builder.go
```diff
@@ -17,238 +17,86 @@
...
 
-func getGzipFileMetadata(gzipFile string, index *compression.GzipZinfo) ([]FileMetadata, compression.Offset, error) {
-	file, err := os.Open(gzipFile)
-	if err != nil {
-		return nil, 0, fmt.Errorf("could not open file for reading: %v", err)
+// NewBuilder creates a `Builder` used to build ztocs.
+func NewBuilder() *Builder {
+	return &Builder{
+		tocBuilder: NewTocBuilder(),
+		zinfoBuilders: map[string]ZinfoBuilder{
```
rdpsin, line 51: Instead of having a map here, can we pass in the `ZinfoBuilder` as an argument to `NewBuilder`?

util/testutil/tar.go
djdongjin: `toc` builder is suppose to work with different compression, and the implementation in this PR supports both gzip and zstd. This PR also has unittest for `toc` builder with zstd.

ztoc/ztoc_builder.go
djdongjin: We don't know beforehand which compression will be used in `IndexBuilder.Build(ctx, img)`. `img` might have layers with both `gzip` and `zstd`, since `compression` is a [per-layer property](https://github.com/awslabs/soci-snapshotter/blob/12b4ceff9a9c82fd588a42a25c3965bc8ef9c270/soci/soci_index.go#L294).

Using a map makes `ztoc.Builder` management simpler since the `ztoc.Builder` will just be part of `IndexBuilder` (1 index builder has 1 ztoc builder which handles layers with different compressions). Also, I don't think there will be many compressions (IIRC, containerd recognize gzip and zstd and others will be just unknow).

If `ZinfoBuilder` is passed as an argument, that means:

1. a `ztoc.Builder` is bind to that `ZinfoBuilder` (e.g., gzip). So the `IndexBuilder` needs to figure out, for each layer, which compression is used and pass that to `NewBuilder` 
2. `IndexBuilder` need to call `NewBuilder` for every layer.
3. `ztoc.Builder` has a `tocBuilder` that supports different compression but a `zinfoBuilder` that supports only one compression.

ztoc/toc_builder.go
djdongjin: Do you mean only for `TocBuilder.TocFromFile` or both `TocBuilder.TocFromFile` and `ZinfoBuilder.ZinfoFromFile`?

I want to keep the index-related change minimum in this PR and only focus on ztoc/compression decouple. Currently the main change on soci index side is basically 1 line:

```golang
// within func (b *IndexBuilder) buildSociLayer

// before
toc, err := ztoc.BuildZtoc(tmpFile.Name(), b.config.spanSize, b.config.buildToolIdentifier)
// after
toc, err := b.ztocBuilder.BuildZtoc(tmpFile.Name(), b.config.spanSize, b.config.buildToolIdentifier, ztoc.WithCompression(compression))
```

I think changing `filename` to `io.Reader` itself is a big change since it comes from soci index builder down to `tocbuilder` and `zinfobuilder` (which down to zinfo compression impl, e.g., `NewGzipZinfoFromFile(gzipFile string, spanSize int64) (*GzipZinfo, error)`).

Will open a separate issue to track that.

rdpsin: For now, just `TocFromFile`. `ZinfoFromFile` is a more involved change since it would require updating the C library. But the Go way is passing around `io.Readers`, not filenames.

djdongjin: If only `TocFromFile` I'd prefer passing filename to both `TocFromFile` and `ZinfoFromFile` now for consistency:

1. The filename is passed from index builder to ztoc builder anyway (and used by zinfo builder).
2. Currently, ztoc builder is a thin warp of zinfo builder and toc builder (no other logic than, e.g., compression support check). This makes logic separation more clearly.
3. If change `TocFromFile` only, the difference will be just moving [`build compress io.Reader from filename` logic](https://github.com/awslabs/soci-snapshotter/blob/62e5d5bc7517ac01779df6b24a5bead8aba261bb/ztoc/toc_builder.go#L79) up to ztoc builder from toc builder.

Agree that passing `io.Reader`s is more Go way but I didn't see much benefit of only changing `Toc` while keeping `Zinfo`. Probably we should change this together starting from index builder to both toc and zinfo.

WDYT @rdpsin

rdpsin: Sure. Can we have an issue to track this?

djdongjin: Sure, will open an issue later with details :)

soci/soci_index.go
```diff
@@ -65,7 +65,7 @@ var (
 	errUnsupportedLayerFormat = errors.New("unsupported layer format")
 )
 
-// Index represents an ORAS/OCI Artifact Manifest
+// Index represents an ORAS/OCI Artifact Manifest.
```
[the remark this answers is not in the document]
djdongjin: SGTM, created #368 to add a `.golangci.yml` for more linter coverage.

ztoc/ztoc_builder.go
```diff
@@ -17,238 +17,86 @@
...
 	}
-	defer file.Close()
+}
 
-	gzipRdr, err := gzip.NewReader(file)
-	if err != nil {
-		return nil, 0, fmt.Errorf("could not create gzip reader: %v", err)
+// BuildZtoc builds a `Ztoc` given the filename of a layer blob. By default it assumes
+// the layer is compressed using `gzip`, unless specified via `WithCompression`.
+func (b *Builder) BuildZtoc(filename string, span int64, buildToolIdentifier string, options ...BuildOption) (*Ztoc, error) {
```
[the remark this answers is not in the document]
djdongjin: SGTM. I can change it in this PR since it's not a big change.

soci/soci_index.go
```diff
@@ -333,6 +339,8 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 
 	ztocDesc.MediaType = SociLayerMediaType
 	ztocDesc.Annotations = map[string]string{
+		// TODO(djdongjin): when we support new compression algorithms (e.g., zstd), we need to
+		// update the annotation value here (e.g., ocispec.MediaTypeImageLayerZstd).
 		IndexAnnotationImageLayerMediaType: ocispec.MediaTypeImageLayerGzip,
```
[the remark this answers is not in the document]
djdongjin: fixed

ztoc/ztoc_builder.go
djdongjin: SGTM.

For `tocBuilder` to support a new algorithm, only a helper func that _creates a tar reader from the compression reader_ is needed and the rest logic is the same.

For `zinfoBuilder`, different algorithms require different implementation (e.g., `GzipZinfoBuilder` and underlying `compression.GzipZinfo`).

I'll add a register to `ztoc.Builder` that accepts a compression name, a tar reader helper, a `zinfoBuilder` instance, and we can have a default register for gzip (e.g., `ztoc.RegisterGzip`).

ztoc/ztoc_builder.go
```diff
@@ -17,238 +17,86 @@
...
-func getGzipFileMetadata(gzipFile string, index *compression.GzipZinfo) ([]FileMetadata, compression.Offset, error) {
-	file, err := os.Open(gzipFile)
-	if err != nil {
-		return nil, 0, fmt.Errorf("could not open file for reading: %v", err)
+// NewBuilder creates a `Builder` used to build ztocs.
+func NewBuilder() *Builder {
+	return &Builder{
+		tocBuilder: NewTocBuilder(),
+		zinfoBuilders: map[string]ZinfoBuilder{
+			"gzip": gzipZinfoBuilder{},
```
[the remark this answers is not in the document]
djdongjin: changed to use const strings for compression names, as strings are returned from containerd and personally I think const strings add sufficient protection/clarification and easy to use.

ztoc/ztoc_builder.go
djdongjin: Added a register func to `ztoc.Builder`. By default `NewBuilder` registers `gzip`, and users can call `builder.RegisterCompressionAlgorithm` to support/register new compressions.

Initially I implmeneted it as an `Option` to `NewBuilder` (i.e., `WithCompressionAlgorithm` which is just a wraper of `RegisterCompressionAlgorithm`) but didn't see much benefit and it's not some config/option that has to be finalized during init. So I only kept `RegisterCompressionAlgorithm`.

ztoc/toc_builder.go
```diff
@@ -0,0 +1,203 @@
...
+	defer os.Remove(uncompressFile.Name())
+
+	uncompressFileSize, err := io.Copy(uncompressFile, compressTarReader)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	// create toc from tar reader.
+	tarSectionReader := io.NewSectionReader(uncompressFile, 0, uncompressFileSize)
+	md, err := metadataFromTarReader(tarSectionReader)
```
djdongjin, line 116: I didn't change this `create .tar temp file then read from it as the tar reader` to keep it consistent with existing impl. But I think we can just pass the `compressTarReader` to `metadataFromTarReader` (as the arg for `tar.NewReader`), and let `metadataFromTarReader` returns the `uncompressFileSize` as well since it has `positionTrackerReader`.

This can  avoid creating yet another temp `.tar`. (in soci index create we create a temp `.tar.gz` as input to ztoc builder)

ztoc/ztoc_builder.go
```diff
@@ -17,238 +17,110 @@
...
+	buildToolIdentifier string
+}
 
-	compressionInfo := CompressionInfo{
-		MaxSpanID:   index.MaxSpanID(),
-		SpanDigests: digests,
-		Checkpoints: checkpoints,
+// NewBuilder creates a `Builder` used to build ztocs. By default it supports gzip,
+// user can register new compression algorithms by calling `RegisterCompressionAlgorithm`.
+func NewBuilder(buildToolIdentifier string) *Builder {
```
Kern--, line 36: FYI @dvnguyen-amzn this moves `buildToolIdentifier` from `builder.Build` to `ztoc.NewBuilder`.

Kern-- (approved):

rdpsin (approved):

PR #350, opened 2023-01-25
Refactor integ test `buildIndex`

### LINKED ISSUE #349: [FEATURE] Refactor Integ test `buildIndex` to use functional args

Our integ tests currenly have 2 forms of `buildIndex`, one that takes no special parameters, and one that takes `minLayerSize` and `spanSize`. 

https://github.com/awslabs/soci-snapshotter/blob/6137c697a6382707397a248824fa5ec2086d7a2f/integration/testutil_soci.go#L55

https://github.com/awslabs/soci-snapshotter/blob/6137c697a6382707397a248824fa5ec2086d7a2f/integration/testutil_soci.go#L60

The first is convenience because we often don't need anything special for the test. However we often only change one of `minLayerSize` and `spanSize` when we call the second. 

- [x] A single `func buildIndex(sh *shell.Shell, src imageInfo, opts ...buildIndexOpts) string` that uses functional opts so that we only need to override the specific parameters that matter for the test. 
This makes the call site clearer since the positional arguments are replaced with function names that highlight specifically what they're changing. It also makes the tests more explicit about which arguments matter (e.g. if they are set, then they're relevant to the test where as before both were always set).

- [ ] Switch to using `defaultSpanSize` and `defaultMinLayerSize` as the defaults. Currently we use 0 as the min layer size. It would be easier to understand our tests if we explicitly set that value when we need it.

We could add 2 more variants for `minLayerSize` and `spanSize` respectively, but that's still less clear and gets messier when we want to add more arguments. E.g. we will probably need and explicit flag for #339 to set whether to use OCI 1.0 or OCI 1.1.

**Additional context**
This is a follow up from https://github.com/awslabs/soci-snapshotter/pull/309#discussion_r1069885289

Kern--:

*Issue #, if available:*
partial implementation of #349 

*Description of changes:*
This change refactors `buildIndex` to use functional arguments.

This makes the call site clearer since the positional arguments are replaced with function names that specifically highlight what they are changing. It also makes it easier to add additional arguments without changing the call sites. Finally, it makes the tests more explicit about which arguments are relvant to the test (since they only need to be set if they are relevant)

*Testing performed:*
`make check && make integration`

### DIFF
--- a/integration/metrics_test.go
+++ b/integration/metrics_test.go
@@ -106,7 +106,7 @@ func TestOverlayFallbackMetric(t *testing.T) {
 			name:  "image with all layers having ztocs and no fs.Mount error results in 0 overlay fallback",
 			image: rabbitmqImage,
 			indexDigestFn: func(sh *shell.Shell, image imageInfo) string {
-				return buildSparseIndex(sh, image, 0, defaultSpanSize)
+				return buildIndex(sh, image, withMinLayerSize(0))
 			},
 			expectedFallbackCount: 0,
 		},
@@ -114,7 +114,7 @@ func TestOverlayFallbackMetric(t *testing.T) {
 			name:  "image with some layers not having ztoc and no fs.Mount results in 0 overlay fallback",
 			image: rabbitmqImage,
 			indexDigestFn: func(sh *shell.Shell, image imageInfo) string {
-				return buildSparseIndex(sh, image, defaultMinLayerSize, defaultSpanSize)
+				return buildIndex(sh, image, withMinLayerSize(defaultMinLayerSize))
 			},
 			expectedFallbackCount: 0,
 		},
@@ -184,7 +184,7 @@ log_fuse_operations = true
 			name:  "image with valid ztocs and index doesn't cause fuse file.read failures",
 			image: rabbitmqImage,
 			indexDigestFn: func(t *testing.T, sh *shell.Shell, image imageInfo) string {
-				return buildSparseIndex(sh, image, 0, defaultSpanSize)
+				return buildIndex(sh, image)
 			},
 			// even a valid index/ztoc produces some fuse operation failures such as
 			// node.lookup and node.getxattr failures, so we only check a specific fuse failure metric.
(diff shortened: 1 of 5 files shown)


integration/metrics_test.go
```diff
@@ -106,15 +106,15 @@ func TestOverlayFallbackMetric(t *testing.T) {
 			name:  "image with all layers having ztocs and no fs.Mount error results in 0 overlay fallback",
 			image: rabbitmqImage,
 			indexDigestFn: func(sh *shell.Shell, image imageInfo) string {
-				return buildSparseIndex(sh, image, 0, defaultSpanSize)
+				return buildIndex(sh, image, withMinLayerSize(0), withSpanSize(defaultSpanSize))
```
djdongjin, line 109: `withMinLayerSize(0), withSpanSize(defaultSpanSize)` can be removed.

integration/metrics_test.go
```diff
@@ -106,15 +106,15 @@ func TestOverlayFallbackMetric(t *testing.T) {
...
 			},
 			expectedFallbackCount: 0,
 		},
 		{
 			name:  "image with some layers not having ztoc and no fs.Mount results in 0 overlay fallback",
 			image: rabbitmqImage,
 			indexDigestFn: func(sh *shell.Shell, image imageInfo) string {
-				return buildSparseIndex(sh, image, defaultMinLayerSize, defaultSpanSize)
+				return buildIndex(sh, image, withMinLayerSize(defaultMinLayerSize), withSpanSize(defaultSpanSize))
```
djdongjin, line 117: `withSpanSize(defaultSpanSize)` can be removed.

integration/metrics_test.go
Kern--: Agreed on `withSpanSize`. I opted to keep `withMinLayerSize` because that is important to this test. I think it's good to be explicit when it's relevant, even if it's a no-op. Thoughts?

djdongjin: SGTM either explicit or implicit.

Meanwhile, maybe we should take both `defaultSpanSize` and `defaultMinLayerSize` as the default for `buildIndex` which is consistent with cli. Then if we anything other than default we should call out (e.g., `withMinLayerSize(0)`).

Currently `buildIndex` uses `defaultSpanSize` as default for span size, but `0` for `minLayerSize` which is confusing, given that we have the `defaultMinLayerSize` and the func name `buildIndex` is generic (i.e., not `buildSparseIndex` or `buildFullIndex`)

hanyuel (dismissed):

rdpsin (dismissed):

integration/metrics_test.go
Kern--: I think we can switch that in a follow up. It will also give us a chance to look at the tests and figure out which should have a `spanSize` set to 0 and which shouldn't.

djdongjin: SGTM. We can open an issue for that or just keep #349 opened.

Kern--: I updated #349 to include updating those.

rdpsin (approved):

djdongjin (approved):

PR #356, opened 2023-01-26
Support Oci 1.0

### LINKED ISSUE #339: Support OCI 1.0
SOCI works by associating a SOCI index with an OCI image. When pulling an image, the SOCI snapshotter will look for an associated SOCI index in the container registry and, if found, use it to lazily load the image instead. SOCI stores indices in the registry as [OCI Artifacts](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/artifact.md) and uses the registry's [Referrers API](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers) to discover the relation between an image and index. 

Artifacts and the Referrers API are defined in v1.1 of the [OCI Image Spec](https://github.com/opencontainers/image-spec) and [OCI Distribution Spec](https://github.com/opencontainers/distribution-spec), respectively. Since v1.1 of these specs is not yet finalized, most registries do not support these features. Once the spec is finalized, there is likely to be a long tail of registries that won't add support for a long time if ever. 

To address this issue, v1.1 of the distribution spec has an [OCI v1.0 compatible fallback](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#backwards-compatibility). Artifacts can be stored using OCI Image manifests, and the response normally sent by the referrers API can be stored in a blob that's tagged with the subject image's digest.

The SOCI snapshotter and CLI works with OCI 1.0 compatible registries that do not support any OCI 1.1 features.

Specifically I think we should
1) Create OCI 1.1 Artifacts by default. OCI 1.0 fallback should be an explicit flag in the CLI/library
2) Automatically handle both OCI 1.1 and 1.0 artifacts in the snapshotter on pull
 
We could support just OCI 1.1. If we went this route we would probably leave many users in the dark and prevent people from utilizing the benefits that SOCI can offer.

**Additional context**
The ORAS go library has supported automatic fallback to OCI 1.0 compatibole mechanisms when pushing/pulling artifacts since [v2.0.0-rc4](https://github.com/oras-project/oras-go/releases/tag/v2.0.0-rc.4). We should get the registry interaction for free. What isn't free is serializing the SOCI index as an OCI Artifact for OCI v1.1 and as an OCI Image for OCI v1.0. Similarly, we will need to be able to deserialize both.

Kern--:
*Issue #, if available:*

closes #339 

*Description of changes:*
Support OCI 1.0
    
    SOCI relies on being able to associate an index with an image. This
    capability was introduced in OCI distribution and image specs version
    1.1 as Artifacts and the Referrers API respectively.
    
    In order to drive adoption, an OCI 1.0 compatible fallback was added
    that encodes artifacts as Image Manifests and stores the Referrers API
    content in an Image Index.
    
    This PR adds the ability to serialize and deserialize SOCI indices as
    either OCI Artifacts or OCI Images. There is no automatic OCI 1.0
    fallback when creating images. The user of the CLI/library must
    explicitly request an OCI 1.0 compatible Image Manifest. The snapshotter
    will be able to automatically load a SOCI index in either serialized
    form.

*Testing performed:*

### DIFF
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -45,6 +45,7 @@ var CreateCommand = cli.Command{
 	ArgsUsage: "[flags] <image_ref>",
 	Flags: append(
 		internal.PlatformFlags,
+		internal.LegacyRegistryFlag,
 		cli.Int64Flag{
 			Name:  spanSizeFlag,
 			Usage: "Span size that soci index uses to segment layer data. Default is 4 MiB",
@@ -96,12 +97,16 @@ var CreateCommand = cli.Command{
 			return err
 		}
 
+		builderOpts := []soci.BuildOption{
+			soci.WithMinLayerSize(minLayerSize),
+			soci.WithSpanSize(spanSize),
+			soci.WithBuildToolIdentifier(buildToolIdentifier),
+		}
+		if cliContext.Bool(internal.LegacyRegistryFlagName) {
+			builderOpts = append(builderOpts, soci.WithLegacyRegistrySupport)
+		}
 		for _, plat := range ps {
-			builder, err := soci.NewIndexBuilder(cs, blobStore,
-				soci.WithMinLayerSize(minLayerSize),
-				soci.WithSpanSize(spanSize),
-				soci.WithBuildToolIdentifier(buildToolIdentifier),
-				soci.WithPlatform(plat))
+			builder, err := soci.NewIndexBuilder(cs, blobStore, append(builderOpts, soci.WithPlatform(plat))...)
 
 			if err != nil {
 				return err
--- a/integration/push_test.go
+++ b/integration/push_test.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/awslabs/soci-snapshotter/util/testutil"
 	"github.com/containerd/containerd/platforms"
 )
@@ -131,6 +132,86 @@ func TestPushAlwaysMostRecentlyCreatedIndex(t *testing.T) {
 					t.Fatalf("incorrect index pushed to remote registry; expected %s, got %s", index, pushedIndex)
 				}
 			}
+		})
+	}
+}
+
+func TestLegacyOCI(t *testing.T) {
+	tests := []struct {
+		name                  string
+		registryImage         string
+		supportLegacyRegistry bool
+		expectError           bool
+	}{
+		{
+			name:                  "OCI 1.1 Artifacts succeed with OCI 1.1 registry",
+			registryImage:         oci11RegistryImage,
+			supportLegacyRegistry: false,
+			expectError:           false,
+		},
+		{
+			name:                  "OCI 1.0 Artifacts succeed with OCI 1.1 registry",
+			registryImage:         oci11RegistryImage,
+			supportLegacyRegistry: true,
+			expectError:           false,
+		},
+		{
+			name:                  "OCI 1.1 Artifacts fail with OCI 1.0 registry",
+			registryImage:         oci10RegistryImage,
+			supportLegacyRegistry: false,
+			expectError:           true,
+		},
+		{
+			name:                  "OCI 1.0 Artifacts succeed with OCI 1.0 registry",
+			registryImage:         oci10RegistryImage,
+			supportLegacyRegistry: true,
+			expectError:           false,
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			regConfig := newRegistryConfig()
+			sh, done := newShellWithRegistry(t, regConfig, withRegistryImageRef(tc.registryImage))
+			defer done()
+
+			if err := testutil.WriteFileContents(sh, defaultContainerdConfigPath, getContainerdConfigYaml(t, false), 0600); err != nil {
+				t.Fatalf("failed to write %v: %v", defaultContainerdConfigPath, err)
+			}
+			if err := testutil.WriteFileContents(sh, defaultSnapshotterConfigPath, getSnapshotterConfigYaml(t, false), 0600); err != nil {
+				t.Fatalf("failed to write %v: %v", defaultSnapshotterConfigPath, err)
+			}
+			rebootContainerd(t, sh, "", "")
+
+			imageName := ubuntuImage
+			copyImage(sh, dockerhub(imageName), regConfig.mirror(imageName))
+			var buildOpts []indexBuildOption
+			if tc.supportLegacyRegistry {
+				buildOpts = append(buildOpts, withLegacyRegistrySupport)
+			}
+
+			indexDigest := buildIndex(sh, regConfig.mirror(imageName), buildOpts...)
+			rawJSON := sh.O("soci", "index", "info", indexDigest)
+			var sociIndex soci.Index
+			if err := soci.UnmarshalIndex(rawJSON, &sociIndex); err != nil {
+				t.Fatalf("invalid soci index from digest %s: %v", indexDigest, rawJSON)
+			}
+			_, err := sh.OLog("soci", "push", "--user", regConfig.creds(), regConfig.mirror(imageName).ref)
+			hasError := err != nil
+			if hasError != tc.expectError {
+				t.Fatalf("unexpected error state: expected error? %v, got %v", tc.expectError, err)
+			} else if hasError {
+				// if we have an error and we expected an error, the test is done
+				return
+			}
+			sh.X("rm", "-rf", "/var/lib/soci-snapshotter-grpc/content/blobs/sha256")
+
+			sh.X("soci", "image", "rpull", "--user", regConfig.creds(), "--soci-index-digest", indexDigest, regConfig.mirror(imageName).ref)
+			if err := sh.Err(); err != nil {
+				t.Fatalf("failed to rpull: %v", err)
+			}
+			checkFuseMounts(t, sh, len(sociIndex.Blobs))
 		})
 	}
 }
--- a/integration/testutil.go
+++ b/integration/testutil.go
@@ -68,8 +68,13 @@ const (
 	dockerLibrary                = "public.ecr.aws/docker/library/"
 	blobStorePath                = "/var/lib/soci-snapshotter-grpc/content/blobs/sha256"
 	containerdBlobStorePath      = "/var/lib/containerd/io.containerd.content.v1.content/blobs/sha256"
+	// Registry images to use in the test infrastructure. These are not intended to be used
+	// as images in the test itself, but just when we're setting up docker compose.
+	oci11RegistryImage = "ghcr.io/oci-playground/registry:v3.0.0-alpha.1"
+	oci10RegistryImage = "docker.io/library/registry:2"
 )
 
+// These are images that we use in our integration tests
 const (
 	alpineImage   = "alpine:latest"
 	nginxImage    = "nginx:latest"
@@ -272,13 +277,27 @@ func (c registryConfig) mirror(imageName string, opts ...imageOpt) imageInfo {
 }
 
 type registryOptions struct {
-	network string
+	network          string
+	registryImageRef string
+}
+
+func defaultRegistryOptions() registryOptions {
+	return registryOptions{
+		network:          "",
+		registryImageRef: oci11RegistryImage,
+	}
 }
 
 type registryOpt func(o *registryOptions)
 
+func withRegistryImageRef(ref string) registryOpt {
+	return func(o *registryOptions) {
+		o.registryImageRef = ref
+	}
+}
+
 func newShellWithRegistry(t *testing.T, r registryConfig, opts ...registryOpt) (sh *shell.Shell, done func() error) {
-	var rOpts registryOptions
+	rOpts := defaultRegistryOptions()
 	for _, o := range opts {
 		o(&rOpts)
 	}
@@ -356,7 +375,7 @@ services:
     volumes:
     - /dev/fuse:/dev/fuse
   registry:
-    image: ghcr.io/oci-playground/registry:v3.0.0-alpha.1
+    image: {{.RegistryImageRef}}
     container_name: {{.RegistryHost}}
     environment:
     - REGISTRY_AUTH=htpasswd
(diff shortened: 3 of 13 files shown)


Kern--:
This is probably ready for review, but it needs to be rebased after https://github.com/awslabs/soci-snapshotter/pull/350

cmd/soci/commands/create.go
```diff
@@ -98,10 +104,11 @@ var CreateCommand = cli.Command{
...
 		for _, plat := range ps {
 			builder, err := soci.NewIndexBuilder(cs, blobStore,
-				soci.WithMinLayerSize(minLayerSize),
-				soci.WithSpanSize(spanSize),
-				soci.WithBuildToolIdentifier(buildToolIdentifier),
-				soci.WithPlatform(plat))
+				append(builderOpts,
+					soci.WithMinLayerSize(minLayerSize),
+					soci.WithSpanSize(spanSize),
+					soci.WithBuildToolIdentifier(buildToolIdentifier),
```
djdongjin, line 110: Given these 3 and `WithLegacyRegistrySupport` are the same and only `plat` is different, should we just put them to `builderOpts` and only append `plat` here. E.g.,

in line 66-70 (also need to move it to after getting minLayerSize and spanSize):

```golang
		builderOpts := []soci.BuildOption{
           soci.WithMinLayerSize(minLayerSize),
		   soci.WithSpanSize(spanSize),
		   soci.WithBuildToolIdentifier(buildToolIdentifier),
         }
		if cliContext.Bool(internal.LegacyRegistryFlagName) {
			builderOpts = append(builderOpts, soci.WithLegacyRegistrySupport)
		}
```

in line 106:

```golang
builder, err := soci.NewIndexBuilder(cs, blobStore, append(builderOpts, soci.WithPlatform(plat)))
```

integration/testutil.go
```diff
@@ -68,6 +68,8 @@ const (
 	dockerLibrary                = "public.ecr.aws/docker/library/"
 	blobStorePath                = "/var/lib/soci-snapshotter-grpc/content/blobs/sha256"
 	containerdBlobStorePath      = "/var/lib/containerd/io.containerd.content.v1.content/blobs/sha256"
+	OCI11RegistryImage           = "ghcr.io/oci-playground/registry:v3.0.0-alpha.1"
+	OCI10RegistryImage           = "docker.io/library/registry:2"
```
djdongjin, line 72: nit: better to put them in the next const block with other `Image`, also mayber `OCI` -> `oci`

integration/testutil.go
```diff
@@ -372,13 +388,15 @@ services:
 		ServiceName     string
 		ImageContextDir string
 		TargetStage     string
+		RegistryImage   string
```
djdongjin, line 391: nit: `RegistryImage` -> `RegistryImageRef` to be consistent.

integration/testutil_soci.go
```diff
@@ -91,12 +96,21 @@ func buildIndex(sh *shell.Shell, src imageInfo, opt ...indexBuildOption) string
 		o(&indexBuildConfig)
 	}
 	opts := encodeImageInfo(src)
+
+	var createArgs []string
+	if indexBuildConfig.supportLegacyRegistry {
+		createArgs = append(createArgs, "--legacy-registry")
+	}
```
djdongjin, line 103: nit: move this to after line 106 (and change `=` to `:=` in line 106) to remove explicit declare in line 100 and colocate `createArgs` logic

soci/soci_index.go
```diff
@@ -94,6 +112,60 @@ type IndexDescriptorInfo struct {
...
+	}
+	var manifest ocispec.Manifest
+	manifest.SchemaVersion = 2
+	manifest.MediaType = ocispec.MediaTypeImageManifest
+	manifest.Config = defaultConfigDescriptor
+	manifest.Layers = i.Blobs
+	manifest.Subject = i.Subject
+	manifest.Annotations = i.Annotations
+	return json.Marshal(manifest)
+}
```
djdongjin, line 167: Should we change the func signature to

```golang
func MarshalIndex(i *Index) ([]byte, error)
```

to be consistent with `UnmarshalIndex`.

For `ztoc`, we also have `func Unmarshal(serializedZtoc io.Reader) (*Ztoc, error)` and `func Marshal(ztoc *Ztoc) (io.Reader, ocispec.Descriptor, error)` (i.e., not `func (ztoc *Ztoc) Marshal() (io.Reader, ocispec.Descriptor, error)`).

soci/soci_index.go
```diff
@@ -130,10 +202,11 @@ func GetIndexDescriptorCollection(ctx context.Context, cs content.Store, img ima
...
 type buildConfig struct {
-	spanSize            int64
-	minLayerSize        int64
-	buildToolIdentifier string
-	platform            ocispec.Platform
+	spanSize              int64
+	minLayerSize          int64
+	buildToolIdentifier   string
+	platform              ocispec.Platform
+	supportLegacyRegistry bool
```
djdongjin, line 209: I think we should use `legacyRegistry` than `supportLegacyRegistry`. `supportLegacyRegistry` indicates that, if false(not supported), then some functionality in soci might missing.

With this PR, we're supporting both `legacyRegistry` (false) and non-`legacyRegistry` (true). And `legacyRegistry` is just an indicator if we're using soci for a `legacyRegistry` or not.

soci/soci_index.go
```diff
@@ -94,6 +112,60 @@ type IndexDescriptorInfo struct {
...
+	}
+	return UnmarshalIndex(b, index)
+}
+
+// UnmarshalIndex deserializes a JSON blob in a byte array
+// into a SOCI index. The blob can be either an OCI 1.1 Artifact
+// or an OCI 1.0 Manifest
+func UnmarshalIndex(b []byte, index *Index) error {
+	err := json.Unmarshal(b, index)
+	if err == nil && index.MediaType == ocispec.MediaTypeArtifactManifest {
```
sbuckfelder, line 131: maybe add a log statement (debug?) that this is a 1.1 artifact?

soci/soci_index.go
```diff
@@ -94,6 +112,60 @@ type IndexDescriptorInfo struct {
...
+		return nil
+	}
+	var manifest ocispec.Manifest
+	err = json.Unmarshal(b, &manifest)
+	if err == nil {
+		fromManifest(manifest, index)
```
sbuckfelder, line 137: maybe add a log statement (debug?) that this is a 1.0 manifest?

soci/soci_index.go
```diff
@@ -94,6 +112,60 @@ type IndexDescriptorInfo struct {
...
+	if err != nil {
+		return err
+	}
+	return UnmarshalIndex(b, index)
+}
+
+// UnmarshalIndex deserializes a JSON blob in a byte array
+// into a SOCI index. The blob can be either an OCI 1.1 Artifact
+// or an OCI 1.0 Manifest
+func UnmarshalIndex(b []byte, index *Index) error {
```
sbuckfelder, line 129: is this public just for testing?

soci/soci_index.go
```diff
@@ -94,6 +112,60 @@ type IndexDescriptorInfo struct {
...
+	index.Blobs = manifest.Layers
+	index.Subject = manifest.Subject
+	index.Annotations = manifest.Annotations
+}
+
+// Marshal serializes a SOCI index into a JSON blob.
+// The JSON blob will be either an OCI 1.1 Artifact or
+// an OCI 1.0 Manifest, depending on how the index was created.
+func (i *Index) Marshal() ([]byte, error) {
+	if i.MediaType == OCIArtifactManifestMediaType {
```
sbuckfelder, line 156: maybe be specific in log or comment which paths are 1.1 and which are 1.0?  Honestly my preference would break this out to make it easier to add future types and be crystal clear.  Into something like :

```
func (i *Index) Marshal() ([]byte, error) {
	if i.MediaType == OCIArtifactManifestMediaType {
              return marshalOCI11(i)
        } else  {
              return marshalOCI10(i)
        }
```

but honestly I'm flexible

soci/soci_index.go
```diff
@@ -94,6 +112,60 @@ type IndexDescriptorInfo struct {
 	CreatedAt time.Time
 }
 
+// DecodeIndex deserializes a JSON blob in an io.Reader
+// into a SOCI index. The blob can be either an OCI 1.1 Artifact
+// or an OCI 1.0 Manifest
+func DecodeIndex(r io.Reader, index *Index) error {
```
sbuckfelder, line 118: any interest in `(i *Index) func DecodeIndex(r io.Reader) error`?

soci/soci_index.go
sbuckfelder, line 129: any interest in `(i *Index) func UnmarshalIndex(b []byte) error`?

sbuckfelder:
Just want to confirm that I understand what this PR does.  Currently soci supports and relies oci 1.1 artifact types, this PR adds support for both CREATING and READING oci 1.0 image types.

The meat of the pr is in the changes to `soci_index.go`.  Where as the creating and reading the index used happen directly with JSON package, there are now abstractions that allow us to inject support for different manifest types.  Adds the following functions
- `func DecodeIndex(r io.Reader, index *Index) error`: for reading from an io.Reader
- `func UnmarshalIndex(b []byte, index *Index) error `: for reading from a byte array
- `func (i *Index) Marshal() ([]byte, error)`: for creating the file from an index object

Within these functions the logic splits out between 1.0 and 1.1 For reading it checks whether the media type is an artifact (1.1) or an image (1.0).  If it is type 1.0 it calls a helper function to add fields to make it feel and act like a 1.1 artifact.  For building it relies on a new flag called `legacy-registry`.  If this is present is builds a 1.0 style soci index.  This flag has been plumbed through.

I haven't had a chance to go through all the tests.  What do they do?

Anything choices that were tricky while making this PR?  Or full steam confidence all the way?

soci/soci_index.go
```diff
@@ -94,6 +112,60 @@ type IndexDescriptorInfo struct {
...
+	err = json.Unmarshal(b, &manifest)
+	if err == nil {
+		fromManifest(manifest, index)
+		return nil
+	}
+	return err
+}
+
+// fromManifest converts and OCI 1.0 Manifest to a SOCI Index
+func fromManifest(manifest ocispec.Manifest, index *Index) {
```
sbuckfelder, line 144: maybe name this `convert10To11`

sbuckfelder:
On the testing side looks like you test both cases for marshall in `TestMarshalIndex` in `soci_index_test.go` and for creation you added `TestLegacyOCI` in `push_test.go`.  All the other test updates were to use the new abstractions.  Correct?

integration/testutil.go
Kern--: +1 to non-exported.

I prefer to keep them out of the next image block because these are not used as SOCI test images, but rather as the base test infrastructure. I can add a comment to make that clear.

soci/soci_index.go
Kern--: I guess so. I'm usually not a huge fan of package functions that operation on objects because it encourages thinking procedurally which was one of the problems we had with the index builder until we refactored it.

That said, serialization is probably not a property of the object itself and we previously removed the concept of a ztocmarshaller because it wasn't adding any value. I think this is the right change here, but I'm not totally sold on this pattern in general.

soci/soci_index.go
Kern--: I modeled this like the `json` package which has both `Unmarshal` and `Decode` so that it's convenient whether you have a `[]byte` or `io.Reader`. Since we have to be able to read the data twice, I'd prefer to leave this exported to avoid the extra copy if you go through `Decode(bytes.NewReader(b))`

integration/testutil.go
djdongjin: SGTM

soci/soci_index.go
Kern--: I think `Index{}.DecodeIndex(r)` is less idiomatic in go than
```
var index Index
Unmarshal(r, &index)
```

The latter at least matches the `json` package. Maybe `Unmarshall` should return a `*Index` instead like `ztoc.Unmarshal` does?

integration/push_test.go
```diff
@@ -134,3 +135,83 @@ func TestPushAlwaysMostRecentlyCreatedIndex(t *testing.T) {
...
+			indexDigest := buildIndex(sh, regConfig.mirror(imageName), buildOpts...)
+			rawJSON := sh.O("soci", "index", "info", indexDigest)
+			var sociIndex soci.Index
+			if err := soci.UnmarshalIndex(rawJSON, &sociIndex); err != nil {
+				t.Fatalf("invalid soci index from digest %s: %v", indexDigest, rawJSON)
+			}
+			_, err := sh.OLog("soci", "push", "--user", regConfig.creds(), regConfig.mirror(imageName).ref)
+			hasError := err != nil
+			if hasError != tc.expectError {
+				t.Fatalf("Unexpected error state: expected error? %v, got %v", tc.expectError, err)
```
djdongjin, line 203: nit: `Unexpected` -> `unexpected`.

soci/soci_index.go
Kern--: It's awkward to put logs here - we would need to pass a context through and this feels like a piece of the core library that other projects could build on.

djdongjin:
@Kern-- I merged #363 which changes to use `defaultMinLayerSize` instead of 0 in `buildIndex`. Just FYI in case you use it and rely on the previously default `0` minLayerSize.

Kern--:
> Just want to confirm that I understand what this PR does.

Yes. I don't want to quote your whole explanation, but it is correct.

> I haven't had a chance to go through all the tests. What do they do?

Your later comment implies you did check this, but just to confirm, 

`TestMarshalIndex` verifies that we can marshal both OCI Artifacts and OCI Image Manifests. The most important piece here is that we remember which type was serialized - if we serialized as a Manifest and read it back as an Artifact, then a subsequent marshalling as an Artifact would change the digest and likely break a bunch of stuff. (e.g. in the snapshotter where we re-serialize the manifest to store on disk for later container launches).

> Anything choices that were tricky while making this PR? Or full steam confidence all the way?

I think the choice of "legacy-registry" as the CLI flag is the most contentious. There's still discussion on the OCI Image spec about whether they should require OCI 1.0 Manifest compatible artifacts by default or if 1.1 Artifacts will go away entirely. I think we can still change this flag later, but we should be thinking about versioning soon and what sort of backwards compatibility we will provide.

sbuckfelder (approved):

djdongjin (approved):

PR #374, opened 2023-01-31
Clean up makefile and move to scripts

### LINKED ISSUE #69: Move build-tool installation logic out of Makefile

Today, the Makefile contains a growing number of rules for installing build-time dependencies. This is an antipattern that hurts maintainability and portability. Makefiles should contain the logic to build project artifacts, and that is it. Dependencies should be documented in *README.md* or *BUILDING.md* and/or managed by a *configure.sh* script.

djdongjin:

*Issue #, if available:* fix #69

*Description of changes:*

1. Move `check`, `install-dependency`, `install-check-tools` related make rules to scripts.
2. Fix existing bugs in these rules:
    1. `install-flatc` uses `make install` instead of `sudo make install` which will fail unless runned with `sudo`.
    2. `cd` into `flatc` and `zlib` to install them but didn't `cd ..` out before `rm`, which doesn't actually clean up downloaded files.

*Testing performed:*

make test, run scripts manually, check GH action output.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,11 +22,11 @@ jobs:
           go-version: '>=1.18.0'
       - run: wget https://github.com/google/flatbuffers/releases/download/v22.9.29/Linux.flatc.binary.g++-10.zip
       - run: unzip Linux.flatc.binary.g++-10.zip
-      - run: make install-check-tools
-      - run: make check-ltag
-      - run: make check-dco
-      - run: make check-lint
-      - run: PATH=$PATH:$(pwd) make check-flatc
+      - run: ./scripts/install-check-tools.sh
+      - run: ./scripts/check-ltag.sh
+      - run: ./scripts/check-dco.sh
+      - run: ./scripts/check-lint.sh
+      - run: PATH=$PATH:$(pwd) ./scripts/check-flatc.sh
   test:
     runs-on: ubuntu-20.04
     steps:
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,6 @@
 CMD_DESTDIR ?= /usr/local
 GO111MODULE_VALUE=auto
 OUTDIR ?= $(CURDIR)/out
-UTIL_CFLAGS=-I${CURDIR}/c -L${OUTDIR} -lz
-UTIL_LDFLAGS=-L${OUTDIR} -lz
 PKG=github.com/awslabs/soci-snapshotter
 VERSION=$(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
 REVISION=$(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
@@ -33,7 +31,7 @@ CMD=soci-snapshotter-grpc soci
 
 CMD_BINARIES=$(addprefix $(OUTDIR)/,$(CMD))
 
-.PHONY: all build check check-ltag check-dco check-lint check-flatc install-check-tools add-ltag install install-cmake install-flatc install-zlib uninstall clean test integration
+.PHONY: all build add-ltag install uninstall clean test integration
 
 all: build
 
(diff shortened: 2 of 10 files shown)


.github/workflows/build.yml
```diff
@@ -22,11 +22,8 @@ jobs:
           go-version: '>=1.18.0'
       - run: wget https://github.com/google/flatbuffers/releases/download/v22.9.29/Linux.flatc.binary.g++-10.zip
       - run: unzip Linux.flatc.binary.g++-10.zip
-      - run: make install-check-tools
-      - run: make check-ltag
-      - run: make check-dco
-      - run: make check-lint
-      - run: PATH=$PATH:$(pwd) make check-flatc
+      - run: ./scripts/install-check-tools.sh
+      - run: PATH=$PATH:$(pwd) ./scripts/check.sh
```
rdpsin, line 26: We had decided to keep all check tests separate to help devs debug which check failed exactly. Can we do the same here, by having all check tests in separate files?

scripts/install-dep.sh
```diff
@@ -0,0 +1,37 @@
...
+
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+echo "Install soci dependencies"
+set -eux -o pipefail
+
```
rdpsin, line 19: np: We should test the existence of these binaries/libraries before installing them (especially zlib, installing a different version may render dependent system libraries unusable).

.github/workflows/build.yml
djdongjin: SGTM. I'm also planning to just remove golangci-lint related stuff. I didn't see other repos have rules/scripts for it and installing/running it is straightforward (the script will be just 1 line anyway). And we should use `golangci/golangci-lint-action` in our CI for it. Any concern with this?

rdpsin: SGTM. But we need to make sure we also lint inside the `cmd` directory; since it's a separate package.

djdongjin: `golangci-lint run` without args uses `./...` and runs in all (sub)directories. https://golangci-lint.run/usage/quick-start/

Will also test in the action run.

Makefile
```diff
@@ -105,16 +58,16 @@ uninstall:
...
-	$(shell go env GOPATH)/bin/ltag $(LTAG_TEMPLATE_FLAG) -v
-
 clean:
 	rm -rf $(OUTDIR)
 
 vendor:
 	@GO111MODULE=$(GO111MODULE_VALUE) go mod tidy
 	@cd ./cmd ; GO111MODULE=$(GO111MODULE_VALUE) go mod tidy
 
+add-ltag:
```
Kern--, line 68: Why shouldn't `add-ltag` also be moved to a script?

scripts/check-flatc.sh
```diff
@@ -0,0 +1,27 @@
...
+
+CUR_DIR="$(dirname "${BASH_SOURCE[0]}")"
+SOCI_SNAPSHOTTER_PROJECT_ROOT="${CUR_DIR}/.."
+FBS_FILE_PATH=${SOCI_SNAPSHOTTER_PROJECT_ROOT}/ztoc/fbs/ztoc.fbs
+
+# check if flatbuffers needs to be generated again
+TMPDIR=$(mktemp -d)
+flatc -o ${TMPDIR} -g ${FBS_FILE_PATH}
+diff -qr ${TMPDIR}/ztoc ${SOCI_SNAPSHOTTER_PROJECT_ROOT}/ztoc/fbs/ztoc || (printf "\n\nThe Ztoc schema seems to be modified. Please run 'make flatc' to re-generate Go files\n\n"; exit 1)
+rm -rf ${TMPDIR}
```
sparr, line 27: I know this is copied from the Makefile, so this isn't a blocker on this PR, but if you could fix this to remove TMPDIR in both cases that would save a little future effort.

Makefile
djdongjin: will move it in next update

scripts/check-flatc.sh
sparr: If not, I can just open issues for later to address this and some other concerns with these targets/scripts that predate this effort.

sparr (dismissed):

scripts/check-flatc.sh
```diff
@@ -0,0 +1,27 @@
...
+
+CUR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+SOCI_SNAPSHOTTER_PROJECT_ROOT="${CUR_DIR}/.."
+FBS_FILE_PATH=${SOCI_SNAPSHOTTER_PROJECT_ROOT}/ztoc/fbs/ztoc.fbs
+
+# check if flatbuffers needs to be generated again
+TMPDIR=$(mktemp -d)
+flatc -o ${TMPDIR} -g ${FBS_FILE_PATH}
+diff -qr ${TMPDIR}/ztoc ${SOCI_SNAPSHOTTER_PROJECT_ROOT}/ztoc/fbs/ztoc || (printf "\n\nThe Ztoc schema seems to be modified. Please run 'make flatc' to re-generate Go files\n\n"; exit 1)
+rm -rf ${TMPDIR}
```
Kern--, line 27: This temp dir doesn't get deleted if the check fails. Looks like that was a problem when it was in the makefile too, but we should probably fix it.

djdongjin: Add the `rm` call before `exit 1` so it's deleted and script exits with 1.

.github/workflows/build.yml
djdongjin: changed back to use script for golangci-lint. turns out it doesn't support multi-mod repo and will just ignore those subdirectories with own `go.mod`: https://github.com/golangci/golangci-lint/issues/828

scripts/check-flatc.sh
rdpsin: This was intentional. The idea (from wmesard@) was that users might be interested in what changed if the check fails.

scripts/check-ltag.sh
```diff
@@ -0,0 +1,25 @@
...
+#   limitations under the License.
+
+set -eux -o pipefail
+
+CUR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+SOCI_SNAPSHOTTER_PROJECT_ROOT="${CUR_DIR}/.."
+
+# check ltag
+pushd ${SOCI_SNAPSHOTTER_PROJECT_ROOT}
+$(go env GOPATH)/bin/ltag -t ${SOCI_SNAPSHOTTER_PROJECT_ROOT}/.headers -check -v || (echo "The files listed above are missing a licence header. Please run make add-ltag"; exit 1)
```
rdpsin, line 24: np: Should update `make add-ltag` to `./scripts/add-ltag`.

scripts/check-flatc.sh
djdongjin: The script will print out the diff if there is any. In most case I think the user will need to run `make flatc` (which will also shows the diff in git). I think it should be sufficient, but I don't have a strong preference since it's just a tmp folder :)

sparr (approved):

rdpsin (approved):

PR #345, opened 2023-01-24
Allow manually remove invalid snapshots on restore

hanyuel:
*Issue #, if available:* closes #93
The problem that #93 tries to fix is that if the snapshotter crashes, it won't be able to restart because the snapshotter won't be able to restore the remote snapshots ([this](https://github.com/awslabs/soci-snapshotter/blob/main/snapshot/snapshot.go#L177) will error out), so it will crash again and again.

*Description of changes:*
cherry-pick https://github.com/containerd/stargz-snapshotter/pull/901

*Testing performed:*
I was not able to reproduce the problem (details above) by killing (SIGKILL-ing) the soci-snapshotter-grpc process when the container is running.  I might have repro'd once... so I was not able to verify this PR fixes the problem. But I think this PR is still worth merging considering 1) it's defensive programming 2) it's cherrypicked from stargz repo.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #383, opened 2023-02-01
Update snapshotter config. Reduce merge conflicts for #345

sparr:
*Description of changes:*
Update config layout from #343 to match incoming changes in #345, which is a cherry pick from stargz and will increase familiarity when users switch snapshotters.

*Testing performed:*
Ran updated automated test, and again with intentional error.

### DIFF
--- a/service/config.go
+++ b/service/config.go
@@ -49,8 +49,8 @@ type Config struct {
 	// ResolverConfig is config for resolving registries.
 	ResolverConfig `toml:"resolver"`
 
-	// MinLayerSize skips remote mounting of smaller layers
-	MinLayerSize int `toml:"min_layer_size"`
+	// SnapshotterConfig is snapshotter-related config.
+	SnapshotterConfig `toml:"snapshotter"`
 }
 
 // KubeconfigKeychainConfig is config for kubeconfig-based keychain.
@@ -74,3 +74,9 @@ type CRIKeychainConfig struct {
 
 // ResolverConfig is config for resolving registries.
 type ResolverConfig resolver.Config
+
+// SnapshotterConfig is snapshotter-related config.
+type SnapshotterConfig struct {
+	// MinLayerSize skips remote mounting of smaller layers
+	MinLayerSize int64 `toml:"min_layer_size"`
+}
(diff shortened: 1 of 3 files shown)


service/config.go
```diff
@@ -74,3 +74,9 @@ type CRIKeychainConfig struct {
 
 // ResolverConfig is config for resolving registries.
 type ResolverConfig resolver.Config
+
+// SnapshotterConfig is snapshotter-related config.
+type SnapshotterConfig struct {
+	// MinLayerSize skips remote mounting of smaller layers
+	MinLayerSize int `toml:"min_layer_size"`
```
djdongjin, line 81: we should change the runtime `MinLayerSize` to `int64` to keep it consistent with 

https://github.com/awslabs/soci-snapshotter/blob/56534c70965d326ed9963f9a0fa5f0ecba9d3bb9/soci/soci_index.go#L217

djdongjin (approved):

snapshot/snapshot.go
```diff
@@ -296,7 +294,7 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 		if o.minLayerSize > 0 {
 			if strVal, ok := base.Labels[source.TargetSizeLabel]; ok {
 				if intVal, err := strconv.Atoi(strVal); err == nil {
```
hanyuel, line 296: ``strVal`` should be converted to ``int64`` via ``strconv.ParseInt(strVal, 10, 64)``

hanyuel (changes requested):

snapshot/snapshot.go
```diff
@@ -103,9 +103,8 @@ type FileSystem interface {
 // SnapshotterConfig is used to configure the remote snapshotter instance
 type SnapshotterConfig struct {
 	asyncRemove bool
-
 	// minLayerSize skips remote mounting of smaller layers
-	minLayerSize int
+	minLayerSize int64
```
sbuckfelder, line 107: why the type change ?

sparr: Consistency with the create-time min layer size option. See review from @djdongjin that prompted the change.

djdongjin (approved):

Issue #355, opened 2023-01-26
Deduplicate temp tar(gz) generation in testutils


djdongjin:
**Is your feature request related to a problem? Please describe.**

We have 2 sets of testutils to generate temp tar gz files:

1. [`ztoc.testutil.go`](https://github.com/awslabs/soci-snapshotter/blob/a553bdb1b8f2cc215236943afb399bf808a38d6a/ztoc/testutils.go):
    1. `buildTempTarGz`: generate TarGz file given N files.
    2. `writeTempTarGz`: write a file to a `tar.Writer`.
    3. `iterDirectory`: iter and write a folder to a `tar.Writer`.
    4. `tempTarGz`: write a folder to a gzip file. (wrapper of `iterDirectory`).
 2. [`util/testutil/tar.go`](https://github.com/awslabs/soci-snapshotter/blob/0784bb3202c4120c6a8dd271350af646aee4b5cb/util/testutil/tar.go):
     1. `BuildTar`: generate Tar **reader** given N **files/folders**.
     2. `BuildTarGz`: generate TarGz **reader** given N files.

Their functionalities are quite similar (so as their own helpers, e.g., struct for representing a file, generate random file/dir to be tar'ed) and we should merge them.

As an example, below is how to represent files to be tar'ed from the two utils:

https://github.com/awslabs/soci-snapshotter/blob/581d3975f537db59786decadd47c924859138d36/ztoc/ztoc_test.go#L164-L167

https://github.com/awslabs/soci-snapshotter/blob/581d3975f537db59786decadd47c924859138d36/fs/backgroundfetcher/background_fetcher_test.go#L88-L94

**Describe the solution you'd like**

Comebine testutils related to `(gzip/...)tar` to `util/testutil/tar.go` that provides the following functionalities:

1. Generate a tar reader (with or w/o compression) from random generated files/folders.
2. Write a tar reader to a temp tar file.

We should separate generating tar reader and writing to tar file because sometimes we want a `io.reader` and sometimes we need a real tar filename.

**Additional context**

Issue #382, opened 2023-02-01
[BUG] SOCI can't read files in the 0th span if gzip flags are set


Kern--:
**Describe the bug**

If a layer sets gzip headers, files can't be read from the 0th span.

In `zinfo.c` we assume that the gzip header is exactly 10 bytes. https://github.com/awslabs/soci-snapshotter/blob/main/compression/zinfo.c#L590-L591

gzip has several flags (`FHCRC`, `FEXTRA`, `FNAME`, `FCOMMENT`) which add additional data following the 10 byte header before compressed data. Since we hardcode the 0th span to start at offset `10`, if these flags are set for a layer, we attempt to interpret this extra data as compressed data which fails. 

**Steps To Reproduce**
1) Create a ztoc for a layer that sets any gzip flag (except FTEXT which doesn't change the length of the header)
2) Attempt to read a file from the 0th span

A concrete example
```
$ ctr image pull --platform linux/arm64 docker.io/library/ubuntu@sha256:f05532b6a1dec5f7a77a8d684af87bc9cd1f2b32eab301c109f8ad151b5565d1
$ soci create --min-layer-size docker.io/library/ubuntu@sha256:f05532b6a1dec5f7a77a8d684af87bc9cd1f2b32eab301c109f8ad151b5565d1
$ soci ztoc get-file sha256:ba1dcd7388f2908850035d5c6bdc031789dc34ef0abc26dad868ade6c8870c1a usr/bin/cat
soci: error extracting data; return code: -3
```

**Expected behavior**
SOCI can read files in the 0th span when the layer has gzip flags set.

**Configuration (please complete the following information):**
 - OS: 
 - Snapshotter Version:
 - Containerd Version: 

**Additional context**
More info on gzip flags https://www.rfc-editor.org/rfc/rfc1952#page-6

PR #388, opened 2023-02-02
Deduplicate temp targz creation in ztoc/testutils.go.

turan18:
Combined duplicate temp targz generation funcs + helpers within ztoc/testutils.go. into util/tar.go . Util/tar.go has 2 functions (BuildTar + BuildTarGz), which both return tar(gz) io.Reader. Since some integration tests require targz file paths to be returned a WriteTarToTempFile function has been added to create a targz file from an io.Reader.

*Issue #, if available:*

Fixes: #355 

*Description of changes:*

*Testing performed:*

make test && make integration

### DIFF
--- a/util/testutil/tar.go
+++ b/util/testutil/tar.go
@@ -66,7 +66,7 @@ func WithPrefix(prefix string) BuildTarOption {
 	}
 }
 
-// BuildTar builds a tar blob
+// BuildTar builds a tar given a list of tar entries and returns an io.Reader
 func BuildTar(ents []TarEntry, opts ...BuildTarOption) io.Reader {
 	var bo BuildTarOptions
 	for _, o := range opts {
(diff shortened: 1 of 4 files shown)


sparr:
Most of the integration test failures appear to be from #382, and the `TestSociCreateSparseIndex` failure looks like a transient network mishap.

sparr (approved):

Kern-- (approved):

util/testutil/tar.go
```diff
@@ -122,8 +125,58 @@ func BuildTarGz(ents []TarEntry, compressionLevel int, opts ...BuildTarOption) i
 	return pr
 }
 
+// WriteTarGz writes the contents of a tar (gz) to a specified path(targz file)
+func WriteTarGz(targzName string, r io.Reader) (*string, error) {
```
djdongjin, line 129: The func is not limited to `TarGz` right? (e.g., both return of `BuildTar` and `BuildTarGz` can be passed to this func which writes to a file). If so suggest to change name to `WriteTarToTempFile`)

djdongjin: also can we change the return type to `string` instead of `*string`?

turan18: Sure

djdongjin (approved):

PR #392, opened 2023-02-03
Futureproof runtime minimum layer size test

### LINKED ISSUE #373: [BUG] Min layer size config option integration test susceptible to future failure
**Describe the bug**
The integration test for #343 will fail if the image used is updated, since it is not currently pinned.

**Steps To Reproduce**
Pin the rabbitmq image to an older version with different layer structure

**Expected behavior**
Test still functions as expected.

**Additional context**
Per @Kern-- in https://github.com/awslabs/soci-snapshotter/pull/343#discussion_r1091221204 a good approach here would be to check the sizes of the layers and choose a min layer size at the median layer size. This would ensure there are some skipped and some non-skipped layers, and allow us to calculate the expected number of fuse mounts. It would also have the nice side effect of testing the case where the min layer size is equal to a specific layer's size (which would fail if the code had a < vs <= mixup).

sparr:
*Issue #, if available:*
Closes #373

*Description of changes:*
Update the integration test for the runtime minimum layer size configuration so that it calculates the median layer size and uses that as the minimum. This makes the test a little more robust and a lot more futureproof.

*Testing performed:*
Ran the test, then again with manual changes to the median index and the layer sizes, confirmed expected results.

### DIFF
--- a/integration/pull_test.go
+++ b/integration/pull_test.go
@@ -38,6 +38,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -146,16 +147,17 @@ func TestOptimizeConsistentSociArtifact(t *testing.T) {
 }
 
 func TestLazyPullWithSparseIndex(t *testing.T) {
+	imageName := rabbitmqImage
+
 	regConfig := newRegistryConfig()
 
 	sh, done := newShellWithRegistry(t, regConfig)
 	defer done()
 
-	const imageName = "rethinkdb@sha256:4452aadba3e99771ff3559735dab16279c5a352359d79f38737c6fdca941c6e5"
-	const imageManifestDigest = "sha256:4452aadba3e99771ff3559735dab16279c5a352359d79f38737c6fdca941c6e5"
-	const minLayerSize = 10000000
-
 	rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, false))
+
+	_, minLayerSize, _ := middleSizeLayerInfo(t, sh, dockerhub(imageName))
+
 	copyImage(sh, dockerhub(imageName), regConfig.mirror(imageName))
 	indexDigest := buildIndex(sh, regConfig.mirror(imageName), withMinLayerSize(minLayerSize))
 
@@ -171,6 +173,10 @@ func TestLazyPullWithSparseIndex(t *testing.T) {
 		sh.Pipe(nil, shell.C("soci", "run", "--rm", "--snapshotter=soci", image, "test", "tar", "-zc", "/usr"), tarExportArgs)
 	}
 
+	imageManifestDigest, err := getManifestDigest(sh, dockerhub(imageName).ref, dockerhub(imageName).platform)
+	if err != nil {
+		t.Fatalf("Failed to get manifest digest: %v", err)
+	}
 	imageManifestJSON := fetchContentByDigest(sh, imageManifestDigest)
 	imageManifest := new(ocispec.Manifest)
 	if err := json.Unmarshal(imageManifestJSON, imageManifest); err != nil {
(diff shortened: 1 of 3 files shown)


integration/pull_test.go
```diff
@@ -796,8 +794,43 @@ min_layer_size=` + minLayerSize + `
...
+	for medianIndex > 0 && snapshotSizes[medianIndex-1] == medianSize {
+		medianIndex--
+	}
+
+	minLayerSizeConfig := `
+[snapshotter]
+min_layer_size=` + strconv.FormatInt(medianSize, 10) + `
+`
+
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, false, minLayerSizeConfig))
```
sparr, line 828: It may be possible to partially reset the environment here instead, to avoid needing to copy the image and build the index again, just making the environment clean for another rpull. I made a moderate effort in that direction but decided this more concise and certain-to-work solution would be acceptable. Feedback on this decision is welcome.

integration/pull_test.go
```diff
@@ -796,8 +794,43 @@ min_layer_size=` + minLayerSize + `
...
+	snapshotListOutput := string(sh.O("ctr", "snapshot", "list"))
+	snapshotList := strings.Split(snapshotListOutput, "\n")
+	for _, snapshotLine := range snapshotList[1 : len(snapshotList)-1] {
+		snapshotDigest := strings.Split(snapshotLine, " ")[0]
+		rawJSON := sh.O("ctr", "snapshot", "--snapshotter", "soci", "info", snapshotDigest)
+		var snapshotInfo snapshots.Info
+		if err := json.Unmarshal(rawJSON, &snapshotInfo); err != nil {
+			t.Fatalf("invalid snapshot info from digest %s: %v", snapshotDigest, rawJSON)
+		}
+		strVal := snapshotInfo.Labels[source.TargetSizeLabel]
```
djdongjin, line 807: We probably should check the `TargetSociIndexDigestLabel` label and skip those not belonging to `indexDigest` to make the test more robust, just in case there is any leftover from other tests or parallel tests.

sparr: Integration tests run in isolated containers, so that shouldn't be a concern. I could add the check, but don't want to unnecessarily complicate the test.

djdongjin: I think it still worth adding the check (given the suggestion below and if it's possible to pull 2 or more images within a single test).

I'd suggest extract the logic (given an image/index, calculate median layer size) to a separate helper func so it can be reused like `indexDigest := buildIndex(sh, imgInfo, withMinLayerSize(medianLayerSize(img)))` to test sparse index. Then we can avoid code that uses fixed imageManifestDigest to ensure sparse index.

https://github.com/awslabs/soci-snapshotter/blob/f3e07639fef7aa7542fad3480dbcbba5eb23e804/integration/pull_test.go#L154-L160

Kern--: The snapshot labels are populated from the manifest. In SOCI we don't have access to the manifest (which is why we have labels), but in integ tests we do.

It might be easier to get the sizes from the manifest like this pull test example:
https://github.com/awslabs/soci-snapshotter/blob/main/integration/pull_test.go#L174-L182

We'd have to pin a manifest version, but I think that's fine.

integration/testutil.go
```diff
@@ -718,3 +719,42 @@ func teardown(cleanups []func() error) error {
 	}
 	return nil
 }
+
+// middleSizeLayerInfo finds a layer not the smallest or largest (if possible), returns index, size, and layer count
+func middleSizeLayerInfo(t *testing.T, sh *shell.Shell, image imageInfo) (int, int64, int) {
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, false))
```
djdongjin, line 725: why is `rebootContainerd` needed here?

sparr: `rebootContainerd` is also secretly "start containerd and/or soci-snapshotter-grpc when they aren't already running". Refactoring to split out start/restart/stop for the two services would be useful in a lot of places.

integration/testutil.go
```diff
@@ -718,3 +719,42 @@ func teardown(cleanups []func() error) error {
...
+
+	sh.O("ctr", "i", "pull", "--platform", platforms.Format(platforms.DefaultSpec()), image.ref)
```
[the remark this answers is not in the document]
sparr: I originally set this based on the use case it was taken from. It needs to match. You're probably right that we could require image.platform to be what is used and expected.

integration/pull_test.go
sparr: Opened #401 based on this and other resolved comment below.

integration/testutil.go
sparr: Opened #401 based on this.

djdongjin: I think `middleSizeLayerInfo`, being a test helper func, shouldn't reboot containerd (and override configs with default) and can assume there is containerd/soci running (similar to `buildIndex` assuming it).

This also might cause unexpected behavior change (e.g. override `additionalConfigs` set up by a test itself, when a test just want to get a value from the helper). Whether or not reboot should be set up by the test itself.

As an exmaple, `TestLazyPullWithSparseIndex` doesn't need to reboot after calling the helper, since it only requires the value. Whereas `TestRpullImageWithMinLayerSize` need to call `rebootContainerd` after `middleSizeLayerInfo`.

Kern-- (approved):

integration/pull_test.go
```diff
@@ -146,14 +147,14 @@ func TestOptimizeConsistentSociArtifact(t *testing.T) {
...
+
 	regConfig := newRegistryConfig()
 
 	sh, done := newShellWithRegistry(t, regConfig)
 	defer done()
 
-	const imageName = "rethinkdb@sha256:4452aadba3e99771ff3559735dab16279c5a352359d79f38737c6fdca941c6e5"
-	const imageManifestDigest = "sha256:4452aadba3e99771ff3559735dab16279c5a352359d79f38737c6fdca941c6e5"
-	const minLayerSize = 10000000
+	_, minLayerSize, _ := middleSizeLayerInfo(t, sh, dockerhub(imageName))
```
djdongjin, line 157: nit: might be safer to move this after line 159 `rebootContainerd` to ensure the test itself provides running containerd/soci

djdongjin (commented):
LGTM, need to resolve conflict

sbuckfelder (approved):
LGTM

djdongjin (approved):

Kern-- (approved):

PR #397, opened 2023-02-03
Fix 0th checkpoint when gzip headers are present

### LINKED ISSUE #382: [BUG] SOCI can't read files in the 0th span if gzip flags are set

Kern--:

*Issue #, if available:*
Fixes #382

*Description of changes:*

The 0th checkpoint points to the start of compressed data immediately following the gzip header. The 0th checkpoint's only data dependency is the length of the gzip header. We erronously believed this to be fixed to 10 bytes and therefore skipped serializing the 0th checkpoint to save about 32 KiB. The gzip file format has a set of flags which encode additional, non-fixed-length metadata in the header. Since we hardcoded the 0th checkpoint's compressed offset to 10, if a gzip file used any of these flags, we would attempt to start decompressing in the middle of the header meant we could not deserialize the 0th span. The issue was only in serializing - during ztoc generation we created the correct 0th checkpoint which means that subsequent checkpoints (and spans) are unaffected.

While we didn't serialize the 0th checkpoint, we did serialize the real number of checkpoints which gives us a mechanism to fix this issue in a backwards compatible manner. We will start serializing the 0th checkpoint for all new ztocs. When deserializing, we can take the length of the blob minus the size of the fixed portion of the blob. If that number divided by the checkpoint size == number of checkpoints - 1, then we have an old blob and we should inject the fixed 0th checkpoint. If it is exactly the number of checkpoints, then we have a new blob and we should not inject the fixed checkpoint.

To ensure that we re-serialize a blob correctly, we add a version to the ztoc. If the ztoc version is 1, skip the 0th checkpoints. If ztoc version > 1 serialize all checkpoints.

*Testing performed:*
`scripts/check-all.sh`


compression/zinfo.c
```diff
@@ -495,19 +494,19 @@ unsigned get_blob_size(struct gzip_zinfo* index)
         return 0;
     }
 
-    unsigned size = decode_int32(index->size); 
+    unsigned size = decode_int32(index->size);
+    if (decode_int32(index->version) == 1) {
+        size--;
+    }
```
sparr, line 500: Perhaps a short comment here, to set a precedent that we explain the version differences where they exist in code?

sbuckfelder (dismissed):
Tests look good in particular

compression/zinfo.h
```diff
@@ -53,9 +53,29 @@
...
+*/
+#define PACKED_CHECKPOINT_SIZE (8 + 8 + 1 + WINSIZE)
+
+/*
+    -  4 bytes, number of checkpoints
+    -  8 bytes, span size
+*/
+#define BLOB_FIXED_SIZE (4 + 8)
+
+
```
djdongjin, line 78: nit: `BLOB_FIXED_SIZE` -> `FIXED_BLOB_SIZE` (more consistent with `PACKED_CHECKPOINT_SIZE`)

djdongjin (approved):

rdpsin:
I wonder if we should update the `ztoc get-file` integ test to have some kind of fixed pattern i.e., if we're choosing 5 files at random, make sure 1 of them is in the first span and 1 of them is in the last span.

turan18:
> I wonder if we should update the `ztoc get-file` integ test to have some kind of fixed pattern i.e., if we're choosing 5 files at random, make sure 1 of them is in the first span and 1 of them is in the last span.

I think that makes sense. Or we could chose a random file(s) from each span

compression/zinfo.h
Kern--: I was trying to use `fixed size` as a noun phrase here to refer to the "fixed size" portion of the blob. I think `FIXED_BLOB_SIZE` makes it sound like `blob size` is a noun phrase, but that means something different. 

To be fair, it would probably be easier to read if both of these 3-word constants placed the noun phrase in the same place, so what about `PACKED_CHECKPOINT_SIZE` and `BLOB_HEADER_SIZE`?

djdongjin: SGTM, that's clear. After all `PACKED_CHECKPOINT_SIZE` is also a fixed size by itself based on the calculation in it's comment. :)

Kern--:
I updated the `ztoc get-file` from reading 10 random files in the archive to reading 1 random file from each span. Happy to increase that number if we think it's valuable.

djdongjin (approved):

rdpsin (approved):

PR #399, opened 2023-02-07
Optimize snapshot prepare

### LINKED ISSUE #398: `snapshot.Prepare` logic is confusing and need refactor/optimize
The code snippet is attached below and the general logic is:

```golang
if target {
	if !localOverride {
		// 1. if-else
		if err := prepareRemoteSnapshot; err != nil {
			// logging, continue to after if block (local snapshot)
		} else {
			// follow up of prepareRemoteSnapshot
			// return
		}
	}

	mounts, err := o.mounts()
	// 2. if-else
	if err = o.prepareLocalSnapshot(); err != nil {
		// logging, continue to after if block (the final return)
	} else {
		// follow up of prepareLocalSnapshot
		// return
	}
}
return o.mounts()
```

This implementation is quite difficult to understand/follow and has duplicated `o.mounts` calls.
1. in the 2 if-else blocks: 
    1. If the logic enters `if`, it skips `else` and continue to after the if-else block.
    2. `else` is the follow up actions of the func called in `if` statement (e.g. `prepareRemoteSnapshot`) and return directly.
    3. in some cases, the flow is: go to `if-1`, skip `else-1`, go to `if-2`, skip `else-2`, return at the end (outside of leading `if`).
2. in case 1.3 (and other cases), `o.mounts` are called twice.

**Expected behavior**

Make the func logic more clear and avoid logic jump (e.g., `if { ... } else { ...; return }`. in this case, we should switch the `if` and `else`, instead of letting `else` split the logic of `if` execution)

https://github.com/awslabs/soci-snapshotter/blob/04a52810a195b8c4ba76736f1092d354fbf687ce/snapshot/snapshot.go#L285-L356

djdongjin:

*Issue #, if available:* fix #398

*Description of changes:*

refactor `snapshot.Prepare`. see #398 for more details.

1. streamline `if-else`
2. avoid duplicated `o.mounts` call

*Testing performed:*

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -282,78 +282,81 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 		}
 	}
 
-	if target, ok := base.Labels[targetSnapshotLabel]; ok {
+	target, ok := base.Labels[targetSnapshotLabel]
+	if !ok {
+		return o.mounts(ctx, s, parent)
+	}
 
-		// NOTE: If passed labels include a target of the remote snapshot, `Prepare`
-		//       must log whether this method succeeded to prepare that remote snapshot
-		//       or not, using the key `remoteSnapshotLogKey` defined in the above. This
-		//       log is used by tests in this project.
-		lCtx := log.WithLogger(ctx, log.G(ctx).WithField("key", key).WithField("parent", parent))
+	// NOTE: If passed labels include a target of the remote snapshot, `Prepare`
+	//       must log whether this method succeeded to prepare that remote snapshot
+	//       or not, using the key `remoteSnapshotLogKey` defined in the above. This
+	//       log is used by tests in this project.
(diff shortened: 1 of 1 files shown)


snapshot/snapshot.go
```diff
@@ -282,78 +282,81 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
...
-				}
-				log.G(lCtx).WithField(remoteSnapshotLogKey, prepareFailed).
-					WithError(err).Warn("failed to internally commit remote snapshot")
-				// Don't fallback here (= prohibit to use this key again) because the FileSystem
-				// possible has done some work on this "upper" directory.
-				return nil, err
-			}
-		}
-
-		mounts, err := o.mounts(ctx, s, parent)
```
djdongjin, line 330: In cases where the current impl calls `o.mounts` twice, I assume it's not on purpose (e.g., between here and final `return o.mounts`, we change `s` or `parent` or something `s`/`parent` rely on so `o.mounts` has to be called again at the end.

rdpsin (approved):

turan18 (approved):

Issue #237, opened 2022-12-15
Create metrics to quantify the performance of background fetch operations


vkuzniet:
**Is your feature request related to a problem? Please describe.**
The background fetch was dramatically redesigned. Old set of metrics became obsolete. 
We need to think if we need any metrics related to background fetch and if the answer is yes, they need to be implemented.

**Describe the solution you'd like**
1. One-pager on whether background fetch related metrics are needed for successful operations of soci snapshotter (including the precise definition of metrics is any).
2. Implement the aforementioned metrics.

rdpsin:
Initially, we will add these 4 metrics:

- Number of spans fetched by bg-fetcher
- Number of errors during span fetch by bg-fetcher
- Time to complete background fetch for a single layer (Maybe some kind of histogram?)
- Periodically emit metrics for number of items in the work queue: `len(workQueue)`

PR #402, opened 2023-02-07
Implement background fetch metrics

hanyuel:
*Issue #, if available:* #237

*Description of changes:*
Added these 4 background fetch metrics:
``background_span_fetch_count`` Number of spans fetched by bg-fetcher
``background_span_fetch_failure_count`` Number of errors during span fetch by bg-fetcher
``operation_duration_background_fetch`` Time in milliseconds to complete background fetch for a layer
``background_fetch_work_queue_size`` Number of items in the work queue of the bg fetcher. Emitted every ``emitMetricPeriod`` seconds (defaulting to 10 sec)

*Testing performed:*
`` curl localhost:6060/metrics | grep background``
Example output:
```
soci_fs_image_operation_count_key{image="",operation_type="background_fetch_work_queue_size"} 18
soci_fs_operation_count{layer="sha256:08b2c94789a92745987c560c2b49f78ebb8b771682c98d378c46b8735fdae9f2",operation_type="background_span_fetch_count"} 3
soci_fs_operation_count{layer="sha256:21933556955c0a0da9c739edbb8e3abc0ec60627975d74448c9bdc035b98a462",operation_type="background_span_fetch_count"} 8
soci_fs_operation_duration_milliseconds_sum{layer="sha256:08b2c94789a92745987c560c2b49f78ebb8b771682c98d378c46b8735fdae9f2",operation_type="background_fetch"} 9000.550487
```

### DIFF
--- a/fs/backgroundfetcher/background_fetcher.go
+++ b/fs/backgroundfetcher/background_fetcher.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	commonmetrics "github.com/awslabs/soci-snapshotter/fs/metrics/common"
 	"github.com/containerd/containerd/log"
 	"golang.org/x/time/rate"
 )
@@ -48,6 +49,13 @@ func WithMaxQueueSize(size int) Option {
 	}
 }
 
+func WithEmitMetricPeriod(period time.Duration) Option {
+	return func(bf *BackgroundFetcher) error {
+		bf.emitMetricPeriod = period
+		return nil
+	}
+}
+
 // An interface for a type to "pause" the background fetcher.
 // Useful for mocking in unit tests.
 type pauser interface {
@@ -63,9 +71,10 @@ func (p defaultPauser) pause(d time.Duration) {
 // A backgroundFetcher is responsible for fetching spans from layers
 // in the background.
 type BackgroundFetcher struct {
-	silencePeriod time.Duration
-	fetchPeriod   time.Duration
-	maxQueueSize  int
+	silencePeriod    time.Duration
+	fetchPeriod      time.Duration
+	maxQueueSize     int
+	emitMetricPeriod time.Duration
 
 	rateLimiter *rate.Limiter
 
@@ -135,14 +144,19 @@ loop:
 }
 
 func (bf *BackgroundFetcher) Run(ctx context.Context) error {
+	ticker := time.NewTicker(bf.emitMetricPeriod)
+	go bf.emitWorkQueueMetric(ctx, ticker)
+
 	for {
 		// Pause the background fetcher if necessary.
 		bf.pause(ctx)
 
 		select {
 		case <-bf.closeChan:
+			ticker.Stop()
 			return nil
 		case <-ctx.Done():
+			ticker.Stop()
 			return nil
 		default:
 		}
@@ -165,6 +179,20 @@ func (bf *BackgroundFetcher) Run(ctx context.Context) error {
 
 		if err := bf.rateLimiter.Wait(ctx); err != nil {
 			return fmt.Errorf("background fetch: error while waiting for rate limiter: %w", err)
+		}
+	}
+}
+
+func (bf *BackgroundFetcher) emitWorkQueueMetric(ctx context.Context, ticker *time.Ticker) {
+	for {
+		select {
+		case <-bf.closeChan:
+			return
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			// background fetcher is at the snapshotter's fs level, so no image digest as key
+			commonmetrics.AddImageOperationCount(commonmetrics.BackgroundFetchWorkQueueSize, "", int32(len(bf.workQueue)))
 		}
 	}
 }
--- a/fs/backgroundfetcher/resolver.go
+++ b/fs/backgroundfetcher/resolver.go
@@ -21,8 +21,10 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/awslabs/soci-snapshotter/compression"
+	commonmetrics "github.com/awslabs/soci-snapshotter/fs/metrics/common"
 	sm "github.com/awslabs/soci-snapshotter/fs/span-manager"
 	"github.com/containerd/containerd/log"
 	"github.com/opencontainers/go-digest"
@@ -46,6 +48,8 @@ type base struct {
 	layerDigest digest.Digest
 	closed      bool
 	closedMu    sync.Mutex
+	// timestamp when background fetch for the layer starts
+	start time.Time
 }
 
 func (b *base) Close() error {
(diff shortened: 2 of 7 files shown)


fs/config/config.go
```diff
@@ -126,4 +126,8 @@ type BackgroundFetchConfig struct {
 	// i.e., the maximum number of span managers that can be queued
 	// in the background fetcher.
 	MaxQueueSize int `toml:"max_queue_size"`
+
+	// EmitMetricPeriod is the amount of interval (in second) at which the background
+	// fetcher emits metrics
+	EmitMetricPeriod int64 `toml:"emit_metric_period_sec"`
```
djdongjin, line 132: `EmitMetricPeriod` -> `EmitMetricPeriodSec` to be consistent with `SilencePeriodMSec` and `FetchPeriodMSec` above

fs/backgroundfetcher/background_fetcher.go
```diff
@@ -168,3 +179,16 @@ func (bf *BackgroundFetcher) Run(ctx context.Context) error {
 		}
 	}
 }
+
+func (bf *BackgroundFetcher) emitMetric() {
+	for {
+		select {
```
djdongjin, line 185: similar to `bf.Run()` we should pass the `ctx` to `emitMetric` and have a `case` for `ctx.Done()`:

```
		case <-ctx.Done():
			return
```

fs/config/config.go
hanyuel: Done.

fs/backgroundfetcher/background_fetcher.go
hanyuel: Done

fs/backgroundfetcher/background_fetcher.go
```diff
@@ -168,3 +179,18 @@ func (bf *BackgroundFetcher) Run(ctx context.Context) error {
 		}
 	}
 }
+
+func (bf *BackgroundFetcher) emitMetric(ctx context.Context) {
```
rdpsin, line 183: np: s/emitMetric/emitWorkQueueMetric/g

fs/backgroundfetcher/resolver.go
```diff
@@ -82,15 +86,21 @@ func (lr *sequentialLayerResolver) Resolve(ctx context.Context) (bool, error) {
 		"spanId": lr.nextSpanFetchID,
 	}).Debug("fetching span")
 
+	if lr.nextSpanFetchID == 0 {
+		lr.start = time.Now()
+	}
 	err := lr.FetchSingleSpan(lr.nextSpanFetchID)
 	if err == nil {
+		commonmetrics.IncOperationCount(commonmetrics.BackgroundSpanFetchCount, lr.layerDigest)
```
rdpsin, line 94: I think this metric should really be part of the bg-fetcher, not resolver.

fs/backgroundfetcher/resolver.go
```diff
@@ -82,15 +86,21 @@ func (lr *sequentialLayerResolver) Resolve(ctx context.Context) (bool, error) {
...
 		lr.nextSpanFetchID++
 		return true, nil
 	}
 	if errors.Is(err, sm.ErrExceedMaxSpan) {
+		commonmetrics.MeasureLatencyInMilliseconds(commonmetrics.BackgroundFetch, lr.layerDigest, lr.start)
 		return false, nil
 	}
 
+	commonmetrics.IncOperationCount(commonmetrics.BackgroundSpanFetchFailureCount, lr.layerDigest)
```
rdpsin, line 103: I think this metric should really be part of the bg-fetcher, not resolver.

fs/backgroundfetcher/resolver.go
```diff
@@ -65,6 +67,8 @@ func (b *base) Closed() bool {
 type sequentialLayerResolver struct {
 	*base
 	nextSpanFetchID compression.SpanID
+	// timestamp when background fetch for the layer starts
+	start time.Time
```
rdpsin, line 71: This should be part of `base` since we want it for each `Resolver`, not just `sequentialLayerResolver`. 

How about adding something like these methods to the interface

```
type Resolver interface {

      Start() // Starts the timer
      
      Duration() time.Time // Get the time
     
}
```

And do `resolver.Start()` in `backgroundfetcher.Add()`.

fs/backgroundfetcher/resolver.go
hanyuel: Why? ``Resolver`` is part of the bg-fetcher

fs/backgroundfetcher/resolver.go
hanyuel: Why do we need ``Resolver`` interface? ``sequentialLayerResolver`` is the only implementation of ``Resolver`` .

rdpsin: For now. In the future, we may have different implementation for Resolver. (For example, a resolver with an ordered list of spans to fetch).

hanyuel: Decided to move ``start`` to ``base`` but keep where ``start`` timestamp is recorded unchanged.  The timestamp when span 0 starts being fetched is a more accurate measurement of when resolving for the layer starts than when the resolver is added to work queue.

fs/backgroundfetcher/background_fetcher.go
```diff
@@ -168,3 +179,18 @@ func (bf *BackgroundFetcher) Run(ctx context.Context) error {
...
+	for {
+		select {
+		case <-bf.closeChan:
+			return
+		case <-ctx.Done():
+			return
+		default:
+			// background fetcher is at the snapshotter's fs level, so no image digest as key
+			commonmetrics.AddImageOperationCount(commonmetrics.BackgroundFetchWorkQueueSize, "", int32(len(bf.workQueue)))
+			time.Sleep(bf.emitMetricPeriod)
```
Kern--, line 193: nit: go has `time.Ticker` for this sort of select block so the sleep doesn't prevent you from receiving on other channels:

```
ticker := time.NewTicker(b.emitMetricPeriod)
select {
    // other cases
    case <- ticker.C:
    // emit metric
}
```

Kern-- (approved):

fs/fs.go
```diff
@@ -185,6 +188,11 @@ func NewFilesystem(root string, cfg config.Config, opts ...Option) (_ snapshot.F
 		bgMaxQueueSize = defaultBgMaxQueueSize
 	}
 
+	bgEmitMetricPeriod := time.Duration(cfg.BackgroundFetchConfig.SilencePeriodMSec) * time.Second
```
kzys, line 191: SilencePeriodMSec + time.Millisecond?

hanyuel: Good catch. will fix

djdongjin (approved):

Kern-- (approved):

Issue #361, opened 2023-01-26
[BUG] Eliminate "error" messages on successful rpull operations


sparr:
**Describe the bug**
A successful normal run of `soci image rpull` can produce log output that appear to be labeled as "error"s.

Examples:
* `{"error":"exec: \"fusermount\": executable file not found in $PATH","key":"default/3/extract-860035799-tI5g sha256:[...]","level":"info","mountpoint":"/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/46/fs","msg":"fusermount not installed; trying direct mount","parent":"","time":"2023-01-25T20:45:24.972341422Z"}`
* `{"error":"skipping mounting layer sha256:[...] as FUSE mount: no ztoc for layer","key":"default/9/extract-93160834-sLdo sha256:[...]","level":"warning","msg":"failed to prepare remote snapshot","parent":"sha256:[...]","remote-snapshot-prepared":"false","time":"2023-01-25T20:45:25.097454762Z"}`

**Steps To Reproduce**
`soci image rpull [...]`

(comprehensive repro steps can be provided if necessary but would be unwieldy)

**Expected behavior**
"info" and "warning" level log output does not contain a field named "error"

Kern--:
One option is to change the logrus error key to something like "context"

Kern--:
We have decided that since this is idiomatic and there are better ways to solve this (e.g. using tooling to look at the log level), we will not implement this.

Issue #362, opened 2023-01-26
[FEATURE] Sort keys in log output for better human readability


sparr:
**Is your feature request related to a problem? Please describe.**
The human readability of the log output from soci-snapshotter-grpc is significantly impaired by the keys being sorted alphabetically, especially with different keys in different entries.

**Describe the solution you'd like**
Sort the output such that the lowest cardinality always-present keys appear first (preferably starting with `"level"`, then the highest cardinality always-present keys, then optional keys.

**Describe alternatives you've considered**
Piping log output through jq with key order specified when I need to read long logs.

**Additional context**
Go's json encoder can't re-order map keys. This could be implemented with a small amount of encoding glue code like https://stackoverflow.com/a/25183429/13675 for ordering and keys, along with https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/encoding/json/stream.go;l=201 for values.
Alternately, a library such as https://gitlab.com/c0b/go-ordered-json or https://github.com/iancoleman/orderedmap could be used.

sbuckfelder:
Could you add an example current output of log and ideal output?

sparr:
Here's an approximate example. I didn't entirely re-arrange the keys, and the specific order of less common keys could be implemented in whatever way is convenient, including keeping the current alphabetical sort. The primary point here is getting the always-present keys and the most important usually-present keys in an order that is easy for a human to read. From my perspective, that would certainly start with `time`, `level`, `msg`, then maybe `error` which is less common but more important, or `key` `parent` `digest` which are more common.

Current:

    {"fetchPeriod":500000000,"level":"info","maxQueueSize":100,"msg":"constructing background fetcher","silencePeriod":30000000000,"time":"2023-01-30T18:53:35.123049386Z"}
    {"key":"default/1/extract-896701523-UN9I sha256:0002c93bdb3704dd9e36ce5153ef637f84de253015f3ee330468dccdeacad60b","level":"info","msg":"preparing filesystem mount at mountpoint=/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/1/fs","parent":"","time":"2023-01-30T18:53:39.907408742Z"}
[...]

PR #410, opened 2023-02-08
Using "context" instead of "error" key when using withError 

turan18:
Using `context` instead of `error` key when logging operations that are not representative of actual errors. This should eliminate extraneous error keys within logs generated from `rpull`

Eg Log:

```
{"context":"exec: \"fusermount\": executable file not found in $PATH","key":"default/1/extract-198840019-FSRo sha256:d543b8cad89e3428ac8852a13cb2dbfaf55b1e10fd95a9753e51faf393d60e81","level":"info","mountpoint":"/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/1/fs","msg":"fusermount not installed; trying direct mount","parent":"","time":"2023-02-08T18:22:39.001486489Z"}
{"context":"skipping mounting layer sha256:5063b4e8e15250a2ed6aaa980e813b2d9145f67a57f1ca884078498fdf36d44b as FUSE mount: no ztoc for layer","key":"default/2/extract-155573583-fJQ7 sha256:91deb04bc1d7ee311492676790144a0d4ca41ce306e3bca19bb04e416ef3cf62","level":"warning","msg":"failed to prepare remote snapshot","parent":"sha256:d543b8cad89e3428ac8852a13cb2dbfaf55b1e10fd95a9753e51faf393d60e81","remote-snapshot-prepared":"false","time":"2023-02-08T18:22:39.158989711Z"}

```

*Issue #, if available:*

Fixes: #361 

*Testing performed:*

### DIFF
--- a/cmd/soci-snapshotter-grpc/main.go
+++ b/cmd/soci-snapshotter-grpc/main.go
@@ -127,6 +127,7 @@ func main() {
 	logrus.SetFormatter(&logrus.JSONFormatter{
 		TimestampFormat: log.RFC3339NanoFixed,
 	})
+	logrus.ErrorKey = "context"
 
 	var (
 		ctx    = log.WithLogger(context.Background(), log.L)


sparr (approved):
Integration test failure is related to upstream repo outage.

snapshot/snapshot.go
```diff
@@ -306,8 +306,8 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 
 		if !localOverride {
 			if err := o.prepareRemoteSnapshot(lCtx, key, base.Labels); err != nil {
-				log.G(lCtx).WithField(remoteSnapshotLogKey, prepareFailed).
-					WithError(err).Warn("failed to prepare remote snapshot")
+				log.G(lCtx).WithField(remoteSnapshotLogKey, prepareFailed).WithField("context", err).
+					Warn("failed to prepare remote snapshot")
```
djdongjin, line 310: Are the `err`s (changed from `WithError(err)` -> `WithField("context", err)`) are all noises (e.g., all `nil`, or `<nil>` or not important/repeated message?)

turan18: The usual `err` returned from `Mount` is `FUSE mount: no ztoc for layer` which is repeated for all layers that do not have ztoc.

turan18: I think that makes sense, I'll add it

Kern--:
There is an easier and system wide way to do this which is to change logrus' ErrorKey:
https://pkg.go.dev/github.com/sirupsen/logrus#ErrorKey

Doing so would mean all errors get logged under the "context" key, but that might be easier than trying to remember whether to use `WithError(err)` or `WithField("context", err)`. I think this is OK because we have a log level to indicate whether we're logging an application-level error.

turan18:
> Doing so would mean all errors get logged under the "context" key, but that might be easier than trying to remember whether to use `WithError(err)` or `WithField("context", err)`. I think this is OK because we have a log level to indicate whether we're logging an application-level error.

While I agree, there are some instances where we might like to explicitly call out errors that were not "expected" (really non-errors) like the one @djdongjin pointed at above. We could potentially have a very hacky workaround wrapper like:

```
func (entry *logrus.Entry) WithContext(err error) *logrus.Entry {
	return entry.WithField("context", value)
}
```
although this would not really be the place for it.

Kern--:
We already decide if a go error is an application error by whether we call `entry.Error` vs `entry.Info`. Does it add additional value to also decide between `WithError` vs `WithField`?

kzys:
I agree with Kern. Generally speaking, it is fine to use WithError even if the level is lower than Error.

containerd/containerd

https://sourcegraph.com/search?q=context:global+repo:containerd/containerd+/WithError.*%28Warn%7CInfo%29f%3F%5C%28/&patternType=standard&sm=1&groupBy=repo

Go in general

https://sourcegraph.com/search?q=context:global+/WithError.*%28Warn%7CInfo%29f%3F%5C%28/+lang:go&patternType=standard&sm=1&groupBy=repo

turan18:
> What we'd like to achieve by using `context` instead of `error`?

I think the issue was seeing extraneous "error" keys in the log that were not truly representative of real errors making logs harder to parse through. I think @Kern-- has a good point, changing the error key to something like context should suffice as the application level log levels already provide enough to determine actual "errors".

kzys:
> making logs harder to parse through

It may be hard if you use grep, but you should check the level key to see whether errors are there or not.

Setting ErrorKey is better than the original, but we still ask operators to see `context` key in SOCI snapshotter, `error` key in containerd.

sparr:
```
{"error":"foo", "level":"error"},
{"error":"foo", "level":"warning"},
{"error":"foo", "level":"info"},
{"error":"foo", "level":"debug"},
```

It is difficult for me to pin down precisely why this pattern concerns me. The best I can do is this much more subjective statement: Seeing this in my logs while trying to track down actual errors, especially seeing hundreds of such lines when there's just a single real failure, slows down my troubleshooting efforts.

Implementing #362 would make this less problematic, possibly enough so, but not entirely.

kzys:
What you want may be https://pkg.go.dev/github.com/sirupsen/logrus#TextFormatter (https://github.com/sirupsen/logrus has its screenshot). You don't have to make JSON eyegrep-friendly.

sparr:
@kzys Can you clarify, are you suggesting we change the codebase to output formatted text logs instead of json by default, or that an individual developer patch their local environment to produce logs more appropriate to their specific situation?

kzys:
Sorry. I wasn't clear.

If your goal is making JSON-formatted logs easier for eyes by renaming and sorting the keys, I recommend to add an option to not use JSON. It wouldn't be hard (e.g. you could just use logrus.TextFormatter), better (e.g. have colors) and keep our JSON schema close to what containerd uses, which most of SOCI snapshotter customers know already.

PR #419, opened 2023-02-10
Emit fuse operation counts as a log

Kern--:

*Issue #, if available:*
N/A

*Description of changes:*
Before this change, we emitted fuse operation counts as prometheus metrics with a label indicating which image is referred to by the metric. This works well when the cardinaltiy of images is low, however it can be quite expensive when the cardinality is high.

This change emits the data as both a metric and a log. This allows low cardinality systems to use rollup to record both aggregate (e.g. P90) metrics and image specific metrics, while allowing high cardinality system to remove the layer label to get aggregate metrics and cheaper logs for image specific information.

*Testing performed:*

### DIFF
--- a/fs/layer/node.go
+++ b/fs/layer/node.go
@@ -160,13 +160,25 @@ func (f *FuseOperationCounter) Inc(op string) {
 	atomic.AddInt32(opCount, 1)
 }
 
-// Run waits for f.waitPeriod to pass before emitting metric for
+// Run waits for f.waitPeriod to pass before emitting a log and metric for each
 // operation in FuseOpsList. Should be started in different goroutine so that it
 // doesn't block the current goroutine.
-func (f *FuseOperationCounter) Run() {
-	<-time.After(f.waitPeriod)
-	for op, opCount := range f.opCounts {
-		commonmetrics.AddImageOperationCount(op, f.imageDigest, atomic.LoadInt32(opCount))
+func (f *FuseOperationCounter) Run(ctx context.Context) {
+	select {
+	case <-ctx.Done():
+		return
+	case <-time.After(f.waitPeriod):
+		for op, opCount := range f.opCounts {
+			// We want both an aggregated metric (e.g. p90) and an image specific metric so that we can compare
+			// how a specific image is behaving to a larger dataset. When the image cardinatlity is small,
+			// we can just include the image digest as a label on the metric itself, however, when the cardinality
+			// is large, this can be very expensive. Here we give consumers options by emitting both logs and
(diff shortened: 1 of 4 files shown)


fs/layer/node.go
```diff
@@ -160,13 +160,25 @@ func (f *FuseOperationCounter) Inc(op string) {
...
+		for op, opCount := range f.opCounts {
+			// We want both an aggregated metric (e.g. p90) and an image specific metric so that we can compare
+			// a how a specific image is behaving to a larger dataset. When the image cardinatlity is small,
+			// including we can just include the image digest as a label on the metric itself, however, when
+			// the cardinality is large, this can be very expensive. Here we give consumers options by emitting both
+			// logs and metrics so that a low cardinality use case can rely on metrics, but a high cardinality use
+			// case can aggregate the metrics across all images, but still get the per-image info via logs.
+			count := atomic.LoadInt32(opCount)
+			commonmetrics.AddImageOperationCount(op, f.imageDigest, count)
+			log.G(ctx).Infof("fust operation found for image %s: %s = %d", f.imageDigest, op, count)
```
rdpsin, line 180: How about using `logrus.WithFields` here instead of `%s, %d`?

Kern--: I thought about doing that, but I wanted it to be easy to grep for this log. `WithContext` can't guarantee the order of the keys.

fs/layer/node.go
```diff
@@ -160,13 +160,25 @@ func (f *FuseOperationCounter) Inc(op string) {
...
-	for op, opCount := range f.opCounts {
-		commonmetrics.AddImageOperationCount(op, f.imageDigest, atomic.LoadInt32(opCount))
+func (f *FuseOperationCounter) Run(ctx context.Context) {
+	select {
+	case <-ctx.Done():
+		return
+	case <-time.After(f.waitPeriod):
+		for op, opCount := range f.opCounts {
+			// We want both an aggregated metric (e.g. p90) and an image specific metric so that we can compare
+			// a how a specific image is behaving to a larger dataset. When the image cardinatlity is small,
```
hanyuel, line 173: nit: typo? ``a how``

hanyuel (approved):

djdongjin (approved):

fs/layer/node.go
rdpsin: I think it may be possible to order keys by chaining `WithField`.

Kern--: `WithField` dumps the data into a map and go is very explicit that map traversal order is undefined. I think @sparr opened an issue specifically about how log key order is random.

PR #425, opened 2023-02-14
Eliminate null errors in resolver

### LINKED ISSUE #375: [BUG] debug output contains null errors
**Describe the bug**
Log output often contains debug messages with `null` and `nil` error values.

**Steps To Reproduce**
`make integration` exposes 121 instances of soci output with `"error":null` and `"level":"debug"` and 43 instances of `error="<nil>"`

**Expected behavior**
Either an appropriate error is surfaced, or the error key is omitted.

sparr:
*Issue #, if available:*
Closes #375 

*Description of changes:*
Make WithError conditional on the presence of an error in resolveFetcher

*Testing performed:*
Ran integration test suite. Confirmed approximately 80 fewer instances of `"error":null`.

### DIFF
--- a/fs/remote/resolver.go
+++ b/fs/remote/resolver.go
@@ -152,7 +152,12 @@ func (r *Resolver) resolveFetcher(ctx context.Context, hosts source.RegistryHost
 		return &remoteFetcher{r}, size, nil
 	}
 
-	log.G(ctx).WithError(handlersErr).WithField("ref", refspec.String()).WithField("digest", desc.Digest).Debugf("using default handler")
+	logger := log.G(ctx)
+	if handlersErr != nil {
+		logger = logger.WithError(handlersErr)
+	}
+	logger.WithField("ref", refspec.String()).WithField("digest", desc.Digest).Debugf("using default handler")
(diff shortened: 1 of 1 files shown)


Kern--:
> Confirmed approximately 80 fewer instances of `"error":null`.

Is this implying that there are no instances of this error after the change?

sparr:
> Is this implying that there are no instances of this error after the change?

No. There are still approx 43 instances, which I noted in #375 are due to us using an old version of containerd. I'll be PRing a bump of that version soon, separately and with mostly unrelated other motivation.

sbuckfelder (approved):
Oh so this was always logging, but sometimes an error wasn't set.

Are we sure we want to log every digest, or just ones with errors?

sparr:
> Are we sure we want to log every digest, or just ones with errors?

I might suggest that the errors show up at level:error and the others at info or debug. I opened #426 today about configurable test output verbosity. However, that's another concern for another issue/PR. I'd like to land this to get it off my plate.

Kern-- (approved):

Issue #385, opened 2023-02-01
[BUG] [INTEG] Dockerhub ratelimiting


turan18:
**Describe the bug**

Most integration tests run in their own container. Each container ( thus each test) usually ends up pulling our test image set from Dockerhub to build soci indices and ztocs.

Eg;
https://github.com/awslabs/soci-snapshotter/blob/75f860e054206960776f5a7a43337d238faddd82/integration/index_test.go#L103

https://github.com/awslabs/soci-snapshotter/blob/75f860e054206960776f5a7a43337d238faddd82/integration/create_test.go#L66

https://github.com/awslabs/soci-snapshotter/blob/6d83eef8cadf1c299050cd539b002fd19a0fe99f/integration/ztoc_test.go#L68

If we continue to add more tests there is a likely possibility of getting rate limited by Dockerhub.

**Expected behavior**

ideally, we would like each test container to have a copy of the image set without having to explicitly pull them. 

One possible way to do this would be to have each of the 4 test images pulled before hand on the users host (either using docker-cli or some containerd client). After pulling the images you can save each as a tarball using something `docker save`. You can then copy these tarballs over to the Dockerfile used to create the test containers like so `COPY img1.tar img2.tar img3.tar img4.tar soci_test_images/`. At this point each container can `docker load` those images from the `soci_test_images/` directly without having to pull them. 

#344 (In review) Introduces a `setup()` function that runs before integration tests, so this would be an ideal place to pull the image set and generate the tarballs.

The only caveat to this approach is that the containers themselves do not currently have docker installed in them. In order to load the image set in the container you will need a dind setup (Docker-in-Docker) as the containerd image import API does not work with docker image tarballs. We can get around this by either:

1. Using `ctr export` or the go client equivalent instead of `docker save` on the host
2. Installing docker within each container (dind)

**Configuration (please complete the following information):**
 - OS: 
 - Snapshotter Version:
 - Containerd Version: 

**Additional context**

sparr:
Could we run a repository on the host (likely in another container) which the test containers could `ctr image pull` from instead of using docker or `ctr import`?

sparr:
Would it be useful/reasonable to have an integration test flag that skips downloading the images if there's already a version of them on the host machine? This would speed up the tests for common usage, and reduce network usage, at the expense of making them less authoritative when run with that flag. That might deserve to be a separate feature request once this one lands.

rdpsin:
[...]

PR #428, opened 2023-02-14
Using `ctr import` instead of pulling images

turan18:
This change allows all images that are needed within our integration tests to be pulled during integration setup rather than test runtime. Since our integration tests run in isolated containers we could docker `COPY` the tarballs for each image (obtained via `ctr export`) for all the test images into our test container, whereby each test container could then `ctr image import` them without having to pull them.

**The main intention behind this PR was to speed up integration test times. Looking at both GitHub runners integ test time as well as local integ test times it appears that importing, surprisingly, does not seem to provide a faster test runtime in most cases. I'm converting this PR to a draft for now, while I explore this issue further**

*Issue #, if available:*

Fixes: #385

*Description of changes:*

*Testing performed:*


Dockerfile
```diff
@@ -24,7 +24,8 @@ ARG RUNC_VERSION
 ARG NERDCTL_VERSION
 COPY . $GOPATH/src/github.com/awslabs/soci-snapshotter
 ENV GOPROXY direct
-RUN apt-get update -y && apt-get install -y libbtrfs-dev libseccomp-dev libz-dev gcc fuse pigz
+RUN mkdir testImages/ && apt-get update -y && apt-get install -y libbtrfs-dev libseccomp-dev libz-dev gcc fuse pigz
+COPY integration/_images/ testImages/
```
sparr, line 28: mkdir should be unnecessary. COPY should create destination directories.

Makefile
```diff
@@ -76,7 +76,7 @@ test:
 integration: build
 	@echo "$@"
 	@echo "SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT)"
-	@GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true go test $(GO_TEST_FLAGS) -v -timeout=0 ./integration
+	@GO111MODULE=$(GO111MODULE_VALUE) sudo SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true go test $(GO_TEST_FLAGS) -v -timeout=0 ./integration
```
sparr, line 79: Why is this necessary?

integration/testutil.go
```diff
@@ -320,9 +332,12 @@ func encodeImageInfoNerdctl(ii ...imageInfo) [][]string {
...
 func copyImage(sh *shell.Shell, src, dst imageInfo) {
 	opts := encodeImageInfoNerdctl(src, dst)
-	sh.
-		X(append([]string{"nerdctl", "pull", "-q", "--platform", platforms.Format(src.platform)}, opts[0]...)...).
-		X("ctr", "i", "tag", src.ref, dst.ref).
+	if src.mirror {
+		sh.X(append([]string{"nerdctl", "pull", "-q", "--platform", platforms.Format(src.platform)}, opts[0]...)...)
+	} else {
+		sh.X(append([]string{"ctr", "i", "import", "--platform", platforms.Format(src.platform)}, fmt.Sprintf("testImages/%s.tar", src.imgName))...)
+	}
```
sparr, line 339: this pattern appears 3 places, can we move it to a helper?

integration/testutil.go
```diff
@@ -703,7 +726,66 @@ func setup() ([]func() error, error) {
...
 
-	return compose.Build(composeYaml, cOpts...)
+	ctx := namespaces.WithNamespace(context.Background(), "soci-testing")
+	client, err := containerd.New(containerdAddress)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to containerd: %w", err)
+	}
+	defer client.Close()
+
+	testImages := []string{alpineImage, ubuntuImage, drupalImage, nginxImage, rabbitmqImage, rabbitmqShaImage, rethinkdbShaImage}
```
sparr, line 736: Future additions to the image constants might not get added here. At minimum, I suggest adding a comment near https://github.com/awslabs/soci-snapshotter/pull/428/files#diff-a1077885a297cb6bcfbde19b03d7c04d94209ca38a3210644c969726aff4d985R82 pointing here. More preferably, move the creation of this slice there.

integration/testutil.go
```diff
@@ -77,13 +81,14 @@ const (
...
-	ubuntuImage   = "ubuntu:23.04"
-	drupalImage   = "drupal:10.0.2"
-	rabbitmqImage = "rabbitmq:3.11.7"
+	alpineImage       = "alpine:3.17.1"
+	nginxImage        = "nginx:1.23.3"
+	ubuntuImage       = "ubuntu:23.04"
+	drupalImage       = "drupal:10.0.2"
+	rabbitmqImage     = "rabbitmq:3.11.7"
+	rabbitmqShaImage  = "rabbitmq@sha256:603be6b7fd5f1d8c6eab8e7a234ed30d664b9356ec1b87833f3a46bb6725458e"
+	rethinkdbShaImage = "rethinkdb@sha256:4452aadba3e99771ff3559735dab16279c5a352359d79f38737c6fdca941c6e5"
```
sbuckfelder, line 90: Why the additional two images?  why different format?

integration/testutil.go
```diff
@@ -703,7 +726,66 @@ func setup() ([]func() error, error) {
...
+			fmt.Println("pulling", ref, "...")
+			i, err = client.Fetch(ctx, ref, remoteOpts...)
+			if err != nil {
+				return nil, fmt.Errorf("failed to pull image %s: %w", ref, err)
+			}
+		} else {
+			fmt.Printf("%s already exists. skipping pulling...\n", i.Name)
+		}
+		fmt.Println("exporting...")
+		tarballPath := filepath.Join(testImagePath, fmt.Sprintf("%s.tar", img[:strings.IndexByte(img, ':')]))
```
sbuckfelder, line 769: isn't this in the ref.imageName ?

sbuckfelder:
Did this work?  What kind of speed pick up did you see on the integration tests?

integration/testutil.go
```diff
@@ -290,14 +295,21 @@ func withPlatform(p spec.Platform) imageOpt {
...
+	mirror    bool
 	creds     string
 	plainHTTP bool
 	platform  spec.Platform
 }
 
 func dockerhub(name string, opts ...imageOpt) imageInfo {
-	i := imageInfo{dockerLibrary + name, "", false, platforms.DefaultSpec()}
+	i := imageInfo{
+		imgName: name[:strings.IndexByte(name, ':')],
```
djdongjin, line 308: this doesn't work if the image name contains a registry port (e.g. `localhost:5000/rabbitmq:latest` in our get start doc).

integration/testutil.go
djdongjin, line 339: can we have an estimate how many times `"nerdctl", "pull", "-q"` and `"ctr", "i", "import"` are called respectively? (e.g., you can run integration test and print `nerdctl-pull-count` (or `ctr-i-import-count`) whenever they're called and check the output.

integration/testutil.go
```diff
@@ -721,8 +803,12 @@ func teardown(cleanups []func() error) error {
 // middleSizeLayerInfo finds a layer not the smallest or largest (if possible), returns index, size, and layer count
 // It requires containerd to be running
 func middleSizeLayerInfo(t *testing.T, sh *shell.Shell, image imageInfo) (int, int64, int) {
-	sh.O("nerdctl", "pull", "-q", "--platform", platforms.Format(image.platform), image.ref)
+	if image.mirror {
+		sh.O("nerdctl", "pull", "-q", "--platform", platforms.Format(image.platform), image.ref)
+	} else {
+		sh.O(append([]string{"ctr", "i", "import", "--platform", platforms.Format(image.platform)}, fmt.Sprintf("testImages/%s.tar", image.imgName))...)
```
djdongjin, line 809: why does only these 3 places have this `if else` logic (e.g. why not other `nerdctl pull` usage)

integration/testutil.go
sparr: They are used in that format in existing tests.

djdongjin (commented):
Personally I feel this complicates our integration test and adds some hidden steps (e.g. 1/ how to add a new image; 2/ when we should use `nerdctl pull` v.s. `ctr i import`; 3/ how to add a new platform of an existing image (e.g. in the testImages slice`)

I'd like to see more concrete improvement from the change (e.g. integration test runtime reduction).

sparr:
> Did this work? What kind of speed pick up did you see on the integration tests?

I ran the test suite as well, but the change in time was below the noise floor for the few iterations I ran.

However, this isn't just about speed. This is also about reducing network usage, both on our end and to avoid throttling on the remote end (which was more likely with docker than ECR)

Makefile
turan18: Since I'm connecting to the containerd-grpc server via the client to pull images on the host.

integration/testutil.go
turan18: Last time I checked nerdctl was called roughly ~50 times or so. I'll run it again and check

integration/testutil.go
sparr: A helper to break apart the image ref into components would be valuable. I also suspect containerd might already have one.

integration/testutil.go
turan18: They are referenced in our pull test and create test. https://github.com/awslabs/soci-snapshotter/blob/main/integration/create_test.go#L59 and https://github.com/awslabs/soci-snapshotter/blob/main/integration/push_test.go#L90

integration/testutil.go
turan18: Good point , although calls to `dockerhub` assume the image to be residing in ECR. Calls to `mirror()` would break though if our local registry happens to contain ':'

integration/testutil.go
turan18: Just ran the integration suite locally here and noted the counts along with the time it took for each operation (pull + import). Here are the results:

pull count: 45
import count: 83
avg pull time: 776.2 ms
avg import time: [4305.6](tel:43058674698) ms

turan18: Looks like imports generally take longer than most pulls , which is very interesting to say the least.

integration/testutil.go
turan18: Actually looking back, this shouldnt break at all since we are splitting the image name+tag IE (rabbitmq:1.x or rabbitmq@sha256) and not the ref IE (localhost:5000/rabbitmq:latest)

integration/testutil.go
sparr: Following up on our meeting regarding this PR. I don't think those comparison statistics convey what they seem to. The scenarios where pull and import are happening could be wildly different, e.g. one might be all small images and the other all large.

Suggestion: Disable this PR and just include the counts and timing collection. Run the integration test suite.
Then compare those results to the results you have here.

PR #430, opened 2023-02-15
Move layer::waiter to layer_test

sbuckfelder:
fs/layer/layer.go contains a struct that creates a waiter that is useful for testing, and is currently only used in testing.  This PR moves the struct and logic from fs/layer/layer.go to fs/layer/layer_test.go.  

This was previously included in a different PR (https://github.com/awslabs/soci-snapshotter/pull/341), which we plan to close as it also adds testing tags which we have decided against.

*Testing performed:*
```make && make check && make test```

### DIFF
--- a/fs/layer/layer.go
+++ b/fs/layer/layer.go
@@ -533,54 +533,6 @@ type layerRef struct {
 	done func()
 }
 
-func newWaiter() *waiter {
-	return &waiter{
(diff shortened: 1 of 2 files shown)


djdongjin (approved):
LGTM

Kern-- (approved):

PR #431, opened 2023-02-15
Relocate some helper funcs in integration test

djdongjin:
*Issue #, if available:*

*Description of changes:*

A couple of integration test clean up, some of which should be potential bugs:

1. (bug) when `t.Parallel()`, we should put `tt := tt` inside `for _, tt := range tests` so it won't be the same `tt` being used across tests (I think we only need this when `t.Parallel()`).
2. (bug) a [non-test func](https://github.com/awslabs/soci-snapshotter/blob/77106683f1a01801ab3502d606337d0523b7af1c/integration/testutil.go#L730) (i.e. in a file without `_test` suffix) uses a [test func](https://github.com/awslabs/soci-snapshotter/blob/77106683f1a01801ab3502d606337d0523b7af1c/integration/create_test.go#L218): not sure why this doesn't cause error. (I saw this from vscode).
3. move `main_test.go` only helpers to `main_test.go`.
4. move some test helpers to testutil or testutil_soci.

*Testing performed:*

make check && make test && make integraion

### DIFF
--- a/integration/create_test.go
+++ b/integration/create_test.go
@@ -22,10 +22,7 @@ import (
 	"testing"
 
 	"github.com/awslabs/soci-snapshotter/soci"
-	shell "github.com/awslabs/soci-snapshotter/util/dockershell"
 	"github.com/containerd/containerd/platforms"
-	"github.com/google/go-cmp/cmp"
-	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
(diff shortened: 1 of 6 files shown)


integration/create_test.go
```diff
@@ -85,7 +82,9 @@ func TestSociCreateSparseIndex(t *testing.T) {
 				}
 			}
 
-			validateSociIndex(t, sh, index, manifestDigest, includedLayers)
+			if err := validateSociIndex(sh, index, manifestDigest, includedLayers); err != nil {
+				t.Fatal(err)
```
sparr, line 86: suggest `t.Fatalf("failed to validate soci index: %v", err)`

sbuckfelder (approved):
No strong opinions on this

turan18 (approved):

PR #442, opened 2023-02-20
Update default retry count for target retry duration of >5s

### LINKED ISSUE #432: [FEATURE] Increase default maximum effective retry duration
When retrying failed requests, we currently wait up to approximately 1-2 seconds total[1]. This is lower than most other services.

Increase default effective retry time to at least 5s.

**Additional context**
This could be accomplished by simply increasing `max_retries` from 5 to 7, although we might also tweak `min_wait_msec`, and we need to make sure `max_wait_msec` is at least `min_wait_msec*(2^max_retries)`

[1] This is based on comments, not code; the actual backoff and jitter code should be read thoroughly and understood and comments/docs corrected if necessary.

sparr:
*Issue #, if available:*
Closes #432 

*Description of changes:*
The default `max_retries` are increased from 5 to 8. This increases the total retry duration from ~1-2 seconds to ~8-16 seconds, which is greater than our planned target of 5 seconds.

Also minor comment updates.

*Testing performed:*
`make integration`


djdongjin:
>  This increases the total retry duration from ~1-2 seconds to ~8-16 seconds, which is greater than our planned target of 5 seconds.

This seems a huge increase, even compared to the planned target (5s). Any reason/benefit?

fs/remote/resolver.go
```diff
@@ -73,7 +73,8 @@ const (
 	defaultValidIntervalSec = 60
 	defaultFetchTimeoutSec  = 300
 
-	defaultMaxRetries  = 5
+	// defaults based on a target total retry time of at least 5s. 30*((2^8)-1)>5000
+	defaultMaxRetries  = 8
```
djdongjin, line 77: suggest to `7` which makes jitter in the range of `[3.8, 7.6]` (5.7 on average) and fits more into the expected 5s.

sparr: @sbuckfelder requested "more than 5 seconds" in planning elsewhere. I interpreted this as a minimum of 5, not a target/average of 5.

~5-10s could be achieved with 8 retries and 20ms minimum, or 7 retries and 40ms minimum. However, changing the minimum could have side effects such as triggering throttling or worsening the average time to success.

Perhaps +100% is not a reasonable jitter range? If we reduce the jitter range then we could have 8-10s instead of 8-16s as the range, without additional changes. @djdongjin?

sparr: I've changed the jitter calculation from +100% to +12.5% by adding a divisor to the jitter function. So the range is now ~8-9s.
Conversation remains open until we decide if this is still too long.

sbuckfelder (dismissed):
Happy with this, we can always change defaults in the next PR.  But overall I think these changes make sense

fs/remote/resolver.go
```diff
@@ -91,11 +92,11 @@ func NewResolver(cfg config.BlobConfig, handlers map[string]Handler) *Resolver {
 	if cfg.MaxRetries == 0 {
 		cfg.MaxRetries = defaultMaxRetries
 	}
-	if cfg.MinWaitMSec == 0 {
-		cfg.MinWaitMSec = defaultMinWaitMSec
+	if cfg.MinWaitMsec == 0 {
+		cfg.MinWaitMsec = defaultMinWaitMSec
```
djdongjin, line 96: I'm okay with either format but we should keep them consistent (e.g. `MinWaitMsec` -> `MinWaitMSec`, or `defaultMinWaitMSec` -> `defaultMinWaitMsec`, same for `defaultMaxWaitMSec`).

fs/remote/resolver.go
```diff
@@ -169,25 +170,26 @@ func (r *Resolver) resolveFetcher(ctx context.Context, hosts source.RegistryHost
...
-	desc        ocispec.Descriptor
-	maxRetries  int
-	minWaitMSec time.Duration
-	maxWaitMSec time.Duration
+	hosts      source.RegistryHosts
+	refspec    reference.Spec
+	desc       ocispec.Descriptor
+	maxRetries int
+	minWait    time.Duration
+	maxWait    time.Duration
```
djdongjin, line 178: I feel it's more intuitive to keep the unit `MSec`. Why do we need to remove this?

fs/config/config.go
```diff
@@ -81,11 +81,11 @@ type BlobConfig struct {
 	FetchTimeoutSec      int64 `toml:"fetching_timeout_sec"`
 	ForceSingleRangeMode bool  `toml:"force_single_range_mode"`
 	MaxRetries           int   `toml:"max_retries"`
-	MinWaitMSec          int   `toml:"min_wait_msec"`
-	MaxWaitMSec          int   `toml:"max_wait_msec"`
+	MinWaitMsec          int   `toml:"min_wait_msec"`
+	MaxWaitMsec          int   `toml:"max_wait_msec"`
```
sparr, line 85: go-staticcheck understands unit suffixes with a single leading capital letter, so it thinks "FooMSec" is "FooMsomething" with a unit of "seconds" instead of "Foo" with a unit of "milliseconds".

fs/remote/resolver.go
sparr: it's a go-staticcheck violation to put a unit in the identifier of a time.Duration. This member conceptually has no unit, and if you dive into the underlying type then it's actually stored in nanoseconds, not milliseconds. https://staticcheck.io/docs/checks#ST1011

djdongjin (approved):

sbuckfelder (approved):

fs/remote/resolver.go
sparr: Resolving based on approval from commenters.

Issue #231, opened 2022-12-13
Migrate decompression from SpanManager to ZInfo


vkuzniet:
**Is your feature request related to a problem? Please describe.**
Per #103, the snapshotter's data path needs to be adjusted to separate ztoc and compression concepts. The complete details on that can be found in the attached design doc.
To achieve this goal we need to implement the following:
1. Create interface Extractor and implement it in GzipZinfo
2. SpanManager should contain the instance of Extractor instead of directly GzipZinfo.

**Describe the solution you'd like**
Implement according to the acceptance criteria in the above section.

**Describe alternatives you've considered**
Keep things as is.

**Additional context**
#103

PR #444, opened 2023-02-21
Refactor span manager to improve comment and reusability

djdongjin:
*Issue #, if available:*

Part of #231, this PR makes refactor-only changes to span manager, so the follow up PR for #231 will be smaller.

*Description of changes:*

1. Remove `fetchSpan` and reuse `fetchWithRetries` (there're just too many "fetch span" related funcs: `FetchSingleSpan`, `fetchWithRetries`, `fetchSpan`, `fetchAndCacheSpan`, etc).
2. Remove `size` arg in `GetSpanContent` (it already has `offsetStart, offsetEnd`, size is redundant).
3. Let `fetchSpanWithRetries` calculate span size and return the buffer. (so every caller doesn't need to calculate span size and make a buffer and pass buffer to `fetchSpanWithRetries`, which is duplicated)
4. Doc/comment improvement.

(in a separate commit https://github.com/awslabs/soci-snapshotter/pull/444/commits/b4022ce41036db1711e8431e89632d6b3aa8ae89)

1. add `err != io.EOF` check to `ReadAt` calls (https://github.com/awslabs/soci-snapshotter/pull/444#discussion_r1114554669)

*Testing performed:*

### DIFF
--- a/fs/span-manager/span_manager.go
+++ b/fs/span-manager/span_manager.go
@@ -60,6 +60,7 @@ var stateTransitionMap = map[spanState][]spanState{
 	uncompressed: {uncompressed},
 }
 
+// Specific error types raised by SpanManager.
 var (
 	ErrSpanNotAvailable           = errors.New("span not available in cache")
 	ErrIncorrectSpanDigest        = errors.New("span digests do not match")
@@ -96,6 +97,7 @@ func (s *span) validateStateTransition(newState spanState) error {
 	return errInvalidSpanStateTransition
 }
 
+// SpanManager fetches and caches spans of a given layer.
 type SpanManager struct {
 	cache                             cache.BlobCache
 	cacheOpt                          []cache.Option
@@ -119,6 +121,8 @@ type spanInfo struct {
 	spanIndexInBuf []compression.Offset
 }
 
+// New creates a SpanManager with given ztoc and content reader, and builds all
+// spans based on the ztoc.
 func New(ztoc *ztoc.Ztoc, r *io.SectionReader, cache cache.BlobCache, retries int, cacheOpt ...cache.Option) *SpanManager {
 	index, err := compression.NewGzipZinfo(ztoc.CompressionInfo.Checkpoints)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


fs/span-manager/span_manager.go
```diff
@@ -442,16 +380,73 @@ func (m *SpanManager) resolveSpanFromCache(s *span, offsetStart, size compressio
 	return nil, ErrSpanNotAvailable
 }
 
-func (m *SpanManager) fetchSpan(buf []byte, spanID compression.SpanID) error {
+// fetchAndCacheSpan fetches a span, uncompresses it and caches the uncompressed
+// span content.
+func (m *SpanManager) fetchAndCacheSpan(spanID compression.SpanID) ([]byte, error) {
+	// fetch compressed span
+	compressedBuf, err := m.fetchSpanWithRetries(spanID)
+	if err != nil && err != io.EOF {
```
sbuckfelder, line 388: what should happen in the case of `err == io.EOF`?

fs/span-manager/span_manager.go
```diff
@@ -442,16 +380,73 @@ func (m *SpanManager) resolveSpanFromCache(s *span, offsetStart, size compressio
...
+	err = s.setState(uncompressed)
+	if err != nil {
+		return nil, err
+	}
+	return uncompSpanBuf, nil
+}
+
+// fetchSpanWithRetries fetches the requested data and verifies that the span digest matches the one in the ztoc.
+// It will retry the fetch and verification m.maxSpanVerificationFailureRetries times.
+// It does not retry when there is an error fetching the data, because retries already happen lower in the stack in httpFetcher.
```
sbuckfelder, line 415: Very nice comment

sbuckfelder (dismissed):
Looks good

fs/span-manager/span_manager.go
djdongjin: From the `ReadAt` doc (https://pkg.go.dev/io#ReaderAt only place where `err` might returned within `fetchSpanWithRetries`).

> If the n = len(p) bytes returned by ReadAt are at the end of the input source, ReadAt may return either err == EOF or err == nil.

This may indicate that we should check `if err != nil && err != io.EOF` in all places where `ReadAt` is called. Meanwhile, we should move the `err != io.EOF` into `fetchSpanWithRetries` when `ReadAt` returns.

djdongjin: @sbuckfelder added `EOF` check to other `ReadAt` calls (in a separate commit https://github.com/awslabs/soci-snapshotter/pull/444/commits/b4022ce41036db1711e8431e89632d6b3aa8ae89). PTAL.

sbuckfelder (approved):

sparr (approved):

fs/span-manager/span_manager.go
```diff
@@ -442,16 +380,75 @@ func (m *SpanManager) resolveSpanFromCache(s *span, offsetStart, size compressio
 	return nil, ErrSpanNotAvailable
 }
 
-func (m *SpanManager) fetchSpan(buf []byte, spanID compression.SpanID) error {
+// fetchAndCacheSpan fetches a span, uncompresses it and caches the uncompressed
+// span content.
+func (m *SpanManager) fetchAndCacheSpan(spanID compression.SpanID) ([]byte, error) {
+	// fetch compressed span
+	compressedBuf, err := m.fetchSpanWithRetries(spanID)
```
Kern--, line 387: I think one subtle difference between this and the previous version is that since we only need `compressedBuf` until `m.uncompressSpan`, the old version didn't escape the function and therefore could be allocated on the stack. Since this `[]byte` is being returned from another function, it causes a heap allocation. I'm not sure how much that really matters, but we should keep that in mind the next time we're taking a look at performance and memory usage.

Kern-- (approved):

fs/span-manager/span_manager.go
djdongjin: got it. So the difference is mainly on "return compressedBuf from `fetchSpanWithRetries` v.s. pass compressedBuf from caller into `fetchSpanWithRetries`". 

Also notice we are already returning (uncompressed, which should be larger) buffer (instead of pass in)  in `uncompressSpan`.

https://github.com/awslabs/soci-snapshotter/pull/446 will clean up several wrappers on span operations, which should make it easier to compare and unify how to take `[]byte buffer` (pass in or return).

Issue #434, opened 2023-02-16
[FEATURE] Make OCI 1.0 Image Manifest the default SOCI CLI format


Kern--:
**Is your feature request related to a problem? Please describe.**
SOCI supports serializing a SOCI index manifest as either an OCI 1.0 Image Manifest or an OCI 1.1 release candidate Artifact Manifest. By default `soci create` uses the Artifact Manifest, but can be overriden using `soci create --legacy-registry`. 

The OCI spec authors are still debating whether the Artifact Manifest will be part of the finalized spec, so we should default to using the Image Manifest to not tie ourselves to a format that may not exist in the future and to support more registries.

**Describe the solution you'd like**
`soci create` creates an image manifest that works with OCI 1.0/1.1 registries.
`soci create --oci-artifact` creates an artifact manifest that works with OCI 1.1 registries.

**Describe alternatives you've considered**
We could leave things the way they are now, but then it will be hard for anyone to use SOCI until the OCI 1.1 spec is finalized and registries add support.

**Additional context**

PR #445, opened 2023-02-21
Create OCI 1.0 Image Manifests for SOCI Indices by default.

turan18:
This change inverts the current behavior of 'soci create' to create OCI 1.0 image manifests for the SOCI index by default. A `--manifest-type` flag has been added to specify the type of manifest to create for SOCI indices (currently either 'artifact' or 'image'). If the flag is not used 'soci create' will default to image manifest.

*Issue #, if available:*

Fixes: #434 

Example:

```
$ soci create --help
NAME:
   soci create - create SOCI index
USAGE:
   soci create [command options] [flags] <image_ref>
OPTIONS:
   ....
   --manifest-type value       Generate either an OCI 1.1 artifact manifest or OCI 1.0 image manifest for the SOCI index.
        (You should use 'artifact' only if you intend on interacting with a registry that support OCI 1.1 artifacts) (default: "image")
   ...


$ soci create -p linux/amd64  docker.io/library/redis:latest
layer sha256:db7c27bf355290eeb889d2e2986ba34352eba3adba6ae44eaef63d5e7dd69898 -> ztoc skipped
layer sha256:51afc2cce3dfa745f19f601b33820e4a960311cd81b0a6a816b1de8e5ae6d7d7 -> ztoc skipped
layer sha256:ac509f65c3e925781d771dc15fd63c2bd795abdc461f29bd9a3acd0a19f65d3f -> ztoc skipped
layer sha256:ab1a1215d5f9ec95bf6981023be29642373dcdd6218d602d1f049d5e5ee3ee29 -> ztoc skipped
layer sha256:817f7e347ebd5d210d4a5923c221e92275d665ced36e12185bc26747e977e47d -> ztoc skipped
layer sha256:bb263680fed18eecdc67f885094df6f589bafc19004839d7fdf141df236a61aa -> ztoc sha256:3bb9b6f663024dc85e7acffe60dff225d3ff32d9ebfef4784760ce4a6214b36b

$ soci index list
DIGEST                                                                     SIZE    IMAGE REF                         PLATFORM       MEDIA TYPE                                    CREATED
sha256:daca991c9c605eb10e1eaa0f4ee608f7ede3b4b3a12ce26317c7e96d92224972    854     docker.io/library/redis:latest    linux/amd64    application/vnd.oci.image.manifest.v1+json    5s ago

$ soci index info sha256:daca991c9c605eb10e1eaa0f4ee608f7ede3b4b3a12ce26317c7e96d92224972 | jq
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "config": {
    "mediaType": "application/vnd.amazon.soci.index.v1+json",
    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
    "size": 2
  },
  "layers": [
    {
      "mediaType": "application/octet-stream",
      "digest": "sha256:3bb9b6f663024dc85e7acffe60dff225d3ff32d9ebfef4784760ce4a6214b36b",
      "size": 1495528,
      "annotations": {
        "com.amazon.soci.image-layer-digest": "sha256:bb263680fed18eecdc67f885094df6f589bafc19004839d7fdf141df236a61aa",
        "com.amazon.soci.image-layer-mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip"
      }
    }
  ],
  "subject": {
    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
    "digest": "sha256:87583c95fd2253658fdd12e765addbd2126879af86a90b34efc09457486b21b1",
    "size": 1573
  },
  "annotations": {
    "com.amazon.soci.build-tool-identifier": "AWS SOCI CLI v0.1"
  }
}
```

*Testing performed:*

### DIFF
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -18,6 +18,7 @@ package commands
 
 import (
 	"errors"
+	"fmt"
 	"os"
 
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/internal"
@@ -30,9 +31,8 @@ import (
 
 const (
 	buildToolIdentifier = "AWS SOCI CLI v0.1"
-
-	spanSizeFlag     = "span-size"
-	minLayerSizeFlag = "min-layer-size"
+	spanSizeFlag        = "span-size"
+	minLayerSizeFlag    = "min-layer-size"
 )
 
 // CreateCommand creates SOCI index for an image
@@ -45,7 +45,7 @@ var CreateCommand = cli.Command{
 	ArgsUsage: "[flags] <image_ref>",
 	Flags: append(
 		internal.PlatformFlags,
-		internal.LegacyRegistryFlag,
+		internal.ManifestTypeFlag,
 		cli.Int64Flag{
 			Name:  spanSizeFlag,
 			Usage: "Span size that soci index uses to segment layer data. Default is 4 MiB",
@@ -107,9 +107,17 @@ var CreateCommand = cli.Command{
 			soci.WithSpanSize(spanSize),
 			soci.WithBuildToolIdentifier(buildToolIdentifier),
 		}
-		if cliContext.Bool(internal.LegacyRegistryFlagName) {
-			builderOpts = append(builderOpts, soci.WithLegacyRegistrySupport)
+
+		manifestType := cliContext.String(internal.ManifestTypeFlagName)
+
+		if manifestType != internal.ImageManifestType && manifestType != internal.ArtifactManifestType {
+			return fmt.Errorf("undefined manifest type: %v. supported manifest types: [%s, %s]", manifestType, internal.ImageManifestType, internal.ArtifactManifestType)
 		}
+
+		if manifestType == internal.ArtifactManifestType {
+			builderOpts = append(builderOpts, soci.WithOCIArtifactRegistrySupport)
+		}
+
 		for _, plat := range ps {
 			builder, err := soci.NewIndexBuilder(cs, blobStore, artifactsDb, append(builderOpts, soci.WithPlatform(plat))...)
 
--- a/cmd/soci/commands/internal/spec.go
+++ b/cmd/soci/commands/internal/spec.go
@@ -19,13 +19,14 @@ package internal
 import "github.com/urfave/cli"
 
 const (
-	LegacyRegistryFlagName = "legacy-registry"
+	ManifestTypeFlagName = "manifest-type"
+	ImageManifestType    = "image"
+	ArtifactManifestType = "artifact"
 )
 
-var LegacyRegistryFlag = cli.BoolFlag{
-	Name: LegacyRegistryFlagName,
-	Usage: `Whether to create the SOCI index for a legacy registry. OCI 1.1 added support for associating artifacts such as soci indices with images.
-     There is a mechanism to emulate this behavior with OCI 1.0 registries by pretending that the SOCI index
-     is itself an image. This option should only be use if the SOCI index will be pushed to a
-     registry which does not support OCI 1.1 features.`,
+var ManifestTypeFlag = cli.StringFlag{
+	Name:  ManifestTypeFlagName,
+	Value: ImageManifestType,
+	Usage: `Generate either an OCI 1.1 artifact manifest or OCI 1.0 image manifest for the SOCI index. 
+        (You should use 'artifact' only if you intend on interacting with a registry that supports OCI 1.1 artifacts)`,
 }
--- a/integration/push_test.go
+++ b/integration/push_test.go
@@ -54,7 +54,7 @@ func TestSociArtifactsPushAndPull(t *testing.T) {
 
 			imageName := ubuntuImage
 			copyImage(sh, dockerhub(imageName, withPlatform(platform)), regConfig.mirror(imageName, withPlatform(platform)))
-			indexDigest := buildIndex(sh, regConfig.mirror(imageName, withPlatform(platform)), withMinLayerSize(0))
+			indexDigest := buildIndex(sh, regConfig.mirror(imageName, withPlatform(platform)), withMinLayerSize(0), withOCIArtifactRegistrySupport)
 			artifactsStoreContentDigest := getSociLocalStoreContentDigest(sh)
 			sh.X("soci", "push", "--user", regConfig.creds(), "--platform", tt.Platform, regConfig.mirror(imageName).ref)
 			sh.X("rm", "-rf", "/var/lib/soci-snapshotter-grpc/content/blobs/sha256")
@@ -108,7 +108,7 @@ func TestPushAlwaysMostRecentlyCreatedIndex(t *testing.T) {
 			copyImage(sh, dockerhub(tc.image), regConfig.mirror(tc.image))
 
 			for _, opt := range tc.opts {
-				index := buildIndex(sh, regConfig.mirror(tc.image), withMinLayerSize(opt.minLayerSize), withSpanSize(opt.spanSize))
+				index := buildIndex(sh, regConfig.mirror(tc.image), withMinLayerSize(opt.minLayerSize), withSpanSize(opt.spanSize), withOCIArtifactRegistrySupport)
 				index = strings.Split(index, "\n")[0]
 				out := sh.O("soci", "push", "--user", regConfig.creds(), regConfig.mirror(tc.image).ref, "-q")
 				pushedIndex := strings.Trim(string(out), "\n")
@@ -123,34 +123,34 @@ func TestPushAlwaysMostRecentlyCreatedIndex(t *testing.T) {
 
 func TestLegacyOCI(t *testing.T) {
 	tests := []struct {
-		name                  string
-		registryImage         string
-		supportLegacyRegistry bool
-		expectError           bool
+		name                    string
+		registryImage           string
+		supportArtifactRegistry bool
+		expectError             bool
 	}{
 		{
-			name:                  "OCI 1.1 Artifacts succeed with OCI 1.1 registry",
-			registryImage:         oci11RegistryImage,
-			supportLegacyRegistry: false,
-			expectError:           false,
+			name:                    "OCI 1.1 Artifacts succeed with OCI 1.1 registry",
+			registryImage:           oci11RegistryImage,
+			supportArtifactRegistry: true,
+			expectError:             false,
 		},
 		{
-			name:                  "OCI 1.0 Artifacts succeed with OCI 1.1 registry",
-			registryImage:         oci11RegistryImage,
-			supportLegacyRegistry: true,
-			expectError:           false,
+			name:                    "OCI 1.0 Artifacts succeed with OCI 1.1 registry",
+			registryImage:           oci11RegistryImage,
+			supportArtifactRegistry: false,
+			expectError:             false,
 		},
 		{
-			name:                  "OCI 1.1 Artifacts fail with OCI 1.0 registry",
-			registryImage:         oci10RegistryImage,
-			supportLegacyRegistry: false,
-			expectError:           true,
+			name:                    "OCI 1.1 Artifacts fail with OCI 1.0 registry",
+			registryImage:           oci10RegistryImage,
+			supportArtifactRegistry: true,
+			expectError:             true,
 		},
 		{
-			name:                  "OCI 1.0 Artifacts succeed with OCI 1.0 registry",
-			registryImage:         oci10RegistryImage,
-			supportLegacyRegistry: true,
-			expectError:           false,
+			name:                    "OCI 1.0 Artifacts succeed with OCI 1.0 registry",
+			registryImage:           oci10RegistryImage,
+			supportArtifactRegistry: false,
+			expectError:             false,
 		},
 	}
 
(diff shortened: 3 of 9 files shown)


djdongjin (commented):
I wonder which of these `indexDigest := buildIndex(sh, imgInfo, withMinLayerSize(0), withOCIArtifactRegistrySupport)` changes are needed. I think we should keep most of our tests in default setting (so no `withOCIArtifactRegistrySupport` nor the previous `withLegacyRegistrySupport`) unless we explicitly test that behavior. Otherwise, we use no-withOCIArtifactRegistrySupport as default but most tests are with `withOCIArtifactRegistrySupport`.

`validateSociIndex` may also be changed to cover this change. You can pass a `artifact bool` as arg or just change to check either imagemanifest or artifactmanifest.

https://github.com/awslabs/soci-snapshotter/blob/8dde69656027f429331db88abb10c5a9dfd3a97b/integration/testutil_soci.go#L121-L123

sbuckfelder:
How about we introduce a `manifest-type` argument, and if the user does not include it the value defaults to legacy-registry?  My reasoning behind this is that it will be easier to maintain as 
1) new manifest types are introduced
2) we need to change the default

soci/soci_index.go
```diff
@@ -256,13 +256,10 @@ func WithPlatform(platform ocispec.Platform) BuildOption {
...
 
-// WithLegacyRegistrySupport sets whether the SOCI Index should be built for legacy registries.
-// OCI 1.1 added support for associating artifacts such as SOCI indices with images. There is a
-// mechanism to emulate this behavior with OCI 1.0 registries by pretending that the SOCI index
-// is itself an image. This option should only be use if the SOCI index will be pushed to a
-// registry which does not support OCI 1.1 features.
-func WithLegacyRegistrySupport(c *buildConfig) error {
-	c.legacyRegistry = true
+// WithOCIArtifactRegistrySupport sets whether the SOCI index should be built for registries that
+// support OCI 1.1 artifacts.
```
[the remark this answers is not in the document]
turan18: There are 4 instances in our integ tests where we push soci indices to an OCI 1.1 registry so I have added in the `withOCIArtifactRegistrySupport` option only for these cases.

cmd/soci/commands/create.go
```diff
@@ -107,9 +109,17 @@ var CreateCommand = cli.Command{
 			soci.WithSpanSize(spanSize),
 			soci.WithBuildToolIdentifier(buildToolIdentifier),
 		}
-		if cliContext.Bool(internal.LegacyRegistryFlagName) {
-			builderOpts = append(builderOpts, soci.WithLegacyRegistrySupport)
+
+		manifestType := cliContext.String(internal.ManifestTypeFlagName)
+
+		if manifestType != imageManifestType && manifestType != artifactManifestType {
+			return fmt.Errorf("Currently only %s and %s manifest types are supported. Undefined manifest type: %v", imageManifestType, artifactManifestType, manifestType)
```
djdongjin, line 116: nit: `fmt.Errorf("undefined manifest type: %v. supported manifest types: [%s, %s]", manifestType, imageManifestType, artifactManifestType)`.

cmd/soci/commands/create.go
```diff
@@ -29,10 +30,11 @@ import (
...
 const (
-	buildToolIdentifier = "AWS SOCI CLI v0.1"
-
-	spanSizeFlag     = "span-size"
-	minLayerSizeFlag = "min-layer-size"
+	buildToolIdentifier  = "AWS SOCI CLI v0.1"
+	spanSizeFlag         = "span-size"
+	minLayerSizeFlag     = "min-layer-size"
+	imageManifestType    = "image"
+	artifactManifestType = "artifact"
```
djdongjin, line 37: these 2 should be moved together with `ManifestTypeFlagName`.

cmd/soci/commands/internal/spec.go
```diff
@@ -19,13 +19,12 @@ package internal
...
 
-var LegacyRegistryFlag = cli.BoolFlag{
-	Name: LegacyRegistryFlagName,
-	Usage: `Whether to create the SOCI index for a legacy registry. OCI 1.1 added support for associating artifacts such as soci indices with images.
-     There is a mechanism to emulate this behavior with OCI 1.0 registries by pretending that the SOCI index
-     is itself an image. This option should only be use if the SOCI index will be pushed to a
-     registry which does not support OCI 1.1 features.`,
+var ManifestTypeFlag = cli.StringFlag{
+	Name:  ManifestTypeFlagName,
+	Value: "image",
```
djdongjin, line 27: "image" -> ImageManifestType

soci/soci_index.go
```diff
@@ -256,13 +256,10 @@ func WithPlatform(platform ocispec.Platform) BuildOption {
...
 
-// WithLegacyRegistrySupport sets whether the SOCI Index should be built for legacy registries.
-// OCI 1.1 added support for associating artifacts such as SOCI indices with images. There is a
-// mechanism to emulate this behavior with OCI 1.0 registries by pretending that the SOCI index
-// is itself an image. This option should only be use if the SOCI index will be pushed to a
-// registry which does not support OCI 1.1 features.
-func WithLegacyRegistrySupport(c *buildConfig) error {
-	c.legacyRegistry = true
+// WithOCIArtifactRegistrySupport sets whether the SOCI index should be built for registries that
+// support OCI 1.1 artifacts by building an artifact manifest instead of an image manifest.
```
djdongjin, line 260: nit:

// WithOCIArtifactRegistrySupport sets whether to build SOCI index as an artifact manifest instead of an image
// manifest. Artifact manifest is released in OCI 1.1 spec and is only supported by OCI 1.1 compatible registries.

integration/testutil_soci.go
```diff
@@ -118,8 +119,8 @@ func buildIndex(sh *shell.Shell, src imageInfo, opt ...indexBuildOption) string
 }
 
 func validateSociIndex(sh *shell.Shell, sociIndex soci.Index, imgManifestDigest string, includedLayers map[string]struct{}) error {
-	if sociIndex.MediaType != ocispec.MediaTypeArtifactManifest {
-		return fmt.Errorf("unexpected index media type; expected = %v, got = %v", ocispec.MediaTypeArtifactManifest, sociIndex.MediaType)
+	if sociIndex.MediaType != ocispec.MediaTypeArtifactManifest && sociIndex.MediaType != ocispec.MediaTypeImageManifest {
+		return fmt.Errorf("unexpected index media type; expected = %v or %v, got = %v", ocispec.MediaTypeArtifactManifest, ocispec.MediaTypeImageManifest, sociIndex.MediaType)
```
djdongjin, line 123: nit: `expected = %v or %v, got = %v` -> `expected: [%v, %v], got: %v`

integration/push_test.go
```diff
@@ -165,11 +165,11 @@ func TestLegacyOCI(t *testing.T) {
 
 			imageName := ubuntuImage
 			copyImage(sh, dockerhub(imageName), regConfig.mirror(imageName))
+
 			var buildOpts []indexBuildOption
-			if tc.supportLegacyRegistry {
-				buildOpts = append(buildOpts, withLegacyRegistrySupport)
+			if !tc.supportLegacyRegistry {
```
djdongjin, line 170: can we invert `supportLegacyRegistry` and change to `tc.supportArtifactRegistry`. Since `supportLegacyRegistry` no longer exists.

soci/soci_index.go
```diff
@@ -448,10 +445,11 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
...
 // NewIndex returns a new index.
-func NewIndex(blobs []ocispec.Descriptor, subject *ocispec.Descriptor, annotations map[string]string, legacy bool) *Index {
-	mediaType := OCIArtifactManifestMediaType
-	if legacy {
-		mediaType = ocispec.MediaTypeImageManifest
+func NewIndex(blobs []ocispec.Descriptor, subject *ocispec.Descriptor, annotations map[string]string, artifact bool) *Index {
+	mediaType := ocispec.MediaTypeImageManifest
+
+	if artifact {
+		mediaType = OCIArtifactManifestMediaType
```
djdongjin, line 452: Since we're using [`image-spec v1.1.0-rc2`](https://github.com/awslabs/soci-snapshotter/blob/7f0c3aa79b204aeb36a70bec40a051bda21c308a/go.mod#L21) and `MediaTypeArtifactManifest` exists in [image-spec](https://github.com/opencontainers/image-spec/blob/19a74bcb54ba211005a68d85c6b359c2947721ce/specs-go/v1/mediatype.go#LL59C2-L59C27), I think we should remove our copy and use `ocispec.MediaTypeArtifactManifest`.

WDYT @Kern--

Kern--: We can. I'm not too rushed because it's not clear that artifact manifests will survive until the release of `v1.1.0`

soci/soci_index.go
turan18: I don't think artifact manifest are officially decided on for 1.1

integration/testutil_soci.go
sparr: I disagree with "[foo, bar]" as shorthand for "foo or bar" in contexts like this. The brackets look more like an array of length two.

turan18: I think the the the `[]` notation is fine if we explicitly mention a plural like **types** (eg: `expected types [a,b]`)

djdongjin (dismissed):

soci/soci_index.go
```diff
@@ -448,10 +443,11 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 }
 
 // NewIndex returns a new index.
-func NewIndex(blobs []ocispec.Descriptor, subject *ocispec.Descriptor, annotations map[string]string, legacy bool) *Index {
-	mediaType := OCIArtifactManifestMediaType
-	if legacy {
-		mediaType = ocispec.MediaTypeImageManifest
+func NewIndex(blobs []ocispec.Descriptor, subject *ocispec.Descriptor, annotations map[string]string, artifact bool) *Index {
```
Kern--, line 446: Sorry, I was being lazy when I first introduced the `legacy` bool to this. We really should have some sort of `newIndexOptions` struct with a functional arg here.

I think there is a reasonably likely case that artifacts never make it into the OCI image spec in which case it would be pretty unfortunate to be stuck with a useless boolean in the function signature. 

If we make it a functional arg, then if artifacts get removed, we could remove just that function and the only code that would need to change is the code where people were using artifacts. If we keep it a bool, then everyone would need to update code to remove the bool.

turan18: Since we already have a buildConfig struct would it make sense to change it to have a `manifestType` attribute instead of an `artifact` bool. Something like 
```
type buildConfig struct {
        ...
	artifactsDb        *ArtifactsDb
	platform           ocispec.Platform
	manifestType       string
}
```
We can set the `manifestType` to be either `ocispec.MediaTypeImageManifest` or `ocispec.MediaTypeArtifactManifest`. That way when creating the Index we can set its `MediaType` field directly to `config.ManifestType`

Kern--: Yeah, I think that could make the index builder structure a bit cleaner. The question is how do we pass that to `soci.NewIndex`? Right now it's a bool. We could switch it to a string which is the media type. I think the risk is that if artifacts don't land in the spec, then we have this string where you just always pass the image manifest string which is weird.

turan18: I think part of the reason behind have the `--manifest-type` argument was to make the CLI extensible/support any new manifest types that may exist in the future. I do think functional arguments would make the function signature a little cleaner but we should remove the `artifact` bool attribute from `buildConfig` if we intend to have a `newIndexOptions` struct since having functionally similar attributes for both would make things repetitive.

Kern--: I think that sounds good, but it's not totally clear to me how we would reduce the repetition. We need some way to pass the info from CLI -> Index builder and from Index builder -> `soci.NewIndex`. Ideally in a way that you only have to pass anything if you're using non-default behavior. I think that would require 2 sets of functional args, but I'm open to reviewing a PR if you have an idea for making it just one set.

turan18: Yeah I think you are right... I cant think of a way of separating the `bool `logic from build config and index config while maintaining a single functional arg.

Kern-- (approved):

djdongjin (approved):
Approve since I don't want to block this PR which is a working solution now.

But I do think it's confusing to have both `buildConfig` and `indexConfig` (e.g., what's the usecase only `indexConfig` is used but not `buildConfig`). Further, what's the usecase `NewIndex` will be called not by `IndexBuilder.Build`?

soci/soci_index.go
```diff
@@ -218,12 +216,24 @@ type buildConfig struct {
 	buildToolIdentifier string
 	artifactsDb         *ArtifactsDb
 	platform            ocispec.Platform
-	legacyRegistry      bool
+	artifactRegistry    bool
+}
+type indexConfig struct {
+	artifact bool
```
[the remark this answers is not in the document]
turan18: There isn't one. Index config just serves as a wrapper to relay index specific configs to `NewIndex`. I agree that it is confusing/repetitive but I think its fine for now. 

I think the path of least resistance would have been to keep only `buildConfig` and replace `artifactRegistry` to `manifestType` as string and pass that string to `NewIndex`. Even though we run the risk of passing type `image` constantly(if artifacts dont go through) we llow room for flexibility in the future if more manifest types are added in (without having to change anything in our soci codepath just the cli) 

Another option would be to attach `newIndex` to the an instance of `indexBuilder` (like you mentioned below). I think this option is fine if we dont plan on having standalone utility for `NewIndex` outside of the the some of the unit tests that call it (which can easily be substituted for something like `IndexBuilder{}.newIndex()` )

PR #446, opened 2023-02-23
Improve span manager API and span state transfer

djdongjin:
*Issue #, if available:*

Part of https://github.com/awslabs/soci-snapshotter/issues/231. This PR make it clear when span manager should check and change state (and when/how to use span state lock), and unify span manager's funcs interface (e.g. all span args are `spanID compression.SpanID` not `*span`) and modularize these funcs (fetch, uncompress, cache).

**From outside, span manager provides 2 functionalities:**

1. Resolve a span without uncompress and without span data request (`FetchSingleSpan`): bg-fetcher makes compressed span available in cache.
2. Resolve a span with uncompress and with span data chunk request (`GetSpanContent`): when there is span data request (e.g., via `GetContents`)

**From inside, span manager does the following operation (state transfer) for a given span:**

1. unreqested -> requested -> fetched -> uncompressed: the whole span workflow from `unrequested` to `uncompressed`
2. unrequested -> requested -> fetched: the whole span workflow without the final `uncompressed` (from bg-fetcher).
3. fetched -> uncompressed: when span data is requested, and the span is compressed in cache (by bg-fetcher).

**For span state, we use 2 race-protect mechanisms (`atomic.Value` and `sync.Mutex`).** In general, we can use `atomic.Value` without `Lock()` only when the state leads to early return:

1. e.g., when bg-fetcher pre-fetches a span via `ResolveSpan`, it can early return if `state != unrequested` which means the data is available or being requested; 
2. e.g. when requesting span data and `state == uncompressed`, it can early return the data from cache directly.

For other cases, we should acquire the `Lock` first (and recheck the state) and then do the operation (fetch, uncompress, etc), update state, release lock.

*Description of changes:*

**Improve span manager API/func based on the above observation:**

1. FetchSingleSpan: if span `unrequested`, fetchAndCacheSpan without uncompress (`unrequested` -> `fetched`); if not, do nothing.
2. GetContents/GetSpanContent: get a chunk of data (muti/single span): do necessary operation to ensure span is uncompressed (i.e. if `unrequested`,  fetchAndCacheSpan; if `fetched`, uncompress and cache span; if `uncompressed`, get from cache) and get a chunk of data from the span.
3. fetchAndCacheSpan: assuming data is `unrequested`, fetch and cache the span. Uncompress is optional (so it can be reused by `FetchSingleSpan` and `GetSpanContent`).

Both 1 and 2 will use 3 to fetch and cache span. `FetchSingleSpan` covers state transfer 1 and 2; `GetSpanContent` covers state transfer 3.

*Testing performed:*

### DIFF
--- a/fs/span-manager/span_manager.go
+++ b/fs/span-manager/span_manager.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"io"
 	"runtime"
-	"strconv"
 	"sync"
 	"sync/atomic"
 
@@ -41,7 +40,7 @@ const (
 	unrequested spanState = iota
 	// A span is in Requested state when it's requested from remote but its content hasn't been returned.
 	requested
-	// A span is in Fetched state when its content is fetched from remote.
+	// A span is in Fetched state when its content is fetched from remote and compressed data is cached.
 	fetched
 	// A span is in Uncompressed state when it's uncompressed and its uncompressed content is cached.
 	uncompressed
@@ -52,12 +51,26 @@ const (
 	defaultSpanVerificationFailureRetries = 3
 )
 
-// map of valid span transtions. Key is the current state and value is valid new states.
+// map of valid span transtions: current state -> valid new states.
+// stateTransitionMap is kept minimum so we won't change state by accident.
+// We should keep it documented when each transition will happen.
 var stateTransitionMap = map[spanState][]spanState{
-	unrequested:  {unrequested, requested},
-	requested:    {requested, fetched},
-	fetched:      {fetched, uncompressed},
-	uncompressed: {uncompressed},
+	unrequested: {
+		// when span starts being fetched; it makes other goroutines aware of this
+		requested,
+	},
+	requested: {
+		// when a span fetch fails; change back to unrequested so other goroutines can request again
+		unrequested,
+		// when bg-fetcher fetches and caches compressed span
+		fetched,
+		// when span data request comes; span is fetched, uncompressed and cached
+		uncompressed,
+	},
+	fetched: {
+		// when span data request comes and span is fetched by bg-fetcher; compressed span is available in cache
+		uncompressed,
+	},
 }
 
 // Specific error types raised by SpanManager.
@@ -78,6 +91,11 @@ type span struct {
 	mu                sync.Mutex
 }
 
+func (s *span) checkState(expected spanState) bool {
+	state := s.state.Load().(spanState)
+	return state == expected
+}
+
 func (s *span) setState(state spanState) error {
 	err := s.validateStateTransition(state)
 	if err != nil {
@@ -94,7 +112,7 @@ func (s *span) validateStateTransition(newState spanState) error {
 			return nil
 		}
 	}
-	return errInvalidSpanStateTransition
+	return fmt.Errorf("%w: %v -> %v", errInvalidSpanStateTransition, state, newState)
 }
 
 // SpanManager fetches and caches spans of a given layer.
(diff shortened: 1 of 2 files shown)


fs/span-manager/span_manager.go
```diff
@@ -173,104 +193,46 @@ func (m *SpanManager) buildAllSpans() {
...
 
-	m.addSpanToCache(fmt.Sprintf("%d", spanID), compressedBuf)
-	return s.setState(fetched)
+	_, err := m.fetchAndCacheSpan(spanID, false)
+	return err
 }
 
+// ResolveSpan ensures a span exists uncompressed in cache by calling
+// `resolveSpan`. (no span data is explicitly requested from `ResolveSpan`).
 func (m *SpanManager) ResolveSpan(spanID compression.SpanID) error {
```
djdongjin, line 223: This existing`ResolveSpan` is only used for testing. I didn't delete it but I don't think it's useful.

fs/span-manager/span_manager.go
```diff
@@ -467,32 +444,41 @@ func (m *SpanManager) uncompressSpan(s *span, compressedBuf []byte) ([]byte, err
...
-	err := m.fetchSpan(compressedBuf, spanID)
-	if err != nil && err != io.EOF {
-		return nil, err
-	}
-
-	err = s.setState(fetched)
-	if err != nil {
-		return nil, err
+// addSpanToCache adds contents of the span to the cache.
+func (m *SpanManager) addSpanToCache(spanID compression.SpanID, contents []byte, opts ...cache.Option) {
```
Kern--: This existed before your change, be we just totally swallow all errors related to caching which will definitely break in the future. I think this method should return an error.

fs/span-manager/span_manager.go
```diff
@@ -331,127 +292,143 @@ func (m *SpanManager) getSpanInfo(offsetStart, offsetEnd compression.Offset) *sp
...
+		m.addSpanToCache(s.id, uncompSpanBuf, m.cacheOpt...)
+		if err := s.setState(uncompressed); err != nil {
 			return nil, err
 		}
 		return bytes.NewReader(uncompSpanBuf[offsetStart : offsetStart+size]), nil
 	}
-	return nil, ErrSpanNotAvailable
+
+	// fetch-uncompress-cache span: span state can only be `unrequested` since
+	// no goroutine will release span state lock in `requested` state
```
Kern--, line 352: This is not true. The lock can be released in the `requested` state if requesting the span fails. 

I think this behavior is still correct though - either we're in the `unrequested` state, or we're in the `requested` state, but need to redrive the request.

fs/span-manager/span_manager.go
djdongjin: yeah agree.

fs/span-manager/span_manager.go
djdongjin: > This is not true. The lock can be released in the requested state if requesting the span fails.

I think the correct behavior should be: if in `requested` state, the lock has to be held by the requesting goroutine; if the goroutine failed to request the span, it should revert the state to `unrequested` (no one is requesting the span now). (the 2nd half is not implemented yet in this PR).

Otherwise, what another goroutine should do if it acquires the lock and the span in `requested` (which means another goroutine is requesting).

> either we're in the unrequested state, or we're in the requested state, but need to redrive the request.

yeah if a goroutine `requested` the span and failed, it should change state back to `unrequested` before releasing the lock.

fs/span-manager/span_manager.go
djdongjin: updated

fs/span-manager/span_manager.go
djdongjin: added the `requested -> unrequested` logic when span fetching fails.

fs/span-manager/span_manager.go
```diff
@@ -304,52 +295,34 @@ func (m *SpanManager) getSpanInfo(offsetStart, offsetEnd compression.Offset) *sp
...
+// which is returned as an `io.Reader`.
+//
+// It resolves the span to ensure it exists and is uncompressed in cache:
+//  1. For `uncompressed` span, directly return the reader from the cache.
+//  2. For `fetched` span, read and uncompress the compressed span from cache, cache and
+//     return the reader from the uncompressed span.
+//  3. For `unrequested` span, fetch-uncompress-cache the span data, return the reader
+//     from the uncompressed span
+//  4. No span state lock will be acquired in `requested` state.
 func (m *SpanManager) GetSpanContent(spanID compression.SpanID, offsetStart, offsetEnd compression.Offset) (io.Reader, error) {
```
sbuckfelder, line 308: Can we make `GetSpanContent` a private function?  Shouldn't every call be going through GetContents? Feels a bit odd that someone outside of the span manager would know the details of a span. I don't think this function is used anywhere other then in GetContents (which is used in fs/fs/reader.go).

sbuckfelder:
Overall I'm happy with this PR.  I do think that GetSpanContents should be a private func.

Longer term I am interested in whether we can move this more into an event driven system with each of the steps being controlled by a private function, which also handle moving state, and the locking.

sbuckfelder (dismissed):

fs/span-manager/span_manager.go
sparr: I'm not a fan of an exported function that's only used for testing.

fs/span-manager/span_manager.go
djdongjin: yeah agree. Updated `GetSpanContent` to private method

fs/span-manager/span_manager.go
djdongjin: updated to private func

sbuckfelder (approved):
LGTM

fs/span-manager/span_manager.go
```diff
@@ -465,15 +461,22 @@ func (m *SpanManager) uncompressSpan(s *span, compressedBuf []byte) ([]byte, err
...
-		w.Close()
+// A non-nil error is returned if the data is not written to the cache.
+func (m *SpanManager) addSpanToCache(spanID compression.SpanID, contents []byte, opts ...cache.Option) error {
+	w, err := m.cache.Add(fmt.Sprintf("%d", spanID), opts...)
+	if err != nil {
+		return err
 	}
+	defer w.Close()
+
+	_, err = w.Write(contents)
```
Kern--, line 472: Should we be checking the number of bytes written here like in the previous version?

Kern-- (commented):
Overall, I think this is good directionally. 

I think we might want to explore what the exported API should be for this as a follow up. It seems a bit weird to me that the two things you can do with a spanmanger are:
1) Get a range of data from the uncompressed stream
2) Ensure that the SpanManager's internal cache has a specific span

The abstractions there feel a bit off.

fs/span-manager/span_manager.go
djdongjin: From the [`io.Writer` interface doc](https://pkg.go.dev/io#Writer):

"It returns the number of bytes written from p (0 <= n <= len(p)) and any error encountered that caused the write to stop early. **Write must return a non-nil error if it returns n < len(p)**".

djdongjin:
> It seems a bit weird to me that the two things you can do with a spanmanger are:
> 
> 1. Get a range of data from the uncompressed stream
> 2. Ensure that the SpanManager's internal cache has a specific span
> 
> The abstractions there feel a bit off.

To me the only functionality that span manager provides to outside is `1. Get a range of data from the uncompressed stream` (without knowing anything about the `span` concept and its state, `GetContents`). 

`2.`/`FetchSingleSpan` is really only to enable bg-fetcher to fetch/cache a specific span. And they're in different packages, so `FetchSingleSpan` is exported.

Probably `backgroudfetcher` should be just part of `span-manager` package, since they're tightly related and both work on the concept of `span`.

Kern-- (approved):

PR #449, opened 2023-02-27
Add guides/conventions/etc to contributing doc

sparr:
*Description of changes:*
Add a section to CONTRIBUTING.md describing some of our CI-enforced coding standards, the coding/style guides we lean toward, and starting a list of our project-specific standards with duration data type norms discussed recently.

Also minor spacing/formatting changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,18 +1,15 @@
 # Contributing Guidelines
 
-Thank you for your interest in contributing to our project. Whether it's a bug report, new feature, correction, or additional
-documentation, we greatly value feedback and contributions from our community.
+Thank you for your interest in contributing to our project. Whether it's a bug report, new feature, correction, or additional documentation, we greatly value feedback and contributions from our community.
 
-Please read through this document before submitting any issues or pull requests to ensure we have all the necessary
-information to effectively respond to your bug report or contribution.
+Please read through this document before submitting any issues or pull requests to ensure we have all the necessary information to effectively respond to your bug report or contribution.
 
 
 ## Reporting Bugs/Feature Requests
 
 We welcome you to use the GitHub issue tracker to report bugs or suggest features.
 
-When filing an issue, please check existing open, or recently closed, issues to make sure somebody else hasn't already
-reported the issue. Please try to include as much information as you can. Details like these are incredibly useful:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -36,21 +34,43 @@ To send us a pull request, please:
...
+GitHub provides additional document on [forking a repository](https://help.github.com/articles/fork-a-repo/) and [creating a pull request](https://help.github.com/articles/creating-a-pull-request/).
+
+
+## Coding conventions and Style guides
+
+In addition to pursuing code correctness, please also keep the following standards, guides, norms, and conventions in mind when drafting contributions.
+
+### Enforced via CI
+
+* [Staticcheck](https://staticcheck.io/)
```
djdongjin, line 46: By default golangci staticcheck seems only check SAxxx checks:

https://golangci-lint.run/usage/linters/#staticcheck

We should mention this or use the golangci link above.

djdongjin: maybe mention `make check` will help you ensure your changes pass these checks including linting, DCO, ltag etc.

sparr: Source for only checking SAxxx checks? Link and https://staticcheck.io/docs/configuration/options/#checks suggest to me that it's all checks minus six specific ST checks. Also, we seem to actually want those excluded checks, so we should probably set this, but that's another issue.

djdongjin: I got this from golangci's staticcheck config: https://golangci-lint.run/usage/linters/#staticcheck

```
    # SAxxxx checks in https://staticcheck.io/docs/configuration/options/#checks
    # Default: ["*"]
    checks: ["all"]
```

The comment mentioned `SAxxxx` checks, but the default value is `all` ... not sure which one is more accurate.

Also, for this staticcheck error, curious how you got this error? (e.g. via golangci run/make check, or others). Since it was not discovered by golangci in previous PRs. https://github.com/awslabs/soci-snapshotter/pull/442#discussion_r1113314577

sparr: I think my local staticcheck might be running `["*"]` by default. I agree that the docs aren't very consistent, but you're right that my previous experience indicates that it must not be running every check since that thing wasn't caught. Also, other docs say that the statichcheck default is `["all", "-ST1..."]` which means that "all" would have to include the ST checks to make that syntax valid.

I'll add some notes to the doc and hopefully we won't get stuck here on diving into that particular bit.

djdongjin (approved):

Kern-- (approved):

Issue #443, opened 2023-02-21
[FEATURE] Make duration types consistent


sparr:
**Is your feature request related to a problem? Please describe.**
Various parts of the codebase use int or int64 to store time durations with specific units, rather than as a `time.Duration`, usually for interoperability with libraries or config files that need to be human readable.

**Describe the solution you'd like**
All places the code and configs store a duration as a number rather than a `time.Duration` use the same type for a given time unit (milliseconds, seconds, etc) where possible. Probably int64 for all of them because that appears to be the trend in more recently touched parts of the codebase, but smaller types for larger units (e.g. `int` or `int32` for seconds or minutes) could be sufficient.

**Describe alternatives you've considered**
Continuing to perform explicit casts when manipulating durations from different parts of the codebase has worked so far, but has costs in development time and mistake risk.

**Additional context**
Partial sample of current counterexamples:
`BlobConfig.*WaitMsec` are `int` when most other millisecond values are `int64`
`FuseConfig.*Timeout` are `int` when most other second values are `int64`
An audit of all such variables/members will be necessary to find other examples.

`latencyBuckets*seconds` are `[]float64` but need to stay that way for compatibility with prometheus

sparr:
Closed by #450

PR #450, opened 2023-02-27
Update duration types to int64

sparr:
*Issue #, if available:*
#443 

*Description of changes:*
Change members, variables, and consts that store time durations to type `int64`. Add necessary explicit type conversions.

Also remove a few magic numbers from time type conversions in favor of constants from the `time` package.

Also make an existing `plus one year` calculation more accurate.

*Testing performed:*
`make integration`

### DIFF
--- a/service/resolver/registry.go
+++ b/service/resolver/registry.go
@@ -41,7 +41,7 @@ import (
 	rhttp "github.com/hashicorp/go-retryablehttp"
 )
 
-const defaultRequestTimeoutSec = 30
+const defaultRequestTimeoutSec int64 = 30
 
 // Config is config for resolving registries.
 type Config struct {
@@ -63,7 +63,7 @@ type MirrorConfig struct {
 	// RequestTimeoutSec is timeout seconds of each request to the registry.
 	// RequestTimeoutSec == 0 indicates the default timeout (defaultRequestTimeoutSec).
 	// RequestTimeoutSec < 0 indicates no timeout.
-	RequestTimeoutSec int `toml:"request_timeout_sec"`
+	RequestTimeoutSec int64 `toml:"request_timeout_sec"`
 }
 
 type Credential func(string, reference.Spec) (string, string, error)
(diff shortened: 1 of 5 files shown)


service/resolver/registry.go
```diff
@@ -80,7 +80,7 @@ func RegistryHostsFromConfig(cfg Config, credsFuncs ...Credential) source.Regist
 			tr := client.StandardClient()
 			if h.RequestTimeoutSec >= 0 {
 				if h.RequestTimeoutSec == 0 {
-					tr.Timeout = defaultRequestTimeoutSec * time.Second
+					tr.Timeout = time.Duration(defaultRequestTimeoutSec) * time.Second
```
djdongjin, line 83: Based on the [Duration](https://pkg.go.dev/time#Duration) doc, it's a nanosecond, so this (`time.Duration(defaultRequestTimeoutSec)`) will change the meaning of 30 second to 30 nanosecond right?

djdongjin: based on how [go produces those time const](https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/time/time.go;l=631) this conversion is not needed. (int * time.Duration is more clear than time.Duration * time.Duration)

sparr: See a few lines below for an example of this pattern.
Yes, `time.Duration(someSeconds)` does temporarily create a duration in nanoseconds, but the immediately following `* time.Second` corrects this.
There are other examples of this pattern at https://github.com/awslabs/soci-snapshotter/blob/f657479e6fdfae8a6a0516e6ba915e6e2dd2e807/fs/fs.go#L152-L162 and https://github.com/awslabs/soci-snapshotter/blob/f657479e6fdfae8a6a0516e6ba915e6e2dd2e807/fs/remote/resolver.go#L140-L141

djdongjin: not aware of `int64 * time.Duration` is invalid... then time.Duration * time.Duration might be the only consistent way of doing that.

Also notice go itself uses both...: https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/time/time.go;l=627-638

djdongjin (approved):

turan18 (approved):

Issue #404, opened 2023-02-08
[BUG]Duplicated shell log in integration test


djdongjin:
**Describe the bug**

The `shell` testutil log everything twice

https://github.com/awslabs/soci-snapshotter/blob/5ba9f953cf3d785f8d430837d9a6f3d9dfe675e2/util/testutil/shell.go#L67-L77

**Steps To Reproduce**

`make integraion` and see logs from shell. Every log from shell is duplicated twice.

```shell
testing: 2023/02/07 16:36:27 >>> Running: [find /var/lib/containerd//. ! -name . -prune -exec rm -rf {} +]
    shell.go:76: >>> Running: [find /var/lib/containerd//. ! -name . -prune -exec rm -rf {} +]
testing: 2023/02/07 16:36:27 >>> Running: [find /var/lib/soci-snapshotter-grpc//. ! -name . -prune -exec rm -rf {} +]
    shell.go:76: >>> Running: [find /var/lib/soci-snapshotter-grpc//. ! -name . -prune -exec rm -rf {} +]
testing: 2023/02/07 16:36:27 >>> Running(returning reader): [/usr/local/bin/soci-snapshotter-grpc --log-level debug --address /run/soci-snapshotter-grpc/soci-snapshotter-grpc.sock]
    shell.go:76: >>> Running(returning reader): [/usr/local/bin/soci-snapshotter-grpc --log-level debug --address /run/soci-snapshotter-grpc/soci-snapshotter-grpc.sock]
testing: 2023/02/07 16:36:27 >>> Running(0/100): [ctr snapshots --snapshotter soci prepare connectiontest-dummy-cfh7s2qltg8o7svcmh80 ]
    shell.go:76: >>> Running(0/100): [ctr snapshots --snapshotter soci prepare connectiontest-dummy-cfh7s2qltg8o7svcmh80 ]
testing: 2023/02/07 16:36:27 >>> Running: [containerd --log-level debug]
    shell.go:76: >>> Running: [containerd --log-level debug]
```

**Expected behavior**

Logs are not duplicated. 

Ideally, we should log where the command is called, instead of where the log is produced (e.g. all the `shell.go:76` above). This helps target errors. Otherwise, usually we need to scroll up and see which test is running and causing the error.

sondavidb:
Closed by #1192

PR #468, opened 2023-03-06
Make containerd/soci log level configurable

### LINKED ISSUE #426: [FEATURE] Make integration test log level configurable
Integration test output is unnecessarily verbose during typical runs.

Provide a mechanism for specifying a default log level (error, warning, info, debug) that will filter test output (e.g. omitting `log.G().Debug()`) and also be passed to processes that are launched via tests (e.g. `soci -log-level`, `soci-snapshotter-grpc --log-level`, `containerd --log-level`, containerd config debug.level).

Manual filtering of test output is time consuming.

djdongjin:
**Issue #, if available:** part of #426

**Description of changes:**

This PR makes the following changes:

1. containerd (config and `containerd --log-level`) log level configurable and `warn` by default.
2. soci-snapshotter (`soci-snapshotter-grpc --log-level`) log level configurable and `debug` by default (no change).
3. add `-q` to `docker(-compose) build` cmd, reduce build output: build produces a lot of output but I didn't find it useful in almost all cases.
4. rename integraion testutils to `util_test.go` and `util_soci_test.go`.

This PR reduces integration test log from 22186 lines to around 10000 lines. (if change soci loglevel to warn, to 7500 lines)

I didn't make a global log level config because I feel we want it different in different components.

**Testing performed:**


sparr:
While this is a good start, I don't think this should close #426, that needs a lot more coverage.

djdongjin:
> While this is a good start, I don't think this should close https://github.com/awslabs/soci-snapshotter/issues/426, that needs a lot more coverage.

The new revision made soci log level configurable (still debug by default). (tested changing to `warn` locally, the output line reduces to 7320.)

The next thing is `logrus.SetLevel()` but checking the output, there are only a couple of lines with `"level":xxxx` after changes made to containerd/soci. So I don't feel that helps much reduce integration output.

Most of the existing 7320 lines come from **output** not **log** (i.e. `fmt.Printf`/etc). They could possibly be minimized either by caller (e.g., our inte tests figure out a way of ignoring cmd exection output) or the callee (e.g., use the `-q` flag in nerdctl/docker), but there isn't a unified config/way of reducing that.

There are some duplication tracked in #404. Fixing that can reduce another ~1500 lines. (The rest are output).

Any other coverage you have in mind?

integration/main_test.go
```diff
@@ -42,7 +42,26 @@ import (
...
+// this can be overwritten by setting up env variables specified by
+// `containerdLogLevelEnv`/`sociLogLevelEnv`.
+var (
+	containerdLogLevel = "warn"
+	sociLogLevel       = "debug"
+)
+
+func validateLogLevel(level string) bool {
+	switch level {
+	case "trace", "debug", "info", "warn", "error", "fatal", "panic":
```
Kern--, line 60: nit: What do you think about replacing this with a logrus check? 
```
_, err := logrus.ParseLevel(level);
return err == nil;
```

Kern-- (approved):

integration/main_test.go
djdongjin: sure, not aware of this :) got the list of value from cli help output 😅

Kern-- (approved):

turan18 (approved):

Issue #477, opened 2023-03-07
[FEATURE] Add flag to print `soci` version


Kern--:
### Description

`soci-snapshotter-grpc --version` prints the snapshotter's version like:

```
$ soci-snapshotter-grpc --version
soci-snapshotter-grpc v0.1.0 3b54ae61448d9a7ffe3624edae926a97f1e553e4
```

`soci` has no equivalent functionality. Adding this would make it easier to track and fix bugs that get reported against the CLI.

### Describe the solution you'd like

A similar command for `soci`

```
$ soci --version
soci v0.1.0 3b54ae61448d9a7ffe3624edae926a97f1e553e4
```

PR #478, opened 2023-03-07
Add `soci --version` to get the CLI version

Kern--:
**Issue #, if available:**
Closes https://github.com/awslabs/soci-snapshotter/issues/477

**Description of changes:**
This change adds a `soci --version` flag to match `soci-snapshotter-grpc --version`.

Since urfave has a specific output format, soci-snapshotter-grpc was updated to match it. The new version format is:

$BINARY_NAME version $TAG $COMMIT

**Testing performed:**
```
$ ./out/soci --version
soci version 303d3db 303d3dbc7ea16376d77ba5279426b9139a147e0b
$ ./out/soci-snapshotter-grpc --version
soci-snapshotter-grpc version 303d3db 303d3dbc7ea16376d77ba5279426b9139a147e0b
```

### DIFF
--- a/cmd/soci-snapshotter-grpc/main.go
+++ b/cmd/soci-snapshotter-grpc/main.go
@@ -119,7 +119,7 @@ func main() {
 		log.L.WithError(err).Fatal("failed to prepare logger")
 	}
 	if *printVersion {
-		fmt.Println("soci-snapshotter-grpc", version.Version, version.Revision)
+		fmt.Println("soci-snapshotter-grpc version", version.Version, version.Revision)
 		return
 	}
 	logrus.SetLevel(lvl)
--- a/cmd/soci/main.go
+++ b/cmd/soci/main.go
@@ -40,6 +40,7 @@ import (
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/image"
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/index"
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/ztoc"
+	"github.com/awslabs/soci-snapshotter/version"
 	"github.com/containerd/containerd/cmd/ctr/commands/run"
 	"github.com/containerd/containerd/defaults"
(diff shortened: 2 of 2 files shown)


sparr:
Can you test/confirm this works with a tag, such as a release version?

sbuckfelder (approved):
lgtm

djdongjin (approved):
LGTM. maybe later we want to have separate server(daemon)/client(soci) versions, similar to containerd.

turan18 (approved):

Kern--:
@sparr as an example:

```
$ git tag v0.1.1
$ make
$ ./out/soci --version
soci version v0.1.1 303d3dbc7ea16376d77ba5279426b9139a147e0b
$ ./out/soci-snapshotter-grpc --version
soci-snapshotter-grpc version v0.1.1 303d3dbc7ea16376d77ba5279426b9139a147e0b
$ git tag -d v0.1.1
```

Kern--:
> maybe later we want to have separate server(daemon)/client(soci) versions, similar to containerd.

This could make sense, but you don't actually need to run the SOCI snapshotter in order to create SOCI indices and currently the CLI doesn't talk directly to the snapshotter. Maybe we can do that opportunistically if the snapshotter is available. I'll create an issue.

djdongjin:
> This could make sense, but you don't actually need to run the SOCI snapshotter in order to create SOCI indices and currently the CLI doesn't talk directly to the snapshotter. Maybe we can do that opportunistically if the snapshotter is available. I'll create an issue.

It's just for convenience and makes debug easier. Otherwise if a user has issue, we'll ask "can you provide output of `soci --version` AND `soci-snapshotter-grpc --version`.

 We can make the cli to ignore snapshotter version if it's not running, and print if it's. An easy way might be just something like `os.Exec("soci-snapshotter-grpc", "--version")` (and ignore err if it doesn't exist).

As a smilar example, `kubectl version` just prints `server: error ....` when there isn't a k8s it can talk to. (I didn't try ctr)

Issue #458, opened 2023-03-01
[Bug] Clean up labels used by SOCI and switch to containerd's


djdongjin:
### Description

SOCI defines a list of labels it uses. However, these labels have the following issues:

1. Located in different places: [config.go](https://github.com/awslabs/soci-snapshotter/blob/aed937970e4b35dac5bf65fd38ad4bc6208c0ef6/fs/config/config.go#L42-L46), [remote.go](https://github.com/awslabs/soci-snapshotter/blob/56534c70965d326ed9963f9a0fa5f0ecba9d3bb9/fs/source/source.go#L79).
2. Labels that are not used at all or deprecated: [`config.TargetImageRefLabel`](https://github.com/awslabs/soci-snapshotter/blob/aed937970e4b35dac5bf65fd38ad4bc6208c0ef6/fs/config/config.go#L43)
3. Labels that mean the same thing but have different keys: [`SociIndexDigestLabel`](https://github.com/awslabs/soci-snapshotter/blob/aed937970e4b35dac5bf65fd38ad4bc6208c0ef6/fs/config/config.go#L46), [`SociIndexDigestLabel`](https://github.com/awslabs/soci-snapshotter/blob/aed937970e4b35dac5bf65fd38ad4bc6208c0ef6/fs/source/source.go#L108).
4. Labels we take from stargz but we didn't really use: [`targetUrls`](https://github.com/awslabs/soci-snapshotter/blob/aed937970e4b35dac5bf65fd38ad4bc6208c0ef6/fs/source/source.go#L97-L103); based on the discussion, this label is used for ipfs only: https://github.com/containerd/nerdctl/pull/2063#discussion_r1121746511

Another issue is that we probably should remove the mention of `com.amazon`? (I'm not sure about this)

https://github.com/awslabs/soci-snapshotter/blob/59b35d4817a4faa82f43bda27a81e4e0c3665daf/soci/soci_index.go#L49-L53

### Steps to reproduce the bug

See the code link above

### Describe the results you expected

1. Resolve the itemized issues.
2. Change to use containerd's label definition for remote snapshotter: https://github.com/containerd/containerd/blob/main/pkg/snapshotters/annotations.go

A similar effort for 2 for stargz can be found here: https://github.com/containerd/stargz-snapshotter/pull/1131

### Host information

1. OS: 
2. Snapshotter Version: 
3. Containerd Version:

Kern--:
[...]

PR #483, opened 2023-03-09
Using containerds default snapshotter labels instead of custom ones

turan18:
Using containerds default snapshotter labels instead of custom ones.
Removing unused labels in config.go 
https://github.com/awslabs/soci-snapshotter/blob/aed937970e4b35dac5bf65fd38ad4bc6208c0ef6/fs/config/config.go#L42-L46

This also changes the snapshotter to use the [FromDefaultLabels()](https://github.com/awslabs/soci-snapshotter/blob/6a716b7f8180f02ae5c03e4350c4f091ad9267ea/fs/source/source.go#L113)  as the source function solely instead of [sourceFromCRILabels()](https://github.com/awslabs/soci-snapshotter/blob/6a716b7f8180f02ae5c03e4350c4f091ad9267ea/service/cri.go#L57). Currently there is no support for layer size labels within the snapshotters package in containerd. This is important since we need a layers size when creating the `io.SectionReader` for the layer in the layer resolver. In order to have kubernetes/CRI support in the future we must either push for layer size label support upstream or re-implement https://github.com/awslabs/soci-snapshotter/pull/234  which will allow us to obtain layer sizes via a network call. 

**Issue #, if available:**

Fixes: #458 

**Testing performed:**


benchmark/stargz_utils.go
```diff
@@ -135,8 +135,7 @@ func (proc *StargzContainerdProcess) StargzRpullImageFromECR(
 		containerd.WithSchema1Conversion,
 		containerd.WithPullUnpack,
 		containerd.WithPullSnapshotter("stargz"),
-		containerd.WithImageHandlerWrapper(source.AppendDefaultLabelsHandlerWrapper(
-			imageRef, "")),
+		containerd.WithImageHandlerWrapper(ctdsnapshotters.AppendInfoHandlerWrapper(imageRef)),
```
djdongjin, line 138: not related to the change. but seems previously we use `source` from soci, not from stargz. Are they the same?

fs/fs.go
```diff
@@ -373,7 +374,7 @@ type filesystem struct {
 }
 
 func (fs *filesystem) MountLocal(ctx context.Context, mountpoint string, labels map[string]string, mounts []mount.Mount) error {
-	imageRef, ok := labels[source.TargetRefLabel]
+	imageRef, ok := labels[ctdsnapshotters.TargetRefLabel]
```
djdongjin, line 377: this will be a behavior change right? since `source.TargetRefLabel` and `ctdsnapshotters.TargetRefLabel` are different strings?

Kern--:
LGTM, just need to address the check failures.

service/service.go
```diff
@@ -96,11 +96,15 @@ func NewSociSnapshotterService(ctx context.Context, root string, config *Config,
...
 	}
+	// The filesytems GetSources function will default to FromDefaultLabels for now,
+	// since there are no size labels for layers within the snapshotters package in
+	// containerd. In order to support kubernetes/CRI we will have to either add
+	// upstream support for layer size labels or make network calls within the
+	// snapshotter to fetch layer size.
 	// Configure filesystem and snapshotter
-	fsOpts := append(sOpts.fsOpts, socifs.WithGetSources(sources(
-		sourceFromCRILabels(hosts),      // provides source info based on CRI labels
+	fsOpts := append(sOpts.fsOpts, socifs.WithGetSources(
```
djdongjin, line 105: Cannot put the comment in func definition

I think we can remove this func (`source`) for now since it's unused and a trivial function, and I feel it's right location should be in `fs/source/source.go`. If we need it later, we can add it there.

djdongjin (approved):
LGTM

benchmark/stargz_utils.go
turan18: Yeah its effectively the same since we  are not setting the prefetch size

Kern-- (approved):

Issue #463, opened 2023-03-03
[FEATURE] Improve the abstraction around ztoc


djdongjin:
### Description

The current structure of ztoc&compression has some issues (see discussion in https://github.com/awslabs/soci-snapshotter/pull/453#discussion_r1122174133):

1. Zinfo is part of ztoc, compression is used to implement zinfo. Thus, compression pkg should be a sub-package of ztoc package.
2. Also, zinfo should be always part of a ztoc object. As an example, spanmanager [holds both ztoc and zinfo objects](https://github.com/awslabs/soci-snapshotter/blob/00474ecf61cfb905a0d6436b516110050bb6435c/fs/span-manager/span_manager.go#L122-L125).
3. Ztoc struct should have the zinfo **struct**, not its [serilized **bytes**](https://github.com/awslabs/soci-snapshotter/blob/878e7d9fbdb67fd030f8ac1b36777a83cd525aea/ztoc/ztoc.go#L59). We should only serilize zinfo when we serilize the whole ztoc.
4. Ztoc/Zinfo should have the compression algorithm used to produce that ztoc/zinfo. This info is already in [ztoc's flatbuf](https://github.com/awslabs/soci-snapshotter/blob/878e7d9fbdb67fd030f8ac1b36777a83cd525aea/ztoc/fbs/ztoc.fbs#LL31C8-L31C8)

### Describe the solution you'd like

Resolve the above issues (or more if found).

### Describe any alternative solutions/features you've considered

Don't change anything.

Kern--:
Copying this discussion out of the PR so we can continue it here.

---
> What direction are we going with this API?
> 
> Are we going to keep embedding `CompressionInfo` into `Ztoc` with this helper? Should we instead embed `ZInfo` into `Ztoc` and just use the ztoc directly instead of `ztoc.Zinfo()`?

I tried embedding `Zinfo` into `Ztoc` but it's a little tricky due to the gzipzinfo c resource release. In some places it will cause the Zinfo creation and release done in different places in some scenarios. https://github.com/awslabs/soci-snapshotter/issues/463#issuecomment-1478335134 has more details.

After digging into the issue, I think both ways have their pros and cons:

Keep zinfo struct in ztoc:

pros:

- simplify zinfo usage: just `ztoc.Zinfo`.

cons:

- whenever a ztoc is created, its zinfo needs to be closed to release the c resource (ideally by the same creator).
[...]

PR #484, opened 2023-03-09
Refactor ztoc

djdongjin:
**Issue #, if available:** part of #463

**Description of changes:**

1. Move compression to ztoc as its sub package: compression/zinfo is part of ztoc, and from our discussion we don't want users to use zinfo separately.
2. Add compression algorithm to our ztoc struct: it is part of ztoc flatbuf, but not our ztoc struct.
3. Change `Ztoc.Compression` and `Ztoc.TOC` to inline sub struct. I don't feel `TOC TOC` provides any other benefit, and instead make it confusing with the field and type being the same same, and making our code verbose (e.g., `spanmanager.ztoc.CompressionInfo.Checkpoints`)

> 3 might be my opinoned preference.

**Testing performed:**

Make check && make test && make integration


sparr:
We're passing around the compression algo as a string in various places here. Not all of it is new code, but it jumped out at me. The fb definition uses a byte (where Gzip == `1`) to store that info. Can you give some insight on that?

ztoc/ztoc.go
```diff
@@ -35,20 +35,22 @@ import (
...
 // CompressionInfo is the "zinfo" part of ztoc including the `Checkpoints` data
 // and other metadata such as all span digests.
 type CompressionInfo struct {
-	MaxSpanID   compression.SpanID //The total number of spans in Ztoc - 1
-	SpanDigests []digest.Digest
-	Checkpoints []byte
+	MaxSpanID            compression.SpanID //The total number of spans in Ztoc - 1
+	SpanDigests          []digest.Digest
+	Checkpoints          []byte
+	CompressionAlgorithm string
```
sparr, line 53: Why a string instead of the byte that's used to store this in the fb?

djdongjin: containerd returns compression algorithms as plain string ("gzip", "zstd"), so we just use the same (string) to keep it consistent with containerd and I didn't feel extra benefit of defining a new type for that ourselves (e.g. `type CompressionAlgorithm string`).

Also fb is only used to (de)serilize `Ztoc` struct from/to flatbuf (and has its own [Compression type definition](https://github.com/awslabs/soci-snapshotter/blob/main/ztoc/fbs/ztoc/CompressionAlgorithm.go)), but Ztoc is used in many places of soci. I don't think we should use fb-specific types within Ztoc. So the conversion is made only in ztoc_marshaler.go.

ztoc/ztoc_marshaler.go
```diff
@@ -132,30 +137,32 @@ func ztocToFlatbuffer(ztoc *Ztoc) (fb []byte, err error) {
 		}
 	}()
 
+	// ztoc - metadata
```
Kern--, line 140: I'm mostly onboard with these changes, but we're referring the the ztoc version as `ztoc - metadata` but `ztoc.Metadata` is actually FileMetadata, not metadata about the ztoc itself. Does it make sense to rename `TOC.Metadata` to `TOC.FileMetadata`?

ztoc/ztoc_marshaler.go
```diff
@@ -238,3 +256,16 @@ func prepareXattrsOffset(me FileMetadata, builder *flatbuffers.Builder) flatbuff
 	xattrs := builder.EndVector(len(me.Xattrs))
 	return xattrs
 }
+
+// compressionAlgorithmToFlatbuf helps convert compression algorithm into flatbuf
+// enum. SOCI/containerd uses lower-case for compression, but our flatbuf capitalizes
+// the first letter. When converting back, we can just `strings.ToLower` so a helper
+// func is not needed in that case.
+func compressionAlgorithmToFlatbuf(algo string) (ztoc_flatbuffers.CompressionAlgorithm, error) {
```
Kern--, line 264: If we stored `algo` as a byte, wouldn't the flatbuffer serialization logic handle this for us? i.e. if you tried to write a value that wasn't present, then it would cause a serialization error.

ztoc/ztoc_marshaler.go
djdongjin: Changed to `FileMetadatas` since it's a slice and also avoid to having the field and type having the same name :)

ztoc/ztoc_marshaler.go
djdongjin: I'm not sure how will that work (also by `byte`, do you mean something like `[]byte("gzip")`?). [flatbuf generates its enum as int](https://github.com/awslabs/soci-snapshotter/blob/main/ztoc/fbs/ztoc/CompressionAlgorithm.go).

Regardless, I feel a conversion is unavoidable and it's more convenient to keep it consistent with containerd (which returns "gzip"/"zstd"). If we use flatbuffer's byte/int, we still need to convert that to string or vice versa when we need to compare.

ztoc/ztoc_marshaler.go
djdongjin: just FYI, this does change Ztoc's API in our library, whereas the previous change won't (both `ztoc.TOC.Metadata` and `ztoc.Metadata` work).

It's okay as long as no one uses soci as a dependency or uses `Ztoc.TOC.Metadata` directly.

Kern--: Is there a concern with the name and type being the same? This is a pretty common pattern in go.

djdongjin: No concern, it's just this is a slice of `FileMetadata`, so I changed to plural `FileMetadatas`

ztoc/ztoc_marshaler.go
Kern--: We discussed offline and decided that this is convenient for now. In the future, we may decide that compression algorithms are actually died to the specific implementation of the algorithm (e.g. a C implementation and a go implementation would have different `CompressionAlgorithm` fields in the ztoc) at which point we can consider representing this differently.

ztoc/ztoc_marshaler.go
djdongjin: changed to `FileMetadata`

turan18 (approved):

Kern-- (approved):

PR #485, opened 2023-03-09
Add ztoc v0.9 tests

Kern--:
**Issue #, if available:**
Related to https://github.com/awslabs/soci-snapshotter/pull/465

**Description of changes:**
We currently have a test that ztoc v0.1 is consistently generated, but we don't have an equivalent test for v0.9 which is what we are currently producing. This also gives us more confidence that the changes in https://github.com/awslabs/soci-snapshotter/pull/465 aren't breaking anything.

**Testing performed:**
`make test`

### DIFF
--- a/ztoc/ztoc_test.go
+++ b/ztoc/ztoc_test.go
@@ -615,7 +615,7 @@ func TestZtocSerialization(t *testing.T) {
 func TestWriteZtoc(t *testing.T) {
 	testCases := []struct {
 		name                    string
-		version                 string
+		version                 Version
 		checkpoints             []byte
 		metadata                []FileMetadata
 		compressedArchiveSize   compression.Offset
@@ -626,15 +626,15 @@ func TestWriteZtoc(t *testing.T) {
 		expSize                 int64
 	}{
 		{
-			name:                    "success write succeeds - same digest and size",
-			version:                 "0.1",
+			name:                    "success write succeeds - same digest and size " + string(Version09),
+			version:                 Version09,
 			checkpoints:             make([]byte, 1<<16),
 			metadata:                make([]FileMetadata, 2),
 			compressedArchiveSize:   2000000,
 			uncompressedArchiveSize: 2500000,
 			maxSpanID:               3,
 			buildTool:               "AWS SOCI CLI",
-			expDigest:               "sha256:9c8effca78ecc82fad49a4d591dd81615555b05eaaae605a621c927ecf6fc1e7",
+			expDigest:               "sha256:eba28fdf50b1b57543f57dd051b2468c1d4f57b64d2006c75aa4de1d03e6c7ec",
 			expSize:                 65928,
 		},
 	}
--- a/cmd/soci/commands/ztoc/info.go
+++ b/cmd/soci/commands/ztoc/info.go
@@ -93,7 +93,7 @@ var infoCommand = cli.Command{
 
 		multiSpanFiles := 0
 		zinfo := Info{
-			Version:   ztoc.Version,
+			Version:   string(ztoc.Version),
 			BuildTool: ztoc.BuildToolIdentifier,
 			Size:      entry.Size,
 			SpanSize:  gzInfo.SpanSize(),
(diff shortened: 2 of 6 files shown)


ztoc/ztoc_test.go
```diff
@@ -637,6 +637,18 @@ func TestWriteZtoc(t *testing.T) {
...
+		{
+			name:                    "success write succeeds - same digest and size v0.9",
+			version:                 "0.9",
+			checkpoints:             make([]byte, 1<<16),
+			metadata:                make([]FileMetadata, 2),
+			compressedArchiveSize:   2000000,
+			uncompressedArchiveSize: 2500000,
+			maxSpanID:               3,
+			buildTool:               "AWS SOCI CLI",
+			expDigest:               "sha256:eba28fdf50b1b57543f57dd051b2468c1d4f57b64d2006c75aa4de1d03e6c7ec",
```
djdongjin, line 649: what's the difference between 0.1 and 0.9? Also how is the `expDigest` generated

Kern--: The difference is just the version string, but that does change the ztoc content and therefore the ztoc digest. I think we could argue that v0.1 isn't valuable, but we do have history of v0.1 in the repo so I thought it might be good to show that every version of ztoc still works.

The digests were generated by running the test and copying the actual digest from the failure message into the expected digest. We have other tests that verify the content is correct, so this is purely a regression test to verify that any changes in ztoc serialization don't break old formats.

djdongjin: Will `Marshal(ztoc)` handle 0.1 or 0.9 differently? If not, I don't feel keep a 0.1 and a 0.9 test here makes much sense.

Actually, if version 0.1 and 0.9 are the same (i.e. except the actual version string we wrote here), why do we need to rename 0.1 as 0.9? It sounds like we bump up a version without any actual change.

Kern--: Looking back at the history, it looks like the "0.1" -> "0.9" was when we migrated from gob to flatbuffers.
https://github.com/awslabs/soci-snapshotter/commit/b32d03ad7406e66acfbae3d1e6bf15863283a18d#diff-2091623917b9834123dba5876c58b90d763085421b421d0904487f73cbbe7a48

This test should have been using "0.9" all along. I'll remove the "0.1" test.

djdongjin: Can we make the `0.9` a const in the ztoc package? I feel that makes the version more visible & reduces potential errors.

djdongjin: Also, we have gzip_zinfo versions (https://github.com/awslabs/soci-snapshotter/blob/3b2cc11be4627b097f4b225bd105100f5f7957d1/compression/gzip_zinfo.h#L56-L57), and in the future maybe zstd_zinfo version(s).

How do we plan to track these versions altogether? a zinfo version change will possibly change zinfo bytes, which changes ztoc.

djdongjin (dismissed):

ztoc/ztoc_test.go
turan18: Thats a good point. I don't believe changing the zinfo version would inherently change the ztoc. As long as the generated index/checkpoints are the same, right?

djdongjin: > As long as the generated index/checkpoints are the same, right?

I feel it mostly will, and the generated bytes will be different. IIRC, in gzip zinfo version 1 we use a const 10 as the header size, but in version 2 it's no longer assumed 10 (depends on the extra flags).

turan18: Ah, yeah you are right. I dont suspect we'll have much more changes within zinfo but it would help to keep track of zinfo version somewhere a little more visible than within the c implementation directly.

turan18 (approved):

turan18 (approved):

djdongjin (approved):

ztoc/ztoc_test.go
Kern--: Created https://github.com/awslabs/soci-snapshotter/issues/494 to think through ztoc versioning

PR #511, opened 2023-03-20
Change ztoc to toc in metadata

### LINKED ISSUE #503: [FEATURE] `./metadata` should only depend on `[] FileMetadata`, not the whole `ztoc.Ztoc`
### Description

Whenever `./metadata` uses the `TOC/[] FileMetadata`, a `Ztoc` is passed to the package.

However, the package should only depend on `[] FileMetadata`, so we should 

Chang every usage/mention of `Ztoc` to `[] FileMetadata` (or `TOC`) to make it clear.

https://github.com/awslabs/soci-snapshotter/blob/94035cc235e0a9aa42d785c42bb03e452b330b04/metadata/reader.go#L81

No change

_No response_

djdongjin:
**Issue #, if available:** fix #503

**Description of changes:**

change `./metadata` pkg to depend on `ztoc.TOC` instead of `ztoc.Ztoc`.

**Testing performed:**

### DIFF
--- a/fs/layer/layer.go
+++ b/fs/layer/layer.go
@@ -322,7 +322,7 @@ func (r *Resolver) Resolve(ctx context.Context, hosts source.RegistryHosts, refs
 			commonmetrics.MeasureLatencyInMilliseconds(commonmetrics.InitMetadataStore, desc.Digest, start)
 		},
 	}
-	meta, err := r.metadataStore(sr, ztoc, append(metadataOpts, metadata.WithTelemetry(&telemetry))...)
+	meta, err := r.metadataStore(sr, ztoc.TOC, append(metadataOpts, metadata.WithTelemetry(&telemetry))...)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


fs/layer/layer.go
```diff
@@ -322,7 +322,7 @@ func (r *Resolver) Resolve(ctx context.Context, hosts source.RegistryHosts, refs
 			commonmetrics.MeasureLatencyInMilliseconds(commonmetrics.InitMetadataStore, desc.Digest, start)
 		},
 	}
-	meta, err := r.metadataStore(sr, ztoc, append(metadataOpts, metadata.WithTelemetry(&telemetry))...)
+	meta, err := r.metadataStore(sr, ztoc.TOC, append(metadataOpts, metadata.WithTelemetry(&telemetry))...)
```
Kern--, line 325: Since `TOC` is embedded in the `Ztoc`, do we actually need to change these call sites? Do you just want to be more clear?

Kern-- (approved):

fs/layer/layer.go
djdongjin: I think it make it more clear that `metadata` only depends on `TOC`. Before looking into the metadata code, I was not sure know how `Ztoc` as a whole is used in it.

djdongjin: also in general I feel it's better to avoid pointer args unless (1) we want to change that object; (2) the object is too large and we want to reduce data copy. This can help reduce accidentally changing the object.

turan18 (approved):

Issue #405, opened 2023-02-08
[DOC] Improve `GETTING_STARTED` doc and provide registry support doc


djdongjin:
**Is your feature request related to a problem? Please describe.**

> With the OCI 1.0 support, SOCI should support more registries I believe?

- [ ] Instead of using a local ORAS registry, we should update `GETTING_STARTED` doc to use a real public registry. This reduces a lot of config burden for users to try soci. (e.g. set up local registry, `--plain-http` everywhere).

- [ ] We should provide a doc listing which registries (1) soci supports natively (e.g. with referrer API); (2) soci supports with OCI 1.0 (e.g., need `--legacy-registry` flag); (3) soci doesn't support. This gives users instructions and headsup in case their registry is not supported instead of wasting their efforts.

**Describe the solution you'd like**

Better documentation, especially `GETTING_STARTED` and `REGISTRY_SUPPORT`.

**Describe alternatives you've considered**

No. We need better documentation.

**Additional context**

I think this is fairly high priority. soci is a open source project. If users cannot try soci easily following a straightforward doc, they probably won't come back.

djdongjin:
@Kern-- I updated this one with a list of doc/subtasks and my estimation for individual docs (just for reference). This one is ready for estimation.

Kern--:
@djdongjin to take BUILD/INSTALL
@turan18 to take DEBUG
@Kern-- to take REGISTRY

sparr:
#736 asks for documenting the authorization process

PR #512, opened 2023-03-21
Add registry doc

Kern--:
The testing is till WIP. I'd like to test a few more well know registries, but I think it's worth reviewing the narrative portion in the meantime.

**Issue #, if available:**
https://github.com/awslabs/soci-snapshotter/issues/405

**Description of changes:**
This change adds a doc to explain how SOCI indexes are stored and retrieved from registries and which registries have been tested for support.

**Testing performed:**
N/A

### DIFF
--- a/docs/registry.md
+++ b/docs/registry.md
@@ -1,0 +1,134 @@
+ # Registry Compatibility with SOCI
+
+SOCI is compatible with most registries. To check if your registry of choice is compatible, see [List of Registry Compatibility](#list-of-registry-compatibility).
+
+For most use-cases, compatibility is the only concern. However, there is a difference in *how* registries work with SOCI that could cause surprising edge cases. The rest of this document is a technical dive into how SOCI indices are stored and retrieved from registries and the surprises you might encounter.
+
+## Registry Requirements
+
+In order for a registry to be compatible SOCI it must support the following features of the OCI distribution and image specs:
+
+1) Accept [OCI Image Manifests](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md) with [subject fields](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md#image-manifest-property-descriptions) and arbitrary config media types.
+This allows the registry to store SOCI indices and is supported by most registries.
+
+2) (optional) Support the [OCI referrers API](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers).
+This adds convenience around retrieving SOCI indices from the registry. If it is not supported, there is a [fallback mechanism](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#unavailable-referrers-api) that works with all registries, but it has a few issues noted in the next section.
+
+## Referrers API vs Fallback
+
+The SOCI snapshotter can retrieve SOCI indices and ztocs either through the [OCI referrers API](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers) or a [Fallback mechanism](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#unavailable-referrers-api). The referrers API is part of the not-yet-released [OCI Distribution Spec v1.1](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md) so registry support is limited. The Fallback is supported by all registries, but has notable edge cases.
+
+The SOCI CLI and soci-snapshotter automatically uses the referrers API if the registry supports it or the fallback mechanism otherwise.
+
+### Referrers API
+
+The [referrers API](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers) is a registry endpoint where an agent can query for all artifacts that reference a given image digest, optionally filtering by artifact type. The registry indexes artifacts for the referrers API when the artifact is pushed. When a container is launched, the soci-snapshotter can query this API to find SOCI indices that reference the digest of the image.
+
+### Fallback
+
+If the referrers API is not available, the OCI distribution spec defines a [fallback mechanism that works with existing registries](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#unavailable-referrers-api). In this mechanism, the contents that would normally be returned by the referrers API are instead put into an [OCI Image Index](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/image-index.md) which is tagged in the registry with the digest of the manifest to which it refers.
+
+For example, imagine you had an image `myregistry.com/image:latest` with digest `sha:123`. If you created and pushed a SOCI index for that image, there would also be a new image index `myregistry.com/image:sha-123` which contains the SOCI index' descriptor. At runtime, the soci-snapshotter will pull the `myregistry.com/image:sha-123` index and apply client side filtering to discover the SOCI index.
+
+For clarity in the rest of this section, we will refer to `myregistry.com/image:sha-123` as the "fallback" (as opposed to image index) to distinguish it from the SOCI index.
+
+An important note here is that the fallback is managed on the *client side* by the tool performing the push. There is therefore a race condition when pushing a SOCI index because the fallback has to be pulled, modified to add the new SOCI index, and then pushed back to the registry. If a second artifact is pushed that references the same image digest, then one modification of the fallback could clobber the other.
+
+To clarify the scope of this problem, the fallback is unique per image digest. Multiple artifacts (SOCI Indices, signatures, etc.) can modify the same fallback. The image digest is generally unique per image/platform pair. As an example of what this means in practice, concurrently creating a SOCI index for the image for platforms `linux/amd64` and `linux/i386` is safe because the image digests will be different. Concurrently creating a SOCI index and signature for an image and platform `linux/amd64` is unsafe because both artifacts will refer to the same image digest.
+
+Since the fallback is managed client side, the registry does not know about the relationship between SOCI indices and the fallback. Deleting a SOCI index will not delete or modify the fallback. It is up to the user to make the necessary modifications or deletions of the fallback when deleting a SOCI index from the registry.
+
+## How SOCI Indices Appear to Registries
+
+Each registry will display information in a slightly different mechanism, but here we show what artifacts might show up in your repository and an explanation of what they are:
+
+
+| Tag         | Type        | Explanation                                                                                                |
+| ----------- | ----------- | ---------------------------------------------------------------------------------------------------------- |
+| latest      | Image       | The actual image                                                                                           |
+| \<untagged> | Image       | The SOCI index manifest. This may appear as type SOCI Index or Other                                       |
+| sha:123     | Image Index | The fallback image index. This will only be present for registries which do not support the referrers API. |
+
+## List of Registry Compatibility
+
+Registries that are not listed have not been tested by the SOCI maintainers or reported by the community, but they may still be compatible SOCI.
+
+| Registry                                                                                  | Compatible? | Mechanism     | Notes                                                |
+| ----------------------------------------------------------------------------------------- | ----------- | ------------- | ---------------------------------------------------- |
+| [Docker Hub](https://hub.docker.com)                                                      | No          | N/A           | Does not support image manifests with subject fields |
+| [Amazon Elastic Container Registry (ECR)](https://aws.amazon.com/ecr/)                    | Yes         | Fallback      |                                                      |
+| [Amazon ECR Public Gallery](https://gallery.ecr.aws)                                      | No          | N/A           | Does not support image manifests with subject fields |
+| [Azure Container Registry](https://azure.microsoft.com/en-us/products/container-registry) | Yes         | Referrers API |                                                      |
+| [GitHub Packages (GHCR)](https://github.com/features/packages)                            | Yes         | Fallback      |                                                      |
+| [Google Cloud Container Registry (GCR)](https://cloud.google.com/container-registry)      | Yes         | Fallback      |                                                      |
+| [Google Cloud Artifact Registry (AR)](https://cloud.google.com/artifact-registry)         | No          | N/A           | Testing the referrers API redirects to login         |
+| [Quay.io](https://quay.io)                                                                | No          | N/A           | Does not support image manifests with subject fields |
+| [Artifactory](https://jfrog.com/artifactory/)                                             | Yes         | Fallback      |                                                      |
+| [Harbor](https://github.com/goharbor/harbor)                                              | Yes         | Fallback      |                                                      |
+| [Distribution](https://github.com/distribution/distribution)                              | Yes         | Fallback      |                                                      |
+| [OCI-playground Distribution](https://github.com/oci-playground/distribution)             | Yes         | Referrers API |                                                      |
+
+### Failure Examples
+
+Below are some slightly redacted examples from the services that don't support the features needed to be compatible SOCI.
+
+**Docker Hub**
+
+```
+$ sudo ./out/soci push --user $USERNAME:$PASSWORD docker.io/####/busybox:latest
+checking if a soci index already exists in remote repository...
+pushing soci index with digest: sha256:d6ebffd218ead37e4862172b4f19491341e72aebc3cc6d9cf1a22297c40fb3c2
+pushing artifact with digest: sha256:cce4c7e12e01b32151d69348fcf52e0db7b44f6df6c23c511fa5c52eaf272c28
+pushing artifact with digest: sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a
+skipped artifact with digest: sha256:acaddd9ed544f7baf3373064064a51250b14cfe3ec604d65765a53da5958e5f5
+successfully pushed artifact with digest: sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a
+successfully pushed artifact with digest: sha256:cce4c7e12e01b32151d69348fcf52e0db7b44f6df6c23c511fa5c52eaf272c28
+pushing artifact with digest: sha256:d6ebffd218ead37e4862172b4f19491341e72aebc3cc6d9cf1a22297c40fb3c2
+soci: error pushing graph to remote: PUT "https://registry-1.docker.io/v2/####/busybox/manifests/sha256:d6ebffd218ead37e4862172b4f19491341e72aebc3cc6d9cf1a22297c40fb3c2": response status code 404: notfound: not found
+```
+
+The index manifest can't be pushed at all.
+
+**Amazon ECR Public Gallery**
+
+```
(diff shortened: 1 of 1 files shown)


docs/registry.md
```diff
@@ -0,0 +1,60 @@
+ # Registry Support for SOCI
+
+SOCI is supported by most registries. The notable exception to this is Duckerhub, which currently does not meet our technical requirements. 
+To check if your registry of choice is supported, see [#list-of-registry-support](List of Registry Support).
```
turan18, line 4: switch the order of link with text [List of...] (# link)

docs/registry.md
```diff
@@ -0,0 +1,60 @@
+ # Registry Support for SOCI
+
+SOCI is supported by most registries. The notable exception to this is Duckerhub, which currently does not meet our technical requirements.
```
djdongjin, line 3: "meet our technical requirements" sounds like we have some non-standard requirement. Is it possible to change to something like "which doesn't fully support OCI 1.0 used by SOCI".

docs/registry.md
```diff
@@ -0,0 +1,60 @@
...
+SOCI is supported by most registries. The notable exception to this is Duckerhub, which currently does not meet our technical requirements. 
+To check if your registry of choice is supported, see [#list-of-registry-support](List of Registry Support).
+
+For most use-cases, support is the only concern. However, there is a difference in *how* registries support SOCI that could cause surprising edge cases. The rest of this document is a technical dive into how SOCI indices are stored and retrieved from registries and the surprises you might encounter.
+
+## Registry Requirements
+
+In order for a registry to support SOCI it must have the following features:
+
+1) Accept [OCI Image Manifests](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md) with [subject fields](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md#image-manifest-property-descriptions) and arbitrary config media types.
```
djdongjin, line 12: Is this part of OCI 1.0 (or prior to 1.0)? If so it might be good to refer a 1.0.0 tag instead of 1.1.0-rc2 which looks like a new feature to be added in OCI 1.1.

docs/registry.md
```diff
@@ -0,0 +1,60 @@
...
+This adds convenience around retrieving SOCI indices from the registry. If it is not supported, there is a [fallback mechanism](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#unavailable-referrers-api) that can be used instead, but it has a few issues noted in the next section. Since the OCI referrers API was defined in the not-yet-released [OCI Distribution Spec v1.1](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md), very few registries support it.
+
+## Referrers API vs Fallback
+
+Registries can support retrieving SOCI artifacts either though the [OCI referrers API](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers) or a [Fallback mechanism](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#unavailable-referrers-api). The referrers API is part of the not-yet-released [OCI Distribution Spec v1.1](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md) so support is limited. The Fallback API is supported by all registries, but has notable edge cases.
```
djdongjin, line 20: nit: SOCI artifacts -> SOCI artifacts (index, ztoc).

though -> through

djdongjin: The 2nd sentence is repeated with the sentence above "2) (optional)..." (copied below). Suggest to remove the one in "2) (optional).."

"Since the OCI referrers API was defined in the not-yet-released [OCI Distribution Spec v1.1](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md), very few registries support it."

docs/registry.md
```diff
@@ -0,0 +1,60 @@
...
+
+The SOCI CLI and soci-snapshotter automatically and seamlessly use the referrers API if the registry supports it or the fallback mechanism otherwise.
```
djdongjin, line 22: nit: "and seamlessly" is unnecessary

docs/registry.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Referrers API
+
+The [referrers API](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers) is a registry endpoint where an agent can query for all artifacts that reference a given image digest, optionally filtering by artifact type. This means that when an container is launched, the soci-snapshotter can query this API to find SOCI indices that reference the digest of the image-to-be-launched. The registry indexes artifacts for the referrers API when the artifact is pushed.
```
djdongjin, line 26: nit: "the image-to-be-launched" -> "the image used by that container"

djdongjin: suggest to switch the 2nd and 3rd sentence. (index happens before query).

also "This means that when" -> "When"

docs/registry.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Fallback
+
+If the referrers API is not available, the OCI spec authors defined a [fallback mechanism that works with existing registries](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#unavailable-referrers-api). In this mechanism, the contents that would normally be returned by the referrers API are instead put into an [OCI Image Index](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/image-index.md) which is tagged in the registry with the digest of the manifest to which it refers.
```
djdongjin, line 30: nit: "authors defined" -> "defines"

docs/registry.md
```diff
@@ -0,0 +1,60 @@
...
+
+For example, imagine you had an image `myregistry.com/image:latest` with digest `sha:123`. If you created and pushed a SOCI index for that image, there would also be a new image index `myregistry.com/image:sha-123` which contains the SOCI index' descriptor. At runtime, the soci-snapshotter will pull the `myregistry.com/image:sha-123` index and apply client side filtering to discover the SOCI index. 
+
+For clarity in the rest of this section, we will refer to `myregistry.com/image:sha-123` as the "fallback" (as opposed to image index) to distinguish it from the SOCI index. Please note that this is not standard terminology. 
+
+An important note here is that the fallback is managed *client side* by the tool performing the push. There is therefore a race condition when pushing a SOCI index because the fallback has to be pulled, modified to add the new SOCI index, and then pushed back to the registry. If a second artifact is pushed that references the same image digest, then one modification of the fallback could clobber the other.
```
djdongjin, line 36: *client side* -> on *client side*

docs/registry.md
```diff
@@ -0,0 +1,60 @@
...
+
+To clarify the scope of this problem, the fallback is unique per image digest. Multiple artifacts (SOCI Indices, signatures, etc.) can modify the same fallback. The image digest is generally unique per image/platform pair. As an example of what this means in practice, concurrently creating a SOCI index for the an image for platforms `linux/amd64` and `linux/i386` is safe because the image digests will be different. Concurrently creating a SOCI index and signature for an image and platform `linux/amd64` is unsafe because both artifacts will refer to the same image digest.
+
+Since the fallback is managed client side, the registry does not know about the relationship between SOCI indices and the fallback. Deleting a SOCI index will not delete or modify the fallback. It is up to the user to make the necessary modifications or deletions of the fallback when deleting a SOCI index from the registry.
+
+## How SOCI Indices appear to registries
```
djdongjin, line 42: make header capitalization consistent

docs/registry.md
Kern--: I'd like to reference the same version of the spec throughout this doc to reduce confusion. In this case, we actually do need the latest doc which defines the subject field.

docs/registry.md
Kern--: I intentionally duplicated some of the wording so that the requirements section is like the summary version and the comparison is the detailed version. I can agree that it's probably not adding enough value, though. I will remove it.

docs/registry.md
djdongjin: I'm okay with that. then we need to change v1.1.0-rc1 to v1.1.0-rc2 (both are used in the doc now)

Kern--: They're actually 2 difference specs. Distribution spec v1.1.0-rc1 and Image spec v1.1.0-rc2.

djdongjin: oh missed the name part 😅

docs/registry.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Fallback
+
+If the referrers API is not available, the OCI distribution spec defines a [fallback mechanism that works with existing registries](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#unavailable-referrers-api). In this mechanism, the contents that would normally be returned by the referrers API are instead put into an [OCI Image Index](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/image-index.md) which is tagged in the registry with the digest of the manifest to which it refers.
+
+For example, imagine you had an image `myregistry.com/image:latest` with digest `sha:123`. If you created and pushed a SOCI index for that image, there would also be a new image index `myregistry.com/image:sha-123` which contains the SOCI index' descriptor. At runtime, the soci-snapshotter will pull the `myregistry.com/image:sha-123` index and apply client side filtering to discover the SOCI index.
+
+For clarity in the rest of this section, we will refer to `myregistry.com/image:sha-123` as the "fallback" (as opposed to image index) to distinguish it from the SOCI index. Please note that this is not standard terminology.
```
sparr, line 34: I suggest finding a way to make/use standard terminology here.

docs/registry.md
```diff
@@ -0,0 +1,118 @@
...
+
+An important note here is that the fallback is managed on the *client side* by the tool performing the push. There is therefore a race condition when pushing a SOCI index because the fallback has to be pulled, modified to add the new SOCI index, and then pushed back to the registry. If a second artifact is pushed that references the same image digest, then one modification of the fallback could clobber the other.
+
+To clarify the scope of this problem, the fallback is unique per image digest. Multiple artifacts (SOCI Indices, signatures, etc.) can modify the same fallback. The image digest is generally unique per image/platform pair. As an example of what this means in practice, concurrently creating a SOCI index for the an image for platforms `linux/amd64` and `linux/i386` is safe because the image digests will be different. Concurrently creating a SOCI index and signature for an image and platform `linux/amd64` is unsafe because both artifacts will refer to the same image digest.
```
sparr, line 38: `for the an image`

sparr, line 38:
```suggestion
To clarify the scope of this problem, the fallback is unique per image digest. Multiple artifacts (SOCI Indices, signatures, etc.) can modify the same fallback. The image digest is generally unique per image/platform pair. As an example of what this means in practice, concurrently creating a SOCI index for the image for platforms `linux/amd64` and `linux/i386` is safe because the image digests will be different. Concurrently creating a SOCI index and signature for an image and platform `linux/amd64` is unsafe because both artifacts will refer to the same image digest.
```

docs/registry.md
```diff
@@ -0,0 +1,118 @@
...
+| [Harbor](https://github.com/goharbor/harbor)                                              | Yes        | Fallback      |                                                      |
+| [Distribution](https://github.com/distribution/distribution)                              | Yes        | Fallback      |                                                      |
+| [OCI-playground Distribution](https://github.com/oci-playground/distribution)             | Yes        | Referrers API |                                                      |
+
+
+
+
+### Failure Examples
+
+Below are some slightly redacted exmples from the services that don't support SOCI.
```
sparr, line 76:
```suggestion
Below are some slightly redacted examples from the services that don't support SOCI.
```

docs/registry.md
```diff
@@ -0,0 +1,118 @@
+ # Registry Support for SOCI
+
+SOCI is supported by most registries. The notable exception to this is Duckerhub, which currently does not support subject fields in image manifests.
```
sparr, line 3: Can we rephrase all the various uses of "registry supports soci" to refer more to the features we depend on and less to us specifically? This phrasing suggests that we want registries to do something special to support us in particular. I want to emphasize that we rely on interoperable standards. Perhaps something like "The features that SOCI relies on are available in most registries" here, but less verbose rephrasing in other instances further down the doc. Perhaps "is compatible with" rather than "supports" or "is supported by".

sbuckfelder, line 3: is duckerhub a thing? or is this a dockerhub typo?

docs/registry.md
```diff
@@ -0,0 +1,118 @@
...
+| sha:123     | Image Index | The fallback image index. This will only be present for registries which do not support the referrers API. |
+
+## List of Registry Support
+
+Registries that are not listed have not been tested by the SOCI maintainers or reported by the community, but they may still support SOCI.
+
+| Registry                                                                                  | Supported? | Mechanism     | Notes                                                |
+| ----------------------------------------------------------------------------------------- | ---------- | ------------- | ---------------------------------------------------- |
+| [Docker Hub](https://hub.docker.com)                                                      | No         | N/A           | Does not support image manifests with subject fields |
+| [Amazon Elastic Container Registry (ECR)](https://aws.amazon.com/ecr/)                    | Yes        | Fallback      |                                                      |
```
sbuckfelder, line 60: Does ECR public support?

docs/registry.md
Kern--: "referrers tag schema" is the closest thing to standard terminology, which I think is more confusing that just using the term "fallback".

docs/registry.md
Kern--: That's a good call out. I'm going to change "support" to "compatible with" when talking about SOCI and only use the term "support" in reference the specific OCI spec features that we rely on.

sparr:
```suggestion
SOCI is supported by most registries. The notable exception to this is Docker Hub, which currently does not support subject fields in image manifests.
```

Kern--: Can you take another look at let me know if you like the new phrasing?

docs/registry.md
```diff
@@ -0,0 +1,134 @@
...
+For most use-cases, compatibility is the only concern. However, there is a difference in *how* registries work with SOCI that could cause surprising edge cases. The rest of this document is a technical dive into how SOCI indices are stored and retrieved from registries and the surprises you might encounter.
+
+## Registry Requirements
+
+In order for a registry to be compatible SOCI it must support the following features of the OCI distribution and image specs:
+
+1) Accept [OCI Image Manifests](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md) with [subject fields](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md#image-manifest-property-descriptions) and arbitrary config media types.
+This allows the registry to store SOCI indices and is supported by most registries.
+
+2) (optional) Support the [OCI referrers API](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers)
```
djdongjin, line 14: missing a `.` after [OCI referrers API](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers)

docs/registry.md
```diff
@@ -0,0 +1,134 @@
...
+This adds convenience around retrieving SOCI indices from the registry. If it is not supported, there is a [fallback mechanism](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#unavailable-referrers-api) that can be used instead, but it has a few issues noted in the next section.
```
djdongjin, line 15: > there is a [fallback mechanism](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#unavailable-referrers-api) that can be used instead

It's unclear to me how this can be used. Is it something a registry needs to support? (i.e., do you mean "If it's not supported, the registry needs to support the fallback mechanism").

docs/registry.md
```diff
@@ -0,0 +1,134 @@
...
+| [Google Cloud Artifact Registry (AR)](https://cloud.google.com/artifact-registry)         | No          | N/A           | Testing the referrers API redirects to login         |
+| [Quay.io](https://quay.io)                                                                | No          | N/A           | Does not support image manifests with subject fields |
+| [Artifactory](https://jfrog.com/artifactory/)                                             | Yes         | Fallback      |                                                      |
+| [Harbor](https://github.com/goharbor/harbor)                                              | Yes         | Fallback      |                                                      |
+| [Distribution](https://github.com/distribution/distribution)                              | Yes         | Fallback      |                                                      |
+| [OCI-playground Distribution](https://github.com/oci-playground/distribution)             | Yes         | Referrers API |                                                      |
+
+
+
+
```
djdongjin, line 73: Unnecessary empty lines (1 empty line is sufficient).

docs/registry.md
```diff
@@ -0,0 +1,134 @@
...
+### Failure Examples
+
+Below are some slightly redacted examples from the services that don't support the features needed to be compatible SOCI.
+
+**Docker Hub**
+```
```
djdongjin, line 79: nit: an empty line before and after the code block, also language annotation (e.g., ```shell/console/bash/etc)

docs/registry.md
```diff
@@ -0,0 +1,134 @@
...
+successfully pushed artifact with digest: sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a
+successfully pushed artifact with digest: sha256:cce4c7e12e01b32151d69348fcf52e0db7b44f6df6c23c511fa5c52eaf272c28
+pushing artifact with digest: sha256:d6ebffd218ead37e4862172b4f19491341e72aebc3cc6d9cf1a22297c40fb3c2
+soci: error pushing graph to remote: PUT "https://registry-1.docker.io/v2/####/busybox/manifests/sha256:d6ebffd218ead37e4862172b4f19491341e72aebc3cc6d9cf1a22297c40fb3c2": response status code 404: notfound: not found
+```
+
+The index manifest can't be pushed at all.
+
+**Amazon ECR Public Gallery**
+```
```
djdongjin, line 95: same here

docs/registry.md
```diff
@@ -0,0 +1,134 @@
...
+$ sudo ./out/soci push --user $USERNAME:$PASSWORD us-east1-docker.pkg.dev/####/busybox:latest
+checking if a soci index already exists in remote repository...
+soci: failed to fetch list of referrers: GET "https://accounts.google.com/v3/signin/identifier?dsh=###&continue=https%3A%2F%2Fconsole.cloud.google.com%2Fartifacts%2Ftags%2Fv2%2Fus-east1%2F####%252Fbusybox%252Freferrers%252Fsha256%2Facaddd9ed544f7baf3373064064a51250b14cfe3ec604d65765a53da5958e5f5...&service=cloudconsole&flowName=WebLiteSignIn&flowEntry=ServiceLogin": failed to decode response: invalid character '<' looking for beginning of value
+```
+
+This is before pushing the SOCI index. Looking up an existing referrer is redirecting to an auth flow.
+
+Pushing a SOCI index works as long as the SOCI CLI receives the `--existing-index allow` flag to skip the check for existing indices. This is because Artifact Registry appears to always redirect requests to the referrers API to an auth flow, despite other authenticated requests to the registry succeeding as expected. Pulling always falls back to OverlayFS due to the redirect.
+
+
```
djdongjin, line 121: nit: 1 empty line

docs/registry.md
sparr: Looks great. Love the phrasing of it as registries supporting features rather than us supporting registries.

djdongjin (approved):

sparr (approved):

PR #513, opened 2023-03-21
Add a zinfo creation func in ztoc

djdongjin:
**Issue #, if available:** related to #463

**Description of changes:**

Add a helper func to ztoc that creates zinfo based on info in the ztoc.

**Testing performed:**

### DIFF
--- a/cmd/soci/commands/ztoc/info.go
+++ b/cmd/soci/commands/ztoc/info.go
@@ -86,7 +86,7 @@ var infoCommand = cli.Command{
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


Kern--:
What direction are we going with this API? 

Are we going to keep embedding `CompressionInfo` into `Ztoc` with this helper? Should we instead embed `ZInfo` into `Ztoc` and just use the ztoc directly instead of `ztoc.Zinfo()`?

Kern-- (approved):
I can agree this is cleaner than what we had before. Let's get this merged and continue the discussion in the issue where it's easier to find.

turan18 (approved):

Issue #30, opened 2022-08-19
[FEATURE] Allow CLI to discover soci artifacts that it doesn't create


Kern--:
**Is your feature request related to a problem? Please describe.**
Currently, `soci` uses a bbolt metadata DB as its source of truth about what SOCI artifacts exist locally. This DB is only updated by `soci create`. That means that any indices/ztocs that are added to the store outside of the `soci create` command are invisible to the inspection tools provided by `soci` (e.g. `soci index list` will not see an index pulled with `soci rpull`). 

**Describe the solution you'd like**
I would like to see a command like:
```
soci rebuild-db
```

which would rediscover all of the soci artifacts in the local storage and add the necessary metadata to the db. 

**Describe alternatives you've considered**
One alternative would be to have the snapshotter update the local DB when it pulls SOCI artifacts. There are a couple of downsides to this approach:
1) SOCI artifacts could be deleted outside of either `soci` or the soci snapshotter which would require a rebuild-like command to avoid a full purge.
2) bbolt locks the DB file while a process is interacting with it. If the soci snapshotter opens the DB at launch, then it must not be running when `soci` is invoked. If the soci snapshotter opens/closes the DB around each artifact fetch, then it may see an unacceptable performance hit.

The second issue could be solved by having the snapshotter own the metadata DB and expose a gRPC (or similar) service that the `soci` command could interact with. This has the downside that the snapshotter is required to be running in order to build SOCI artifacts which is not desirable in cases like a managed index builder where the images will never be run locally.

**Additional context**

Issue #423, opened 2023-02-14
[FEATURE] Support using containerd's content store


Kern--:
**Is your feature request related to a problem? Please describe.**
Today, we store soci artifacts (index manifests/ztocs) in the snapshotter's state directory (under `/var/lib/soci-snapshotter-grpc`). While we know how to map an index manifest to ztocs, we don't have a good way to map ztocs to index manifests that reference them. Since we don't have this, we don't offer an CLI mechanism for cleaning up SOCI artifacts without potentially orphaning ztocs or deleting ztocs out from under a manifest. The easiest way to clean up old ztocs is to just blow away the whole snapshotter state directory at `/var/lib/soci-snapshotter-grpc`.

It would be nice we could use containerd's content store and garbage collector instead. This would automatically clean up ztocs when all the referencing index manifests are deleted. 

**Describe the solution you'd like**
A list of acceptance criteria:
1) The SOCI CLI/library supports using containerd's content store for SOCI artifacts by default
2) The SOCI CLI has a flag to use the old, local directory to put artifacts under `/var/lib/soci-snapshotter-grpc`
3) ztocs are automatically deleted when the last index manifest that references them is deleted
4) Index manifests are deleted only when using `soci index remove`

Since this is a pretty large refactor, I think it can be broken down into several milestones (I'm leaving them in this issue for now, but we can split them out of this isn't working).

- [ ] Create a containerd-backed implementation of `content.Storage` [ref](https://github.com/oras-project/oras-go/blob/main/content/storage.go#L43) keeping in mind the need to set the correct gc labels
- [ ] Update `soci create` to use the new storage with a temporary flag (e.g. `--containerd-storage`)
- [ ] Update `soci index *` to use the new storage  with a temporary flag
- [ ] Update `soci ztoc *` to use the new storage with a temporary flag
- [ ] Update `soci push` to use the new storage with a temporary flag
- [ ] Update the snapshotter use the new storage
- [ ] Default to the new storage with a new flag to use the old storage (e.g. `--local-storage`)

**Describe alternatives you've considered**
We could alternatively build our own ztoc garbage collection setup on top of the local store.

**Additional context**
Information about how content flows through containerd:
https://github.com/containerd/containerd/blob/main/docs/content-flow.md

Information about containerd's garbage collection to make sure indices/ztocs don't get prematurely deleted.
https://github.com/containerd/containerd/blob/main/docs/garbage-collection.md

ningziwen:
[...]

PR #527, opened 2023-03-24
Add `rebuild-db` command to rebuild artifacts db

turan18:
**Issue #, if available:**

Fixes: #30 

**Description of changes:**

Added `rebuild-db` command to CLI to rebuild/sync artifacts db with local content store. 

TODO: 
 * [x] add integration tests
 * [x] determine if getting platform for an index is feasible

**Testing performed:**

```shell
$ sudo ./soci image rpull --platform linux/amd64  dkr.ecr.us-west-2.amazonaws.com/rabbitmq:latest
fetching sha256:f5f012e4... application/vnd.docker.distribution.manifest.v2+json
fetching sha256:7cb0caa8... application/vnd.docker.container.image.v1+json 

$ sudo ./soci index list
DIGEST    SIZE    IMAGE REF    PLATFORM    MEDIA TYPE    CREATED

$ sudo ./soci ztoc list
DIGEST    SIZE    LAYER DIGEST

$ sudo ./soci rebuild-db

$ sudo ./soci index list
DIGEST                                                                     SIZE    IMAGE REF                                                       PLATFORM    MEDIA TYPE                                    CREATED
sha256:a6494e77b9d51f160e9c74a282b1779421f980f3dd3d130402c2222aa8f38e40    1570    .dkr.ecr.us-west-2.amazonaws.com/rabbitmq:latest    linux/amd64     application/vnd.oci.image.manifest.v1+json    9s ago

$ sudo ./soci ztoc list
DIGEST                                                                     SIZE       LAYER DIGEST
sha256:0d83470e087681a426a7219adf9c9e02b96583df6e3e045e9277c815b92392b9    1090344    sha256:328b9d3248edeb3ae6e7f9c347bcdb5632c122be218e6ecd89543cca9c8f1997
sha256:1a6457d173b415902ee571f068009fe288e5e12d8ce5fd80a0288ba8b0c2f4eb    1152224    sha256:df6635ed1257a768a4cf0fba31ebc5c8a6a03ae7d5b9b079bfd9df9eb89e0f81
sha256:1ae8016952be5b522dbc5a695ad3044dc52ef9af9fb1840c8c6442b694ecb579    1235272    sha256:e50ba1435db6125cd01ba5740e1899412c8dfa2f8ce8526ea32574742d3a0f41 
```

### DIFF
--- a/cmd/soci/commands/rebuild_db.go
+++ b/cmd/soci/commands/rebuild_db.go
@@ -1,0 +1,50 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package commands
+
+import (
+	"path/filepath"
+
+	"github.com/awslabs/soci-snapshotter/fs/config"
+	"github.com/awslabs/soci-snapshotter/soci"
+	"github.com/containerd/containerd/cmd/ctr/commands"
+	"github.com/urfave/cli"
+	"oras.land/oras-go/v2/content/oci"
+)
+
+var RebuildDBCommand = cli.Command{
+	Name:  "rebuild-db",
+	Usage: `rebuild the artifacts database. You should use this command after "rpull" so that indices/ztocs can be discovered by commands like "soci index list".`,
+	Action: func(cliContext *cli.Context) error {
+		client, ctx, cancel, err := commands.NewClient(cliContext)
+		if err != nil {
+			return err
+		}
+		defer cancel()
+		containerdContentStore := client.ContentStore()
+		artifactsDb, err := soci.NewDB(soci.ArtifactsDbPath())
+		if err != nil {
+			return err
+		}
+		blobStore, err := oci.New(config.SociContentStorePath)
+		if err != nil {
+			return err
+		}
+		blobStorePath := filepath.Join(config.SociContentStorePath, "blobs")
+		return artifactsDb.SyncWithLocalStore(ctx, blobStore, blobStorePath, containerdContentStore)
+	},
+}
--- a/cmd/soci/main.go
+++ b/cmd/soci/main.go
@@ -90,6 +90,7 @@ func main() {
 		commands.CreateCommand,
 		commands.PushCommand,
 		run.Command,
+		commands.RebuildDBCommand,
 	}
 
 	if err := app.Run(os.Args); err != nil {
--- a/integration/rebuild-db_test.go
+++ b/integration/rebuild-db_test.go
@@ -1,0 +1,97 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package integration
+
+import (
+	"bytes"
+	"fmt"
+	"path/filepath"
+	"testing"
+
+	"github.com/awslabs/soci-snapshotter/soci"
+)
+
+func TestRebuildArtifactsDB(t *testing.T) {
+	regConfig := newRegistryConfig()
+	sh, done := newShellWithRegistry(t, regConfig)
+	defer done()
+	rebootContainerd(t, sh, "", "")
+	img := rabbitmqImage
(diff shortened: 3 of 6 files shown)


cmd/soci/commands/rebuild_db.go
```diff
@@ -0,0 +1,155 @@
...
+	"time"
+
+	"github.com/awslabs/soci-snapshotter/fs/config"
+	"github.com/awslabs/soci-snapshotter/soci"
+	"github.com/containerd/containerd/cmd/ctr/commands"
+	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/platforms"
+	"github.com/opencontainers/go-digest"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
```
djdongjin, line 35: seems `ocispec` is a more common name than `v1` in our repo.

cmd/soci/commands/rebuild_db.go
```diff
@@ -0,0 +1,155 @@
...
+			return err
+		}
+		blobStore, err := oci.New(config.SociContentStorePath)
+		if err != nil {
+			return err
+		}
+		blobStorePath := filepath.Join(config.SociContentStorePath, "blobs")
+		// check if artifact db contains all artifacts in content store, if not add them
+		err = filepath.WalkDir(blobStorePath, func(path string, d fs.DirEntry, err error) error {
+			if !d.IsDir() {
```
djdongjin, line 61: nit: can we change to `if d.IsDir() { return nil }` to remove some indent?

cmd/soci/commands/rebuild_db.go
```diff
@@ -0,0 +1,155 @@
...
+				}
+			}
+			return nil
+		})
+		return err
+	},
+}
+
+// addHashPrefix adds either a sha256 or sha512 prefix to a string
+func addHashPrefix(name string) string {
```
Kern--, line 150: I think this is what `godigest.NewDigestFromBytes` is for: https://pkg.go.dev/github.com/opencontainers/go-digest#NewDigestFromBytes

cmd/soci/commands/rebuild_db.go
```diff
@@ -0,0 +1,155 @@
...
+				}
+				f, err := os.Open(path)
+				if err != nil {
+					return err
+				}
+				defer f.Close()
+				var sociIndex soci.Index
+				err = soci.DecodeIndex(f, &sociIndex)
+				// skip: file is either a ztoc or empty config
+				if err != nil || info.Size() < 10 {
```
djdongjin, line 74: should the `info.Size() < 10` check more close to where `info` is created (line 63)

cmd/soci/commands/rebuild_db.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/platforms"
+	"github.com/opencontainers/go-digest"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/urfave/cli"
+	"oras.land/oras-go/v2/content/oci"
+)
+
+var RebuildDBCommand = cli.Command{
+	Name:  "rebuild-db",
```
sparr, line 41: This is only our second command with two words and a `-` in it. Are we intentionally moving toward more verbose command names?

turan18: No specific reason, just going with the suggested name in #30

cmd/soci/commands/rebuild_db.go
```diff
@@ -0,0 +1,155 @@
...
+			return err
+		}
+		// check if artifact db contains artifacts that no longer exist in content store, if so delete them
+		err = artifactsDb.Walk(func(ae *soci.ArtifactEntry) error {
+			existsInContentStore, err := blobStore.Exists(context.Background(), v1.Descriptor{MediaType: ae.MediaType, Digest: digest.Digest(ae.Digest)})
+			if err != nil {
+				return err
+			}
+			if !existsInContentStore {
+				err := artifactsDb.RemoveArtifactEntryByDigest(ae.Digest)
```
djdongjin, line 138: nit: if err := artifactsDb.RemoveArtifactEntryByDigest(ae.Digest); err != nil

cmd/soci/commands/rebuild_db.go
turan18: We already have the hash encoding as its the filename. We just need to determine the hash algorithm (sha256 vs 512) and we can do this obtaining the length of the encoding.

cmd/soci/commands/rebuild_db.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/urfave/cli"
+	"oras.land/oras-go/v2/content/oci"
+)
+
+var RebuildDBCommand = cli.Command{
+	Name:  "rebuild-db",
+	Usage: `rebuild the artifacts database. You should use this command after "rpull" so that indices/ztocs can be discovered by commands like "soci index list".`,
+	Action: func(cliContext *cli.Context) error {
```
sparr, line 42: Strongly request the functional logic here be broken out into its own function that could be called from tests, other commands, etc. Having program logic within command functions has been a problem point with containerd in the past.

turan18: I agree. The reason I was a little hesitant at first is that we might not eventually need this logic at all once/ if #423 happens, so I thought it might be a good idea to keep the core logic in `cmd` for now (with the exception of `RemoveOldArtifacts` since that required some explicit db logic)

I was originally thinking of having some exported func like `SyncDbWithStore` in our `soci` package how does that sound?

sparr: That seems viable. Most of the logic here would move there.

sparr (approved):
LGTM now

Kern-- (approved):

PR #537, opened 2023-03-31
Using rhttp client when fetching artifacts instead of oras http client.

turan18:
**Issue #, if available:**

**Description of changes:**

Switching to using `rhttp` client when fetching artifacts. Default timeout of `2s`.

**Testing performed:**


fs/artifact_fetcher.go
```diff
@@ -66,32 +71,69 @@ func newArtifactFetcher(refspec reference.Spec, localStore, remoteStore content.
...
-		}, nil
+		},
 	}
-
-	repo.Client = authClient
+	// set rhttp retry logic
+	tr := authClient.Client.Transport
+	if rt, ok := tr.(*rhttp.RoundTripper); ok {
+		rt.Client.RetryMax = 8
+		rt.Client.RetryWaitMin = 30
```
Kern--, line 131: This is in nanoseconds. It needs to be multiplied by `time.millisecond`.

fs/artifact_fetcher.go
```diff
@@ -66,32 +71,69 @@ func newArtifactFetcher(refspec reference.Spec, localStore, remoteStore content.
...
+		rt.Client.RetryWaitMax = 300000
```
Kern--, line 132: Same as above. Multiply by `time.Millisecond`.

fs/artifact_fetcher.go
```diff
@@ -66,32 +71,75 @@ func newArtifactFetcher(refspec reference.Spec, localStore, remoteStore content.
...
+func jitter(duration time.Duration, divisor int64) time.Duration {
+	return time.Duration(rand.Int63n(int64(duration)/divisor) + int64(duration))
+}
+
+func backoffStrategy(min, max time.Duration, attemptNum int, resp *http.Response) time.Duration {
+	delayTime := rhttp.DefaultBackoff(min, max, attemptNum, resp)
+	return jitter(delayTime, 8)
+}
+
+func retryStrategy(ctx context.Context, resp *http.Response, err error) (bool, error) {
```
djdongjin, line 86: can we just create a `util/http/retry.go` and move these three there from https://github.com/awslabs/soci-snapshotter/blob/main/fs/remote/resolver.go, instead of making a copy.

fs/artifact_fetcher.go
```diff
@@ -66,32 +71,75 @@ func newArtifactFetcher(refspec reference.Spec, localStore, remoteStore content.
...
 
-	authClient := auth.DefaultClient
-	authClient.Cache = auth.DefaultCache
-	authClient.Credential = func(_ context.Context, host string) (auth.Credential, error) {
-		username, secret, err := dockerconfig.DockerCreds(host)
-		if err != nil {
-			return auth.EmptyCredential, err
-		}
-		if username == "" && secret != "" {
+	const defaultTimeout = 2 * time.Second
```
djdongjin, line 103: can we move this const to the beginning of the file and add some comment about the value / what it is used for.

fs/artifact_fetcher.go
```diff
@@ -66,32 +71,75 @@ func newArtifactFetcher(refspec reference.Spec, localStore, remoteStore content.
...
+		},
 	}
-
-	repo.Client = authClient
+	// set rhttp retry logic
+	tr := authClient.Client.Transport
+	if rt, ok := tr.(*rhttp.RoundTripper); ok {
+		rt.Client.RetryMax = 8
+		rt.Client.RetryWaitMin = time.Millisecond * 30
+		rt.Client.RetryWaitMax = time.Millisecond * 300000
```
djdongjin, line 138: where do these value come from?

sparr: I recognize these as the current defaults for the other retry logic. This is something else we could read/share there.

util/http/retry.go
```diff
@@ -0,0 +1,47 @@
+package http
```
Kern--, line 17: nit: missing ltag

fs/artifact_fetcher.go
```diff
@@ -72,26 +77,46 @@ func newRemoteStore(refspec reference.Spec) (*remote.Repository, error) {
...
-		if err != nil {
-			return auth.EmptyCredential, err
-		}
-		if username == "" && secret != "" {
+	rhttpClient := rhttp.NewClient()
+
+	// set the dial timeout and response header timeout to 1 second
+	innerTransport := rhttpClient.HTTPClient.Transport
+	if t, ok := innerTransport.(*http.Transport); ok {
+		t.DialContext = (&net.Dialer{
```
Kern--, line 85: Can you make this same change in `service/resolver/registry.go`? I think that's too aggressive as well.

Kern-- (approved):

djdongjin (approved):

PR #538, opened 2023-04-04
Ignore ErrAlreadyExists when storing index

djdongjin:
**Issue #, if available:**

The `ErrAlreadyExists` should be handled internally by `FetchSociArtifacts` instead of returning to the caller, in which case the caller might ignore the error but the returned soci index is `nil`, causing `nil` dereference. 

**Description of changes:**

Handle `ErrAlreadyExists` in `FetchSociArtifacts` by ignoring it and still returning the index.

**Testing performed:**

make test && make integration

### DIFF
--- a/fs/artifact_fetcher.go
+++ b/fs/artifact_fetcher.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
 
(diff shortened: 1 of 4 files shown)


turan18 (approved):

fs/artifact_fetcher.go
```diff
@@ -215,7 +217,10 @@ func FetchSociArtifacts(ctx context.Context, refspec reference.Spec, indexDesc o
 			if local {
 				return nil
 			}
-			return fetcher.Store(ctx, blob, rc)
+			if err := fetcher.Store(ctx, blob, rc); err != nil && !errors.Is(err, errdef.ErrAlreadyExists) {
+				return err
```
turan18, line 221: nit `return fmt.Errorf("unable to store ztoc in local store: %w", err)`

djdongjin: fix, PTAL @turan18

Kern-- (approved):

turan18 (approved):

PR #499, opened 2023-03-14
Implement network retry integration tests

sparr:
**Issue #, if available:**
Resolves #358 

**Description of changes:**
Added a new set of integration tests that interact with containers with different retry settings while disabling/enabling network access to the registry.

**Testing performed:**
Running with multiple variations of test configurations.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #558, opened 2023-04-12
Create framework to support multiple log monitors

### LINKED ISSUE #520: [FEATURE] Abstract monitoring of log files for soci and containerd
### Description

`testutil.RemoteSnapshotMonitor` and the incoming `testutil.FatalMonitor` implement scanning/monitoring of the soci log output for particular patterns. Before a third such behavior is implemented we should abstract this to a more generally usable form.

At minimum, a log monitor method that can be passed in a function to do the scanning (like `testutil.RemoteSnapshotMonitor.ScanLog` or `testutil.ScanFatalLog`).
Ideally, a single method implementing the outer loop logic shared by them, and a way to register the specific comparisons/checks unique to those functions, so a single log scanner can apply all of the necessary rules in one pass.

Continuing to reproduce the code in question would be faster in the short term but more fragile.

_No response_

sparr:
**Issue #, if available:**
Resolves #520
Also needed for next steps on #499 

**Description of changes:**
Adds a new LogMonitor struct that manages multiple functions to scan stdout/stderr coming from a process. Currently only used for soci-snapshotter-grpc, but likely to be used for containerd later as well.
RemoteSnapshotMonitor and FatalMonitor are now modified (and the latter renamed) to register a function with LogMonitor rather than doing their own scanning directly. This should ensure that every log line (generated after the registration) reaches both of them, rather than one occasionally consuming logs meant for the other. It will also support the need for #499 to monitor logs for network retry, failure, and success.

**Testing performed:**
`make integration` and some manual tweaking of parameters to force failures

### DIFF
--- a/integration/util_test.go
+++ b/integration/util_test.go
@@ -624,7 +624,7 @@ func getReferrers(sh *dockershell.Shell, regConfig registryConfig, imgName, dige
 	return &index, nil
 }
 
-func rebootContainerd(t *testing.T, sh *shell.Shell, customContainerdConfig, customSnapshotterConfig string) *testutil.RemoteSnapshotMonitor {
+func rebootContainerd(t *testing.T, sh *shell.Shell, customContainerdConfig, customSnapshotterConfig string) *testutil.LogMonitor {
 	var (
 		containerdRoot    = "/var/lib/containerd/"
 		containerdStatus  = "/run/containerd/"
@@ -664,8 +664,12 @@ func rebootContainerd(t *testing.T, sh *shell.Shell, customContainerdConfig, cus
 		t.Fatalf("failed to create pipe: %v", err)
 	}
 	reporter := testutil.NewTestingReporter(t)
-	testutil.FatalMonitor(reporter, outR, errR)
-	var m *testutil.RemoteSnapshotMonitor = testutil.NewRemoteSnapshotMonitor(reporter, outR, errR)
+	var m *testutil.LogMonitor = testutil.NewLogMonitor(reporter, outR, errR)
+
+	err = testutil.LogConfirmStartup(m)
+	if err != nil {
+		t.Fatalf("snapshotter startup failed: %v", err)
+	}
 
 	// make sure containerd and soci-snapshotter-grpc are up-and-running
 	sh.Retry(100, "ctr", "snapshots", "--snapshotter", "soci",
--- a/util/testutil/shell.go
+++ b/util/testutil/shell.go
@@ -88,6 +88,49 @@ func (r *TestingReporter) Stderr() io.Writer {
 	return TestingL.Writer()
 }
 
+// LogMonitor manages a list of functions that should scan lines coming from stdout and stderr Readers
+type LogMonitor struct {
+	monitorFuncs map[string]func(string)
+}
+
+// NewLogMonitor creates a LogMonitor for a given pair of stdout and stderr Readers
+func NewLogMonitor(r shell.Reporter, stdout, stderr io.Reader) *LogMonitor {
+	m := &LogMonitor{}
+	m.monitorFuncs = make(map[string]func(string))
+	go m.ScanLog(io.TeeReader(stdout, r.Stdout()))
+	go m.ScanLog(io.TeeReader(stderr, r.Stderr()))
+	return m
+}
+
+// Add registers a new log monitor function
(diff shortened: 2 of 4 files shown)


sparr:
Also one totally unrelated fix to avoid a warning about a capitalized error message

integration/util_test.go
```diff
@@ -664,8 +664,12 @@ func rebootContainerd(t *testing.T, sh *shell.Shell, customContainerdConfig, cus
...
 	}
 	reporter := testutil.NewTestingReporter(t)
-	testutil.FatalMonitor(reporter, outR, errR)
-	var m *testutil.RemoteSnapshotMonitor = testutil.NewRemoteSnapshotMonitor(reporter, outR, errR)
+	var m *testutil.LogMonitor = testutil.NewLogMonitor(reporter, outR, errR)
+
+	err = testutil.LogConfirmStartup(m)
+	if err != nil {
+		t.Fatalf("snapshotter startup failed: %v", err)
+	}
```
djdongjin, line 672: `if err := testutil.LogConfirmStartup(m); err != nil { ...`

util/testutil/shell.go
```diff
@@ -88,41 +88,79 @@ func (r *TestingReporter) Stderr() io.Writer {
...
+type LogMonitor struct {
+	monitorFuncs map[string]func(string)
+}
+
+// NewLogMonitor creates a LogMonitor for a given pair of stdout and stderr Readers
+func NewLogMonitor(r shell.Reporter, stdout, stderr io.Reader) *LogMonitor {
+	m := &LogMonitor{}
+	m.monitorFuncs = make(map[string]func(string))
+	go m.ScanLog(io.TeeReader(stdout, r.Stdout()))
+	go m.ScanLog(io.TeeReader(stderr, r.Stderr()))
```
djdongjin, line 101: When/how these goroutines will exit?

util/testutil/shell.go
```diff
@@ -143,38 +181,37 @@ func (m *RemoteSnapshotMonitor) CheckAllRemoteSnapshots(t *testing.T) {
...
 }
 
-// FatalMonitor scans logs streamed from the specified io.Reader for fatal errors during startup
-func FatalMonitor(r shell.Reporter, stdout, stderr io.Reader) {
+// LogConfirmStartup registers a LogMonitor function to scan until startup succeeds or fails
+func LogConfirmStartup(m *LogMonitor) error {
 	errs := make(chan error, 1)
-	go scanFatalLog(io.TeeReader(stdout, r.Stdout()), errs)
-	go scanFatalLog(io.TeeReader(stderr, r.Stderr()), errs)
+	m.Add("startup", monitorStartup(errs))
```
djdongjin, line 187: Why is `RemoteSnapshotMonitor` a separete struct and `monitorStartup` just a function?

util/testutil/shell.go
```diff
@@ -88,41 +88,79 @@ func (r *TestingReporter) Stderr() io.Writer {
...
+func NewLogMonitor(r shell.Reporter, stdout, stderr io.Reader) *LogMonitor {
+	m := &LogMonitor{}
+	m.monitorFuncs = make(map[string]func(string))
+	go m.ScanLog(io.TeeReader(stdout, r.Stdout()))
+	go m.ScanLog(io.TeeReader(stderr, r.Stderr()))
+	return m
+}
+
+// Add registers a new log monitor function
+func (m *LogMonitor) Add(name string, monitorFunc func(string)) error {
```
turan18, line 106: Could we have this accept an interface instead eg: 

```go
type LogMonitor interface {
   Monitor() error
}
```
m.monitorFuncs[name] = m.Monitor

Kern-- (approved):

turan18 (approved):

djdongjin (approved):

util/testutil/shell.go
sparr: `for scanner.Scan() {` will terminate when the scanner reaches EOF. I believe EOF will occur on stdout and stderr when the spawned process terminates.

util/testutil/shell.go
sparr: Per code review meeting, I expect to have a better answer to this question after implementing 1-2 more impending uses cases for this framework.

integration/util_test.go
sparr: Good catch. I will note this for a fix in a followup PR.

PR #578, opened 2023-04-18
Update pull mode doc

### LINKED ISSUE #488: [Bug] Update the `pull-modes.md` doc
### Description

The doc is no longer accurate:

https://github.com/awslabs/soci-snapshotter/blob/main/docs/pull-modes.md

(fixed) ~**Item 1:**~

~> "For lazy loading the snapshotter expects the user to pass --soci-index-digest, which should be the sha256 of the SOCI index manifest. If it’s not passed, the snapshotter will download and decompress the layers just as the default snapshotter does."~

~1. The  `--soci-index-digest` is optional as we use the referral API.~
~2. The `the container image will be pulled in non-SOCI mode` error message is no longer in our codebase.~

**Item 2:**

> unable to fetch SOCI artifacts: , which will indicate that the container image will not be lazily loaded.

1. It's not clear that, if the image will not be lazily loaded, what will happen instead?

**Item 3:**

> There can be a case, where there’s no zTOC for a specific layer. In this case, there will be an error with the text {"error":"failed to resolve layer which will indicate that the specific layer will be synchronously downloaded at launch time.

1. What is the case where "no zTOC for a specific layer"?
2. Is the case a normal case or a real error?

_No response_

Make the pull doc more accurate. For each step, we should:

1. have what's happening (e.g., lazy loading or not); why it's happening (e.g. because a user-input or a previous step leads to here); how we can know it's happening (e.g. error/log message).
2. check if the log/error message is still up-to-date.

Ideally, we should have a diagram to show the whole workflow.

### Host information

1. OS: 
4. Snapshotter Version: 
5. Containerd Version: 

_No response_

djdongjin:
**Issue #, if available:** fix #488

Items in the issue:

Item1: already fixed.
Item2: added that if a image won't be lazily loaded, the entire image pull will fall back to normal pull (like overlayfs).
Item3: mentioned that "no zTOC for a specific layer" is a normal case and when it happens.

**Description of changes:**

Update the `pull-modes` doc to specify what will happen during `rpull`. 

Also fixed an issue in the getting-started doc where an unexpected `--user` arg is passed to `ctr run`. `ctr run` doesn't require registry auth and the `--user` flag is to pass user-id to the container.

**Testing performed:**

Go through the doc and check the log.

### DIFF
--- a/docs/pull-modes.md
+++ b/docs/pull-modes.md
@@ -2,8 +2,8 @@
 
 soci-snapshotter is a remote snapshotter. It is able to lazily load the contents
 of a container image when a *SOCI index* is present in the remote registry. If
-a SOCI index is not found, it will download and uncompress the layers at launch
-time, just like the default snapshotter does.
+a SOCI index is not found, it will download and uncompress the image layers at
+launch time, just like the default snapshotter does.
 
 SOCI indices can also be "sparse", meaning that any individual layer may not be
 indexed. In that case, that layer will be downloaded at launch time, while the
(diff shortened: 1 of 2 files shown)


sparr (approved):

docs/pull-modes.md
```diff
@@ -12,24 +12,54 @@ indexed layers will be lazily loaded.
...
 a normal overlay layer if it's not.
 
-When you pull an image, there can be the following cases:
+Overall, lazily pulling a container image with soci-snapshotter
+(via the `soci image rpull` command) involves the following steps:
 
-1. For lazy loading the snapshotter accepts an optional flag `--soci-index-digest`,
-which is the sha256 of the SOCI index manifest. If
-not provided, the snapshotter will use the OCI distribution-spec's
+## Step 1: specify SOCI index digest
```
Kern--, line 18: Nit

```suggestion
## Step 1: Optionally specify a SOCI index digest
```

docs/pull-modes.md
```diff
@@ -12,24 +12,54 @@ indexed layers will be lazily loaded.
...
+> validate if this step is successful or not, and [the debug doc](./debug.md#common-scenarios)
+> for the common scenarios where `rpull` might fail and how to debug/fix them.
+
+## Step 2: fetch SOCI artifacts
+
+During `rpull`, on the first layer mount there will be an attempt to download
+and parse the SOCI manifest. If this doesn’t go well, there will be the following
+error in the log: `unable to fetch SOCI artifacts:`, indicating that the
+container image will not be lazily loaded. In this case, the snapshotter will
+fallback to default snapshotter configured (eg: overlayfs) entirely.
```
Kern--, line 42: nit: the wording here is a bit awkward.
```suggestion
fallback to default snapshotter (eg: overlayfs) for all layers.
```

turan18 (approved):

PR #585, opened 2023-04-24
Add Benchmark Workloads

sbuckfelder:
This PR adds the necessary Dockerfiles and scripts to generate the benchmark workloads.

**TESTING**
Manually built the workloads


Kern--:
The set of images that we're benchmarking has diverged from this list. I think we should close this PR and publish those ones instead.

Issue #581, opened 2023-04-20
[FEATURE] Deduplicate setup of retryable http clients


Kern--:
### Description

Currently the ArtifactFetcher (used to pull the SOCI artifacts) and RegistryHosts (used to pull spans) have similar, but not identical setup of retryable http clients. We should consolidate those setups to at least use the same methods, but allow each piece to configure the options separately. 

### Describe the solution you'd like

ArtifactFetcher and RegistryHosts use the same method to setup http clients

PR #597, opened 2023-04-27
Move http client creation to shared method

Kern--:
**Issue #, if available:**
closes https://github.com/awslabs/soci-snapshotter/issues/581

**Description of changes:**
We set up an http client in 2 places - one in the artifactFetcher that's used to pull the SOCI artifacts, and another in the registry hosts which is used to pull spans. This change centralizes their http client setup so that we are using the same defaults in both places.

I didn't attempt to solve all the networking issues here. There is still logic in fs/remote/resolver.go that applies blob config to overwrite the default timeouts/retries. There are also still 2 implementations of networking in the containerd style and the oras-go style. We should eventually consolidate these as much as possible.

**Testing performed:**

integ tests have been flaky on my setup. I want to run in CI to see if I can get easier to parse error messages.


util/http/retry.go
```diff
@@ -27,13 +28,101 @@ import (
...
 
-// defaults based on a target total retry time of at least 5s. 30*((2^8)-1)>5000
 const (
-	DefaultMaxRetries        = 8
-	DefaultMinWaitMsec int64 = 30
-	DefaultMaxWaitMsec int64 = 300000
+	// DefaultDialTimeoutMsec is the default number of milliseconds before timeout while connecting to a remote endpoint. See `TimeoutConfig.DialTimeout`.
+	DefaultDialTimeoutMsec = 3000
+	// DefaultResponseHeaderTimeoutMsec is the default number of milliseconds before timeout while waiting for response header from a remote endpoint. See `TimeoutConfig.ResponseHeaderTimeout`.
+	DefaultResponseHeaderTimeoutMsec = 3000
```
djdongjin, line 35: These 2 are changed from previous 1 sec right?

util/http/retry.go
```diff
@@ -27,13 +28,101 @@ import (
...
+	// DefaultRequestTimeoutMsec is the default number of milliseconds that the entire request can take before timeout. See `TimeoutConfig.RequestTimeout`.
+	DefaultRequestTimeoutMsec = 30_000
```
djdongjin, line 37: Where was this one used previously?

util/http/retry.go
turan18: Yeah - we are bumping the granular timeouts from 1 - 3 seconds.

util/http/retry.go
turan18: This was the default overall `http.Timeout` defined here https://pkg.go.dev/net/http#DefaultTransport

util/http/retry.go
```diff
@@ -27,13 +28,101 @@ import (
...
+// RetryableClientConfig is the complete config for a retryable http client
+type RetryableClientConfig struct {
+	TimeoutConfig
+	RetryConfig
+}
+
+// NewRetriableClientConfig creates a new config with default values.
+// Users of `NewRetryableClient` should use this method to get a new
+// config and then overwrite values if desired.
+func NewRetriableClientConfig() RetryableClientConfig {
```
turan18, line 84: nit: `NewRetriableClientConfig` -> `NewRetryableClientConfig`

turan18: Also, not a part of your change, but could we change `fetching artifacts: Retrying request` to just `Retrying request` within `RetryStrategy`

util/http/retry.go
Kern--: Yeah. Would it be better to have one commit that sets them to 1s and a second commit that increases them to 3s?

util/http/retry.go
Kern--: Yeah, just making the overall default timeout explicit.

util/http/retry.go
djdongjin: Both work for me :) it's not a big change (1 -> 3) after all.

djdongjin (dismissed):
LGTM on CI green

turan18 (approved):

sparr (dismissed):
CI failure appears to be known intermittent network retry problem

turan18 (approved):

sparr (approved):

Issue #584, opened 2023-04-21
[FEATURE] Don't write uncompressed layer to disk when building ztoc


Kern--:
### Description

Currently, building a ztoc for a layer requires more than 3x the layer size of disk space. This is because we copy the [compressed layer into /tmp](https://github.com/awslabs/soci-snapshotter/blob/d458679afaf4009015f69a80a774ee3b7ac58c85/soci/soci_index.go#L410-L415) and [write the uncompressed layer into /tmp](https://github.com/awslabs/soci-snapshotter/blob/d458679afaf4009015f69a80a774ee3b7ac58c85/ztoc/toc_builder.go#L103-L109). We have 2 copies of the compressed layer and 1 copy of the uncompressed layer for each layer in the image. 

https://github.com/awslabs/soci-snapshotter/issues/366 would resolve this issue, but it's blocked on either https://github.com/awslabs/soci-snapshotter/issues/76 or https://github.com/awslabs/soci-snapshotter/issues/481 which are fairly large lifts.

In the meantime, we can at least stop copying the uncompressed archive to disk. We currently do so to convert an `io.Reader` into an `io.SectionReader` for use in `ztoc.positionTrackerReader`. `ztoc.positionTrackerReader` only ever reads starting from the [current position](https://github.com/awslabs/soci-snapshotter/blob/main/ztoc/toc_builder.go#L194) which means that this could be downgraded to an `io.Reader`. We can use the `io.Reader` we already had rather than write it to disk first.

### Describe the solution you'd like

`ztoc.TOCBuilder` doesn't write the uncompressed layer to disk

### Describe any alternative solutions/features you've considered

https://github.com/awslabs/soci-snapshotter/issues/366 solves this as well, but we can deliver this quickly.

PR #598, opened 2023-04-27
Don't save uncompressed layer to disk

Kern--:
**Issue #, if available:**
closes https://github.com/awslabs/soci-snapshotter/issues/584

**Description of changes:**

Before this change, when building a TOC, the SOCI build library would decompress each layer and write the uncompressed tar file to disk. It would then open that temporary file and build a TOC. The reason for this was so that it could use a `io.SectionReader`, but it never actually read anywhere but the next avilable bytes in the stream.

This change skips writing to disk and replaces `io.SectionReader` with `io.Reader` since that's effectively what we were already doing.

**Testing performed:**


ztoc/toc_builder.go
```diff
@@ -99,30 +99,16 @@ func (tb TocBuilder) getFileMetadata(algorithm, filename string) ([]FileMetadata
...
-		return nil, 0, err
-	}
-	return md, compression.Offset(uncompressFileSize), nil
+	return md, uncompressFileSize, nil
 }
 
 // metadataFromTarReader reads every file from tar reader `sr` and creates
 // `FileMetadata` for each file.
-func metadataFromTarReader(sr *io.SectionReader) ([]FileMetadata, error) {
+func metadataFromTarReader(sr io.Reader) ([]FileMetadata, compression.Offset, error) {
```
djdongjin, line 111: nit: `sr` is no longer a `SectionReader`, can change to `r` or `reader`

ztoc/ztoc_test.go
```diff
@@ -301,37 +302,45 @@ func TestZtocGeneration(t *testing.T) {
...
 			},
-			spanSize:  65535,
-			targzName: "testcase0.tar.gz",
-			buildTool: "AWS SOCI CLI",
+			spanSize: 65535,
+			// uncompressedArchiveSize is part of regression testing. The value comes
+			// from running the test before refactoring the way we track this value
+			// to verify that the refactor did not change anything. It doesn't directly
+			// validate that this number is correct.
+			uncompressedArchiveSize: 11975168,
```
djdongjin, line 310: Can we calculate the uncompressed archive size using a helper (read gzip, uncompress, and get the size). I think it's okay even if we need to write the uncompressed tar to /tmp, since it's just for testing.

This makes the test difficult to understand and add more testcases in the future (e.g. for other compression/tar support).

Kern--: This test already takes quite a long time, so I'd prefer not to add another decompression round or disk roundtrip if we can avoid it.

We can switch from using
```
tarReader := testutil.BuildTarGz(tc.tarEntries, gzip.DefaultCompression)
```

to something like this
```
tarReader := testutil.BuildTar(tc.tarEntries)
countReader := positionTrackerReader{r: tarReader}
gzipReader := testutil.BuildGzip(tarReader, gzip.DefaultCompression)
```

But this is using the exact same counting method as the ztoc itself. Does that actually test what we want?

djdongjin: > This test already takes quite a long time, so I'd prefer not to add another decompression round or disk roundtrip if we can avoid it.

Will this add much overhead? I don't have very strong preference but if not, I would prefer calculating from the uncompressed tar file.

> But this is using the exact same counting method as the ztoc itself.

Similarly, since the PR will change the uncompression size calculation to using `positionTrackerReader`, what if we want to add a new testcase later?

djdongjin: also just FYI I'm making these `ztoc_test.go` tests generic to different compressions as part of the uncompression layer support work. So far I've made them support both gzip and tar. This also makes it easier to test zstd when we work on that.

(https://github.com/awslabs/soci-snapshotter/commit/c0c6f68d18a50cdd08f4bb6e489da4566ff624f0#diff-a969359eb5e2603a81addbef098db6684a517aad11bc68c5aeb9e91702a37ae1).

Kern--: What do you think about removing these test changes for now, but keeping https://github.com/awslabs/soci-snapshotter/issues/584 open and adding the checks once your changes land?

djdongjin: I'm okay with either keeping these test changes (I can rebase in my change) or removing.

just want to figure out how to add checks/new tests after switching to use `positionTrackerReader` to calculate uncompressed archive size (as in your PR). Since once it lands we can no longer do `The value comes from running the test before refactoring the way we track`.

Kern--: I have to push another version to change the variable name for your other comment, so I'll remove the tests changes. I'm going to keep #584 open and I'll add a new check after your changes.

sparr (approved):
nit: would cast compression.Offset to a size-like type

turan18 (approved):

djdongjin (approved):

PR #600, opened 2023-04-28
Use docker auth config for `soci push` without `--user`

### LINKED ISSUE #572: [FEATURE] Use docker credentials in SOCI CLI
### Description

`soci-snapshotter-grpc` can load credentials through either the docker credential tooling or the kubernetes credential tooling. The SOCI CLI, on the other hand, requires passing `--user $USERNAME:$PASSWORD` like `ctr`. This path is unfamiliar for customers and unlike their normal container build/push workflows. We should support at least docker credentials in the CLI.

`soci push` can use docker credentials.

It should support both credentials directly in `~/.docker/config.json` and credential helpers.

_No response_

_No response_

sparr:
**Issue #, if available:**
Closes #572 

**Description of changes:**
When `soci push` is used without `--user`, try to read credentials from docker config file, or cred stores/helpers configured there

Also per feedback during review, I moved a lot of code outside of the per-platform loop.

**Testing performed:**
Pushed to ECR using creds stored in `~/.docker/config.json` `auths` section. Pushed to ECR using creds retrieved from a custom creds helper. Confirmed push fails with wrong or missing creds not supplied at the command line.

### DIFF
--- a/cmd/soci/commands/push.go
+++ b/cmd/soci/commands/push.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"sort"
 	"strings"
@@ -30,6 +31,7 @@ import (
 	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/containerd/containerd/cmd/ctr/commands"
 	"github.com/containerd/containerd/reference"
+	dockercliconfig "github.com/docker/cli/cli/config"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/urfave/cli"
 	oraslib "oras.land/oras-go/v2"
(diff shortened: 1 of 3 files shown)


cmd/soci/commands/push.go
```diff
@@ -109,11 +116,30 @@ if they are available in the snapshotter's local content store.
...
+				return err
+			}
+
+			var username string
 			var secret string
-			if i := strings.IndexByte(username, ':'); i > 0 {
-				secret = username[i+1:]
-				username = username[0:i]
+			var ac dockercliconfigtypes.AuthConfig
+			if cliContext.Bool("dockerauth") {
```
Kern--, line 127: I wonder if the default should be to just figure out credentials. For now, that would mean use docker credentials if `--user` isn't specified, but in the future maybe there would be other ways to authenticate.

It might still be useful to have this flag to be explicit about it and later add a debug flag which will log what credentials it's using.

What do you think? Is that too magical?

sparr: I think that if we're going to be intentional about this decision we should probably mimic what nerdctl does, since I think it's the most common such tool that supports docker auth and isn't docker (where it's obviously implicit). I have not investigated nerdctl's behavior.

Kern--: I think it's worth taking a look. I'd rather not introduce a flag that we might immediately remove.

sparr: It looks like nerdctl uses docker config automatically iff username and password aren't specified on the command line. I'll switch to that behavior.

cmd/soci/commands/push.go
```diff
@@ -109,11 +112,28 @@ if they are available in the snapshotter's local content store.
 				return indexDescriptors[i].CreatedAt.Before(indexDescriptors[j].CreatedAt)
 			})
 
-			username := cliContext.String("user")
+			refspec, err := reference.Parse(ref)
```
turan18, line 115: Could we move this outside of the loop?

Kern-- (approved):

cmd/soci/commands/push.go
turan18: re: Actually I think we can move the getting credentials outside the loop entirely.

sparr: This was actually pre-existing code that I moved, but yes, this and half the rest of the loop can be moved outside the loop. I'll do that.

turan18 (approved):

Kern-- (approved):

PR #601, opened 2023-04-28
Skip indexing layers with unrecognized compression

### LINKED ISSUE #591: [FEATURE] Skip indexing unknown compression types
### Description

If the SOCI index builder encounters an unknown compression type it will fail the index build. We should skip unknown layers by default with a flag.

We should skip indexing unknown layer types and indicate why we skipped that layer like:

`ztoc skipped - unsupported compression: zstd`
`ztoc skipped - layer is smaller than min-layer-size: 1000000`

If the builder skips all ztocs, it should return an error

Alternatively, we could fail by default and require a flag to skip the layers. This would add the benefit that it's very explicit when lazy loading isn't going to work, but adds friction when onboarding. 

Similar to https://github.com/awslabs/soci-snapshotter/issues/580

sparr:
**Issue #, if available:**
Closes #591

**Description of changes:**
Replace fatal error on non-gzip compression type with simply skipping the layer.
Emit fatal error if all layers are skipped, including for pre-existing skip conditions which currently is mostly min-layer-size.

**Testing performed:**
Significant troubleshooting and some updates to existing tests.

### DIFF
--- a/soci/soci_index.go
+++ b/soci/soci_index.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/awslabs/soci-snapshotter/ztoc"
@@ -35,7 +36,6 @@ import (
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 
-	"golang.org/x/sync/errgroup"
 	orascontent "oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/errdef"
 )
@@ -296,28 +296,57 @@ func (b *IndexBuilder) Build(ctx context.Context, img images.Image) (*IndexWithM
 		return nil, err
 	}
 
+	// attempt to build a ztoc for each layer
 	sociLayersDesc := make([]*ocispec.Descriptor, len(manifest.Layers))
-	eg, ctx := errgroup.WithContext(ctx)
-	for i, l := range manifest.Layers {
-		i, l := i, l
-		eg.Go(func() error {
-			desc, err := b.buildSociLayer(ctx, l)
-			if err != nil {
-				return fmt.Errorf("could not build zTOC for layer %s: %w", l.Digest.String(), err)
-			}
-			sociLayersDesc[i] = desc
-			return nil
-		})
-	}
-	if err := eg.Wait(); err != nil {
-		return nil, err
+	errChan := make(chan error)
+	go func() {
+		var wg sync.WaitGroup
+		for i, l := range manifest.Layers {
+			wg.Add(1)
+			go func(i int, l ocispec.Descriptor) {
+				defer wg.Done()
+				desc, err := b.buildSociLayer(ctx, l)
+				if err != nil {
+					if err != errUnsupportedLayerFormat {
+						errChan <- err
+					}
+					return
+				}
+				if desc != nil {
+					// index layers must be in some deterministic order
+					// actual layer order used for historic consistency
+					sociLayersDesc[i] = desc
+				}
+			}(i, l)
+		}
+		wg.Wait()
+		close(errChan)
+	}()
+
+	errs := make([]error, 0, len(manifest.Layers))
+
+	for err := range errChan {
+		errs = append(errs, err)
 	}
 
-	ztocsDesc := make([]ocispec.Descriptor, 0, len(manifest.Layers))
+	if len(errs) > 0 {
+		errWrap := errors.New("errors encountered while building soci layers")
+		for _, err := range errs {
+			errWrap = fmt.Errorf("%w; %v", errWrap, err)
+		}
+
+		return nil, errWrap
+	}
+
+	ztocsDesc := make([]ocispec.Descriptor, 0, len(sociLayersDesc))
 	for _, desc := range sociLayersDesc {
 		if desc != nil {
 			ztocsDesc = append(ztocsDesc, *desc)
 		}
+	}
+
+	if len(ztocsDesc) == 0 {
+		return nil, errors.New("no ztocs created, all layers either skipped or produced errors")
 	}
 
 	annotations := map[string]string{
@@ -346,8 +375,8 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 		return nil, errNotLayerType
 	}
 	// check if we need to skip building the zTOC
-	if skipBuildingZtoc(desc, b.config) {
-		fmt.Printf("layer %s -> ztoc skipped\n", desc.Digest)
+	if skip, reason := skipBuildingZtoc(desc, b.config); skip {
+		fmt.Printf("ztoc skipped - layer %s (%s) %s\n", desc.Digest, desc.MediaType, reason)
 		return nil, nil
 	}
 
@@ -365,8 +394,9 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 	}
 
 	if !b.ztocBuilder.CheckCompressionAlgorithm(compressionAlgo) {
-		return nil, fmt.Errorf("layer %s (%s) is compressed in an unsupported format. expect: [tar, gzip, unknown] but got %q: %w",
-			desc.Digest, desc.MediaType, compressionAlgo, errUnsupportedLayerFormat)
+		fmt.Printf("ztoc skipped - layer %s (%s) is compressed in an unsupported format. expect: [tar, gzip, unknown] but got %q\n",
+			desc.Digest, desc.MediaType, compressionAlgo)
+		return nil, errUnsupportedLayerFormat
 	}
 
 	ra, err := b.contentStore.ReaderAt(ctx, desc)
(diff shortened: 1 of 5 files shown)


soci/soci_index.go
```diff
@@ -333,24 +333,46 @@ func (b *IndexBuilder) Build(ctx context.Context, img images.Image) (*IndexWithM
...
+	wg.Wait()
+
+errLoop:
+	for {
+		select {
+		case err := <-errChan:
+			fmt.Print(err)
+		default:
+			break errLoop
+		}
```
djdongjin, line 368: How does this section of code work? E.g., when will the `for loop` and `break errLoop` ends and the code proceeds to, e.g., line 371

djdongjin: will it break and proceed, or break and go to errLoop? (what the `errLoop` label does?)

Kern--: It will break and proceed. Without the `errLoop:` label, the break statement would break out of `select`, but continue to loop in the `for`.

See https://go.dev/doc/effective_go#switch

soci/soci_index.go
```diff
@@ -333,24 +333,46 @@ func (b *IndexBuilder) Build(ctx context.Context, img images.Image) (*IndexWithM
...
 
-	sociLayersDesc := make([]*ocispec.Descriptor, len(manifest.Layers))
-	eg, ctx := errgroup.WithContext(ctx)
-	for i, l := range manifest.Layers {
-		i, l := i, l
-		eg.Go(func() error {
+	sociLayersDesc := make([]*ocispec.Descriptor, 0, len(manifest.Layers))
+	var mu sync.Mutex
+	var wg sync.WaitGroup
+	errChan := make(chan error, len(manifest.Layers))
```
djdongjin, line 339: This change (336-368) adds: (1) a mutex (extra overhead); (2) a waitgroup (manually add/done); (3) error channel (extra complication).

Personally I still prefer the existing `errgroup` approach which is simpler by using index instead of mutex + append when saving `sociLayersDesc`.

I suggest add a smilar `errs` slice (`make(error, 0, len(manifest.Layers))`) that stores all errors, and you can for-loop and print non-nil errors after `eg.Wait()`. This can avoid all the three additions above.

djdongjin (dismissed):

soci/soci_index.go
sparr: I think that `errgroup` can only produce the first error emitted by any goroutine in the group. We need to collect all of the errors. At the moment we're just printing them, which could happen in the goroutine and not get returned and avoid that need, but I anticipate we will actually collect them for error aggregation soon.

sparr: I would find it misleading to use an errgroup that doesn't actually capture errors.
The latest code eliminates the mutex, but still has a bit more code than the errgroup.

soci/soci_index.go
```diff
@@ -334,23 +335,45 @@ func (b *IndexBuilder) Build(ctx context.Context, img images.Image) (*IndexWithM
...
-		eg.Go(func() error {
+		wg.Add(1)
+		go func(i int, l ocispec.Descriptor) {
+			defer wg.Done()
 			desc, err := b.buildSociLayer(ctx, l)
 			if err != nil {
-				return fmt.Errorf("could not build zTOC for layer %s: %w", l.Digest.String(), err)
+				if err != errUnsupportedLayerFormat {
+					errChan <- err
+				}
```
djdongjin, line 350: why do we only send (and print) non-`errUnsupportedLayerFormat` errors?

soci/soci_index.go
```diff
@@ -334,23 +335,45 @@ func (b *IndexBuilder) Build(ctx context.Context, img images.Image) (*IndexWithM
...
-	if err := eg.Wait(); err != nil {
-		return nil, err
+
+	wg.Wait()
+
+errLoop:
+	for {
+		select {
+		case err := <-errChan:
+			fmt.Print(err)
```
djdongjin, line 366: `fmt.Print` -> `fmt.Println`

soci/soci_index.go
```diff
@@ -334,23 +335,45 @@ func (b *IndexBuilder) Build(ctx context.Context, img images.Image) (*IndexWithM
...
+			defer wg.Done()
 			desc, err := b.buildSociLayer(ctx, l)
 			if err != nil {
-				return fmt.Errorf("could not build zTOC for layer %s: %w", l.Digest.String(), err)
+				if err != errUnsupportedLayerFormat {
+					errChan <- err
+				}
+				return
 			}
+			atomic.AddUint64(&layerCount, 1)
```
djdongjin, line 353: It seems a skipped layer due to `errUnsupportedLayerFormat` won't increment this `layerCount`, but a skipped layer due to `min-layer-size` will increment it (line 414 returns `nil`). Is this expected?

soci/soci_index.go
sparr: `errUnsupportedLayerFormat` is no longer an actual error condition at this point; it should not be propagated or interrupt execution. It gets sent through the errChan to be printed so that we have potentially useful troubleshooting output.

soci/soci_index.go
sparr: No. Good catch.

soci/soci_index.go
```diff
@@ -405,8 +428,11 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 	}
 
 	if !b.ztocBuilder.CheckCompressionAlgorithm(compressionAlgo) {
-		return nil, fmt.Errorf("layer %s (%s) is compressed in an unsupported format. expect: [tar, gzip, unknown] but got %q: %w",
-			desc.Digest, desc.MediaType, compressionAlgo, errUnsupportedLayerFormat)
+		if compressionAlgo != compression.Gzip {
+			fmt.Printf("ztoc skipped - layer %s (%s) is compressed in an unsupported format. expect: [tar, gzip, unknown] but got %q",
+				desc.Digest, desc.MediaType, compressionAlgo)
+			return nil, errUnsupportedLayerFormat
+		}
```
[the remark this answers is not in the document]
sparr: This may be a mistake in merge conflict resolution after the other half of uncompressed layer support landed while my PR was in draft. I will revisit this behavior and justify or correct it.

soci/soci_index.go
```diff
@@ -365,8 +387,9 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 	}
 
 	if !b.ztocBuilder.CheckCompressionAlgorithm(compressionAlgo) {
-		return nil, fmt.Errorf("layer %s (%s) is compressed in an unsupported format. expect: [tar, gzip, unknown] but got %q: %w",
-			desc.Digest, desc.MediaType, compressionAlgo, errUnsupportedLayerFormat)
+		fmt.Printf("ztoc skipped - layer %s (%s) is compressed in an unsupported format. expect: [tar, gzip, unknown] but got %q",
```
Kern--, line 390: Now that an unsupported compression algorithm causes a skip instead of an error, I think it would make sense to move this whole check into `skipBuildingZtoc` and return this a reason rather than logging it directly.

also nit: This is missing `\n`

soci/soci_index.go
```diff
@@ -346,8 +368,8 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 		return nil, errNotLayerType
 	}
 	// check if we need to skip building the zTOC
-	if skipBuildingZtoc(desc, b.config) {
-		fmt.Printf("layer %s -> ztoc skipped\n", desc.Digest)
+	if skip, reason := skipBuildingZtoc(desc, b.config); skip {
+		fmt.Printf("ztoc skipped - layer %s (%s) %s\n", desc.Digest, desc.MediaType, reason)
```
Kern--, line 372: I think we might want to keep the format:
```
fmt.Printf("layer %s -> ztoc skipped: %s\n",  desc.Digest, reson)
```

soci/soci_index.go
```diff
@@ -297,29 +297,51 @@ func (b *IndexBuilder) Build(ctx context.Context, img images.Image) (*IndexWithM
...
+				sociLayersDesc[i] = desc
+			}
+		}(i, l)
 	}
-	if err := eg.Wait(); err != nil {
-		return nil, err
+
+	wg.Wait()
+
+errLoop:
```
Kern--, line 324: Before, if we got an error building a ztoc, we would exit with error and stop building the index. Now we just log the errors and continue. Shouldn't we still exit?

soci/soci_index.go
henry118: How about something like the following?

Pros
1. smaller channel buffer
2. start printing errors as soon as they are sent thru.

```
ch := make(chan error, 1)

go func() {
  var wg sync.WaitGroup
  for ... {
    wg.Add(1)
    go func() {
      defer wg.Done()
      ...
      if err != nil {
        ch <- err
      }
    }()
  }
  wg.Wait()
  close(ch)
}()

for err := range ch {
  fmt.Println(err)
}
```

sparr: I don't think that would start printing errors immediately. the wg.Wait() would block until the whole group was finished, but if two things produce errors then their channel write would block because of the buffer size.

djdongjin: wg.Wait() (and channel close) are in a separate goroutine. The main goroutine will just go from `ch := make(..)` to the for loop over error channel until the channel is closed (after wg.Wait())

soci/soci_index.go
sparr: The `ztoc skipped - ` format is taken from your suggestion in #591.

sparr:
Check failure is due to `golangci-lint` exhausting available memory and being killed, after upgrading to golang 1.20. I will re-implement my change not using 1.20 features and open a new issue to get the environment upgraded to use 1.20 successfully.

soci/soci_index.go
sparr: I have implemented most of Henry's suggestion, although the error accumulation is still difference per Kern's feedback below that we should be returning an error. I decided to wrap all the errors, rather than just print them and then return a single generic error.

henry118 (approved):

soci/soci_index.go
henry118: IMO it's fine to keep it as it is. We can open another PR if we believe this logic gets more complicated and needs refactor in the future. This PR does not need to be blocked because of it.

soci/soci_index.go
henry118: Agreed with conforming the feature request.

turan18 (approved):

PR #603, opened 2023-05-02
Support uncompressed image layers

### LINKED ISSUE #579: [FEATURE] Support uncompressed layers
### Description

soci currently does not support layers that are uncompresed.

Support these layers

_No response_

error encountered with an uncompressed layer:
```
soci: could not build zTOC for layer sha256:[cut]: layer sha256:[cut] (application/vnd.docker.image.rootfs.diff.tar) must be compressed by gzip, but got "unknown": unsupported layer format
```

command to build an uncompressed layer:
```shell
docker buildx build --output type=image,compression=uncompressed
```

Current workarounds are to rebuild the image with compressed layers, or to exclude the layer from indexing via `soci create --min-layer-size`

Implementation here should be done in concert with #519 to avoid duplication of effort on adding support for a new "compressor"

djdongjin:
**Issue #, if available:** close #579

**Description of changes:**

This PR adds uncompressed image layer (tar) support to SOCI. It has the following changes:

1. Add `Uncompressed` to flatbuf definition as a new compression enum value (also regenerate flatbuf code).
2. Let soci index builder recognize uncompressed layer and call ztoc builder to build ztoc for tar.
3. Support toc builder and zinfo builder for tar image layer.
4. Implement tar_zinfo.
5. Refactor all ztoc_test to be compression-agnostic, and run these tests for gzip ztoc and tar ztoc.

**Testing performed:**

make test && make integration

manually created an image with uncompressed tar layer and go through the getting_start doc using the image.


ztoc/ztoc_test.go
```diff
@@ -391,15 +440,20 @@ func TestZtocGeneration(t *testing.T) {
 }
 
 func TestZtocSerialization(t *testing.T) {
+	for _, tc := range testZtocs {
+		testZtocSerialization(t, tc.compressionAlgo, tc.tarGenerator)
+	}
+}
+
+func testZtocSerialization(t *testing.T, compressionAlgo string, generator tarGenerator) {
```
sparr, line 448: I think this naming scheme may be confusing to read later.

djdongjin: I think it's common to have a short exported func for preprocessing/sanity check/etc and put detailed logic in a unexported func with the same name. We have similar tests in other test cases (e.g. [here](https://github.com/awslabs/soci-snapshotter/blob/fc45d633a63e4d8512ac2e4b264bd3a19e5a9d47/fs/reader/reader_test.go#L66)).

Any suggestion to improve the readability or reduce confusion while able to test all various ztoc compressions?

soci/soci_index.go
```diff
@@ -395,8 +395,13 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 	if err != nil {
 		return nil, fmt.Errorf("could not determine layer compression: %w", err)
 	}
-	if compressionAlgo != compression.Gzip {
-		return nil, fmt.Errorf("layer %s (%s) must be compressed by gzip, but got %q: %w",
+	// `images.DiffCompression` returns "" if it's a uncompressed layer (tar).
+	if compressionAlgo == "" || compressionAlgo == "unknown" {
+		compressionAlgo = compression.Uncompressed
+	}
```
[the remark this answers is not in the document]
Kern--: I think we should have a separate "unknown" compression algo witch we treat as uncompressed when creating a zinfo. Eventually, we may want to test the actual file to see what compression it is to resolve this, but if we collapse it to uncompressed here then it will be harder to do that later. 

"unknown" is distinct from "unrecognized" -> the latter comes back as an error.

ztoc/compression/tar_zinfo.go
```diff
@@ -0,0 +1,198 @@
...
+// are only to santisfy the `Zinfo` interface and equal to their `uncompressed`-equivalent.
+type TarZinfo struct {
+	version  int32
+	spanSize int64
+	size     int64
+}
+
+// newTarZinfo creates a new instance of `TarZinfo` from serialized bytes.
+func newTarZinfo(zinfoBytes []byte) (*TarZinfo, error) {
+	if len(zinfoBytes) != tarZinfoLength {
```
Kern--, line 46: Should we model this as a flatbuf? We didn't for gzip, but that was probably a mistake.

ztoc/compression/tar_zinfo.go
```diff
@@ -0,0 +1,198 @@
...
+	zinfoBytes := buf.Bytes()
+	if len(zinfoBytes) != tarZinfoLength {
+		return nil, fmt.Errorf("invalid tar zinfo bytes length, expected: %d, actual: %d", len(zinfoBytes), tarZinfoLength)
+	}
+	return zinfoBytes, nil
+}
+
+// MaxSpanID returns the max span ID.
+func (i *TarZinfo) MaxSpanID() SpanID {
+	return SpanID(i.size / i.spanSize)
```
Kern--, line 111: Is this off by one if the size is a multiple of the spanSize?

djdongjin: you mean if `i.size % i.spanSize == 0` then we should off by one right?

yeah I think you're right.

ztoc/compression/tar_zinfo.go
djdongjin: You mean model a corresponding table in `ztoc.fbs` (like `table TarZinfo`)?

I'm not sure if `fbs` itself supports interface abstraction (i.e., a single `Zinfo` points to either a `GzipZinfo` or a `TarZinfo` table, or `ZstdZinfo` in the future).

Also this probably requires more (possibly breaking) changes on the fbs definition and `flatbuf <-> Ztoc struct` conversion. And we'll face one zinfo `GzipZinfo` being `[]byte` and another one `TarZinfo` being a strut or a fbs table.

WDYT?

soci/soci_index.go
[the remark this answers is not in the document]
djdongjin: Sorry I didn't quite understand the suggestion here.

I think we want to recognize the following media types as uncompresed layer:

```golang
// docker:
MediaTypeDockerSchema2Layer            = "application/vnd.docker.image.rootfs.diff.tar"
MediaTypeDockerSchema2LayerForeign     = "application/vnd.docker.image.rootfs.foreign.diff.tar"
// containerd/oci
MediaTypeImageLayer = "application/vnd.oci.image.layer.v1.tar"
MediaTypeImageLayerNonDistributable = "application/vnd.oci.image.layer.nondistributable.v1.tar"
```

For the first 2, `images.DiffCompression` returns `unknown`: https://github.com/containerd/containerd/blob/a7ceac8b63575b08e65f74c852f7dd8b3407ad64/images/mediatypes.go#LL77C24-L77C24

For the last 2, it returns `""`: https://github.com/containerd/containerd/blob/a7ceac8b63575b08e65f74c852f7dd8b3407ad64/images/mediatypes.go#LL93C10-L93C10

So in these 2 cases they both map to `uncompressed` ztoc.

Maybe we can change this to a switch similar to the `DiffCompression` func, e.g.,

```golang
switch mediaType {
case MediaTypeDockerSchema2Layer, MediaTypeDockerSchema2LayerForeign:
    if compressAlgo == "unknown" {
        compressAlgo = "uncompressed"
    }
case MediaTypeImageLayer, MediaTypeImageLayerNonDistributable:
    if compressAlgo == "" {
        compressAlgo = "uncompressed"
    }
}
```

WDYT?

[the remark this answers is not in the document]
Kern--: I think we should keep our options open here. Containerd does auto-detection of the compression algorithm for "unkown" so I think we should try to do the same. For now, we can skip the auto detection and just use no compression, but we should revisit that later.

```
"gzip" => compression.Gzip
"uncompressed" => compression.Uncompressed
"unknown" => compression.Unknown
```

then when creating a zinfo:
```
case Gzip:
	return newGzipZinfo(zinfoBytes)
case Uncompressed:
case Unknown:
	return newTarZinfo(zinfoBytes)
```

ztoc/compression/tar_zinfo.go
Kern--: No, I mean modeling it in compression. We could still serialize it into the ztoc as `[]byte`, but serializing/deserializing into that `[]byte` would be easier to evolve over time.

soci/soci_index.go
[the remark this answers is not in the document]
djdongjin: update unknown handling in a separete commit. PTAL @Kern--

Kern-- (approved):

turan18 (approved):

PR #609, opened 2023-05-03
Add uncompressed archive size check to ztoc tests

Kern--:
**Issue #, if available:**
closes https://github.com/awslabs/soci-snapshotter/issues/584

**Description of changes:**
Adds an uncompressed archive size verification to the ztoc build tests. 

This also fixes a bug in the position tracker where it wouldn't record the number of bytes read on `io.EOF` and thus could truncate the file.

**Testing performed:**
`make test`

### DIFF
--- a/ztoc/toc_builder.go
+++ b/ztoc/toc_builder.go
@@ -184,9 +184,7 @@ type positionTrackerReader struct {
 
 func (p *positionTrackerReader) Read(b []byte) (int, error) {
 	n, err := p.r.Read(b)
-	if err == nil {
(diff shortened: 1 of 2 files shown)


sparr (approved):
I am again motivated toward a refactor to make our use of Offset and Size-like types more intuitive, but that is not a blocker on this PR.

Also, good catch on EOF handling.

> Callers should always process the n > 0 bytes returned before considering the error err. Doing so correctly handles I/O errors that happen after reading some bytes and also both of the allowed EOF behaviors.

djdongjin (approved):

Issue #608, opened 2023-05-03
[FEATURE] Make HTTP timeouts configurable


turan18:
### Description

We currently set custom [HTTP timeouts](https://github.com/awslabs/soci-snapshotter/blob/main/util/http/retry.go#L33-L35). Instead of having hard coded timeouts we should find a way to be able to make them configurable using SOCIs config file. Something similar to the [blob config](https://github.com/awslabs/soci-snapshotter/blob/main/fs/config/config.go#L77-L79). 

### Describe the solution you'd like

A place in the `config.toml` to define HTTP timeouts eg: 

```toml

[timeouts]
  request_timeout=x
  dial_timeout=x
  response_header_timeout=x

``` 

### Describe any alternative solutions/features you've considered

Should we have an overarching config that covers timeouts and retries (having only 1 config instead of a separate blob config)? This would mean that the artifact fetcher and span fetcher share the timeout and retry policy. If so this may be blocked on #224.

Kern--:
We should have a top level config for timeouts and retries.
That config should use the same names (where applicable) as the blob config so that in the future we can replace the blob config with an embedded top level config.

PR #618, opened 2023-05-10
Make HTTP timeouts and retries configurable

turan18:
**Issue #, if available:**

Resolves: #608 

**Description of changes:**

This PR contains 2 changes. The first one moves all the config structs to a separate `config` package. The second moves all the parsing of the configs to the newly created `config` package and also adds an `http` toml property to make the `rhttp` client used to fetch artifacts + spans configurable.

`http` toml property. 

```toml

[http]

[http.timeout]
	dial_timeout_msec=3_000
	response_header_timeout_msec=3_000
	request_timeout_msec=30_000 
[http.retry]
	max_retries=8
	min_wait_msec=30
	max_wait_msec=300_000

```

**Testing performed:**

### DIFF
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -1,0 +1,96 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+/*
+   Copyright 2019 The Go Authors. All rights reserved.
+   Use of this source code is governed by a BSD-style
+   license that can be found in the NOTICE.md file.
+*/
+
+package config
+
+// Config (root) defaults
+const (
+	defaultMetricsNetwork = "tcp"
+)
+
+// ServiceConfig defaults
+const (
+	defaultImageServiceAddress = "/run/containerd/containerd.sock"
+)
+
+// FSConfig defaults
+const (
+	defaultFuseTimeoutSec = 1
+
+	// defaultBgSilencePeriodMsec specifies the amount of time the background fetcher will wait once a new layer comes in
+	// before (re)starting fetches.
+	defaultBgSilencePeriodMsec = 30_000
+
+	// defaultBgFetchPeriodMsec specifies how often the fetch will occur.
+	// The background fetcher will fetch a single span every `defaultFetchPeriod`.
+	defaultBgFetchPeriodMsec = 500
+
+	// defaultBgMaxQueueSize specifies the maximum size of the bg-fetcher work queue i.e., the maximum number
+	// of span managers that can be queued. In case of overflow, the `Add` call
+	// will block until a span manager is removed from the workqueue.
+	defaultBgMaxQueueSize = 100
+
+	// defaultBgMetricEmitPeriodSec is the default amount of interval at which the background fetcher emits metrics
+	defaultBgMetricEmitPeriodSec = 10
+
+	// defaultMountTimeoutSec is the amount of time Mount will time out if a layer can't be resolved.
+	defaultMountTimeoutSec = 30
+
+	// defaultFuseMetricsEmitWaitDurationSec is the amount of time the snapshotter will wait before emitting the metrics for FUSE operation.
+	defaultFuseMetricsEmitWaitDurationSec = 60
+
+	defaultValidIntervalSec = 60
+
+	defaultFetchTimeoutSec = 300
+
+	// defaultDialTimeoutMsec is the default number of milliseconds before timeout while connecting to a remote endpoint. See `TimeoutConfig.DialTimeout`.
+	defaultDialTimeoutMsec = 3_000
+	// defaultResponseHeaderTimeoutMsec is the default number of milliseconds before timeout while waiting for response header from a remote endpoint. See `TimeoutConfig.ResponseHeaderTimeout`.
+	defaultResponseHeaderTimeoutMsec = 3_000
+	// defaultRequestTimeoutMsec is the default number of milliseconds that the entire request can take before timeout. See `TimeoutConfig.RequestTimeout`.
+	defaultRequestTimeoutMsec = 30_000
+
+	// defaults based on a target total retry time of at least 5s. 30*((2^8)-1)>5000
+
+	// defaultMaxRetries is the default number of retries that a retryable request will make. See `RetryConfig.MaxRetries`.
+	defaultMaxRetries = 8
+	// defaultMinWaitMsec is the default minimum number of milliseconds between attempts. See `RetryConfig.MinWait`.
+	defaultMinWaitMsec = 30
+	// defaultMaxWaitMsec is the default maximum number of milliseconds between attempts. See `RetryConfig.MaxWait`.
+	defaultMaxWaitMsec = 300_000
+)
--- a/config/fs.go
+++ b/config/fs.go
@@ -38,15 +38,9 @@
 
 package config
 
-const (
-	// Default path to OCI-compliant CAS
-	SociContentStorePath = "/var/lib/soci-snapshotter-grpc/content/"
+import "time"
 
-	// Default path to snapshotter root dir
-	SociSnapshotterRootPath = "/var/lib/soci-snapshotter-grpc/"
-)
-
-type Config struct {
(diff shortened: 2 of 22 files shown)


fs/config/config.go
```diff
@@ -58,6 +58,8 @@ type Config struct {
 	MountTimeoutSec                int64  `toml:"mount_timeout_sec"`
 	FuseMetricsEmitWaitDurationSec int64  `toml:"fuse_metrics_emit_wait_duration_sec"`
 
+	// RetryableHTTPClientConfig is the complete config for a retryable http client
+	RetryableHTTPClientConfig `toml:"http"`
```
djdongjin, line 62: nit: do we want to change `http` to `rhttp` to make it more clear?

fs/config/config.go
```diff
@@ -126,3 +128,34 @@ type BackgroundFetchConfig struct {
 	// fetcher emits metrics
 	EmitMetricPeriodSec int64 `toml:"emit_metric_period_sec"`
 }
+
+type RetryableHTTPClientConfig struct {
```
djdongjin, line 132: Since they're exported structs, can we add a brief comment to reduce some fmt warnings?

fs/config/config.go
```diff
@@ -126,3 +128,34 @@ type BackgroundFetchConfig struct {
...
+type TimeoutConfig struct {
+	// DialTimeout is the maximum duration that connection can take before a request attempt is timed out.
+	DialTimeout int64 `toml:"dial_timeout_msec"`
+	// ResponseHeaderTimeout is the maximum duration waiting for response headers before a request attempt is timed out.
+	// This starts after the entire request body is uploaded to the remote endpoint and stops when the request headers
+	// are fully read. It does not include reading the body.
+	ResponseHeaderTimeout int64 `toml:"response_header_timeout_msec"`
+	// RequestTimeout is the maximum duration before the entire request attempt is timed out. This starts when the
+	// client starts the connection attempt and ends when the entire response body is read.
+	RequestTimeout int64 `toml:"request_timeout_msec"`
```
sparr, line 148: Can we get "Msec" appended to these struct members, similar to the RetryConfig?

turan18: Sure.

Kern--: Even better: Is it possible to do something with go-toml such that we can actually represent these as durations and do the conversion while serializing/deserializing?

fs/config/config.go
turan18: I stuck with `http`, I think its seems a little more understandable to the user than `rhttp`. Also the `http.retries` implies that the client is "retryable".

fs/config/config.go
turan18: So apparently v1 of `go-toml` supports serializing/deserializing `time.Duration`s but v2 of the package does not - https://github.com/pelletier/go-toml/issues/767. We have a couple options:

1. Create a custom `time.Duration` type that implements `UnmarshalText` - this way we can be compatible with v2 if we ever decide to upgrade - example: https://github.com/pelletier/go-toml/issues/767#issuecomment-1121299700
2. Stick to using v1 with built-in `time.Duration` support
3. Stick to using `int64` to represent duration's (what we currently do). This would make it the responsibility of the users of the config to handle converting them to `time.Duration`s.

config/root.go
```diff
@@ -0,0 +1,100 @@
...
+	// DebugAddress is a Unix domain socket address where the snapshotter exposes /debug/ endpoints.
+	DebugAddress string `toml:"debug_address"`
+
+	// MetadataStore is the type of the metadata store to use.
+	MetadataStore string `toml:"metadata_store" default:"db"`
+}
+type configParser func(*Config)
+
+func NewConfigFromToml(cfgPath string) (*Config, error) {
+	defaultConfigPath := "/etc/soci-snapshotter-grpc/config.toml"
```
djdongjin, line 77: can we move this to the above `const` block as well? 

(nit: also do we want to move the const block to `config/defaults.go`?)

service/plugin/plugin.go
```diff
@@ -58,7 +59,7 @@ import (
 
 // Config represents configuration for the soci snapshotter plugin.
 type Config struct {
-	service.Config
+	config.ServiceConfig
```
djdongjin, line 62: what's the difference between this two config? (`plugin.Config` v.s. `config.ServiceConfig`) (e.g., can we move this to the `config` pkg as well?)

util/http/retry.go
```diff
@@ -42,82 +43,62 @@ const (
...
+	if cfg.RequestTimeout == 0 {
+		cfg.RequestTimeout = DefaultRequestTimeoutMsec
 
-// RetryableClientConfig is the complete config for a retryable http client
-type RetryableClientConfig struct {
-	TimeoutConfig
-	RetryConfig
-}
+	}
+	if cfg.MaxRetries == 0 {
```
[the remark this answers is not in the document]
sparr: I haven't looked through all the new code paths since I last touched this, but at the time the retries value was compared to number of attempts with a <=, so setting max retries to -1 (or any other negative number) would cause zero retries to happen.

config/fs.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	if cfg.BlobConfig.MaxRetries == 0 {
+		cfg.BlobConfig.MaxRetries = cfg.RetryableHTTPClientConfig.RetryConfig.MaxRetries
+	}
+	if cfg.BlobConfig.MinWaitMsec == 0 {
+		cfg.BlobConfig.MinWaitMsec = cfg.RetryableHTTPClientConfig.RetryConfig.MinWaitMsec
+	}
+	if cfg.BlobConfig.MaxWaitMsec == 0 {
+		cfg.BlobConfig.MaxWaitMsec = cfg.RetryableHTTPClientConfig.RetryConfig.MaxWaitMsec
+	}
```
sparr, line 237: Can we use (or make) a constructor for BlobConfig that does this, and call it from here?

config/fs.go
```diff
@@ -0,0 +1,239 @@
...
+		cfg.RetryableHTTPClientConfig.RetryConfig.MaxRetries = defaultMaxRetries
+
+	}
+	if cfg.RetryableHTTPClientConfig.RetryConfig.MinWaitMsec == 0 {
+		cfg.RetryableHTTPClientConfig.RetryConfig.MinWaitMsec = defaultMinWaitMsec
+
+	}
+	if cfg.RetryableHTTPClientConfig.RetryConfig.MaxWaitMsec == 0 {
+		cfg.RetryableHTTPClientConfig.RetryConfig.MaxWaitMsec = defaultMaxWaitMsec
+	}
```
sparr, line 216: Same question as for BlobConfit. Separate constructors would be more readable and reusable.

config/fs.go
```diff
@@ -0,0 +1,239 @@
...
+	RequestTimeoutMsec int64
+}
+
+// RetryableHTTPClientConfig is the complete config for a retryable http client
+type RetryableHTTPClientConfig struct {
+	TimeoutConfig
+	RetryConfig
+}
+
+func parseFSConfig(cfg *Config) {
```
sparr, line 155: Instead of this monolithic function that constructs all the nested structs, can we have separate reusable constructors for each struct and call them from here?

service/plugin/plugin.go
turan18: I'm not sure why we need `plugin.go`. It looks like its purpose to register the snapshotter with containerd and CRI. Although that is already done in `main.go` within the snapshotter - https://github.com/awslabs/soci-snapshotter/blob/main/cmd/soci-snapshotter-grpc/main.go#L230. Also the package itself is never imported. https://github.com/search?q=repo%3Aawslabs%2Fsoci-snapshotter+github.com%2Fawslabs%2Fsoci-snapshotter%2Fservice%2Fplugin&type=code

sparr (approved):
Great work, this has come a long way.

sparr (approved):
Still LGTM.

config/defaults.go
```diff
@@ -0,0 +1,95 @@
...
+
+// ServiceConfig defaults
+const (
+	defaultImageServiceAddress = "/run/containerd/containerd.sock"
+)
+
+// FSConfig defaults
+const (
+	defaultFuseTimeoutSec = 1
+	// Amount of time the background fetcher will wait once a new layer comes in
```
Kern--, line 54: nit: I think it's useful to use the doc comment format (the first word is the name of the identifier) even for unexported constants since it makes it easier to export them later.

config/defaults.go
```diff
@@ -0,0 +1,95 @@
...
+	defaultMountTimeoutSec = 30
+
+	// Amount of time the snapshotter will wait before emitting the metrics for FUSE operation.
+	defaultFuseMetricsEmitWaitDurationSec = 60
+
+	defaultValidIntervalSec = 60
+
+	defaultFetchTimeoutSec = 300
+
+	// DefaultDialTimeoutMsec is the default number of milliseconds before timeout while connecting to a remote endpoint. See `TimeoutConfig.DialTimeout`.
```
Kern--, line 80: nit: doc comment name doesn't match variable name since it's not exported any more.

config/root.go
```diff
@@ -0,0 +1,101 @@
+/*
```
Kern--, line 1: nit: the general convention is to name the main file of a package after the package (i.e. this should be `config/config.go`)

fs/fs.go
```diff
@@ -75,35 +75,9 @@ import (
...
-	defaultBgMetricEmitPeriod = 10 * time.Second
-
-	// Amount of time Mount will time out if a layer can't be resolved.
-	defaultMountTimeout = 30 * time.Second
-
-	// Amount of time the snapshotter will wait before emitting the metrics for FUSE operation.
-	defaultFuseMetricsEmitWaitDuration = 60 * time.Second
-)
-
 var (
```
Kern--, line 78: nit: can these be `const`?

Kern-- (approved):

config/defaults.go
turan18: Agreed. I'll fix it, this was mainly copied over from `fs.go`.

fs/fs.go
turan18: No, since `SelectFirstPolicy` is a func.

Kern-- (approved):

PR #620, opened 2023-05-11
Run integration tests on arm64

### LINKED ISSUE #394: [FEATURE] Run integration tests against ARM64
Currently, our automated tests only run on x86_64 machines. We should also run tests against ARM64.

A matrix of architectures on which to build and test SOCI in our github actions yaml file. We can use something like https://github.com/uraimo/run-on-arch-action to provide the alternate architectures via QEMU. 

1) We could alternatively use self-hosted runners with native arch support. This might be limiting in the future since only X86_64, i386, and ARM64 are supported today. https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#architectures

2) We could write our own qemu wrapper script to build/run to make it easier to run against other architectures locally.

**Additional context**
N/A

sparr:
**Issue #, if available:**
Resolves #394 

**Description of changes:**
Add a test matrix axis for architecture. Use the existing steps for amd64, new steps using https://github.com/uraimo/run-on-arch-action/tree/v2/ for other arch.

**Testing performed:**
None locally. Using Github workflows for testing.

This draft PR has a subset of the full matrix configured to test specific changes prior to the final PR with the full matrix and all steps enabled.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,47 +6,147 @@ on:
   pull_request:
     branches: [ main ]
 
+env:
+  GO_VERSION: '1.18.10'
+
 jobs:
-  check:
-    runs-on: ubuntu-20.04
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          # check-dco will check the last 20 commits, but commit ranges
-          # exclude the start commit in the result, but need that commit
-          # in order to calculate the range. i.e. HEAD~20..HEAD includes
-          # 20 commits, but including HEAD it needs 21 commits.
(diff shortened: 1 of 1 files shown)


sparr:
The current state of this PR is that we are compiling containerd 1.7.0 from source on the runner, in an attempt to get around the bug fixed in https://github.com/containerd/containerd/pull/7636. However this did not resolve the issue, so our next idea is to compile docker from source (since it uses containderd packages during build), but that seems prohibitive and unlikely to be the final blocker. Discussion is ongoing about when and how to proceed.

Issue #475, opened 2023-03-07
[FEATURE] Statically link binaries


Kern--:
### Description

Currently, `soci` and `soci-snapshotter-grpc` are dynamically linked against `zlib` and `glibc`. This requires consumers of SOCI to have a compatible version of `libc` and `libz` in order to use either program and means that there is no guarantee that our released binaries will work for a particular user.

### Describe the solution you'd like

We should offer a mechanism to build static binaries and include those static binaries in our releases. This will remove the need for most runtime dependencies and allow our release binaries to run on any linux system.

### Describe any alternative solutions/features you've considered

We can alternatively be careful to build releases against older versions of glibc and keep our zlib api surface small to avoid ABI changes and unknown symbols.

Kern--:
This should also include an analysis on whether statically linking is the right thing to do. 

For example, go and nerdctl build static binaries by default, but containerd [discourages it](https://github.com/containerd/containerd/blob/main/BUILDING.md#static-binaries)

Kern--:
Using a combination of #476  and adding `-linkmode "external" -extldflags "-static"` to ldflags (as noted by https://mt165.co.uk/blog/static-link-go/) looks pretty good.

dyanmically linked:
```
$ ls -lh out
total 66M
-rwxr-xr-x 1 admin admin 26M Mar  8 23:01 soci
-rwxr-xr-x 1 admin admin 41M Mar  8 23:01 soci-snapshotter-grpc
```

statically linked:
```
$ ls -hl out
total 64M
-rwxr-xr-x 1 admin admin 22M Mar  8 23:51 soci
-rwxr-xr-x 1 admin admin 42M Mar  8 23:51 soci-snapshotter-grpc

# just to convince myself
$ ldd ./out/soci
	not a dynamic executable
$ ldd ./out/soci-snapshotter-grpc 
	not a dynamic executable
```

Kern--:
I do see a bunch of warnings, though. I'm not sure if these are things that we don't call or if we still implicitly depend on glibc.

```
/usr/bin/ld: /tmp/go-link-3718045094/000024.o: in function `mygetgrouplist':
/_/os/user/getgrouplist_unix.go:15: warning: Using 'getgrouplist' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-3718045094/000023.o: in function `mygetgrgid_r':
[...]

PR #623, opened 2023-05-12
add static built binaries

henry118:
Fix: #475 

This patch enables build statically linked binaries with the following command:

```
make STATIC=1
```

Tested on alpine docker image for both commands:

<details>
  <summary>soci</summary>

```
/soci # ./soci create docker.io/library/ubuntu:latest
layer sha256:dbf6a9befcdeecbb8813406afbd62ce81394e3869d84599f19f941aa5c74f3d1 -> ztoc sha256:372417343d4b4b741e6b940c5829a7cb5d934c3edcae0af504e6ff9d5e029cdd
```
</details>

<details>
  <summary>soci-snapshotter-grpc</summary>

```
/soci # ./soci-snapshotter-grpc
{"level":"info","msg":"starting soci-snapshotter-grpc","revision":"3f21ac241646e219b4986578a6f6e59a98d94f64","time":"2023-05-12T05:28:57.964955078Z","version":"3f21ac2"}
{"error":"failed to mount overlay: operation not permitted","level":"fatal","msg":"snapshotter is not supported","time":"2023-05-12T05:28:57.965572192Z"}
```
The snapshotter failed to due to missing overlay fs. But the binary was able to run.
</details>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,19 @@ OUTDIR ?= $(CURDIR)/out
 PKG=github.com/awslabs/soci-snapshotter
 VERSION=$(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
 REVISION=$(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
-GO_LD_FLAGS=-ldflags '-s -w -X $(PKG)/version.Version=$(VERSION) -X $(PKG)/version.Revision=$(REVISION) $(GO_EXTRA_LDFLAGS)'
+
+GO_BUILDTAGS ?=
+ifneq ($(STATIC),)
+	GO_BUILDTAGS += osusergo netgo static_build
+endif
+GO_TAGS=$(if $(GO_BUILDTAGS),-tags "$(strip $(GO_BUILDTAGS))",)
+
+GO_LD_FLAGS=-ldflags '-s -w -X $(PKG)/version.Version=$(VERSION) -X $(PKG)/version.Revision=$(REVISION) $(GO_EXTRA_LDFLAGS)
+ifneq ($(STATIC),)
+	GO_LD_FLAGS += -extldflags "-static"
+endif
+GO_LD_FLAGS+='
+
 SOCI_SNAPSHOTTER_PROJECT_ROOT ?= $(shell pwd)
 LTAG_TEMPLATE_FLAG=-t ./.headers
 FBS_FILE_PATH=$(CURDIR)/ztoc/fbs/ztoc.fbs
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -20,7 +20,19 @@ OUTDIR ?= $(CURDIR)/out
...
-GO_LD_FLAGS=-ldflags '-s -w -X $(PKG)/version.Version=$(VERSION) -X $(PKG)/version.Revision=$(REVISION) $(GO_EXTRA_LDFLAGS)'
+
+GO_BUILDTAGS ?=
+ifneq ($(STATIC),)
+	GO_BUILDTAGS += osusergo netgo static_build
+endif
+GO_TAGS=$(if $(GO_BUILDTAGS),-tags "$(strip $(GO_BUILDTAGS))",)
+
+GO_LD_FLAGS=-ldflags '-s -w -X $(PKG)/version.Version=$(VERSION) -X $(PKG)/version.Revision=$(REVISION) $(GO_EXTRA_LDFLAGS)
+ifneq ($(STATIC),)
```
Kern--, line 31: Can we merge some of these sections to be easier to read? 

Ideally we would only need one static check that sets several variables and then one assignment to `GO_LD_FLAGS`.

henry118: Tags and linker flags are different things. So I think it's actually clearer to keep them separate as different variables.

Kern--: They don't need to be the same variable, I'd just like to see less context needed across lines. Something like:
```
BUILDTAGS ?=
OPTIONAL_LD_FLAGS=
 ifneq ($(STATIC),)
 	GO_BUILDTAGS += osusergo netgo static_build
        OPTIONAL_LD_FLAGS+=-extldflags "-static"
 endif
GO_TAGS=$(if $(GO_BUILDTAGS),-tags "$(strip $(GO_BUILDTAGS))",)
GO_LD_FLAGS=-ldflags '-s -w -X $(PKG)/version.Version=$(VERSION) -X $(PKG)/version.Revision=$(REVISION) $(OPTIONAL_LD_FLAGS) $(GO_EXTRA_LDFLAGS)'
```

henry118: TBH I don't see how your approach is better/clearer than mine. You have to introduce another variable for the sake of combining the logic into a single if condition. which could lead to more readability issues once we have more flags to be customized in the future. Instead grouping logics per individual variable looks cleaner and more understandable, right?

Kern--: I guess I can buy keeping the if separate to group by tags and ldflags instead. The particular piece that I don't like about the current change is that Lines 30-34 are stateful. I think it's worth introducing a new variable there and interpolating with a single assignment to `GO_LD_FLAGS` rather than appending directly.

henry118: It's a common pattern in Makefiles though. That's the reason of the `+=` operator I believe.

Kern--: We discussed a bit offline. I have this preference, but it's not a blocker to merging the change.

Kern-- (approved):

Kern--:
Just for comparison 

Without `STATIC=1`
```
total 65M
-rwxr-xr-x 1 admin admin 25M May 15 18:56 soci
-rwxr-xr-x 1 admin admin 40M May 15 18:56 soci-snapshotter-grpc
```

With `STATIC=1`
```
total 60M
-rwxr-xr-x 1 admin admin 20M May 15 18:55 soci
-rwxr-xr-x 1 admin admin 41M May 15 18:55 soci-snapshotter-grpc
```

djdongjin (approved):

Makefile
```diff
@@ -20,7 +20,19 @@ OUTDIR ?= $(CURDIR)/out
...
+	GO_LD_FLAGS += -extldflags "-static"
+endif
```
sparr, line 33: Could we instead put this into a var and then expand that var in GO_LD_FLAGS? I think that makes the string manipulation easier to understand.

sparr: This is not a blocking request.

sparr: Resolving as already addressed in previous conversation.

PR #648, opened 2023-06-02
Update resolver and refactor code

coderbirju:
**Issue #, if available:**

Problem being addressed - Make the benchmark framework work with any registry supported by SOCI.
1. Refactor parts of the benchmarker that directly call on ECR functionality
    i. We should not have this for instance  - [benchmarkTests.go](https://github.com/awslabs/soci-snapshotter/blob/main/benchmark/benchmarkTests.go#L113)
        
2. Make the benchmarker runnable on any registry. EG: If a user were to run the benchmarker locally with an image that resides in Azure/GCR. In order to do this we will have to find a way to support reading credentials for multiple registries.
    1. One way we can do this is to use some external auth store like the Docker authorizer 
        i. You can see examples of this in the soci cli  - [push.go](https://github.com/awslabs/soci-snapshotter/blob/main/cmd/soci/commands/push.go#LL120C10-L120C25)
        
**Description of changes:**

- Refactored the resolver to use credentials from docker credential store. 
- Renamed aws_utils.go to utils.go to make it more generic
- Removed the use of aws_secret file

**Testing performed:**
- Ran the `make benchmarks-comp` command to verify that the comparison tests are working


sparr:
Use `git commit --signoff` or manually add a signoff line to your commit message to get past the build check failure.

turan18:
It looks like the first commit may have been a left over commit. We can squash this down to 1 commit.

coderbirju:
> It looks like the first commit may have been a left over commit. We can squash this down to 1 commit.

Squashed the commits into one and made reformatted the commit message

coderbirju:
@turan18 @sparr. I closed the pull request by mistake, please review once again

benchmark/framework/utils.go
```diff
@@ -18,35 +18,45 @@ package framework
...
-	secretByteArray, err := os.ReadFile(awsSecretFile)
-	secret := string(secretByteArray)
+func GetResolver(ctx context.Context, imageRef string) remotes.Resolver {
+
+	var username string
+	var secret string
+	refspec, err := reference.Parse(imageRef)
 	if err != nil {
-		panic("Cannot read aws ecr login password")
+		panic("Cannot read registry login password")
```
turan18, line 50: Can we change this to `failed to parse image ref`

benchmark/framework/utils.go
```diff
@@ -18,35 +18,45 @@ package framework
...
 	if err != nil {
-		panic("Cannot read aws ecr login password")
+		panic("Failed to parse image ref")
+	}
+	cf := dockercliconfig.LoadDefaultConfigFile(io.Discard)
+	if cf.ContainsAuth() {
+		if ac, err := cf.GetAuthConfig(refspec.Hostname()); err == nil {
+			username = ac.Username
+			secret = ac.Password
+		}
```
sparr, line 57: Can we fall back to AWS credentials if the registry host is in AWS? Possibly in a way that can be extended later with additional credential sources?

coderbirju: I'm not sure I understand.. credentials are fetched based on the registry host provided by the imageRef, the user is required to have valid credentials stored in the docker config before attempting to run the benchmark tests.

Kern--: I think it's ok to defer to the docker configuration for credentials. To convert AWS credentials (e.g. an instance profile) into ECR credentials, you would configure something like [amazon-ecr-credential-helper](https://github.com/awslabs/amazon-ecr-credential-helper)

Kern--:
nit: this isn't something you need to do for this PR, but I recommend this article on writing git commit messages: https://cbea.ms/git-commit/

This style and variations on this style are pretty common now and help to make logs more readable.

Kern-- (approved):

sparr (approved):

Issue #486, opened 2023-03-09
[FEATURE] Support for podman


iain-macdonald:
### Description

Hi soci-snapshotter maintainers,

We would like to use the SOCI Snapshotter to reduce image pull time for containers running in our remote execution service at [BuildBuddy](https://www.buildbuddy.io/). I’ve managed to get the snapshotter to work in our environment but it requires some modification because we use podman to pull and run container images in production. I’m curious if you would be open to merging a pull request adding support for podman to the snapshotter? It’s worth noting that such support already exists for the stargz-snapshotter on which the soci-snapshotter is based.

I have a forked repository with the changes to support podman at https://github.com/iain-macdonald/soci-snapshotter – the gist of these changes are to add a new binary, [soci-store](https://github.com/iain-macdonald/soci-snapshotter/blob/main/cmd/soci-store/main.go) that mounts a fuse that uses the snapshotter to fetch image layers on demand. This is mostly self-contained but required a couple of other changes in the code, most notably [exposing the layer-to-ztoc mapping in snapshot.go](https://github.com/iain-macdonald/soci-snapshotter/blob/main/snapshot/snapshot.go#L103) to the store so that it has that information.

There are a couple of things I still want to clean up in the forked repo, but I’m mostly just curious if you’re interested in accepting this or if we should plan on maintaining this fork.

Thanks for all of your hard work on this, it looks like it’s going to help us out quite a bit!
Iain

Kern--:
Hey, cool! I'm glad you got it working, but I think podman support is out of scope for this specific project. Our core use-case is containerd (hence building around a containerd snapshotter) and we don't really have the experience with podman to properly maintain an integration. That said, I think there could be a lot of value in working to separate the core SOCI FS piece from the snapshotter implementation so that we can have separate integrations with various runtimes. I think that could be useful for sharing ideas between SOCI and stargz as well if they're interested. 

In the short term, we would be willing to consider the patches you need that would allow you to consume SOCI as a library rather than requiring a fork so long as you're ok with unstable nature of our library APIs.

iain-macdonald:
Thanks for the reply, Kern. If we'll still have to maintain a fork or patch on our side then refactoring the code as you describe isn't as high a priority for me. If I get around to it some time, I'll reach out again. In the mean time, I'll close this feature request. And if you change your mind, let me know!

PR #649, opened 2023-06-02
Add podman support to soci-snapshotter

iain-macdonald:
This PR does a few things:
1. Adds support for podman by porting the [store/fs.go ](https://github.com/containerd/stargz-snapshotter/blob/main/store/fs.go) file (and dependencies) from the stargz-snapshotter into the soci-snapshotter and hooking it up to a new binary that was added, the soci-store.
2. Adds a hook that makes the snapshotter attempt to retrieve artifacts from the local filesystem before the local OCI store and remote container registry. This is used in our set up to support streaming images from registries that we can't write to, the executor will fetch the artifacts elsewhere and store them in the right place locally.
3. Adds a hook that looks up the image-to-index mapping from the local filesystem before the remote container registry. This is, again, used to support streaming images from registries that we can't write to. Note that the mapping is from the hash of the image's config which changes when the image contents change, not its tag which may not change.

This is pretty similar to the version of the soci-snapshotter we have running in production now, though I rebased it to the [awslabs/soci-snapshotter](https://github.com/awslabs/soci-snapshotter) head and cleaned it up a bit.


sparr:
Thanks for the contribution. Did you intend to close the PR? We're looking at making some changes around our use of the OCI content store right now, so this is a good time to be discussing other use cases for different paradigms there.

iain-macdonald:
Hey, I meant to open this PR against a fork of this repo to support our use case which is described in the description, but the GitHub UI fooled me. Feel free to ignore it. If it's helpful for you, I would be happy to provide some more details about our use of the snapshotter and why the local OCI content store isn't necessary for us. I [checked with the team](https://github.com/awslabs/soci-snapshotter/issues/486) a few months ago about adding podman support and was told that it's out of scope for this project, so we're planning to create a fork for our own use, but happy to share more information if it would be helpful for you all.

Sorry for the confusion!

sparr:
No problem. Happy to have more info, and to see the results of your fork.

Issue #655, opened 2023-06-06
[Bug] Blob resolver extra roundtrips


turan18:
### Description

There is exactly one [remotes.Resolve](https://github.com/awslabs/soci-snapshotter/blob/main/fs/layer/layer.go#LL385C26-L385C27) operation per SOCI eligible layer in an image. `remote.Resolve` returns an instance of a `Blob` which stores an instance of an [httpFetcher](https://github.com/awslabs/soci-snapshotter/blob/main/fs/remote/resolver.go#L114) struct which contains the client used to fetch spans. It also contains a [url](https://github.com/awslabs/soci-snapshotter/blob/main/fs/remote/resolver.go#L238) field that is supposed to contain the "real"/redirected URL of the blob so the client does not have to make extra round-trip(s) to fetch the blob (**given that the registry uses a separate backend for storage like ECR**). Currently, this does not happen. What ends up happening is we make a full [Roundtrip](https://github.com/awslabs/soci-snapshotter/blob/main/fs/remote/resolver.go#L304) with a given client to the suggested URL, [**which by default follows up to 10 redirects**](https://pkg.go.dev/net/http#Get). If the request is successful we set the URL to be the [original URL](https://github.com/awslabs/soci-snapshotter/blob/main/fs/remote/resolver.go#L314). 

Although not directly harmful, we do end up incurring a performance hit (most likely in the order of milliseconds) every time we want to make requests to fetch a span (eg: when running the container or background fetching) since we have to make an additional roundtrip(s) to get the blob. **Note: as mentioned above this is only an issue for registries that have 1 or more storage backends**

### Describe the results you expected

We should set the URL to be that last `URL` that a request was sent to to get a 200 response. 
```go
res, err := tr.RoundTrip(req)
if err != nil {
     return "", fmt.Errorf("failed to request: %w", err)
}
if res.Status == 200 {
 url := res.Request.URL.String()
}else{
  ....
}
```

### Host information

1. OS: 
2. Snapshotter Version: 
3. Containerd Version:

Kern--:
For reference, this used to work correctly until stargz added retryable requests:
https://github.com/containerd/stargz-snapshotter/commit/ce4fb004ae05f7e8420fcbe0c714428dc306fc45#diff-f170c75b92738ce5b1a2384c22a812a13f826d52c99d082f070a6cacca129861L62

[...]

PR #667, opened 2023-06-14
Fix blob resolver redirection

turan18:
There is a single retryable `http.Client` embedded as the inner transport of the `http.Client` used to fetch spans and SOCI artifacts. The default behavior of the embedded client is to follow up to 10 redirects. When resolving a layer we attempt to follow any redirects to get the actual location of the blob. However, since by default the client follows redirects we end up making a full `RoundTrip` and end up setting the redirected url to be the original url. We should instead set the redirected url to be the last url that a request was sent to get a 200 response.

**Issue #, if available:**

Resolves: #655 

**Description of changes:**

**Testing performed:**

### DIFF
--- a/fs/remote/resolver.go
+++ b/fs/remote/resolver.go
@@ -292,10 +292,7 @@ func redirect(ctx context.Context, blobURL string, tr http.RoundTripper, timeout
 	}()
 
 	if res.StatusCode/100 == 2 {
-		url = blobURL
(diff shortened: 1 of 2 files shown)


Kern--:
I think we can close this as superseded by #977, right?

turan18:
> I think we can close this as superseded by https://github.com/awslabs/soci-snapshotter/pull/977, right?

Yup, closing.

PR #680, opened 2023-06-20
dependabot ignore k8s.io deps

sparr:
**Description of changes:**
Merging #678 led to dependabot updating #670 #671 #672, which may have been opened in response to #636 being merged.
We ignore k8s dependencies in bump-deps.sh "since they use the latest go version/features that may not be in the go version soci uses" so they should also be ignored by dependabot.

Also made minor regex, comment, and trailing whitespace updates.

**Testing performed:**
Ran update script in fresh go environment locally.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,18 +1,27 @@
 version: 2
 updates:
-  
+ 
   # Automatic upgrade for go modules.
   - package-ecosystem: "gomod"
     directory: "/"
     schedule:
       interval: "daily"
-      
+    ignore:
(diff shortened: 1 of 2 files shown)


turan18 (approved):

Kern--:
What's the condition for undoing this change? IIRC it was dropping support for containerd 1.4.

We should have an issue tracking upgrading k8s dependencies so that this doesn't become a permanent fix.

sparr:
I don't think I was part of that discussion. I'll open an issue to remove this, blocked on... something we can cover during further planning.

Issue #245, opened 2022-12-20
Setup public SOCI indices for benchmarking


Kern--:
**Is your feature request related to a problem? Please describe.**

The benchmarking framework currently requires the caller to [supply a set of images with SOCI artifacts](https://github.com/awslabs/soci-snapshotter/blob/main/Makefile#L134). This makes benchmarking in CI difficult (as we need a place to put the artifacts and authentication to those artifacts) and makes it challenging to repeat benchmarking results across different people. 

**Describe the solution you'd like**

It would be nice if we had a set of public, reference SOCI indices that we could use both in CI and across contributors. This would be a consistent way for developers to verify the impact of changes and for people to independently verify benchmarking results. 

**Describe alternatives you've considered**
https://github.com/awslabs/soci-snapshotter/issues/239 suggested allowing local registries in the benchmarking framework which would solve the CI problem, but not the consistency problem.

**Additional context**

PR #685, opened 2023-06-22
Add cli flags to benchmark tests

coderbirju:
This commit adds three new cli flags to the perfomanceTest and comparisionTests of the benchmark frame work. It also adds default workloads into the binary itself.

- -count allows users to specify the number of runs the benchmark should run
- -show-commit will tag the current commit hash of the soci project to the benchmark results 
- -f accepts a path to the csv file describing custom image details to run the benchmarker on

**Issue #, if available:**

**Description of changes:**
- Added GetDefaultWorkloads() function to utils.go provides hardcoded workloads to the benchmark framework
- Added flags to the main functions of comparison and performance tests

**Testing performed:**
- Verified that the benchmarker runs on default workloads without having to provide any config files
- Verified that the benchmark tests can be customized through the use of flags

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,11 +94,13 @@ integration: build
 
 benchmarks:
 	@echo "$@"
-	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests . && sudo ../bin/PerfTests $(COMMIT) ../singleImage.csv 10
+	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests . && sudo ../bin/PerfTests
+	@cd benchmark/comparisonTest ;  GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/CompTests . && sudo ../bin/CompTests
 
-benchmarks-comp:
+build-benchmarks:
 	@echo "$@"
-	@cd benchmark/comparisonTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/CompTests . && sudo ../bin/CompTests $(COMMIT) ../singleImage.csv 10
+	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests .
+	@cd benchmark/comparisonTest ;  GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/CompTests .
 
 benchmarks-stargz:
 	@echo "$@"
--- a/benchmark/comparisonTest/main.go
+++ b/benchmark/comparisonTest/main.go
@@ -20,7 +20,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"os/exec"
 	"testing"
 
 	"github.com/awslabs/soci-snapshotter/benchmark"
@@ -33,30 +32,35 @@ var (
 
 func main() {
 
-	var numberOfTests int
-	var configCsv string
-	var showCom bool
-	var parseFileAccessPatterns bool
+	var (
+		numberOfTests int
+		configCsv     string
+		showCom       bool
+		imageList     []benchmark.ImageDescriptor
+		err           error
+		commit        string
+	)
 
-	flag.BoolVar(&parseFileAccessPatterns, "parseFileAccess", false, "Parse fuse file access patterns.")
 	flag.BoolVar(&showCom, "show-commit", false, "tag the commit hash to the benchmark results")
 	flag.IntVar(&numberOfTests, "count", 5, "Describes the number of runs a benchmarker should run. Default: 5")
-	flag.StringVar(&configCsv, "f", "../imageList.csv", "Path to a csv file describing image details in this order ['Name','Image ref', 'Ready line', 'manifest ref'].")
+	flag.StringVar(&configCsv, "f", "default", "Path to a csv file describing image details in this order ['Name','Image ref', 'Ready line', 'manifest ref'].")
 
 	flag.Parse()
 
-	var commit string
-
 	if showCom {
-		commit, _ = getCommitHash()
+		commit, _ = benchmark.GetCommitHash()
 	} else {
 		commit = "N/A"
 	}
 
-	imageList, err := benchmark.GetImageListFromCsv(configCsv)
-	if err != nil {
-		errMsg := fmt.Sprintf("Failed to read csv file %s with error:%v\n", configCsv, err)
-		panic(errMsg)
+	if configCsv == "default" {
+		imageList = benchmark.GetDefaultWorkloads()
+	} else {
+		imageList, err = benchmark.GetImageListFromCsv(configCsv)
+		if err != nil {
+			errMsg := fmt.Sprintf("Failed to read csv file %s with error:%v\n", configCsv, err)
+			panic(errMsg)
+		}
 	}
 
 	err = os.Mkdir(outputDir, 0755)
@@ -100,13 +104,4 @@ func main() {
 		Drivers:   drivers,
 	}
 	benchmarks.Run(ctx)
-}
-
-func getCommitHash() (string, error) {
-	cmd := exec.Command("git", "rev-parse", "HEAD")
-	output, err := cmd.Output()
-	if err != nil {
-		return "", err
-	}
-	return string(output), nil
 }
--- a/benchmark/comparisonTest/main.go
+++ b/benchmark/comparisonTest/main.go
@@ -17,9 +17,10 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"os"
-	"strconv"
+	"os/exec"
 	"testing"
 
 	"github.com/awslabs/soci-snapshotter/benchmark"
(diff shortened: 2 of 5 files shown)


benchmark/performanceTest/main.go
```diff
@@ -29,22 +28,30 @@ import (
...
 
 func main() {
-	parseFileAccessPatterns := flag.Bool("parseFileAccess", false, "Parse fuse file access patterns.")
+	var numberOfTests int
+	var configCsv string
+	var showCom bool
+	var parseFileAccessPatterns bool
+	var commit string
+
+	flag.BoolVar(&parseFileAccessPatterns, "parseFileAccess", false, "Parse fuse file access patterns.")
```
turan18, line 41: Doesn't have to be a part of this PR but maybe we can remove the common initialization steps between both binaries into another package.

turan18 (changes requested):

benchmark/comparisonTest/main.go
```diff
@@ -17,31 +17,51 @@
...
 
 func main() {
-	commit := os.Args[1]
-	configCsv := os.Args[2]
-	numberOfTests, err := strconv.Atoi(os.Args[3])
-	if err != nil {
-		errMsg := fmt.Sprintf("Failed to parse number of test %s with error:%v\n", os.Args[3], err)
-		panic(errMsg)
+
+	var numberOfTests int
```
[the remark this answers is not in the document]
coderbirju: Changed

benchmark/utils.go
```diff
@@ -55,3 +56,43 @@ func GetImageListFromCsv(csvLoc string) ([]ImageDescriptor, error) {
...
+func GetCommitHash() (string, error) {
+	cmd := exec.Command("git", "rev-parse", "HEAD")
+	output, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+	return string(output), nil
+}
+
+func GetDefaultWorkloads() ([]ImageDescriptor, error) {
```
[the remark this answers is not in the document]
coderbirju: Updated the code to reflect these changes

benchmark/performanceTest/main.go
```diff
@@ -29,22 +28,30 @@ import (
 )
 
 var (
-	outputDir = "./output"
+	outputDir = "../performanceTest/output"
 )
 
 func main() {
-	parseFileAccessPatterns := flag.Bool("parseFileAccess", false, "Parse fuse file access patterns.")
+	var numberOfTests int
```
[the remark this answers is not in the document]
coderbirju: Updated the code to reflect these changes

benchmark/performanceTest/main.go
coderbirju: Will add this in the next milestone along with the CI pipeline

benchmark/comparisonTest/main.go
```diff
@@ -17,31 +17,51 @@
...
+	var parseFileAccessPatterns bool
+
+	flag.BoolVar(&parseFileAccessPatterns, "parseFileAccess", false, "Parse fuse file access patterns.")
+	flag.BoolVar(&showCom, "show-commit", false, "tag the commit hash to the benchmark results")
+	flag.IntVar(&numberOfTests, "count", 5, "Describes the number of runs a benchmarker should run. Default: 5")
+	flag.StringVar(&configCsv, "f", "default", "Path to a csv file describing image details in this order ['Name','Image ref', 'Ready line', 'manifest ref'].")
+
+	flag.Parse()
+
+	var commit string
```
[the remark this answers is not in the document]
coderbirju: Updated

Makefile
```diff
@@ -94,11 +94,8 @@ integration: build
 
 benchmarks:
 	@echo "$@"
-	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests . && sudo ../bin/PerfTests $(COMMIT) ../singleImage.csv 10
-
-benchmarks-comp:
-	@echo "$@"
-	@cd benchmark/comparisonTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/CompTests . && sudo ../bin/CompTests $(COMMIT) ../singleImage.csv 10
+	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests .
+	@cd benchmark/comparisonTest ;  GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/CompTests .
```
sparr, line 98: This seems to build the tests but no longer run them. Is there another workflow that runs them?

coderbirju: No, there are none. Currently the user has to run the binaries themselves. Do you see value in adding a make target that runs both the benchmarks with the default values? I feel it wouldn't incentivize a user from customizing their benchmark runs

sparr: How useful is it to run with default values and no customization? If there is value to the user there, we should make it easy, and a `make` target is our usual approach for that.

benchmark/utils.go
```diff
@@ -55,3 +56,55 @@ func GetImageListFromCsv(csvLoc string) ([]ImageDescriptor, error) {
...
+	if err != nil {
+		return "", err
+	}
+	return string(output), nil
+}
+
+func GetDefaultWorkloads() []ImageDescriptor {
+	var imageList []ImageDescriptor
+	imageList = append(imageList,
+		ImageDescriptor{
```
sparr, line 72: While we're here, can we rename this? "image descriptor" already means something pretty specific in the container ecosystem.

Makefile
Kern--: We currently have `make test` and `make integration` which run tests. I think it would be good to keep `make benchmarks` similar. We could potentially pull the build out into `make build-benchmarks` which I think makes it pretty clear.

coderbirju: It would probably be the only thing a user would use, since it's straight forward and easy to run. 
However I feel that we should also add some documentation to let users know that the benchmark runs can be customizable..

benchmark/comparisonTest/main.go
```diff
@@ -17,31 +17,52 @@
...
+		numberOfTests           int
+		configCsv               string
+		showCom                 bool
+		parseFileAccessPatterns bool
+		imageList               []benchmark.ImageDescriptor
+		err                     error
+		commit                  string
+	)
+
+	flag.BoolVar(&parseFileAccessPatterns, "parseFileAccess", false, "Parse fuse file access patterns.")
```
Kern--, line 45: to match other flag formats:

```suggestion
	flag.BoolVar(&parseFileAccessPatterns, "parse-file-access", false, "Parse fuse file access patterns.")
```

benchmark/utils.go
```diff
@@ -55,3 +56,55 @@ func GetImageListFromCsv(csvLoc string) ([]ImageDescriptor, error) {
...
+	output, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+	return string(output), nil
+}
+
+func GetDefaultWorkloads() []ImageDescriptor {
+	var imageList []ImageDescriptor
+	imageList = append(imageList,
```
sparr, line 71: Why append instead of just returning a slice literal?

benchmark/utils.go
coderbirju: Are we using ImageDescriptor somewhere else in SOCI?

benchmark/utils.go
```diff
@@ -55,3 +56,55 @@ func GetImageListFromCsv(csvLoc string) ([]ImageDescriptor, error) {
 	}
 	return images, nil
 }
+
+func GetCommitHash() (string, error) {
+	cmd := exec.Command("git", "rev-parse", "HEAD")
```
Kern--, line 61: nit: This can be a follow up, but this can be wrong if you don't build right before running the benchmarks. I wonder if we should embed a version into the benchmark binary instead like we do for the SOCI CLI/snapshotter.

Makefile
coderbirju: Added ```make build-benchmarks``` and ```make benchmarks``` runs both the benchmarks with default values.

benchmark/utils.go
coderbirju: Agreed. Changed to use slice literal

benchmark/comparisonTest/main.go
coderbirju: Updated.

benchmark/utils.go
coderbirju: Understood, Would like to discuss this more and possibly change it in the next code commit.

Makefile
```diff
@@ -94,11 +94,13 @@ integration: build
 
 benchmarks:
 	@echo "$@"
-	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests . && sudo ../bin/PerfTests $(COMMIT) ../singleImage.csv 10
+	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests . && sudo ../bin/PerfTests
+	@cd benchmark/comparisonTest ;  GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/CompTests . && sudo ../bin/CompTests
```
sparr, line 98: This looks like it's going to fail because it won't be back in the original directory before doing the second `cd`

docs/build.md
coderbirju, line 1: @Kern-- Added some updates to the documentation. Could we maybe close [ #245 ](https://github.com/awslabs/soci-snapshotter/issues/245) ?

turan18 (approved):

Kern-- (approved):

sparr (approved):

PR #688, opened 2023-06-26
Replace fmt.Sprintf's with strconv.Format's

austinvazquez:
**Issue #, if available:**
N/A

**Description of changes:**
Non-functional (performance)

Replaces usage of `fmt.Sprintf` for itoa to builtin `strconv.Format`.

**Testing performed:**
```
var numToConvert uint = 100000000

func BenchmarkFmt(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = fmt.Sprintf("%d", numToConvert)
	}
}

func BenchmarkStrconv(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = strconv.FormatUint(uint64(numToConvert), 10)
	}
}
```

```
BenchmarkFmt-8           7690650               154.8 ns/op
BenchmarkStrconv-8      21786693                53.33 ns/op
```

### DIFF
--- a/fs/source/source.go
+++ b/fs/source/source.go
@@ -183,7 +183,7 @@ func AppendDefaultLabelsHandlerWrapper(indexDigest string, wrapper func(images.H
 							c.Annotations = make(map[string]string)
 						}
 
-						c.Annotations[TargetSizeLabel] = fmt.Sprintf("%d", c.Size)
+						c.Annotations[TargetSizeLabel] = strconv.FormatInt(c.Size, 10)
 						c.Annotations[TargetSociIndexDigestLabel] = indexDigest
 
 						remainingLayerDigestsCount := len(strings.Split(c.Annotations[ctdsnapshotters.TargetImageLayersLabel], ","))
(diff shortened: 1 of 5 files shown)


austinvazquez:
More of a micro-optimization, but I figured if anyone appreciated a small performance diff it would be the SOCI folks. :)

fs/source/source.go
```diff
@@ -181,13 +181,13 @@ func AppendDefaultLabelsHandlerWrapper(indexDigest string, wrapper func(images.H
...
 
-						c.Annotations[TargetSizeLabel] = fmt.Sprintf("%d", c.Size)
+						c.Annotations[TargetSizeLabel] = strconv.FormatInt(c.Size, 10)
 						c.Annotations[TargetSociIndexDigestLabel] = indexDigest
 
 						var layerSizes string
 						for _, l := range children[i:] {
 							if images.IsLayerType(l.MediaType) {
-								ls := fmt.Sprintf("%d,", l.Size)
+								ls := strconv.FormatInt(l.Size, 10)
```
Kern--, line 190: this is missing a trailing comma in the resulting string

austinvazquez: Oh nice catch. Let me revert it here.

sparr: `FormatInt` + `+` might still be faster, but no need to benchmark or make that change here.

austinvazquez: Alternatively could have done something like this:

```go
ls := strconv.FormatInt(l.Size, 10) + ","
```

which from profiling is still faster even with string concatenation; however, was hoping to avoid such micro optimizations and only focus on intuitive swaps. It was more of a fun project, not really expecting a noticeable performance diff.

austinvazquez:
Hey, folks going to close this for now to let it bake some more.

PR #694, opened 2023-06-29
Add support for containerd content store

### LINKED ISSUE #423: [FEATURE] Support using containerd's content store

sparr:
**Issue #, if available:**
Closes #423 

**Description of changes:**
This PR adds support for storing SOCI artifacts in the containerd content store (usually found on disk at `/var/lib/containerd/io.containerd.content.v1.content`). This support is optional, enabled for the cli via `soci --namespace foo ...` and for the snapshotter via the following config:
```
[content_store]
type = "containerd"
```

Both places also take a `namespace` parameter, which is passed on and honored by containerd, but soci's ArtifactsDB doesn't understand or honor namespaces so commands like `soci index list` also will not.

This feature adds a dependency on a running containerd service for most `soci` CLI operations and also access to content by `soci-snapshotter-grpc`.

Support for multiple content stores is accomplished by abstracting to a new `store.Store` interface, for which there are two implementations, one wrapping the prior use of the OCI content store, the other wrapping the use of containerd's content store via the GRPC api. Similar abstraction was also required for integration tests and helpers that perform behind-the-scenes access to the content store in order to test content functionality.

Most tests still run on the default content store, which is soci for now, but some integration tests run on both stores to verify functionality of the new store:
- TestSociCreate/test_create_for_drupal
- TestOptimizeConsistentSociArtifact
- TestRebuildArtifactsDB
- TestOverlayFallbackMetric

This implementation is optional and non-default. The primary reason behind this is uncertainty regarding potential database deadlocks. When containerd is communicating with the soci snapshotter, that may happen within a write lock of the underlying metadata and content store database. If soci then attempts to write content into the content store, that would block and deadlock. Further research and investigation into why this is not happening is required before stating that this feature is certainly safe.

**Testing performed:**
`make test`, `make integration`, both including new tests, and also separately both with the default store set to containerd.

### DIFF
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -23,9 +23,9 @@ import (
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/internal"
 	"github.com/awslabs/soci-snapshotter/config"
 	"github.com/awslabs/soci-snapshotter/soci"
+	"github.com/awslabs/soci-snapshotter/soci/store"
 	"github.com/containerd/containerd/cmd/ctr/commands"
 	"github.com/urfave/cli"
-	"oras.land/oras-go/v2/content/oci"
 )
 
 const (
@@ -85,7 +85,8 @@ var CreateCommand = cli.Command{
 		} else if err != nil {
 			return err
 		}
-		blobStore, err := oci.New(config.SociContentStorePath)
+
+		ctx, blobStore, err := store.NewContentStore(ctx, store.WithType(store.ContentStoreType(cliContext.GlobalString("content-store"))), store.WithNamespace(cliContext.GlobalString("namespace")))
 		if err != nil {
 			return err
 		}
--- a/cmd/soci/commands/rebuild_db.go
+++ b/cmd/soci/commands/rebuild_db.go
@@ -19,16 +19,15 @@ package commands
 import (
 	"path/filepath"
 
-	"github.com/awslabs/soci-snapshotter/config"
 	"github.com/awslabs/soci-snapshotter/soci"
+	"github.com/awslabs/soci-snapshotter/soci/store"
 	"github.com/containerd/containerd/cmd/ctr/commands"
 	"github.com/urfave/cli"
-	"oras.land/oras-go/v2/content/oci"
 )
 
 var RebuildDBCommand = cli.Command{
 	Name:  "rebuild-db",
-	Usage: `rebuild the artifacts database. You should use this command after "rpull" so that indices/ztocs can be discovered by commands like "soci index list".`,
+	Usage: `rebuild the artifacts database. You should use this command after "rpull" so that indices/ztocs can be discovered by commands like "soci index list", and after "index rm" when using the containerd content store so that deleted orphaned zTOCs will be forgotten`,
 	Action: func(cliContext *cli.Context) error {
 		client, ctx, cancel, err := commands.NewClient(cliContext)
 		if err != nil {
@@ -40,11 +39,17 @@ var RebuildDBCommand = cli.Command{
 		if err != nil {
 			return err
 		}
-		blobStore, err := oci.New(config.SociContentStorePath)
+		ctx, blobStore, err := store.NewContentStore(ctx, store.WithType(store.ContentStoreType(cliContext.GlobalString("content-store"))), store.WithNamespace(cliContext.GlobalString("namespace")))
 		if err != nil {
 			return err
 		}
-		blobStorePath := filepath.Join(config.SociContentStorePath, "blobs")
+
+		contentStorePath, err := store.GetContentStorePath(store.ContentStoreType(cliContext.GlobalString("content-store")))
+		if err != nil {
+			return err
+		}
+
+		blobStorePath := filepath.Join(contentStorePath, "blobs")
 		return artifactsDb.SyncWithLocalStore(ctx, blobStore, blobStorePath, containerdContentStore)
 	},
 }
--- a/cmd/soci/main.go
+++ b/cmd/soci/main.go
@@ -40,6 +40,7 @@ import (
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/image"
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/index"
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/ztoc"
+	"github.com/awslabs/soci-snapshotter/config"
 	"github.com/awslabs/soci-snapshotter/version"
 	"github.com/containerd/containerd/cmd/ctr/commands/run"
 	"github.com/containerd/containerd/defaults"
@@ -78,6 +79,11 @@ func main() {
 		cli.BoolFlag{
 			Name:  "debug",
 			Usage: "enable debug output",
+		},
+		cli.StringFlag{
+			Name:  "content-store",
+			Usage: "use a specific content store (soci or containerd)",
+			Value: config.DefaultContentStoreType,
 		},
 	}
 
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -45,7 +45,7 @@ const (
 
 // ServiceConfig defaults
 const (
-	defaultImageServiceAddress = "/run/containerd/containerd.sock"
+	DefaultImageServiceAddress = "/run/containerd/containerd.sock"
 )
 
 // FSConfig defaults
@@ -93,4 +93,7 @@ const (
 	defaultMinWaitMsec = 30
 	// defaultMaxWaitMsec is the default maximum number of milliseconds between attempts. See `RetryConfig.MaxWait`.
 	defaultMaxWaitMsec = 300_000
+
+	// DefaultContentStore chooses the soci or containerd content store as the default
+	DefaultContentStoreType = "soci"
 )
--- a/config/fs.go
+++ b/config/fs.go
@@ -38,6 +38,10 @@
 
 package config
 
+import (
+	"github.com/containerd/containerd/namespaces"
+)
+
 type FSConfig struct {
 	HTTPCacheType                  string `toml:"http_cache_type"`
 	FSCacheType                    string `toml:"filesystem_cache_type"`
@@ -58,6 +62,8 @@ type FSConfig struct {
 	FuseConfig `toml:"fuse"`
 
 	BackgroundFetchConfig `toml:"background_fetch"`
+
+	ContentStoreConfig `toml:"content_store"`
 }
 
 // BlobConfig is config for layer blob management.
@@ -152,6 +158,12 @@ type RetryableHTTPClientConfig struct {
 	RetryConfig
 }
 
+// ContentStoreConfig chooses and configures the content store
+type ContentStoreConfig struct {
+	Type      string `toml:"type"`
+	Namespace string `toml:"namespace"`
+}
+
 func parseFSConfig(cfg *Config) {
 	// Parse top level fs config
 	if cfg.MountTimeoutSec == 0 {
@@ -161,7 +173,7 @@ func parseFSConfig(cfg *Config) {
 		cfg.FuseMetricsEmitWaitDurationSec = defaultFuseMetricsEmitWaitDurationSec
 	}
 	// Parse nested fs configs
-	parsers := []configParser{parseFuseConfig, parseBackgroundFetchConfig, parseRetryableHTTPClientConfig, parseBlobConfig}
+	parsers := []configParser{parseFuseConfig, parseBackgroundFetchConfig, parseRetryableHTTPClientConfig, parseBlobConfig, parseContentStoreConfig}
 	for _, p := range parsers {
 		p(cfg)
 	}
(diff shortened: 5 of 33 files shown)


sparr:
https://github.com/awslabs/soci-snapshotter/actions/runs/5455837761 contains tests run with `DefaultContentStoreType` changed to `ContainerdContentStore`

cmd/soci/main.go
```diff
@@ -74,11 +76,17 @@ func main() {
 		cli.DurationFlag{
 			Name:  "timeout",
 			Usage: "timeout for commands",
+			Value: 30 * time.Second,
```
Kern--, line 79: nit: do we want a timeout by default? I believe CTR does not (although it does have a dial timeout to containerd itself which we should inherit)

fs/artifact_fetcher.go
```diff
@@ -177,24 +178,32 @@ func FetchSociArtifacts(ctx context.Context, refspec reference.Spec, indexDesc o
...
-			Digest: indexDesc.Digest,
-			Size:   cw.Size(),
-		}, bytes.NewReader(b))
+		err = localStore.Push(ctx, desc, bytes.NewReader(b))
 
 		if err != nil && !errors.Is(err, errdef.ErrAlreadyExists) {
 			return nil, fmt.Errorf("unable to store index in local store: %w", err)
 		}
+
+		err = store.LabelGCRoot(ctx, localStore, desc)
```
Kern--, line 197: Do we have a race if containerd cleans up the content between the `Push` and `LabelGCRoot`? I think this is what the lease service is for.

config/fs.go
```diff
@@ -58,6 +63,8 @@ type FSConfig struct {
 	FuseConfig `toml:"fuse"`
 
 	BackgroundFetchConfig `toml:"background_fetch"`
+
+	ContentStoreConfig store.Config `toml:"content_store"`
```
turan18, line 67: Should this be moved to the root config?

sparr: That's a good question. On its own merits, I decided not, but the consolidation of the other configs happened while I was working on this. Did we make plans for future configs to also be consolidated?

soci/store/store.go
```diff
@@ -0,0 +1,337 @@
...
+type ContentStoreType string
+
+const (
+	ContainerdContentStoreType ContentStoreType = "containerd"
+	SociContentStoreType       ContentStoreType = "soci"
+	DefaultContentStoreType    ContentStoreType = SociContentStoreType
+)
+
+// GetContentStoreTypes returns a slice of all supported content store types
+func GetContentStoreTypes() []ContentStoreType {
```
Kern--, line 66: nit to be more go-idomatic:

>Go doesn't provide automatic support for getters and setters. There's nothing wrong with providing getters and setters yourself, and it's often appropriate to do so, but it's neither idiomatic nor necessary to put Get into the getter's name.

https://go.dev/doc/effective_go

```suggestion
func ContentStoreTypes() []ContentStoreType {
```

util/testutil/util.go
```diff
@@ -137,3 +137,7 @@ func RandomDigest() string {
 	d := digest.FromBytes(RandomByteData(10))
 	return d.String()
 }
+
+func TrimSha256Prefix(s string) string {
```
Kern--, line 141: nit: I think this is the same as `Digest.Parse(s).Encoded()`

fs/artifact_fetcher.go
sparr: We do have a lease, created when the content store is created. Per discussion I will confirm that the lease is dropped after the push+label and not held so long it expires.

cmd/soci/main.go
sparr: I believe that I encountered situations where our zero timeout was causing operations to immediately abort. I will more thoroughly explore those and document them here.

util/testutil/util.go
sparr: It is, but `Digest.Parse()` requires error handling and `TrimSha256Prefix` is used inline in multiple places.
This function is now new to this PR, just moved. I'm happy to refactor the call sites to use `...Encoded()` and handle any errors.

config/fs.go
sparr: Soliciting input from @Kern--

fs/artifact_fetcher.go
sparr: Latest push moves leases out to a new "Batch" concept which is currently directly analogous to a containerd lease, but I don't want to use that terminology to avoid committing too much to that implementation. A batch is created in each function that pushes content and labels it, and then expired at the end of that function.

cmd/soci/main.go
sparr: That seems to have been an artifact of the development process. This is not necessary and I have removed it.

config/fs.go
turan18: Since we have a [metadata config](https://github.com/awslabs/soci-snapshotter/blob/main/config/config.go#L74) in the root I thought it might make sense to move this there as well. Nonetheless, I don't have a strong preference and I don't think its a blocker for merging.

cmd/soci/commands/rebuild_db.go
```diff
@@ -19,16 +19,15 @@ package commands
...
+	"github.com/awslabs/soci-snapshotter/soci/store"
 	"github.com/containerd/containerd/cmd/ctr/commands"
 	"github.com/urfave/cli"
-	"oras.land/oras-go/v2/content/oci"
 )
 
 var RebuildDBCommand = cli.Command{
 	Name:  "rebuild-db",
-	Usage: `rebuild the artifacts database. You should use this command after "rpull" so that indices/ztocs can be discovered by commands like "soci index list".`,
+	Usage: `rebuild the artifacts database. You should use this command after "rpull" so that indices/ztocs can be discovered by commands like "soci index list", and after "index rm" when using the containerd content store so that deleted orphaned zTOCs will be forgotten`,
```
turan18, line 30: Just want to clarify - If we `soci index rm` an index than it's children will be _eligible_ for deletion. Is there a scenario where we delete an index and then rebuild the db but the ztocs still persists since the GC hasn't run yet?

sparr: Yes. Garbage collection frequency and triggers are controlled by containerd. Here are some scenarios where garbage collection wouldn't be triggered:
1. The user has specified custom gc configuration for their containerd that adds delays, lowers frequency, etc
2. Containerd started less than 100ms ago, that's the default startup delay before the first gc
3. Garbage collection ran recently; default containerd configuration allows gc to take 2% of total runtime so if gc takes 1ms then there's a 49ms gap where no gc will take place.

sparr: There are mechanisms to perform synchronous delete and garbage collection for images and leases, but not for content. https://github.com/containerd/containerd/blob/main/docs/garbage-collection.md#synchronous-garbage-collection
If that ever gets extended to include content, we could take advantage of it.

soci/store/store.go
```diff
@@ -0,0 +1,356 @@
...
+	if err != nil {
+		return ctx, nil, err
+	}
+	switch contentStoreType {
+	case ContainerdContentStoreType:
+		return NewContainerdStore(ctx, storeConfig)
+	case SociContentStoreType:
+		return NewSociStore(ctx)
+	default:
+		return ctx, nil, ErrUnknownContentStoreType(contentStoreType)
```
turan18, line 172: nit: Isn't this check done above in `CanonicalizeContentStoreType`?

soci/store/store.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+}
+
+// GetContentStoreBlobPath returns the bottom level directory for the content store, e.g. "/blobs/sha256".
+func GetContentStoreBlobPath(contentStoreType ContentStoreType) (string, error) {
+	contentStorePath, err := GetContentStorePath(contentStoreType)
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(contentStorePath, "blobs", "sha256"), nil
```
turan18, line 152: nit: Since this is only used in tests maybe we can move it to util/testutl.

soci/store/store.go
sparr: Yes and it's omitted everywhere else that CanonicalizeContentStoreType gets used, but right here I need a default return either in the case or after the case to satisfy the compiler even though we know execution will never reach it.

soci/store/store.go
sparr: Good call. This is very much an implementation detail of the content stores, and not something we want to export.

soci/store/store.go
turan18: I see. Maybe we can just `break`?

cmd/soci/commands/rebuild_db.go
turan18: Cool thanks for the clarification

soci/store/store.go
sparr: I've made the unreachable return more unique and explicit.

sparr:
Further discussion with other soci and containerd devs makes me pretty confident that the database deadlock concerns were based on a misconception. The bbolt database(s) that get locked when containerd calls the snapshotter api (metadate, snapshotter) are not the same bbolt database that is written to by this feature (content store). Once this is merged I will continue efforts to plan and implement changing the default content store.

integration/run_test.go
```diff
@@ -422,3 +423,78 @@ func TestRootFolderPermission(t *testing.T) {
 		t.Fatal("non-root user should be able to `ls /`")
 	}
 }
+
+func TestRunInContentStore(t *testing.T) {
+	imageName := helloImage
+	// regConfig := newRegistryConfig()
```
turan18, line 429: nit: Left over from testing

config/fs.go
Kern--: The consolitdation is in memory only. Was it an accident? We embed each subconfig so they end up appearing on the root config, but we still have the hierarchy in toml.

I think the default content store is a property of the root config. If we were configuring the content store itself, then I think it makes sense as a sub key in toml:

e.g.
```
content_store = "containerd"

[contentstore.containerd]
    address = "..."
```

thoughts?

sparr: I'm wary of adding top level options, but we don't have a standard for where to put config so I'm not entirely opposed. It does make the parsing a little more complex.

sparr: Will consolidate source file and package configs per discussion.
Will not change config struct.

config/defaults.go
```diff
@@ -95,4 +93,7 @@ const (
 	defaultMinWaitMsec = 30
 	// defaultMaxWaitMsec is the default maximum number of milliseconds between attempts. See `RetryConfig.MaxWait`.
 	defaultMaxWaitMsec = 300_000
+
+	// defaultContentStore chooses the soci or containerd content store as the default
```
turan18, line 97: nit: `DefaultContentStore`

cmd/soci/commands/create.go
```diff
@@ -85,7 +85,8 @@ var CreateCommand = cli.Command{
 		} else if err != nil {
 			return err
 		}
-		blobStore, err := oci.New(config.SociContentStorePath)
+
+		ctx, blobStore, err := store.NewContentStore(ctx, store.WithType(store.ContentStoreType(cliContext.GlobalString("content-store"))), store.WithNamespace(cliContext.GlobalString("namespace")))
```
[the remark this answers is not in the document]
sparr: I agree. I intend to do a larger const refactor soon, and will pick these up then.

turan18 (approved):
Looks good once squashed.

PR #711, opened 2023-07-06
Revert removal of layer size fetch, conditional on size annotation unavailable

### LINKED ISSUE #536: [FEATURE] Don't require layer size annotations
### Description

In https://github.com/awslabs/soci-snapshotter/commit/86ddfcdb2c521586177bf3c33eed7e7dcb516f86 we switched from fetching layer sizes to expecting the layer size to be passed to the snapshotter in the form of labels. This works well when the image is launched via the SOCI CLI, but doesn't work well with other tools. E.g. the containerd CRI implementation passes some labels, but not size: https://github.com/containerd/containerd/commit/c1b7bcf395d1334333e254ead5b6cd7dfa84cbb9.

If we took size as an optional label and went to the registry otherwise, we could get the benefit we have today while also gaining kubernetes support. It might be worthwhile to add the layer size label to containerd as well, but that will only help with containerd >= 2.0 since we already missed the last 1.x release.

The snapshotter does not require layer size labels.

We don't have to make this change, but if we don't, then we can't support kubernetes until a version of containerd with the layer size label is published.

_No response_

sparr:
**Issue #, if available:**
fixes #536

**Description of changes:**
Partial revert of https://github.com/awslabs/soci-snapshotter/commit/86ddfcdb2c521586177bf3c33eed7e7dcb516f86

When setting up `newHTTPFetcher`, if the descriptor Size is missing (due to the layer size annotation not being available for the layer) then run the previously-removed function to get the size from the remote host.

**Testing performed:**
Needs new integration test. PR is Draft until then.

### DIFF
--- a/fs/remote/resolver.go
+++ b/fs/remote/resolver.go
@@ -66,7 +66,6 @@ import (
 	rhttp "github.com/hashicorp/go-retryablehttp"
 	digest "github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
-	"github.com/pkg/errors"
 )
 
 const (
@@ -336,13 +335,13 @@ func getLayerSize(ctx context.Context, hf *httpFetcher) (int64, error) {
 	// HEAD request (2020).
 	req, err = http.NewRequestWithContext(ctx, "GET", hf.url, nil)
 	if err != nil {
-		return 0, errors.Wrapf(err, "failed to make request to the registry")
+		return 0, fmt.Errorf("failed to make request to the registry: %w", err)
 	}
 	req.Close = false
 	req.Header.Set("Range", "bytes=0-1")
 	res, err = hf.tr.RoundTrip(req)
 	if err != nil {
-		return 0, errors.Wrapf(err, "failed to request")
+		return 0, fmt.Errorf("failed to request: %w", err)
 	}
 	defer func() {
 		io.Copy(io.Discard, res.Body)
--- a/fs/remote/resolver.go
+++ b/fs/remote/resolver.go
@@ -66,6 +66,7 @@ import (
 	rhttp "github.com/hashicorp/go-retryablehttp"
 	digest "github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


fs/remote/resolver.go
```diff
@@ -303,6 +310,56 @@ func redirect(ctx context.Context, blobURL string, tr http.RoundTripper, timeout
...
+	// HEAD request (2020).
+	req, err = http.NewRequestWithContext(ctx, "GET", hf.url, nil)
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to make request to the registry")
+	}
+	req.Close = false
+	req.Header.Set("Range", "bytes=0-1")
+	res, err = hf.tr.RoundTrip(req)
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to request")
```
sparr, line 345: We seem to have migrated away from `errors.Wrap` in favor of `fmt.Errorf("%w")` since this code was written and removed. I will update to more recent conventions unless someone has input on why this should remain different.

Kern--: `errors.Wrapf` was from before `fmt.Errorf` was part of the standard library. We should use the more recent convention.

Kern-- (commented):
I think we also need to have a check that if the layer sizes label for adjacent layers is empty that we don't error and fetch those ahead of time. 

If I understand it correctly, that code should error if the layer size is set but it's length doesn't match the number of layer digests.

Kern--:
Yeah, I just tested and we need to update `source.FromDefaultLabels` so that it doesn't require `source.TargetSizeLabel` or `source.targetImageLayersSizeLabel`. If they're present, we should use them, but if not, we should just set it to 0 and this code will fetch from the remote.

I think we will want to try to get the size labels upstreamed, though, because it adds several round trips to the pull that don't really need to happen. 

On the plus side, with this patch + the change listed above + the fix for https://github.com/awslabs/soci-snapshotter/issues/725, I was able to successfully rpull a SOCI-fied image with an unmodified nerdctl: `sudo nerdctl pull --snapshotter stargz $IMAGE`

I had to pretend it was a stargz snapshotter to get nerdctl to add the remote labels.

fs/source/source.go
```diff
@@ -111,36 +111,38 @@ func FromDefaultLabels(hosts RegistryHosts) GetSources {
 			return nil, err
 		}
 
+		var targetSize int64
```
Kern--, line 114: nit: the following line will initialize to 0 so this is redundant.

Kern-- (approved):

fs/source/source.go
sparr: The following line creates the string version, not the int.

turan18 (approved):

fs/source/source.go
Kern--: Whoops. Reading too quickly.

PR #718, opened 2023-07-07
added xattr flag when creating soci index

sbuckfelder:
_DO NOT MERGE: Testing still needs to be added, but please feel free to review what is present_

### Why
The listxattr and getxattr fuse operations are quite costly, but the vast majority of files do not have any extended attributes.  Given this we can disable xattrs in the fuse file system.

### What
This PR does two main things:
1) Adds an option to build a soci index that knows at the layer level whether xattrs are present
2) Adds the ability to read this new field in the soci index and disable xattrs in the fuse file system if xattrs are not present in that layer.

**BUILDING THE INDEX**
This PR adds a new **xattr-optimization** flag to the create command (*cmd/soci/commands/create.go*).  This allows the user to opt into a new flow which checks if any files in a given layer have an xattr.  If none of the files have an xattr a new annotation is added to the soci index ocispec.  (*soci/soci_index.go*)

This new feature is turned off by default.

**DISABLING XATTRS IN FUSE**
When the layer is resolved it checks the ocispec for the new annotation (*fs/layer/layer.go*).  If not found it defaults to the original behavior.  This new information is exposed via a new function on the layer interface .  If the annotation indicates there are no xattrs present the fuse.MountOptions are updated to DisableXAttrs (*fs/fs.go*)

### Testing
_Still To Come_

### DIFF
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -32,6 +32,7 @@ const (
 	buildToolIdentifier = "AWS SOCI CLI v0.1"
 	spanSizeFlag        = "span-size"
 	minLayerSizeFlag    = "min-layer-size"
+	xattrFlag           = "xattr-optimization"
 )
 
 // CreateCommand creates SOCI index for an image
@@ -54,6 +55,10 @@ var CreateCommand = cli.Command{
 			Usage: "Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10 MiB.",
 			Value: 10 << 20,
 		},
+		cli.BoolFlag{
+			Name:  xattrFlag,
+			Usage: "If set to true layers without xattr will have xattr disabled, yielding a perfomrance benefit. Does not effect layers where is xattrs are present. Default is False.",
+		},
 	),
 	Action: func(cliContext *cli.Context) error {
 		srcRef := cliContext.Args().Get(0)
@@ -75,6 +80,7 @@ var CreateCommand = cli.Command{
 		}
 		spanSize := cliContext.Int64(spanSizeFlag)
 		minLayerSize := cliContext.Int64(minLayerSizeFlag)
+		xattr := cliContext.Bool(xattrFlag)
 		// Creating the snapshotter's root path first if it does not exist, since this ensures, that
 		// it has the limited permission set as drwx--x--x.
 		// The subsequent oci.New creates a root path dir with too broad permission set.
@@ -104,6 +110,7 @@ var CreateCommand = cli.Command{
 			soci.WithMinLayerSize(minLayerSize),
 			soci.WithSpanSize(spanSize),
 			soci.WithBuildToolIdentifier(buildToolIdentifier),
+			soci.WithXAttrOptimization(xattr),
 		}
 
 		for _, plat := range ps {
--- a/fs/layer/layer.go
+++ b/fs/layer/layer.go
@@ -45,6 +45,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+        "strconv"
 	"sync"
 	"time"
 
@@ -61,6 +62,7 @@ import (
 	"github.com/awslabs/soci-snapshotter/metadata"
 	"github.com/awslabs/soci-snapshotter/util/lrucache"
 	"github.com/awslabs/soci-snapshotter/util/namedmutex"
+	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/awslabs/soci-snapshotter/ztoc"
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/reference"
@@ -102,6 +104,9 @@ type Layer interface {
 
 	// ReadAt reads this layer.
 	ReadAt([]byte, int64, ...remote.Option) (int, error)
+
+        //True if none of the files in the layer has an xattr
+        NoXAttr() bool
 
 	// Done releases the reference to this layer. The resources related to this layer will be
 	// discarded sooner or later. Queries after calling this function won't be serviced.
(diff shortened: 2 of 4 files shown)


cmd/soci/commands/create.go
```diff
@@ -32,6 +32,7 @@ const (
 	buildToolIdentifier = "AWS SOCI CLI v0.1"
 	spanSizeFlag        = "span-size"
 	minLayerSizeFlag    = "min-layer-size"
+	xattrFlag           = "xattr-optimization"
```
sparr, line 35: I imagine we will eventually have more than one of this sort of opt-in optimization available. Instead of a top level flag for each, perhaps something like `--optimizations xattr,foo,bar`?

cmd/soci/commands/create.go
```diff
@@ -54,6 +55,10 @@ var CreateCommand = cli.Command{
 			Usage: "Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10 MiB.",
 			Value: 10 << 20,
 		},
+		cli.BoolFlag{
+			Name:  xattrFlag,
```
sparr, line 59: Other flag names have the whole cli flag name duplicated. I think we should be consistent.
```suggestion
			Name:  xattrOptimizationFlag,
```

cmd/soci/commands/create.go
```diff
@@ -54,6 +55,10 @@ var CreateCommand = cli.Command{
...
+			Usage: "If set to true layers without xattr will have xattr disabled, yielding a perfomrance benefit. Does not effect layers where is xattrs are present. Default is False.",
```
sparr, line 60: nit: "affect"

cmd/soci/commands/create.go
Kern--: Do we want this to be configurable long term? As I understand it, we don't have total confidence that this is safe to do (we think it is, but haven't rigorously tested) and so we shouldn't do it by default. I think we should put it behind a flag that makes it more clear that it might go away in the future like `experimental-xattr-optimization`.

If we decide to keep it optional in the future, we can design the actual flag name.

soci/soci_index.go
```diff
@@ -224,6 +229,14 @@ func WithMinLayerSize(minLayerSize int64) BuildOption {
 	}
 }
 
+// WithXAttr
```
turan18, line 232: nit: ` WithXAttrOptimization`

fs/layer/layer.go
```diff
@@ -103,6 +105,9 @@ type Layer interface {
 	// ReadAt reads this layer.
 	ReadAt([]byte, int64, ...remote.Option) (int, error)
 
+        //True if none of the files in the layer has an xattr
+        NoXAttr() bool
```
Kern--, line 109: nit: I think the positive version is easier to read. I'm ok with it as is if there's pushback on that.
```suggestion
        HasXattrs() bool
```

fs/layer/layer.go
```diff
@@ -338,9 +343,9 @@ func (r *Resolver) Resolve(ctx context.Context, hosts source.RegistryHosts, refs
 	if err != nil {
 		return nil, fmt.Errorf("failed to read layer: %w", err)
 	}
-
+        xattr :=  getXAttrAnnotation(sociDesc)
```
Kern--, line 346:
```suggestion
        hasXattrs :=  hasXAttrAnnotation(sociDesc)
```

fs/layer/layer.go
```diff
@@ -517,6 +529,19 @@ func (l *layer) isClosed() bool {
 	return closed
 }
 
+func getXAttrAnnotation(desc ocispec.Descriptor) bool {
```
Kern--, line 532:
```suggestion
func hasXAttrAnnotation(desc ocispec.Descriptor) bool {
```

soci/soci_index.go
```diff
@@ -51,9 +51,13 @@ const (
 	IndexAnnotationImageLayerDigest = "com.amazon.soci.image-layer-digest"
 	// IndexAnnotationBuildToolIdentifier is the index annotation for build tool identifier
 	IndexAnnotationBuildToolIdentifier = "com.amazon.soci.build-tool-identifier"
+	// IndexAnnotationXattrPresent is the index annotation if the layer has
+        // extended attributes
+	IndexAnnotationXattrPresent = "com.amazon.soci.xattr-present"
```
Kern--, line 56: We use this to decide if we want to disable xattrs, right? The name seems backwards.

I'm not totally sold on this. The effect of this is to disable xattrs. The state is that there are no xattrs. What do you think?
```suggestion
	IndexAnnotationXattrPresent = "com.amazon.soci.disable-xattrs"
```

cmd/soci/commands/create.go
turan18: I think `--disable-xattrs` makes sense - we can then talk about how the optimizations in the description.

soci/soci_index.go
```diff
@@ -203,6 +207,7 @@ type buildConfig struct {
 	buildToolIdentifier string
 	artifactsDb         *ArtifactsDb
 	platform            ocispec.Platform
+        xattr               bool
```
Kern--, line 210:
```suggestion
        disableXattr               bool
```

Kern-- (commented):
Mostly, I think we need to align on has xattrs/does not have xattrs/disable xattrs. 

We should probably be consistent. Even my comments on this aren't consistent, so I don't think it really matters which we choose in the end.

cmd/soci/commands/create.go
sparr: I predict that, long term, we will have *some* optional optimizations. My proposal here is meant to accommodate that, regardless of what they end up being, without us repeatedly changing our command line flag list.

fs/layer/layer.go
sparr: I agree that it's easier to read, but I think there is maintenance and bug prevention value in arranging it for the default to be False so that things Just Work when we default-construct something containing this bool.

Kern--:
Question I'm not sure how to answer: How does this interact with overlayfs xattrs?

Kern--:
Taking a step back and considering https://github.com/awslabs/soci-snapshotter/issues/971, I think the real issue we have is that checking opaqueness is expensive. Each check goes through bbolt to see if a file exists. Since the lower layers are immutable, I think we could probably just push opaqueness up into the in-memory nodes. 

I'm going to run some quick hard-coded benchmarks (normal, disableXattrs hard coded to true, isOpaque hard coded to false) to see how that compares.

Kern--:
Nope, looks like I was wrong:

image

Disabling xattrs entirely really is much more performant than moving opaqueness into memory.

austinvazquez:
Follow-up PR #1021 has been merged. Thanks to @sbuckfelder for the initial research and design here.

PR #721, opened 2023-07-12
Remove trailing '/' from proto's go_package

iain-macdonald:
I thought this would append the name of the proto file to the package and generate something comprehensible, but it actually just uses it as-is which makes it un-importable in go because packages can't end in a '/'.


iain-macdonald:
Argh, sorry, meant to open this against buildbuddy-io/soci-snapshotter. Please ignore.

PR #723, opened 2023-07-12
Add new workflow and benchmark job

coderbirju:
This commit adds comparison-test workflow which runs the comparsion tests once every 2 days. This commit also adds a new make target 'benchmarks-test' which is invoked in the benchmark workflow. The make target runs the performance tests and outputs the results onto the console.

**Issue #, if available:**

**Description of changes:**

**Testing performed:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,10 @@ build-benchmarks:
 	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests .
 	@cd benchmark/comparisonTest ;  GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/CompTests .
 
+benchmarks-perf-test:
+	@echo "$@"
+	@cd benchmark/performanceTest ; sudo rm -rf output ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests . && sudo ../bin/PerfTests -show-commit
+
 benchmarks-stargz:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -102,6 +102,11 @@ build-benchmarks:
 	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests .
 	@cd benchmark/comparisonTest ;  GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/CompTests .
 
+benchmarks-test:
```
Kern--, line 105: Can we call this `benchmarks-perf-test` to be clear about which specific tests are running.

Makefile
```diff
@@ -102,6 +102,11 @@ build-benchmarks:
...
+	@echo "$@"
+	@cd benchmark/performanceTest ; sudo rm -rf output ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests . && sudo ../bin/PerfTests -show-commit
+	@cd benchmark/performanceTest ; cat output/results.json
```
turan18, line 108: For consistency we should have the `benchmark` workflow cat the results  similar to to the comparison test workflow.

Makefile
coderbirju: Updated

Makefile
coderbirju: Done

Kern-- (approved):

PR #750, opened 2023-07-31
Add shell script to check regression

coderbirju:
This commit adds a 'check_regression.sh' script that accepts two results json files and compares the p90 of all the stats and checks for regression. Regression once identified is echo'd on to the console and the shell exits. The workflow that uses this shell script will handle the exit.

The script accepts two files as cli arguments, it compares the second file (newer) p90 value against the p90 values of the first file (older) and checks if the new p90 value is greater than 110% of the older values.
**Usage:** ` ./check_regression.sh <path_to_past>.json <path_to_current>.json ` 

**Json format:** Same as the results generated by the benchmark framework
```json
{
    "commit": "N/A",
    "benchmarkTests": [
     {
      "testName": "SociFullECR-public-ffmpeg",
      "numberOfTests": 1,
      "fullRunStats": {
       "BenchmarkTimes": [
        7.681071871
       ],
       "stdDev": 0,
       "mean": 7.681071871,
       "min": 7.681071871,
       "pct25": 7.681071871,
       "pct50": 7.681071871,
       "pct75": 7.681071871,
       "pct90": 7.681071871,
       "max": 7.681071871
      },
      "pullStats": {
       "BenchmarkTimes": [
        7.523
       ],
       "stdDev": 0,
       "mean": 7.523,
       "min": 7.523,
       "pct25": 7.523,
       "pct50": 7.523,
       "pct75": 7.523,
       "pct90": 7.523,
       "max": 7.523
      },
      "lazyTaskStats": {
       "BenchmarkTimes": [
        0.005
       ],
       "stdDev": 0,
       "mean": 0.005,
       "min": 0.005,
       "pct25": 0.005,
       "pct50": 0.005,
       "pct75": 0.005,
       "pct90": 0.005,
       "max": 0.005
      },
      "localTaskStats": {
       "BenchmarkTimes": [
        0.005
       ],
       "stdDev": 0,
       "mean": 0.005,
       "min": 0.005,
       "pct25": 0.005,
       "pct50": 0.005,
       "pct75": 0.005,
       "pct90": 0.005,
       "max": 0.005
      }
     }    
   ]
}
```

**Output:**
```
Testing - 'SociFullECR-public-ffmpeg'
Testing - 'SociFullECR-public-tensorflow'
Testing - 'SociFullECR-public-tensorflow_gpu'
Testing - 'SociFullECR-public-node'
Testing - 'SociFullECR-public-busybox'
Testing - 'SociFullECR-public-mongo'
Testing - 'SociFullECR-public-rabbitmq'
ERROR: fullRunStats - Current P90 value (12.328233926) exceeds the 110% threshold (11.3611) of the past P90 value (10.328233926)
Comparison failed. Regression detected.
```

**Issue #, if available:**

**Description of changes:**
 
- Added a script check_regression.sh
 
**Testing performed:**

### DIFF
--- a/scripts/check_regression.sh
+++ b/scripts/check_regression.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+# Check if two arguments are provided (paths to past.json and current.json)
+if [ $# -ne 2 ]; then
+    echo "Usage: $0 <path_to_past.json> <path_to_current.json>"
+    exit 1
+fi
+
+# Extract the file paths from command-line arguments
+past_json_path="$1"
+current_json_path="$2"
+
+# Read the contents of past.json and current.json into variables
+past_data=$(cat "$past_json_path")
+current_data=$(cat "$current_json_path")
+
+# Function to compare P90 values for a given statistic
+compare_stat_p90() {
+    local past_value="$1"
+    local current_value="$2"
+    local stat_name="$3"
+
+    # Calculate 110% of the past value
+    local threshold=$(calculate_threshold "$past_value")
+
+    # Compare the current value with the threshold
+    if (( $(awk 'BEGIN {print ("'"$current_value"'" > "'"$threshold"'")}') )); then
+        echo "ERROR: $stat_name - Current P90 value ($current_value) exceeds the 110% threshold ($threshold) of the past P90 value ($past_value)"
+        return 1
+    fi
+
+    return 0
+}
+
+calculate_threshold() {
+    local past_value="$1"
+    awk -v past="$past_value" 'BEGIN { print past * 1.1 }'
+}
+
+# Loop through each object in past.json and compare P90 values with current.json for all statistics
+compare_p90_values() {
+    local past_json="$1"
+    local current_json="$2"
+
+    local test_names=$(echo "$past_json" | jq -r '.benchmarkTests[].testName')
+
+    # Use a flag to indicate if any regression has been detected
+    local regression_detected=0
+
+    for test_name in $test_names; do
+        echo "Checking for regression in '$test_name'"
+        for stat_name in "fullRunStats" "pullStats" "lazyTaskStats" "localTaskStats"; do
+            local past_p90=$(echo "$past_json" | jq -r --arg test "$test_name" '.benchmarkTests[] | select(.testName == $test) | .'"$stat_name"'.pct90')
+            local current_p90=$(echo "$current_json" | jq -r --arg test "$test_name" '.benchmarkTests[] | select(.testName == $test) | .'"$stat_name"'.pct90')
+
+            # Call the compare_stat_p90 function
+            compare_stat_p90 "$past_p90" "$current_p90" "$stat_name" || regression_detected=1
+        done
+    done
+
+    # Check if any regression has been detected and return the appropriate exit code
+    return $regression_detected
+}
+
+# ... (remaining code)
+
+# Call compare_p90_values and store the exit code in a variable
+compare_p90_values "$past_data" "$current_data"
+exit_code=$?
+
+# Check the return status and display appropriate message
+if [ $exit_code -eq 0 ]; then
+    echo "Comparison successful. No regressions detected, all P90 values are within the acceptable range."
+else
+    echo "Comparison failed. Regression detected."
+fi
+
+# Set the final exit code to indicate if any regression occurred
+exit $exit_code


scripts/check_regression.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Function to compare P90 values for a given statistic
+compare_stat_p90() {
+    local past_value="$1"
+    local current_value="$2"
+    local test_name="$3"
+    local stat_name="$4"
+
+    # Calculate 110% of the past value
+    local threshold=$(awk -v past="$past_value" 'BEGIN { print past * 1.1 }')
```
turan18, line 25: If the 10% threshold is the same for the different events (pull, lazyTask, localTask) we can move the calculation out either into the `compare_p90_values()` function or into a standalone function `get_threshold()`.

scripts/check_regression.sh
```diff
@@ -0,0 +1,65 @@
...
+
+    # Compare the current value with the threshold
+    if (( $(awk 'BEGIN {print ("'"$current_value"'" > "'"$threshold"'")}') )); then
+        echo "Test '$test_name' - Threshold: $threshold"
+        echo "ERROR: Test '$test_name' - Current P90 value ($current_value) exceeds 110% of past $stat_name P90 value ($past_value)"
```
turan18, line 30: nit:  echo "ERROR: $stat_name - Current P90 value ($current_value) exceeds the 110% threshold ($threshold) of the past P90 value ($past_value)"

scripts/check_regression.sh
coderbirju: Moved the threshold calculation to a different function.

scripts/check_regression.sh
```diff
@@ -0,0 +1,65 @@
...
+    local current_value="$2"
+    local test_name="$3"
+    local stat_name="$4"
+
+    # Calculate 110% of the past value
+    local threshold=$(awk -v past="$past_value" 'BEGIN { print past * 1.1 }')
+
+    # Compare the current value with the threshold
+    if (( $(awk 'BEGIN {print ("'"$current_value"'" > "'"$threshold"'")}') )); then
+        echo "Test '$test_name' - Threshold: $threshold"
```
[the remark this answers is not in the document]
coderbirju: Updated

scripts/check_regression.sh
coderbirju: Updated

scripts/check_regression.sh
```diff
@@ -0,0 +1,70 @@
...
+
+    local test_names=$(echo "$past_json" | jq -r '.benchmarkTests[].testName')
+
+    for test_name in $test_names; do
+        echo "Testing - '$test_name'"
+        for stat_name in "fullRunStats" "pullStats" "lazyTaskStats" "localTaskStats"; do
+            local past_p90=$(echo "$past_json" | jq -r --arg test "$test_name" '.benchmarkTests[] | select(.testName == $test) | .'"$stat_name"'.pct90')
+            local current_p90=$(echo "$current_json" | jq -r --arg test "$test_name" '.benchmarkTests[] | select(.testName == $test) | .'"$stat_name"'.pct90')
+
+            compare_stat_p90 "$past_p90" "$current_p90" "$test_name" "$stat_name" || return 1
```
turan18, line 54: Instead of returning here I think we should echo that we saw a regression keep but keep iterating through the rest of the image set/stats. This way we don't miss any other regressions.


We can note that we saw a regression with some var and return it at the end:

```
local regression = 0

if [ $? -eq 1 ]; then
$regression = 1
...


if [ $regression -eq 1 ]; then
    echo "Comparison failed. Regression detected."
    exit $regression
```

coderbirju: Yup I think it's a good idea to identify all regressions at once. Updated

coderbirju:
> nit: can you add these strict settings? https://github.com/awslabs/soci-snapshotter/blob/main/scripts/add-ltag.sh#L17

@Kern-- I have excluded the -x flag for this script, the -x command does the following
> Enables a mode of the shell where all executed commands are printed to the terminal.
Running with these checks I noticed that a lot of unwanted logs are printed onto the console, making it hard to read and identify which image had a regression. I think going with the -x flag will make these logs harder to read in our current use case.

scripts/check_regression.sh
```diff
@@ -0,0 +1,86 @@
...
+
+# Loop through each object in past.json and compare P90 values with current.json for all statistics
+compare_p90_values() {
+    local past_json="$1"
+    local current_json="$2"
+
+    local test_names=$(echo "$past_json" | jq -r '.benchmarkTests[].testName')
+
+    # Use a flag to indicate if any regression has been detected
+    local regression_detected=false
```
turan18, line 51: nit: we should set this to be 0 or 1 so we can return it as an exit code

sparr:
Why `sudo` in the usage instructions?

scripts/check_regression.sh
```diff
@@ -0,0 +1,82 @@
...
+
+# Read the contents of past.json and current.json into variables
+past_data=$(cat "$past_json_path")
+current_data=$(cat "$current_json_path")
+
+# Function to compare P90 values for a given statistic
+compare_stat_p90() {
+    local past_value="$1"
+    local current_value="$2"
+    local test_name="$3"
```
turan18, line 23: nit: We don't need test_name here anymore

scripts/check_regression.sh
```diff
@@ -0,0 +1,82 @@
...
+    local past_json="$1"
+    local current_json="$2"
+
+    local test_names=$(echo "$past_json" | jq -r '.benchmarkTests[].testName')
+
+    # Use a flag to indicate if any regression has been detected
+    local regression_detected=0
+
+    for test_name in $test_names; do
+        echo "Testing - '$test_name'"
```
turan18, line 54: nit: `echo "Checking for regression in $test_name ..."`

turan18 (approved):
LGTM, barring nits.

sparr (approved):
nit: `# ... (remaining code)`

scripts/check_regression.sh
```diff
@@ -0,0 +1,81 @@
...
+
+    # Check if any regression has been detected and return the appropriate exit code
+    return $regression_detected
+}
+
+# ... (remaining code)
+
+# Call compare_p90_values and store the exit code in a variable
+compare_p90_values "$past_data" "$current_data"
+exit_code=$?
```
Kern--, line 71: Does this work with `set -e`? Won't it exit when `compare_p90_values` returns 1?

turan18: It should work since its the first part of `||`

re: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html
> The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test in an if statement, part of any command executed in a && or || list except the command following the final && or ||

turan18 (approved):

scripts/check_regression.sh
coderbirju: > The set -e option instructs bash to immediately exit if any command [1] has a non-zero exit status. You wouldn't want to set this for your command-line shell, but in a script it's massively helpful. In all widely used general-purpose programming languages, an unhandled runtime error
whether that's a thrown exception in Java, or a segmentation fault in C, or a syntax error in Python - immediately halts execution of the program; subsequent lines are not executed.

My understanding is that, If a function returns a non-zero value, it is considered a regular function return value, and the script will continue executing without terminating due to set -e.

I tested it out by running the workflow in my fork (console logs below), it seems to be working as expected and not exiting after the first regression is identified.

```
Checking for regression in 'SociFullECR-public-ffmpeg'
Checking for regression in 'SociFullECR-public-tensorflow'
ERROR: fullRunStats - Current P90 value (20.328798902) exceeds the 110% threshold (13.5617) of the past P90 value (12.328798902)
ERROR: localTaskStats - Current P90 value (2.526) exceeds the 110% threshold (1.6786) of the past P90 value (1.526)
Checking for regression in 'SociFullECR-public-tensorflow_gpu'
ERROR: pullStats - Current P90 value (39.567) exceeds the 110% threshold (38.0237) of the past P90 value (34.567)
ERROR: localTaskStats - Current P90 value (2.6) exceeds the 110% threshold (1.7787) of the past P90 value (1.617)
Checking for regression in 'SociFullECR-public-node'
Checking for regression in 'SociFullECR-public-busybox'
Checking for regression in 'SociFullECR-public-mongo'
Checking for regression in 'SociFullECR-public-rabbitmq'
Checking for regression in 'SociFullECR-public-redis'
Comparison failed. Current P90 values exceed 110% of the corresponding past values.
```

PR #751, opened 2023-07-31
Add shell script for benchmark visualization

coderbirju:
This commit adds a visualization_data_converter shell script that accepts a results json file and converts it to a benchmark result compatible format for the visualization tooland this commits also adds 'benchmark_visualization' workflow which calls the shell script on every code merge and updates the data file in the gh-pages branch.

**Issue #, if available:**

**Description of changes:**

**Testing performed:**


turan18:
The PR only adds the script, could we remove the `adds 'benchmark_visualization' workflow which calls the shell script on every code merge and updates the data file in the gh-pages branch.` from the description and add a sample input and output aswell?

scripts/visualization_data_converter.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+
```
Kern--, line 3: nit: can you add these strict settings? https://github.com/awslabs/soci-snapshotter/blob/main/scripts/add-ltag.sh#L17

Here's an explanation of what they do: https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425

coderbirju: Added. I'm assuming we need to add these strict settings to the regression check script as well.

sparr:
It's hard to know whether this script does what it should do without having it in some other context, or some instructions for what to do with it.

coderbirju:
> It's hard to know whether this script does what it should do without having it in some other context, or some instructions for what to do with it.

This PR mainly adds scripts that will be used in the workflows, the script on it's own isn't necessarily of any use other than in context of the workflow.

Kern-- (approved):

scripts/visualization_data_converter.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+set -eux -o pipefail
+
+if [ $# -ne 2 ]; then
+    echo "Usage: $0 <path_to_results.json> <path_to_output_dir.json>"
```
turan18, line 6: nit: remove the the .json from path_to_output_dir

turan18 (approved):

scripts/visualization_data_converter.sh
```diff
@@ -0,0 +1,73 @@
...
+    "extra": "P90"
+  }]'
+
+  # Save the JSON content to the output file
+  echo "$json_content" > "$output_file"
+}
+
+# Parse the JSON using jq
+commit=$(jq -r '.commit' "$input_file")
+tests=$(jq -r '.benchmarkTests | length' "$input_file")
```
sparr, line 57: Do we state a dependency on jq anywhere? Are we just assuming the github actions runners have it? (fortunately they do)
I think this should get called out as an optional dependency in the docs, or at least mentioned in one of the doc creation issues.

scripts/visualization_data_converter.sh
```diff
@@ -0,0 +1,73 @@
...
+  },
+  {
+    "name": "'"$test_name"'-pullTaskDuration",
+    "unit": "Seconds",
+    "value": '"$pull_task_value"',
+    "extra": "P90"
+  }]'
+
+  # Save the JSON content to the output file
+  echo "$json_content" > "$output_file"
```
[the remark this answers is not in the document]
sparr: This is more than a nit but not a blocker. I'm resolving since we have approvals to merge.

scripts/visualization_data_converter.sh
coderbirju: Added an issue for updating the documentation - #764

Issue #744, opened 2023-07-27
[FEATURE] Use custom user-agent when fetching spans


Kern--:
### Description

The SOCI snapshotter uses ranged gets for retrieving spans within a layer. This behavior may look weird to registries since it's unlike the way most other container runtimes pull images. It would be good to use a custom soci-snapshotter user agent to at least point curious registry owners towards this project.

### Describe the solution you'd like

Set a custom user agent: `soci-snapshotter`

Issue #759, opened 2023-08-01
[FEATURE] Design Unified HTTP Client / Authorization


turan18:
### Description

This issue tracks the design doc around unifying the HTTP clients (fetching artifacts and fetching spans) and our authorization logic across both.

sparr:
Can you link the other issues for those specific things if they exist and label them blocked on this one?

turan18:
Sure, I was meaning to yesterday and I forgot.

turan18:
The unified client has been merged: #977. The next step is to create a technical doc regarding our networking design.

PR #762, opened 2023-08-03
Add custom User-Agent when fetching spans

turan18:
**Issue #, if available:**

Resolves: #744 

**Description of changes:**

Add custom `User-Agent` when fetching spans. This could provide some useful information to registry owners.

Add extra link flags to `make test` to ensure that the version is set in our unit tests.

**Testing performed:**

### DIFF
--- a/fs/remote/resolver.go
+++ b/fs/remote/resolver.go
@@ -413,6 +413,7 @@ func (f *httpFetcher) fetch(ctx context.Context, rs []region, retry bool) (multi
 	}
 	req.Header.Add("Range", fmt.Sprintf("bytes=%s", ranges[:len(ranges)-1]))
 	req.Header.Add("Accept-Encoding", "identity")
+	req.Header.Add("User-Agent", socihttp.UserAgent)
 	req.Close = false
 
 	// Recording the roundtrip latency for remote registry GET operation.
--- a/fs/remote/resolver_test.go
+++ b/fs/remote/resolver_test.go
@@ -296,7 +296,7 @@ func TestRetry(t *testing.T) {
 	}
 
 	if tr.retryCount != 4 {
-		t.Fatalf("unxpected retryCount; expected=4 got=%d", tr.retryCount)
+		t.Fatalf("unexpected retryCount; expected=4 got=%d", tr.retryCount)
 	}
 }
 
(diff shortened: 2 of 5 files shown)


fs/remote/resolver_test.go
```diff
@@ -335,3 +335,33 @@ func (r *retryRoundTripper) RoundTrip(req *http.Request) (res *http.Response, er
 	}
 	return
 }
+
+func TestCustomUserAgent(t *testing.T) {
+	// Note: The version is set at link time, so the version will be
+	// "<unknown>" when running unit tests unless
```
turan18, line 341: The version is set at link time, so the version will be "<unknown>" when running unit tests unless we include the vcs information in `go test` like we do for `go build`.

util/http/retry.go
```diff
@@ -18,17 +18,23 @@ package http
...
 
 	"github.com/awslabs/soci-snapshotter/config"
+	"github.com/awslabs/soci-snapshotter/version"
 	"github.com/containerd/containerd/log"
 	rhttp "github.com/hashicorp/go-retryablehttp"
 	"github.com/sirupsen/logrus"
 )
 
+var (
+	UserAgent = fmt.Sprintf("soci-snapshotter/%s", version.Version)
```
turan18, line 35: Both the version/revision are set at [link time](https://github.com/awslabs/soci-snapshotter/blob/main/Makefile#L30). Unless, we also add the ld flags to `go test`, the `version.Version` will not be set and will remain " < undefined > " in our unit tests.

sparr: Good catch. I think it should be the same as it would be for a real build from the same source/commit.

Kern--: I agree with Clarence. I wonder if we should just have another test that verifies `version.Version != "<undefined>"`

turan18: Cool then I think we can switch to having ldflags present in go test.

fs/remote/resolver.go
```diff
@@ -413,6 +413,7 @@ func (f *httpFetcher) fetch(ctx context.Context, rs []region, retry bool) (multi
 	}
 	req.Header.Add("Range", fmt.Sprintf("bytes=%s", ranges[:len(ranges)-1]))
 	req.Header.Add("Accept-Encoding", "identity")
+	req.Header.Add("User-Agent", socihttp.UserAgent)
```
Kern--, line 416: Do we want to set this on all requests to the registry? I think we probably do.

We can defer that for now, but we should mention it in https://github.com/awslabs/soci-snapshotter/issues/759

turan18: All span requests? I think so. 

> We can defer that for now, but we should mention it in https://github.com/awslabs/soci-snapshotter/issues/759

Sure, I can mention that.

Kern-- (approved):

sparr (approved):

Issue #686, opened 2023-06-23
[Bug] OAuth Bearer tokens aren't refreshed correctly


iain-macdonald:
### Description

Hi SOCI-Snapshotter maintainers,

I'm reporting a buggy interaction between the [soci-snapshotter fs/remote/resolver.go](https://github.com/awslabs/soci-snapshotter/blob/ffd306b79f413cb2cd6ebfc05cadfb13cf4d325d/fs/remote/resolver.go#L238) code that resolves remote resources and the [containerd remotes/docker/authorizer.go](https://github.com/containerd/containerd/tree/main/remotes/docker) code that authenticates remote container repository requests. The code in authorizer.go doesn't keep track of OAuth token expiration times, relying on the caller to inform it of failed requests in [AddResponses](https://github.com/containerd/containerd/blob/1a5eaa9ad023470bddde2429f90a600325febfe4/remotes/docker/authorizer.go#L143), which it uses to determine when to invalidate OAuth tokens. The logic for doing that is in [invalidAuthorization](https://github.com/containerd/containerd/blob/1a5eaa9ad023470bddde2429f90a600325febfe4/remotes/docker/authorizer.go#L331) which invalidates tokens if there is an "error" in the auth challenge and the two more recent provided HTTP Responses are for the same HTTP method and URL. In other words, the client has to receive two successive authorization failures for the same URL and only then will the authorizer will refresh its cached auth token.

[...]

PR #763, opened 2023-08-03
Fix bearer token invalidation bug by calling AddResponses() twice on 401 errors

iain-macdonald:
**Issue #, if available:**
https://github.com/awslabs/soci-snapshotter/issues/686

**Description of changes:**
There's a pretty detailed description of the bug that this PR fixes at [https://github.com/awslabs/soci-snapshotter/issues/686](https://github.com/awslabs/soci-snapshotter/issues/686). This fix addresses the first sentence of the second paragraph there:

> However, the [soci-snapshotter code that provides HTTP Responses to the authorizer](https://github.com/awslabs/soci-snapshotter/blob/ffd306b79f413cb2cd6ebfc05cadfb13cf4d325d/fs/remote/resolver.go#L256) only ever provides the most recent HTTP Response, so the logic above is never exercised. Instead, it always hits the n == 1 [condition in invalidAuthorization](https://github.com/containerd/containerd/blob/1a5eaa9ad023470bddde2429f90a600325febfe4/remotes/docker/authorizer.go#L338).

Instead of just calling `AddResponses()` with the one response, this PR changes that logic to provide the same 401 response twice to `AddResponses()` to force invalidation of the expired token and then adds a second call to `AddResponses()` with just once 401 response to fetch a fresh bearer token before proceeding to retry fetching resources from the remote registry.

**Testing performed:**
We have been using this fix in production at BuildBuddy to stream public container images from remote registries without incident: 

Kern also kindly offered to add an integration test for this bug in https://github.com/awslabs/soci-snapshotter/issues/686.

### DIFF
--- a/fs/remote/resolver.go
+++ b/fs/remote/resolver.go
@@ -258,7 +258,23 @@ func (tr *transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	if resp.StatusCode == http.StatusUnauthorized {
 		log.G(ctx).Infof("Received status code: %v. Refreshing creds...", resp.Status)
 
-		// prepare authorization for the target host using docker.Authorizer
+		// Prepare authorization for the target host using docker.Authorizer.
+		// The docker authorizer only refreshes OAuth tokens after two
+		// successive 401 errors for the same URL. Rather than issue the same
+		// request multiple times to tickle the token-refreshing logic, just
+		// provide the same response twice to trick it into refreshing the
+		// cached OAuth token. Call AddResponses() twice, first to invalidate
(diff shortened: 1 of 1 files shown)


sparr:
> * 9958f9a "Fix bearer token invalidation bug by calling AddResponses() twice on 401 errors" ... FAIL
  - FAIL - does not have a valid DCO

Could you amend this commit to include a signoff? `git commit -s` will do so automatically.

iain-macdonald:
Oh shoot, sorry about that. Just added one, confirmed it works with `make check` and re-pushed.

Kern-- (approved):
Thank you!

PR #769, opened 2023-08-07
Add regression check automation

coderbirju:
Needs #770 #785 

This commit adds benchmark_regression_test.yml workflow that runs a regression test by comparing the benchmark results of the current branch with the previous benchmark results of the code in main branch.

**Issue #, if available:**

**Description of changes:**
 
- Add a new workflow benchmark_regression_test.yml that runs on all new PR requests.
 
**Testing performed:**
- Verified expected behaviour. [Previous workflow run](https://github.com/coderbirju/soci-snapshotter/actions/runs/5788304422)


sparr:
The failure is because there is no prior benchmark to download. How will that cycle be started?

coderbirju:
> The failure is because there is no prior benchmark to download. How will that cycle be started?

We should merge #770 before this PR, It should be able to fetch the required files from the uploaded artifacts

PR #784, opened 2023-08-14
Populate content/index.json manifest list

### LINKED ISSUE #647: [Bug] `content/index.json` has empty manifest list
### Description

`/var/lib/soci-snapshotter-grpc/content/index.json` contains `"manifests":[]` where it seems there should be a list of the manifests present in the content store.

1. `soci create` for any image that produces an index and ztocs
2. `cat /var/lib/soci-snapshotter-grpc/content/index.json`

the newly created index manifest listed in index.json

### Host information

N/A

**WARNING**
Resolving this situation will slightly increase the startup time of `soci` and `soci-snapshotter-grpc` while they read in each existing manifest.

Resolving this situation would allow our existing use of `oci.New()` to successfully `loadIndexFile` and then `graph.IndexAll`. Note that the "index" in "index.json" and "IndexFile" relates to the subject of this issue involving a list of manifests, while the "index" in "IndexAll" is a separate concept referring to an internal mapping between manifests and zTOCs.

https://github.com/opencontainers/image-spec/blob/main/image-layout.md#indexjson-file describes the expected contents of this file.

https://pkg.go.dev/oras.land/oras-go/pkg/content#OCI.SaveIndex is the method to export this file after changes have been made to the `Store.index`

sparr:
**Issue #, if available:**
Closes #647 

**Description of changes:**
`Tag` newly `Push`ed content in the soci content store, then do `SaveIndex` after performing content store operations to enumerate tagged content in `content/index.json`'s `manifests` array.

**Testing performed:**
Existing unit and integration tests.
Manual creation of an index and check for a populated manifests array.
Needs new tests.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -141,10 +141,12 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 			return docker.ConfigureDefaultRegistries(docker.WithPlainHTTP(docker.MatchLocalhost))(refspec.Hostname())
 		})
 	}
-	ctx, store, err := store.NewContentStore(ctx, store.WithType(store.ContentStoreType(cfg.ContentStoreConfig.Type)), store.WithNamespace(cfg.ContentStoreConfig.Namespace))
+	ctx, store, done, err := store.NewContentStore(ctx, store.WithType(store.ContentStoreType(cfg.ContentStoreConfig.Type)), store.WithNamespace(cfg.ContentStoreConfig.Namespace))
 	if err != nil {
 		return nil, fmt.Errorf("cannot create content store: %w", err)
 	}
+	// FIXME: This runs far too soon. Writing index.json for the soci content store should happen after each change to the store and/or when the snapshotter shuts down.
+	defer done(ctx)
 
 	var bgFetcher *bf.BackgroundFetcher
 
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -86,10 +86,11 @@ var CreateCommand = cli.Command{
 			return err
 		}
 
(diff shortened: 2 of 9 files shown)


fs/fs.go
```diff
@@ -141,10 +141,12 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 			return docker.ConfigureDefaultRegistries(docker.WithPlainHTTP(docker.MatchLocalhost))(refspec.Hostname())
 		})
 	}
-	ctx, store, err := store.NewContentStore(ctx, store.WithType(store.ContentStoreType(cfg.ContentStoreConfig.Type)), store.WithNamespace(cfg.ContentStoreConfig.Namespace))
+	ctx, store, done, err := store.NewContentStore(ctx, store.WithType(store.ContentStoreType(cfg.ContentStoreConfig.Type)), store.WithNamespace(cfg.ContentStoreConfig.Namespace))
 	if err != nil {
 		return nil, fmt.Errorf("cannot create content store: %w", err)
 	}
+	// FIXME: This runs far too soon. Writing index.json for the soci content store should happen after each change to the store and/or when the snapshotter shuts down.
+	defer done(ctx)
```
sparr, line 149: All the other cases are single CLI commands which perform some operation then clean up after themselves. Here we are creating the filesystem for the snapshotter which will be used for an indefinite period. This `done` should happen when the filesystem is closed, probably when the snapshotter is exiting. An equivalent operation (which wouldn't really be "done") may also be needed after each new content creation. Either or both of these options would require putting the function into the filesystem struct (or somewhere else persistent) to be called later.

I am seeking input on the desirable path forward here.

sparr: This operation could be moved into `Push` but would then happen far more often than is reasonable for sequential operations.

PR #785, opened 2023-08-14
Add baseline benchmark workflow

coderbirju:
This commit adds baseline comparison logic to the
benchmark_regression_test workflow. The baseline artifact is uploaded via the benchmark_baseline workflow once on the 20th of each month. The baseline artifact comparison will be skipped if the baseline result artifact is not available for download

**Issue #, if available:**

**Description of changes:**

- Adds benchmark_baseline.yml
 
**Testing performed:**

### DIFF
--- a/.github/workflows/benchmark_baseline.yml
+++ b/.github/workflows/benchmark_baseline.yml
@@ -1,0 +1,27 @@
+name: Update Benchmark Baseline
+
+
+on:
+    schedule:
+      - cron: '0 0 */20 * *'
+
+jobs:
+    benchmark:
+        runs-on: ubuntu-20.04
+        steps:
+          - name: Checkout main branch
+            uses: actions/checkout@v3
+            with:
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark_baseline.yml
```diff
@@ -0,0 +1,27 @@
+name: Update Benchmark Baseline
+
+
+on:
+    schedule:
+      - cron: '0 0 */20 * *'
```
sparr, line 6: Could we also run on `workflow_dispatch` so it can be manually triggered?

coderbirju:
Updated the commit message to reflect the change in the logic of running this once every month on the 20th. 
The logic for running it every 20 days involves running the workflow everyday and checking if it's the 20th day, which in my opinion is not necessary here as the aim is to have a sufficiently old benchmark result and running it once a month will fit the purpose reasonably well.

PR #788, opened 2023-08-15
Retain original cache.Get error in span manager

Kern--:
**Issue #, if available:**
N/A

**Description of changes:**
Before this change, the span manager would replace the error received from `m.cache.Get` with a generic `ErrSpanNotAvailable`. The way we use the cache is really just as an abstraction of disk storage for span data so we don't generally expect `m.cache.Get` to throw an error. If it does, we should keep that context.

Multiple error wraps is a go 1.20 feature, so this is probably the first change that requires >go 1.18. 

**Testing performed:**


turan18 (approved):

swagatbora90 (approved):
LGTM

Looks like error wrap is available since Go1.13 https://go.dev/blog/go1.13-errors. what am i missing?

Kern--:
Error wrapping was added in go 1.13, but multiple error wrapping wasn't added until go 1.20.

E.g. if you had `io.EOF` and you wanted to add a `soci.InsufficientData`, you had to do `fmt.Errorf("%v: %w", soci.InsufficientData, io.EOF)` and `errors.Is(err, soci.InsufficientData)` would be false.

PR #790, opened 2023-08-17
Update check regression script

coderbirju:
This commit updates the regression check script to skip the initial value in all BenchmarkTimes array of the benchmark results json file to calculate a new p90. We use this new p90 to identify regression, this change was made to combat the skewed p90 metrics we were seeing due to the slow starts of the benchmark pull times which were affecting the overall regression comparison. Skipping the first value allows us to have a more uniform comparison, remove github environment noise and we'd be able to identify true regression in our code

**Issue #, if available:**

**Description of changes:**

- Updated the check_regression.sh file

**Testing performed:**

### DIFF
--- a/scripts/check_regression.sh
+++ b/scripts/check_regression.sh
@@ -22,12 +22,12 @@ compare_stat_p90() {
     local current_value="$2"
     local stat_name="$3"
 
-    # Calculate 110% of the past value
+    # Calculate 125% of the past value
     local threshold=$(calculate_threshold "$past_value")
 
     # Compare the current value with the threshold
-    if (( $(awk 'BEGIN {print ("'"$current_value"'" > "'"$threshold"'")}') )); then
-        echo "ERROR: $stat_name - Current P90 value ($current_value) exceeds the 110% threshold ($threshold) of the past P90 value ($past_value)"
+    if (( $(echo "$current_value > $threshold" |bc -l) )); then
+        echo "ERROR: $stat_name - Current P90 value ($current_value) exceeds the 125% threshold ($threshold) of the past P90 value ($past_value)"
         return 1
     fi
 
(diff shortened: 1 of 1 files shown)


scripts/check_regression.sh
```diff
@@ -36,7 +37,19 @@ compare_stat_p90() {
 
 calculate_threshold() {
     local past_value="$1"
-    awk -v past="$past_value" 'BEGIN { print past * 1.1 }'
+    awk -v past="$past_value" 'BEGIN { print past * 1.5 }'
```
turan18, line 40: Should we reduce this back to 10%?

coderbirju: I think we can go to 15 % instead of 50%, yeah I can make this change

scripts/check_regression.sh
```diff
@@ -36,7 +37,19 @@ compare_stat_p90() {
...
+
+}
+
+calculate_p90_after_skip() {
```
Kern--, line 44: nit: can you add a comment that this calculates the p90 ignoring the first result because we generally see an outlier in the first result?

scripts/check_regression.sh
```diff
@@ -52,8 +65,10 @@ compare_p90_values() {
     for test_name in $test_names; do
         echo "Checking for regression in '$test_name'"
         for stat_name in "fullRunStats" "pullStats" "lazyTaskStats" "localTaskStats"; do
-            local past_p90=$(echo "$past_json" | jq -r --arg test "$test_name" '.benchmarkTests[] | select(.testName == $test) | .'"$stat_name"'.pct90')
-            local current_p90=$(echo "$current_json" | jq -r --arg test "$test_name" '.benchmarkTests[] | select(.testName == $test) | .'"$stat_name"'.pct90')
+            local past_p90_array=$(echo "$past_json" | jq -r --arg test "$test_name" '.benchmarkTests[] | select(.testName == $test) | .'"$stat_name"'.BenchmarkTimes')
```
turan18, line 68: nit: past_p90_array -> past_array

scripts/check_regression.sh
coderbirju: sure

scripts/check_regression.sh
```diff
@@ -22,12 +22,13 @@ compare_stat_p90() {
...
 
-    # Calculate 110% of the past value
+
+    # Calculate 150% of the past value
     local threshold=$(calculate_threshold "$past_value")
 
     # Compare the current value with the threshold
-    if (( $(awk 'BEGIN {print ("'"$current_value"'" > "'"$threshold"'")}') )); then
-        echo "ERROR: $stat_name - Current P90 value ($current_value) exceeds the 110% threshold ($threshold) of the past P90 value ($past_value)"
+    if (( $(echo "$current_value > $threshold" |bc -l) )); then
```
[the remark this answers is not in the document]
coderbirju: Yeah, we need the -l flag to enable the math library since we're comparing floating point numbers here, it wouldn't be required if we were comparing just numbers

Kern-- (approved):

turan18 (approved):

PR #792, opened 2023-08-18
Deterministically close open span cache file descriptors

turan18:
The snapshotter stores fetched spans in a cache either in memory or on disk. When reading from the cache on disk we use a [Finalizer](https://github.com/awslabs/soci-snapshotter/blob/main/fs/span-manager/span_manager.go#L401 ) construct to close the open file descriptors when the Go garbage collector sees that the fd is no longer being referenced. The issue with this is that we don't have control over when the GC runs (although it's possible), and so the process could hold on too open fds for a unknown amount of time causing a sort of leak. On systems where the snapshotter is bounded by a ulimit in the number of open files, this can cause the snapshotter span cache get calls to fail, causing `file.Read` failures for the running container/process. This change wraps the readers returned by the cache in `io.ReadCloser`'s, so we can deterministically close the files once the content has been read from them.

**Issue #, if available:**

**Description of changes:**

**Testing performed:**

To test the change we can set a `ulimit` on the snapshotter and then pull an image and run a container with and without the change. In theory, we should see `failed to read the file: span not available in cache` errors without the change and none with it.

### Determine ulimit

We can determine `nofile ulimit` by dry running a container and curling the metrics endpoint to examine the amount of open fd's when idle (not running the container), when running the container, and a little while after the container fetches the spans.

```shell

$ soci image rpull public.ecr.aws/soci-workshop-examples/rabbitmq:latest
# run a script that continuously calls the metrics endpoint for around 20s and dumps the results in a file
$ ./collector.sh
# start container 
$ ctr run -d --snapshotter soci --net-host public.ecr.aws/soci-workshop-examples/rabbitmq:latest sociExample
$ cat fd.txt
process_open_fds 46
...
process_open_fds 79
...
process_open_fds 118
...
process_open_fds 46
```
It looks like the idle open fd is around 46 and can spike up to 100+ when running the container without the change. We have also ran the exact test with the change and the open fd's seem to be fairly stable and only ever goes up to around 47.

### Running test

Without change
``` 
$ sudo prlimit --pid soci --nofile=70:70 
$ soci image rpull public.ecr.aws/soci-workshop-examples/rabbitmq:latest
$ ctr run -d --snapshotter soci --net-host public.ecr.aws/soci-workshop-examples/rabbitmq:latest sociExample
# Cat logs
$ cat ~/soci_log | grep "span not available in cache"
{"digest":"sha256:99803d4b97f3db529ae9ca4174b0951afac6b309e7deaa8ec3214c584e02b3a8","error":"file.Read: failed to read the file: span not available in cache: failed to open blob file for \"0\": open /var/lib/soci-snapshotter-grpc/soci/spancache/3753186881/0: too many open files","fetchedPercent":0,"fetchedSize":0,"level":"error","msg":"statFile error","size":28578563,"time":"2023-08-18T14:39:10.708256919Z"}
{"digest":"sha256:778d7f6440dcc8d94f29eab78819fdbd296217cbeee40364f4cc9e9fc41790a3","error":"file.Read: failed to read the file: span not available in cache: failed to open blob file for \"0\": open /var/lib/soci-snapshotter-grpc/soci/spancache/1121699285/0: too many open files","fetchedPercent":0,"fetchedSize":0,"level":"error","msg":"statFile error","size":54901810,"time":"2023-08-18T14:39:11.976654170Z"}
{"digest":"sha256:778d7f6440dcc8d94f29eab78819fdbd296217cbeee40364f4cc9e9fc41790a3","error":"file.Read: failed to read the file: span not available in cache: failed to open blob file for \"0\": open /var/lib/soci-snapshotter-grpc/soci/spancache/1121699285/0: too many open files","fetchedPercent":0,"fetchedSize":0,"level":"error","msg":"statFile error","size":54901810,"time":"2023-08-18T14:39:11.976737217Z"}
{"digest":"sha256:778d7f6440dcc8d94f29eab78819fdbd296217cbeee40364f4cc9e9fc41790a3","error":"file.Read: failed to read the file: span not available in cache: failed to open blob file for \"0\": open /var/lib/soci-snapshotter-grpc/soci/spancache/1121699285/0: too many open files","fetchedPercent":0,"fetchedSize":0,"level":"error","msg":"statFile error","size":54901810,"time":"2023-08-18T14:39:11.976799974Z"}
```
With change
``` 
$ sudo prlimit --pid soci --nofile=70:70 
$ soci image rpull public.ecr.aws/soci-workshop-examples/rabbitmq:latest
$ ctr run -d --snapshotter soci --net-host public.ecr.aws/soci-workshop-examples/rabbitmq:latest sociExample
# Cat logs
$ cat ~/soci_log | grep "span not available in cache"
```

### DIFF
--- a/fs/span-manager/span_manager.go
+++ b/fs/span-manager/span_manager.go
@@ -38,6 +38,30 @@ var (
 	ErrExceedMaxSpan       = errors.New("span id larger than max span id")
 )
 
+type MultiReaderCloser struct {
+	c []io.Closer
+	io.Reader
+}
+
+func (mrc *MultiReaderCloser) Close() error {
+	errs := []error{}
+	for _, c := range mrc.c {
+		if err := c.Close(); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return errors.Join(errs...)
+}
+
+type SectionReaderCloser struct {
+	c io.Closer
+	*io.SectionReader
+}
+
+func (src *SectionReaderCloser) Close() error {
+	return src.c.Close()
+}
+
 // SpanManager fetches and caches spans of a given layer.
 type SpanManager struct {
 	cache                             cache.BlobCache
@@ -145,10 +169,11 @@ func (m *SpanManager) resolveSpan(spanID compression.SpanID) error {
 
 // GetContents returns a reader for the requested contents. The contents may be
 // across multiple spans.
-func (m *SpanManager) GetContents(startUncompOffset, endUncompOffset compression.Offset) (io.Reader, error) {
+func (m *SpanManager) GetContents(startUncompOffset, endUncompOffset compression.Offset) (io.ReadCloser, error) {
 	si := m.getSpanInfo(startUncompOffset, endUncompOffset)
 	numSpans := si.spanEnd - si.spanStart + 1
 	spanReaders := make([]io.Reader, numSpans)
+	spanClosers := make([]io.Closer, numSpans)
 
 	eg, _ := errgroup.WithContext(context.Background())
 	var i compression.SpanID
@@ -161,14 +186,14 @@ func (m *SpanManager) GetContents(startUncompOffset, endUncompOffset compression
 				return err
 			}
 			spanReaders[j] = r
+			spanClosers[j] = r
 			return nil
 		})
 	}
 	if err := eg.Wait(); err != nil {
 		return nil, err
 	}
-
-	return io.MultiReader(spanReaders...), nil
+	return &MultiReaderCloser{spanClosers, io.MultiReader(spanReaders...)}, nil
 }
 
 // getSpanInfo returns spanInfo from the offsets of the requested file
@@ -216,7 +241,7 @@ func (m *SpanManager) getSpanInfo(offsetStart, offsetEnd compression.Offset) *sp
 //  3. For `unrequested` span, fetch-uncompress-cache the span data, return the reader
 //     from the uncompressed span
 //  4. No span state lock will be acquired in `requested` state.
-func (m *SpanManager) getSpanContent(spanID compression.SpanID, offsetStart, offsetEnd compression.Offset) (io.Reader, error) {
+func (m *SpanManager) getSpanContent(spanID compression.SpanID, offsetStart, offsetEnd compression.Offset) (io.ReadCloser, error) {
 	s := m.spans[spanID]
 	size := offsetEnd - offsetStart
 
@@ -240,6 +265,7 @@ func (m *SpanManager) getSpanContent(spanID compression.SpanID, offsetStart, off
 		if err != nil {
 			return nil, err
 		}
+		defer r.Close()
 
 		// read compressed span
 		compressedBuf, err := io.ReadAll(r)
@@ -260,7 +286,7 @@ func (m *SpanManager) getSpanContent(spanID compression.SpanID, offsetStart, off
 		if err := s.setState(uncompressed); err != nil {
 			return nil, err
 		}
-		return bytes.NewReader(uncompSpanBuf[offsetStart : offsetStart+size]), nil
+		return io.NopCloser(bytes.NewReader(uncompSpanBuf[offsetStart : offsetStart+size])), nil
 	}
 
 	// fetch-uncompress-cache span: span state can only be `unrequested` since
@@ -270,7 +296,7 @@ func (m *SpanManager) getSpanContent(spanID compression.SpanID, offsetStart, off
 		return nil, err
 	}
 	buf := bytes.NewBuffer(uncompBuf[offsetStart : offsetStart+size])
-	return io.Reader(buf), nil
+	return io.NopCloser(buf), nil
 }
 
 // fetchAndCacheSpan fetches a span, uncompresses the span if `uncompress == true`,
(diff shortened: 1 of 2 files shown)


fs/span-manager/span_manager.go
```diff
@@ -38,6 +38,29 @@ var (
...
 
+type MultiReaderCloser struct {
+	c []io.Closer
+	io.Reader
+}
+
+func (mrc *MultiReaderCloser) Close() error {
+	for _, c := range mrc.c {
+		if err := c.Close(); err != nil {
+			return err
```
Kern--, line 49: I think we want to be sure that we at least try to close all the inner Closers before returning to minimize a leak in the case of error. Can we switch this to a multi-error that we append to and only return after processing all the closers?

fs/span-manager/span_manager.go
```diff
@@ -161,14 +185,14 @@ func (m *SpanManager) GetContents(startUncompOffset, endUncompOffset compression
...
+			spanClosers[j] = r
 			return nil
 		})
 	}
 	if err := eg.Wait(); err != nil {
 		return nil, err
 	}
-
-	return io.MultiReader(spanReaders...), nil
+	return &MultiReaderCloser{spanClosers, io.MultiReader(spanReaders...)}, nil
```
Kern--, line 195: nit: to mimic the API of `io.MultiReader(spanReaders...)` would it make sense to add a `multiReadCloser(spanReadClosers...)` function here which basically just returns what you've written here?

This is definitely not blocking, just food for thought.

fs/span-manager/span_manager.go
```diff
@@ -145,10 +168,11 @@ func (m *SpanManager) resolveSpan(spanID compression.SpanID) error {
 
 // GetContents returns a reader for the requested contents. The contents may be
 // across multiple spans.
-func (m *SpanManager) GetContents(startUncompOffset, endUncompOffset compression.Offset) (io.Reader, error) {
+func (m *SpanManager) GetContents(startUncompOffset, endUncompOffset compression.Offset) (io.ReadCloser, error) {
 	si := m.getSpanInfo(startUncompOffset, endUncompOffset)
 	numSpans := si.spanEnd - si.spanStart + 1
 	spanReaders := make([]io.Reader, numSpans)
+	spanClosers := make([]io.Closer, numSpans)
```
Kern--, line 175: nit: can we change the type of `spanReaders` from `[]io.Reader` to `[]io.ReadCloser` so we don't have to track 2 arrays?

fs/span-manager/span_manager.go
turan18: Agh,  good catch. Will do.

fs/span-manager/span_manager.go
turan18: Yeah, I tried that at first but the compiler complained about mismatched types since `io.MultiReader` strictly accepts `[]io.Reader`.

fs/span-manager/span_manager.go
turan18: We can but since I can't pass in a `[]io.ReadCloser` to `io.MultiReader` I'll have to either pass both slices or only pass the reader slice and have to loop again through the readers  in `multiReadCloser` to get a slice of `io.Closer` to pass to `MultiReaderCloser`.

fs/span-manager/span_manager.go
turan18: I've updated to use `errors.Join`, introduced in Go 1.20 - https://pkg.go.dev/errors#Join. I could of have implemented a separate wrapper Error type or just used `strings.Join` and then returned a new error, but since we already use multi error wrapping introduced in 1.20 (#788),  I thought this made sense.

fs/span-manager/span_manager.go
Kern--: Oh right. Slices don't work like that.

fs/span-manager/span_manager.go
Kern--: Yeah, this has the same problem as my previous comment. Feel free to ignore.

fs/span-manager/span_manager.go
Kern--: We're using [go-multierror](https://github.com/hashicorp/go-multierror) elsewhere ([e.g.](https://github.com/awslabs/soci-snapshotter/blob/8c6880c62279317ca0e66629f70f70ce99babcc5/fs/reader/reader.go#L191)). I do like the idea of being able to get rid of that as a direct dependency, though. I'll create an issue to replace those with this style.

Kern-- (approved):

fs/span-manager/span_manager.go
Kern--: https://github.com/awslabs/soci-snapshotter/issues/793

fs/span-manager/span_manager.go
```diff
@@ -38,6 +38,30 @@ var (
 	ErrExceedMaxSpan       = errors.New("span id larger than max span id")
 )
 
+type MultiReaderCloser struct {
+	c []io.Closer
+	io.Reader
```
austinvazquez, line 43: Would this need to be pointer embedded?

turan18: I don't think so since it's an interface, which also handles pointers.

turan18: It was need for `io.SectionReader` since it returns a concrete pointer type.

austinvazquez: Ah, I see. Yeah was reader `io.SectionReader` and trying to understand the difference. Thanks for answering.

Issue #696, opened 2023-06-30
[Bug] Integration tests don't work on non-x86


turan18:
### Description

Our integration tests fail on other architectures. Here are some things we need to do to:

- [x] Fix a bug in our Dockerfile where `TARGETARCH` is not honored since it is not scoped locally. This causes us to use only amd64 artifacts. - #695 
- [ ] Remove dependency on ghcr.io/oci-playground/registry:v3.0.0-alpha.1 since its only available for amd64 architectures. Since the notion of OCI Artifact Manifests have seemingly been rejected from the v1.1.0 Image Spec I don't think we need a v1.1.0 compliant registry at least until the v1.1.0 spec has been finalized. We can then remove any tests that use this image - https://github.com/awslabs/soci-snapshotter/blob/801acef5adc5f7976fa7b482e1b1f829e6ee6097/integration/push_test.go#L133 and then set the default registry to be registry:2 here https://github.com/awslabs/soci-snapshotter/blob/801acef5adc5f7976fa7b482e1b1f829e6ee6097/integration/util_test.go#L423
- [ ] We have 2 tests that specifically use an amd64 variant of rabbitmq- [TestPullWithAribtraryBlobInvalidZtocFormat()](https://github.com/awslabs/soci-snapshotter/blob/184d1715fe4985936018f8013dd81c54019ae4e4/integration/pull_test.go#L415) and [TestPushAlwaysMostRecentlyCreatedIndex()](https://github.com/awslabs/soci-snapshotter/blob/184d1715fe4985936018f8013dd81c54019ae4e4/integration/push_test.go#L84). We should change them to use tags that support multiple architectures. 

### Steps to reproduce the bug

Run the integration tests on non x86 arch.

### Describe the results you expected

Integration tests should work with different architectures.

### Host information

1. OS: 
2. Snapshotter Version: 
3. Containerd Version:

sparr:
> We have 2 tests that specifically use an amd64 variant of rabbitmq- [...] We should change them to use tags that support multiple architectures.

Those tags should be historic and static, not versions that might see backports or other new pushes.

Kern--:
For the second issue, I think we can just skip that test on ARM for now.

sparr:
[...]

PR #806, opened 2023-08-28
Fix integration test failures on non-x86

turan18:
There were several issues that caused integration tests to fail on non-x86. The first one was our dependency GHCR registry image which is only supported on amd64. The GHCR registry was needed when we supported artifact manifests, but those have been seemingly removed from the OCI 1.1 spec proposal, so we no longer need it. We also had 2 tests that relied on a pinned amd64 variant of rabbitmq. Those have been replaced with a pinned multi-arch index.

**Issue #, if available:**

Fixes: #696 

**Description of changes:**

1. Removed our dependency on the GHCR registry.
2. Use pinned index instead of a pinned manifest

**Testing performed:**

Spin up an ARM64 instance and run integration tests.

```shell
$ uname -p
aarch64
$ make integration 
....
     --- PASS: TestSociZtocGetFile/Ztoc_and_each_file_exists,_file_contents_redirected_to_output_file#03 (0.09s)
PASS
ok  	github.com/awslabs/soci-snapshotter/integration	1032.225s

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,5 +44,3 @@ RUN curl -sSL --output /tmp/nerdctl.tgz https://github.com/containerd/nerdctl/re
     rm -f /tmp/nerdctl.tgz
 
 FROM registry:2 AS registry2
-
-FROM ghcr.io/oci-playground/registry:v3.0.0-alpha.1 AS registry3alpha1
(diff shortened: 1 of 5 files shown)


Kern-- (approved):
We should add a follow up issue to add a new integration test against the referrers API.

turan18:
> We should add a follow up issue to add a new integration test against the referrers API.

Created https://github.com/awslabs/soci-snapshotter/issues/808

PR #833, opened 2023-09-12
Protect node.ents and node.entsCached with a mutex in fs/layer/node.go

iain-macdonald:
**Issue #, if available:**
None

**Description of changes:**
We run a modified version of the soci-snapshotter in production at BuildBuddy and recently enabled the golang race detector in our dev environment. It flagged racy accesses to `node.ents` and `node.entsCached`, I've attached the [logs](https://github.com/awslabs/soci-snapshotter/files/12591454/datarace.txt) for line numbers (which should be unmodified). This PR protects access to those two fields with a mutex.

**Testing performed:**
In addition to the PR checks, I pulled and ran an indexed image and it seemed to work fine, but didn't do any additional testing beyond that.

### DIFF
--- a/fs/layer/node.go
+++ b/fs/layer/node.go
@@ -252,11 +252,13 @@ func (fs *fs) inodeOfID(id uint32) (uint64, error) {
 // node is a filesystem inode abstraction.
 type node struct {
 	fusefs.Inode
-	fs         *fs
-	id         uint32
-	attr       metadata.Attr
+	fs   *fs
+	id   uint32
+	attr metadata.Attr
+
 	ents       []fuse.DirEntry
 	entsCached bool
+	entsMu     sync.Mutex
 }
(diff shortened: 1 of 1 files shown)


Kern--:
I'd like to take a quick look at benchmarking data before merging this to see if we should be using a `sync.Value` or a more complicated variant of `sync.Once` that allows retries on errors.

Kern-- (approved):
Sorry it took me a while to get back to this. I'm not seeing any significant difference in our benchmark images.

iain-macdonald:
> Sorry it took me a while to get back to this. I'm not seeing any significant difference in our benchmark images.

Thanks, Kern! Looks like I need another approver to merge. Is there anyone you'd suggest?

turan18 (approved):

PR #850, opened 2023-09-29
Fix integrations tests with cgroupv2

### LINKED ISSUE #735: [Bug] Integration tests don't work with cgroupv2
### Description

When running the integration tests on a host machine that uses cgroupsv2, the tests fail with errors about cgroupv2 being in an invalid state.

It looks like Docker had this same problem and has a script to initialize cgroupv2 for nesting properly: https://github.com/moby/moby/pull/41065

Run any test that tries to run a container on a host that's configured for cgroupsv2. Observe container failures due to cgroups being in an invalid state;

```
$ STATIC=1 GO_TEST_FLAGS='-run TestLazyPull' make integration
...
ctr: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: unable to apply cgroup configuration: cannot enter cgroupv2 "/sys/fs/cgroup/default" with domain controllers -- it is in an invalid state: unknown
```

The tests should pass when the host is configured for cgroupv2.

### Host information

1. OS: AL2023
2. Snapshotter Version: soci-snapshotter-grpc version 08dcb20 08dcb2097d0389019ffa2531b784f3910cd7451a
3. Containerd Version: 1.6.19

_No response_

Kern--:
**Issue #, if available:**
closes #735 

**Description of changes:**
With cgroupv2, dind doesn't work out of the box because the inner docker process is in the (containerized) root cgroup so it can't create the inner container's cgroups because doing so would make the inner docker a process on an interior cgroup node. cgroupv2 only allows processes on the leaf nodes.

The solution is to move docker to a child cgroup (called init) so that the container can exist as a sibling.

**Testing performed:**
`make check && make integration` on an al2023 instance which uses cgroupv2. 

Before this change the tests failed with:
```
ctr: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: unable to apply cgroup configuration: cannot enter cgroupv2 "/sys/fs/cgroup/default" with domain controllers -- it is in an invalid state: unknown
```

after this change, the tests pass

### DIFF
--- a/integ_entrypoint.sh
+++ b/integ_entrypoint.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+#
+#   Copyright The Soci Snapshotter Authors.
+
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+
+#       http://www.apache.org/licenses/LICENSE-2.0
+
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+set -x
+
+# copied from moby/moby https://github.com/moby/moby/blob/52c92be4c5f94cdb93f775099d42c0ee6b66ccc7/hack/dind#L28-L38
+# Apache v2.0 License
+# Copyright 2012-2017 Docker, Inc.
+# cgroup v2: enable nesting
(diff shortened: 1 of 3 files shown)


sondavidb (approved):

integ_entrypoint.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
turan18, line 1: nit: Should this be under `.scripts/`?

turan18 (approved):

integ_entrypoint.sh
Kern--: I didn't put it under `scripts` because those are generally project level scripts and this is really integration test code. I put it in the root to live next to the Dockerfile.

Personally, I think we should move both the Dockerfile and setup script into `integration` somewhere.

Issue #494, opened 2023-03-13
ztoc versioning needs clarification


Kern--:
Also, we have gzip_zinfo versions (https://github.com/awslabs/soci-snapshotter/blob/3b2cc11be4627b097f4b225bd105100f5f7957d1/compression/gzip_zinfo.h#L56-L57), and in the future maybe zstd_zinfo version(s).

How do we plan to track these versions altogether? a zinfo version change will possibly change zinfo bytes, which changes ztoc. (e.g. a 0.9 ztoc with version1 gzip_zinfo != a 0.9 ztoc with version2 gzip_zinfo, but to user they're the same 0.9 ztoc).

_Originally posted by @djdongjin in https://github.com/awslabs/soci-snapshotter/pull/485#discussion_r1131783912_

Kern--:
Another example of clarification needed: https://github.com/awslabs/soci-snapshotter/pull/853#discussion_r1347794062

`FileMetadata.XAttrs` in the flatbuffer actually contains PAX records. Should/Could a major version change fix the naming?

PR #853, opened 2023-10-02
Support xattrs

### LINKED ISSUE #812: [Bug] Xattrs don't work
### Description

Currently, the TOC builder dumps all [`tar.Header.PAXRecords`](https://pkg.go.dev/archive/tar@go1.20.6#Header) into the `Xattrs` field of `FileMetadata`.

https://github.com/awslabs/soci-snapshotter/blob/ec1e62326578318ace5048c487b7e155a068dd4b/ztoc/toc_builder.go#L150

This is directly used to respond to `GetXAttr` and `ListXAttr`
https://github.com/awslabs/soci-snapshotter/blob/ec1e62326578318ace5048c487b7e155a068dd4b/fs/layer/node.go#L516-L520

https://github.com/awslabs/soci-snapshotter/blob/ec1e62326578318ace5048c487b7e155a068dd4b/fs/layer/node.go#L541-L543

According to the [`tar.Header` documentation](https://pkg.go.dev/archive/tar@go1.20.6#Header), there is a deprecated `Xattrs` field which says:

> // Xattrs stores extended attributes as PAX records under the
> // "SCHILY.xattr." namespace.
> //
> // The following are semantically equivalent:
> //  h.Xattrs[key] = value
> //  h.PAXRecords["SCHILY.xattr."+key] = value

So at a minimum, `FileMetadata.Xattrs` should only contain the data from `PAXRecords` where the key is prefixed with "SCHILY.xattr", but even this is not totally right because there are other ways to encode xattrs into a tar file (example issue in the image spec highlighting this problem https://github.com/opencontainers/image-spec/issues/725).

It might be useful to take a look around at some of the common tools for building images (docker, buildkit, podman, kaniko?) to see what (if any) they support.

TBD

TBD

### Host information

1. OS: 
2. Snapshotter Version: 
3. Containerd Version: 

_No response_

Kern--:
**Issue #, if available:**

Fixes #812 

**Description of changes:**
Before this change, SOCI stored all PAX header records as linux xattrs. PAX header records are a generic key-value pair for TAR files, not specifically linux xattrs. While go does support linux xattrs by prefixing them with SCHILY.xattr, since we didn't parse them back to linux xattrs, they did not behave correctly with SOCI. The most likely way users would experience this is that file capabilities don't work with SOCI.

This change keeps all PAX header records in the ztoc format, but parses out just the linux xattrs without the prefix when creating the filesystem metadata from a ztoc.

Docker, buildkit, buildah/podman, and kaniko all use the go tarHeader.Xattrs to add xattrs which uses the `SCHILY.xattr.` prefix. While there are technically other ways to encode xattrs (e.g. `LIBARCHIVE.xattr.`) it doesn't seem common.

**Testing performed:**

I also re-performed the security capability test from #812 to verify that it works with SOCI now.

### DIFF
--- a/ztoc/fbs/ztoc.fbs
+++ b/ztoc/fbs/ztoc.fbs
@@ -22,7 +22,7 @@ table FileMetadata {
 	devmajor : long;		// Major device number (valid for TypeChar or TypeBlock)
 	devminor : long;		// Minor device number (valid for TypeChar or TypeBlock)
 
-	xattrs : [Xattr];
+	xattrs : [Xattr];       // Raw PAXRecords from the tar file. The name is wrong, but changing it is backwards incompatible
 }
 
 enum CompressionAlgorithm : byte { Gzip = 1, Uncompressed }
--- a/ztoc/toc_builder.go
+++ b/ztoc/toc_builder.go
@@ -150,7 +150,7 @@ func metadataFromTarReader(r io.Reader) ([]FileMetadata, compression.Offset, err
 			ModTime:            hdr.ModTime,
 			Devmajor:           hdr.Devmajor,
 			Devminor:           hdr.Devminor,
-			Xattrs:             hdr.PAXRecords,
+			PAXHeaders:         hdr.PAXRecords,
 		}
 		md = append(md, metadataEntry)
 		// The next file's tar header can be found immediately after the current file + padding
--- a/fs/layer/util_test.go
+++ b/fs/layer/util_test.go
@@ -264,7 +264,7 @@ func testExistenceWithOpaque(t *testing.T, factory metadata.Store, opaque Overla
 			},
 			want: []check{
 				hasOpaque("foo/"),
-				hasNodeXattrs("foo/", "SCHILY.xattr.foo", "bar"),
+				hasNodeXattrs("foo/", "foo", "bar"),
 				fileNotExist("foo/.wh..wh..opq"),
 			},
 		},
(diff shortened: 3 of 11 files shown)


ztoc/toc_builder.go
```diff
@@ -150,7 +150,7 @@ func metadataFromTarReader(r io.Reader) ([]FileMetadata, compression.Offset, err
 			ModTime:            hdr.ModTime,
 			Devmajor:           hdr.Devmajor,
 			Devminor:           hdr.Devminor,
-			Xattrs:             hdr.PAXRecords,
+			PAXHeaders:         hdr.PAXRecords,
```
turan18, line 153: This wouldn't affect backwards content addressability, right?

turan18 (approved):

ztoc/toc_builder.go
Kern--: Yeah. It's backwards compatible with serialized ztocs because it doesn't change the serialization. The library did have an API change that we might want to document, though.

ztoc/fbs/ztoc.fbs
```diff
@@ -22,7 +22,7 @@ table FileMetadata {
 	devmajor : long;		// Major device number (valid for TypeChar or TypeBlock)
 	devminor : long;		// Minor device number (valid for TypeChar or TypeBlock)
 
-	xattrs : [Xattr];
+	xattrs : [Xattr];       // Raw PAXRecords from the tar file. The name is wrong, but changing it is backwards incompatible
```
sondavidb, line 25: Should we ever make a major version change, would we want to look into correcting this?

sondavidb (approved):

ztoc/fbs/ztoc.fbs
Kern--: Added a note to https://github.com/awslabs/soci-snapshotter/issues/494. We need to think more about ztoc versioning in general.

Issue #808, opened 2023-08-29
[Bug] Test against referrers API


turan18:
### Description

We recently had a change that removed our dependency on the GHCR registry in our integration tests since it was only supported on amd64 - #806. The GHCR registry provided support for the referrers API, allowing us to test pushing and pulling artifacts with it enabled - https://github.com/awslabs/soci-snapshotter/blob/ec1e62326578318ace5048c487b7e155a068dd4b/integration/push_test.go#L135 We should look into alternative registries that have support for multiple architectures and the referrers API, in the meantime. https://github.com/project-zot/zot seems to be a good candidate. Ideally, we'd like to be able to test against the fallback and the referrers API without having to build 2 separate registries. We should look into if its possible to disable the referrers API in Zot.

### Describe the results you expected

A test that pushes/pulls against a registry that supports the referrers API.

### Host information

1. OS: 
2. Snapshotter Version: 
3. Containerd Version:

sondavidb:
Closed with #856

PR #856, opened 2023-10-04
Add test against referrers API

sondavidb:
**Issue #, if available:**
#808

**Description of changes:**
#806 removed our test against the referrers API. This test uses a multi-platform image (zot) to reintroduce this test.

**Testing performed:**
Ran `GO_TEST_FLAGS='-run TestLegacyOCI -count=1' make integration` on local machine

### DIFF
--- a/integration/util_test.go
+++ b/integration/util_test.go
@@ -45,6 +45,7 @@ import (
 	"math/big"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 	"testing"
@@ -74,6 +75,7 @@ const (
 	// Registry images to use in the test infrastructure. These are not intended to be used
 	// as images in the test itself, but just when we're setting up docker compose.
 	oci10RegistryImage = "registry2:soci_test"
+	oci11RegistryImage = "ghcr.io/project-zot/zot-linux-" + runtime.GOARCH + ":v2.0.0-rc6"
 )
 
 // These are images that we use in our integration tests
@@ -175,7 +177,8 @@ services:
    - REGISTRY_HTTP_ADDR={{.RegistryHost}}:443
    - REGISTRY_STORAGE_DELETE_ENABLED=true
   volumes:
-   - {{.AuthDir}}:/auth:ro
+   - {{.HostVolumeMount}}/auth:/auth:ro
+   - {{.HostVolumeMount}}/etc/zot/config.json:/etc/zot/config.json:ro
 {{.NetworkConfig}}
 `
 const composeRegistryAltTemplate = `
@@ -206,7 +209,7 @@ services:
     - REGISTRY_HTTP_ADDR={{.RegistryHost}}:443
     - REGISTRY_STORAGE_DELETE_ENABLED=true
     volumes:
-    - {{.AuthDir}}:/auth:ro
+    - {{.HostVolumeMount}}/auth:/auth:ro
   registry-alt:
     image: {{.RegistryAltImageRef}}
     container_name: {{.RegistryAltHost}}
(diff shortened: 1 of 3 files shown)


integration/util_test.go
```diff
@@ -84,6 +85,7 @@ const (
 	ubuntuImage   = "ubuntu:23.04"
 	drupalImage   = "drupal:10.0.2"
 	rabbitmqImage = "rabbitmq:3.11.7"
+	zotImage      = "ghcr.io/project-zot/zot-linux-" + runtime.GOARCH + ":v2.0.0-rc6"
```
turan18, line 88: nit: Can we use string format notation:  `fmt.Sprintf`

Also to keep some consistency, can we rename it to something like `oci11RegistryImage` and move it up to where the `oci10RegistryImage` is initialized?

integration/util_test.go
```diff
@@ -483,6 +486,32 @@ networks:
 `, nw)
 	}
 
+	zotConfigFileStr := fmt.Sprintf(`
```
turan18, line 489: nit: Could we move this up to where the other configs are declared at the top of the file and use `testutil.ApplyTextTemplate` to fill in the host address portion

integration/util_test.go
sondavidb: Sure, I can rename the variable. I can't use string format notation as it's a constant so it'll unfortunately have to look a tad ugly, unless we can find some workaround.

Kern--:
Can you run `make check` and fix the errors reported by golangci-lint? You might need to run `scripts/install-check-tools.sh` if you haven't already.

integration/util_test.go
turan18: Ah, right. I think it's fine then.

turan18 (approved):

integration/util_test.go
```diff
@@ -483,6 +512,17 @@ networks:
 `, nw)
 	}
 
+	zotConfigFile, err := testutil.ApplyTextTemplate(zotConfigTemplate, zotConfigStruct{
```
sondavidb, line 515: Just want to note, this is called every single time a new shell is created. It is inconsequential operationally — most shells will just ignore this file — but my test is the only one that needs this config, adding an option for this (in a separate commit, perhaps?) seems to make more sense logically.

turan18: Yeah, personally I think we should move away from passing registry specific configs as environment variables in the [template](https://github.com/awslabs/soci-snapshotter/blob/main/integration/util_test.go#L169-#L176) and instead do something similar to what we did with Zot (having a separate config file and volume mounting). The Docker registry supports [reading](https://github.com/distribution/distribution/blob/main/docs/configuration.md#overriding-the-entire-configuration-file) a config from a file, so this should be possible. 

We can then volume mount the specific config file for the registry that gets passed in. I don't think it's really a blocker for this though.

sondavidb: Makes sense, I suppose we can modify the header to add options depending on which registry config we needed instead of attaching the same variables for every image. I could look into refactoring that in the future.

integration/util_test.go
```diff
@@ -176,6 +178,7 @@ services:
    - REGISTRY_STORAGE_DELETE_ENABLED=true
   volumes:
    - {{.AuthDir}}:/auth:ro
+   - {{.AuthDir}}/config.json:/etc/zot/config.json
```
Kern--, line 181: Does this really have anything to do with the AuthDir?

sondavidb: My reasoning for placing it here was that it helped point the zot image to the right auth files, and thus it somewhat made sense.

I could alternatively make another tempdir (mabye call it config?) and put the config files in there. That way if we ever need to use any config files in the future it could be placed there. But it would also be largely unused and I don't know how expensive an extra tempdir for every single shell ever created in our testing suite would be. What do you think?

Kern--: I think it might be good to change `AuthDir` to a generic `HostVolumeDir` that contains `auth/*` and `zot/config.json` so we can put arbitrary data that we want to mount from shell container -> registry container. 

If it’s really static across all shells, maybe we could bake it in to the integ test docker file.

sondavidb: It's static across all shells but it's also unnecessary. I think we could probably redo this function a bit so that we aren't loading an extra zot config for every single time we get a shell with a registry. Currently it's probably fine but if our testing suite ends up encompassing more images that need more files, it might get a bit clunky.

sondavidb: Ok, I made some changes to help make the separation of ideas more clear. Hopefully this is better?

Kern-- (approved):

Issue #844, opened 2023-09-21
[Bug] Replace `soci create/push` with `nerdctl push` in our documentation


turan18:
### Description

[`nerdctl` v1.6.0 ](https://github.com/containerd/nerdctl/releases/tag/v1.6.0) adds support for creating and pushing SOCI indexes. We should replace mentions of `soci create/push` with `nerdctl push` in our documentation. Note: The `nerdctl` CLI just shells out to the `soci` CLI, so we should mention that the `soci` CLI still needs to be installed. 

### Describe the results you expected

Replacing references of `soci create/push` with `nerdctl push`.

### Host information

1. OS: 
2. Snapshotter Version: 
3. Containerd Version:

ningziwen:
Should we add as one option instead of replacing, as nerdctl is not a must-have to use soci. (Containerd is a must-have though).

Kern--:
We need some way to manage container images in containerd. Right now, our alternative is `ctr`. While `nerdctl` is technically optional, the alternative (`ctr`) is pretty inappropriate for production use-cases.

If we eventually have a containerd-less option (or docker containerd support is default), then we can probably have options, but I think making nerdctl + SOCI CLI the only recommendation in our docs is reasonable.

PR #861, opened 2023-10-05
Update doc using nerdctl

sondavidb:
**Issue #, if available:**
#844 

**Description of changes:**
With [nerdctl v1.6.0](https://github.com/containerd/nerdctl/releases/tag/v1.6.0) natively supporting SOCI, we can make our documentation workflow a bit smoother. Replaced instances of `soci create` and `soci push` with `nerdctl push` and updated/restructured documentation as such. In particular, the Getting Started guide got a fair amount of restructuring, so I would definitely appreciate a sanity check on if the ordering is still logical.

Additionally, I ran `doctoc` on each changed file.

**Testing performed:**
For the Getting Started guide, I ensured it still ran as expected. I was unable to test the output of the commands in Registry.md due to not having access to those registry types, so admittedly the output may not be accurate.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -10,10 +10,8 @@ with the SOCI snapshotter.
 - [Dependencies](#dependencies)
 - [Install the SOCI snapshotter](#install-the-soci-snapshotter)
 - [Push an image to your registry](#push-an-image-to-your-registry)
-- [Create and push SOCI index](#create-and-push-soci-index)
-  - [Create SOCI index](#create-soci-index)
+  - [About the SOCI index](#about-the-soci-index)
   - [(Optional) Inspect SOCI index and zTOC](#optional-inspect-soci-index-and-ztoc)
-  - [Push SOCI index to registry](#push-soci-index-to-registry)
 - [Run container with the SOCI snapshotter](#run-container-with-the-soci-snapshotter)
   - [Configure containerd](#configure-containerd)
   - [Start the SOCI snapshotter](#start-the-soci-snapshotter)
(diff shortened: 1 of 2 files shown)


docs/registry.md
```diff
@@ -88,7 +88,7 @@ Below are some slightly redacted examples from the services that don't support t
 **Docker Hub**
 
 ```
-$ sudo ./out/soci push --user $USERNAME:$PASSWORD docker.io/####/busybox:latest
+$ sudo nerdctl push docker.io/####/busybox:latest
```
Kern--, line 91: We might want to leave this file alone because the output is the actual shell output I got when I tested these registries.

sondavidb: My only concern is that leaving it in would potentially suggest to users that this is the preferred way of using the CLI when our preferred way is to not use it at all.

I can remove it, but I would definitely like to modify this doc to use nerdctl and show the proper expected output.

Kern--: If we're going to change the command, I'd like to actually run them and use real output.

sondavidb: Makes sense. I can try and set up these registries and get the proper output, if that doesn't work out I'll revert this change.

sondavidb: Will be reverting the changes. Would it perhaps be worth mentioning in the registry file that we generally suggest using `nerdctl push` instead of `soci push` as a disclaimer? Or is it niche enough that we don't need to bother?

Kern-- (approved):

docs/registry.md
Kern--: No, I don't think it's necessary. We can update these next time we test these registries.

turan18 (approved):

PR #864, opened 2023-10-06
Add go-fuse logger

Kern--:
**Issue #, if available:**

**Description of changes:**
soci-snapshotter-gprc overrides the default logger to be a logrus logger with debug as the write level. If the snapshotter is invoked with `--log-level debug`, then logs from e.g. `log.PrinLn` are visibile.

go-fuse uses this logging method to output a complete transaction of fuse operations keeping track of each request/response with a unique id. The ID is only unique within a single server, so trying to parse the logs is ambiguous since you don't know which server a response corresponds to.

go-fuse recently added support for per-server loggers. This change uses that functionality to add the layer digest to the logs from each fuse server so request/response pairs can be unambiguously matched.

It also changes the log level of go-fuse logs to trace since there really is no debug reason for enabling them.

**Testing performed:**

I also manually ran the snapshotter with `--log-level debug` to make sure there are no go-fuse logs and then again with `--log-level trace` to see the logs.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -45,6 +45,7 @@ package fs
 import (
 	"context"
 	"fmt"
+	golog "log"
 	"os/exec"
 	"sync"
 	"syscall"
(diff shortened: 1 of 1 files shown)


fs/fs.go
```diff
@@ -494,10 +495,16 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 		NegativeTimeout: &fs.negativeTimeout,
 		NullPermissions: true,
 	})
+	// Pass in a logger to go-fuse with the layer digest
+	// The go-fuse logs are useful for tracing exactly what's happening at the fuse level.
+	logger := log.L.
```
austinvazquez, line 500: Sanity check, we want to use the standard logger here instead of the context logger?

turan18 (approved):

fs/fs.go
Kern--: I think so. This logger is attached to the context of the layer, not the context of the Mount operation (specifically, I don't think we care about the mountpoint in the fuse logs).

sondavidb (approved):

Issue #758, opened 2023-08-01
[FEATURE] Document configuration file options


sparr:
### Description

Add to the documentation an example config.toml with all available options and the default values.

### Any additional context or information about the feature request

Forked from #405

Kern--:
We should also add a test that loads this example. As long as the snapshotter starts up, the test passes.

sondavidb:
#871 has been merged. Once #884 is implemented we can close this.

Issue #851, opened 2023-10-02
[FEATURE] Refactor FS Structure


sondavidb:
### Description

The current code structure is confusing and is largely inherited from stargz, which in turn attempts to mimic containerd ideas. Due to fundamental differences between SOCI and stargz, we should heavily look into how we structure our codebase to be more clearly aligned with how we want to structure SOCI.

This would avoid problems outlined in various PRs, most recently #849, which mutates many file headers to pass a FSConfig all the way down to the layer level, not to mention the many paths it could have taken to get there when it should all be relatively streamlined. If we had clearer abstractions and better mechanisms, it could be a trivial change.

This PR is largely meant to track our plans for design structure for this project going forward.

PR #871, opened 2023-10-16
Add example config TOML

sondavidb:
**Issue #, if available:**
#758 

**Description of changes:**
Added an example TOML in config/config.toml documenting most of the available variables that can be changed in the TOML file. I omitted some variables that are known to be unused (particularly the `Config` struct in in service/plugin/plugin.go), but there are probably still some variables in this example that have no effect. If any of them are known to do nothing, they can be changed in a later PR.

I additionally left some comments attempting to clarify what certain variables do. For example, `MaxWaitMsec=0` under `[http]` will actually set the variable to the default value stored in the codebase for MaxWaitMsec, which in this case would be 300,000. This makes sense in this context (as waiting 0ms would never allow it to happen), but for something like `min_layer_size=0` under `[snapshotter]`, where a user might want no minimum layer size, it could cause some confusion. I did my best to outline this in the comments, but perhaps this confusion could be cleared in the future (e.g. by defaulting numeric values to -1 instead of 0).

**Testing performed:**
Created a new integration test in `run_test.go` called `TestRunWithDefaultConfig`, which creates a snapshotter base shell and reboots containerd with the example TOML. If it cannot parse the example TOML, it fails.

I also verified the TOML structure by getting the config and marshalling it with gotoml (right before line 178 in cmd/soci-snapshotter-grps/main.go [here](https://github.com/awslabs/soci-snapshotter/blob/31a8bcc683f7522d146d0505d034558909e53433/cmd/soci-snapshotter-grpc/main.go#L178)), then printing to stdout.

### DIFF
--- a/cmd/soci-snapshotter-grpc/main.go
+++ b/cmd/soci-snapshotter-grpc/main.go
@@ -174,6 +174,7 @@ func main() {
 	if err != nil {
 		log.G(ctx).WithError(err).Fatalf("failed to configure metadata store")
 	}
+
 	fsOpts = append(fsOpts, fs.WithMetadataStore(mt))
 	rs, err := service.NewSociSnapshotterService(ctx, *rootDir, &cfg.ServiceConfig,
 		service.WithCredsFuncs(credsFuncs...), service.WithFilesystemOptions(fsOpts...))
--- a/config/config.toml
+++ b/config/config.toml
@@ -1,0 +1,126 @@
+# An example config showing all of the toml variables used.
+
+# Copy to /etc/soci-snapshotter-grpc/config.toml
+# to use on your system.
+
+# NOTE: Many variables set to zero are just an indicator
+# to use the built-in default. These values may change over time,
+# which is why the config uses zeroes.
+
+# TODO: Can we put these inside dedicated TOML vars to refer to them?
+
+# config/fs.go FSConfig
+http_cache_type=""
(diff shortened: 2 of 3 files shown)


sondavidb:
Created #872 to try and clean up TOML variable usage, as it's pretty unclear what is and is not actually used.

sondavidb:
After making some changes I realized that `disable_verification` being forced in the function header for getSnapshotterConfigToml causes TestRunWithDefaultConfig to fail. I Just commented it out in the example and left a comment that it can be enabled when used in `/etc/soci-snapshotter-grpc-config.toml`.

Perhaps another case to prioritize #851 a bit more?

Kern-- (approved):

turan18:
LGTM. I do think we need a `config.md` that describes what each config does, as well as other nuanced things like the fact that `0` is used to as a placeholder for some configs.

sondavidb:
Created #884 to properly document this change

I noted this already but for greater transparency — this implementation has one bug I managed to do a workaround for. `TestRunWithDefaultConfig` cannot run unless I comment out the `disable_verification` variable because otherwise it believes it has been assigned multiple times. Not sure exactly how to fix it elegantly but this is worth noting since it could be a little confusing.

I also quite frankly don't like how I implemented the test (copying the file from x location) but as the alternative is maintaining two identical copies of the same logic I suppose this is the best way to do this.

Issue #841, opened 2023-09-18
[FEATURE] Stop using `ctr` code in CLI


Kern--:
### Description

The SOCI CLI is built using flags and helper methods directly from [`ctr`](https://github.com/containerd/containerd/tree/main/cmd/ctr). Often this ends up pulling in many flags that aren't needed or used which leaves the user guessing as to which ones are actually connected ([See this discussion for an example](https://github.com/awslabs/soci-snapshotter/discussions/839)).

For one example, `soci push` has the following flags which are not connected to anything:

```
   --skip-verify, -k               Skip SSL certificate validation
   --refresh value                 Refresh token for authorization server
   --hosts-dir value               Custom hosts configuration directory
   --tlscacert value               Path to TLS root CA
   --tlscert value                 Path to TLS client certificate
   --tlskey value                  Path to TLS client key
   --http-dump                     Dump all HTTP request/responses when interacting with container registry
   --http-trace                    Enable HTTP tracing for registry interactions
   --label value                   Labels to attach to the image
   --snapshotter value             Snapshotter name. Empty value stands for the default value. [$CONTAINERD_SNAPSHOTTER]  
   --max-concurrent-uploads value  Max concurrent uploads. Default is 10 (default: 10)
```

`ctr` is meant to be a debugging tool for containerd which means that it makes [no guarantees about stability](https://github.com/containerd/containerd/blob/main/RELEASES.md#public-api-stability). 

### Describe the solution you'd like

We should remove the direct references to CTR code and copy the flags that we use into our codebase.

We shouldn't bring over flags that aren't currently connected, but we should create issues for flags that would be useful to implement. We shouldn't confine ourselves to the specific flag names, but instead address the features the flags add.

sondavidb:
Fixed with #998

PR #874, opened 2023-10-17
Remove most ctr refs from CLI

sondavidb:
**Issue #, if available:**
#841 (partial fix)

**Description of changes:**
Removed many references to ctr. Only things not removed were the `run` and the `image rpull` commands. Both were more challenging to port over — `run` is entirely delegated to ctr and may not even be worth the effort to port over, while `image rpull` uses more than just the containerd client, which might require larger scale changes should we focus on this later. All other references to ctr should be gone.

**Testing performed:**
Minimal testing on some basic operations (create, push, ztoc, ls) and ensured they remained functionally the same. I suppose the testing suite will perform some more robust tests as they use the CLI.


cmd/soci/commands/internal/client.go
```diff
@@ -0,0 +1,68 @@
...
+	gocontext "context"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/log"
+	"github.com/containerd/containerd/namespaces"
+	"github.com/containerd/containerd/pkg/epoch"
+	"github.com/urfave/cli"
+)
+
+// Largely taken from containerd/cmd/ctr/commands/client.go
```
Kern--, line 29: If this was copied, then this file needs the containerd copyright notice. I think this applies to all the files in `internal`, right?

For example, see https://github.com/awslabs/soci-snapshotter/blob/main/snapshot/snapshot.go#L1-L31.

sondavidb: Sure. What about the files that use the same containerd flags?

Kern--: Yeah, I think they're word-for-word identical, aren't they? We should add the containerd copyright.

turan18:
Overall LGTM, besides the small nit. Can we update the commit message to mention which specific commands still reference the `commands` package?

sondavidb:
> Overall LGTM, besides the small nit. Can we update the commit message to mention which specific commands still reference the `commands` package?

I think it would make the commit msg a little too verbose.

Kern-- (approved):

turan18 (approved):

Issue #447, opened 2023-02-24
[FEATURE] Automate Releases


Kern--:
**Is your feature request related to a problem? Please describe.**
As we start thinking about releasing binaries of the SOCI snapshotter, it would be good to start thinking about how to make the process repeatable to make sure we don't miss an artifacts or details.

**Describe the solution you'd like**
Ideally we would have a github actions pipeline that automatically created a github release with all the relevant artifacts whenever we push a tag.

**Describe alternatives you've considered**
N/A

**Additional context**
References:
Containerd's workflow https://github.com/containerd/containerd/blob/main/.github/workflows/release.yml
nerdctl's workflow https://github.com/containerd/nerdctl/blob/main/.github/workflows/release.yml

Kern--:
Another reference is the way Finch is doing it:
https://github.com/runfinch/finch/blob/main/.github/workflows/release-please.yaml

In particular they use [release please](https://github.com/googleapis/release-please) to automate the releases including generating the changelog.

Kern--:
Our branching strategy is defined here: https://github.com/awslabs/soci-snapshotter/blob/main/RELEASES.md#releases-1
The release artifacts are defined here: https://github.com/awslabs/soci-snapshotter/blob/main/RELEASES.md#release-artifacts

The last piece is the actual process which is:
1) Create the branch if necessary
2) Use [scripts/build_third_party_licenses.sh](https://github.com/awslabs/soci-snapshotter/blob/main/scripts/build_third_party_licenses.sh) to create the final third party licenses file. Commit it and open a PR to merge it into the release branch.
3) Create and push signed tag against the license file commit
4) Automation should take over here and create the releases according to our documentation.
5) [Maybe] Update the release change log if we take an approach like nerdctl which doesn't generate it automatically



We should probably add this documentation somewhere, but not in the releases doc because these instructions are things that only maintainers care about.

djdongjin:
fyi: nerdctl's new release doc for maintainers: https://github.com/containerd/nerdctl/blob/main/MAINTAINERS_GUIDE.md

Kern--:
A couple more notes that should go into the maintainer's version of the release doc:

1) I've been using ubuntu 18.04 when building dynamically linked binaries so that the glibc version is "relatively old" and likely to work for most people.
2) I've been building with the latest version of go at the time of the release.
3) The tag needs to be created before building so that the binary version reports correctly
4) I verify that the static binarys are actually static with ldd.

[...]

PR #876, opened 2023-10-18
Add release automation on tag push

sondavidb:
**Issue #, if available:**
#447

**Description of changes:**
Created a new workflow for automating part of the release process. On tag push, it will test the build on AMD64, create the binaries + sha256 checksums, and create a draft release with a changelog. Note that testing and building are run in parallel, which was an intentional design choice, but if we feel that the tests should pass before building the binaries, that can be changed.

ARM64 releases must still be created manually. Technically we could build the binaries with ARM64 here anyway but it would be considered poor practice to test and build in separate environments. (Still waiting for GH Actions to natively support ARM64 :v)

**Testing performed:**
I pushed a tag and ensured the workflow passed. You can see it running successfully [here](https://github.com/sondavidb/soci-snapshotter/actions/runs/6657291217/job/18091686729). The workflow was modified to include a THIRD_PARTY_LICENSES file, but otherwise is identical to the one in this PR. The workflow was triggered with the following:

```
git tag -s -a v999.999.999 -m "Release v999.999.999"
git push origin v999.999.999
```

The created release can be seen [here](https://github.com/sondavidb/soci-snapshotter/releases/tag/v999.999.999) (release will be deleted after PR is merged). Note that the changelog is empty. My guess is that this is due to previous testing with other tags making GitHub erroneously believe the last release was the tag previously pushed and deleted. This should not be an issue for official releases, but if a release needs to be modified for whatever reason, please bear in mind it may generate an empty changelog, so we should ensure we keep the original generated changelog.

To be explicit, this is not an official v999.999.999 release :P

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -1,0 +1,84 @@
+name: Release
+
+on:
+  create:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+env:
+  GO_VERSION: '1.20.6'
+
+permissions:
+  contents: write
+  deployments: write
+
+jobs:
+  # Any way we can just call build.yml?
+  test:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - run: make
+      - run: make test
+      
+  integration:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        containerd: ["1.6.19", "1.7.0"]
+    env:
+      DOCKER_BUILD_ARGS: "CONTAINERD_VERSION=${{ matrix.containerd }}"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - run: make integration
+
+  generate-artifacts:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v4
+    - name: Setup and export variables
+      run: |
+        export release_tag=${GITHUB_REF#refs/*/} # Strip down to raw tag name
+        export release_version=${release_tag/v/} # Remove v from tag name
+
+        echo "release_tag=${release_tag}">> $GITHUB_ENV
+        echo "dynamic_binary_name=soci-snapshotter-${release_version}-linux-amd64.tar.gz" >> $GITHUB_ENV
+        echo "static_binary_name=soci-snapshotter-${release_version}-linux-amd64-static.tar.gz" >> $GITHUB_ENV
+        
+        mkdir release
+    - name: Create release binaries
+      run: make RELEASE_TAG=${{ env.release_tag }} release
+    - uses: actions/upload-artifact@v3
+      with:
+        name: artifacts
+        path: release/
+        if-no-files-found: error
+
+    outputs:
+      dynamic_binary_name: ${{ env.dynamic_binary_name }}
+      static_binary_name: ${{ env.static_binary_name }}
+
+  create-release:
+    needs: generate-artifacts
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/download-artifact@v3
+        with:
+          name: artifacts
+      - uses: softprops/action-gh-release@v1
+        with:
+          draft: true
+          prerelease: false
+          generate_release_notes: true
+          files: |
+            ${{ needs.generate-artifacts.outputs.dynamic_binary_name }}
+            ${{ needs.generate-artifacts.outputs.dynamic_binary_name }}.sha256sum
+            ${{ needs.generate-artifacts.outputs.static_binary_name }}
+            ${{ needs.generate-artifacts.outputs.static_binary_name }}.sha256sum
--- a/scripts/create-releases.sh
+++ b/scripts/create-releases.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+#   Copyright The Soci Snapshotter Authors.
+
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


.github/workflows/releases.yml
```diff
@@ -0,0 +1,103 @@
+name: Releases workflow
```
austinvazquez, line 1: non-functional: `Release` just to be consistent with other workflow naming.

.github/workflows/releases.yml
```diff
@@ -0,0 +1,103 @@
...
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+env:
+  GO_VERSION: '1.20.6'
+
+permissions:
+  contents: write
+  deployments: write
+
+jobs:
+  # TODO: Any way we can just call build.yml?
```
austinvazquez, line 15: Technically possible, but no sure if there are trade-offs / effort required.
https://docs.github.com/en/actions/using-workflows/reusing-workflows

.github/workflows/releases.yml
```diff
@@ -0,0 +1,103 @@
...
+      fail-fast: false
+      matrix:
+        containerd: ["1.6.19", "1.7.0"]
+    env:
+      DOCKER_BUILD_ARGS: "CONTAINERD_VERSION=${{ matrix.containerd }}"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20.6'
```
austinvazquez, line 38: nit: use `${{ env.GO_VERSION }}`

.github/workflows/releases.yml
```diff
@@ -0,0 +1,103 @@
...
+        
+        mkdir release
+    - name: Create dynamic binary
+      run: |
+        make build
+        cp NOTICE.md THIRD_PARTY_LICENSES out/
+        cd out/
+        tar -czvf ../release/${{ env.dynamic_binary_name }} *
+        cd ../
+        rm -rf out/
```
austinvazquez, line 62: Is it worth moving some of this logic into a Make target or bash script for local runs? Thinking out loud, it might be helpful for troubleshooting. (Likewise for static build)

austinvazquez (commented):
I noticed the test release did not have the release version in the artifacts. Has that been resolved?

.github/workflows/releases.yml
sondavidb: Yeah, it seemed like it relies on a particular naming scheme as well as a few other refactors I was hesitant to change in this PR. I think it might not be a bad change in the future to make, though.

.github/workflows/releases.yml
sondavidb: I'm not against it but it's also only a few lines of pretty basic code that requires some basic preset conditions. I think this is good enough for now but if the process ever becomes more convoluted we should certainly revisit the idea.

That said, if we were to do that sooner rather than later, maybe the entire release logic should be put there.

sondavidb:
> I noticed the test release did not have the release version in the artifacts. Has that been resolved?

That would be a bug, it looks like you can't refer to GH environment variables within the same step that they're assigned. Fixed that up by just using a local variable.

.github/workflows/releases.yml
austinvazquez: Can we remove the TODO for now? It seems like there are some unknowns with this effort that need to be ironed out.

.github/workflows/releases.yml
```diff
@@ -0,0 +1,103 @@
...
+
+  generate-artifacts:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v4
+    - name: Setup and export variables
+      run: |
+        export release_tag=${GITHUB_REF#refs/*/} # Strip down to raw tag name
+        export release_version=${release_tag/v/} # Remove v from tag name
+        echo "release_version=$release_version" >> $GITHUB_ENV
```
austinvazquez, line 49: nit: is `release_version` still needed as a GitHub environment variable? Usage appears to be only scoped to this step.

sondavidb: Good catch, forgot to remove it from the previous action I tried using

.github/workflows/releases.yml
sondavidb: Is the new comment alright? I would like to still note that this is something we should look into to reduce redundant code.

austinvazquez (approved):
LGTM. Thanks.

.github/workflows/releases.yml
austinvazquez: Maybe this is fine. I might be over-indexing here, but was just concerned a community member may come along and see the TODO and take some initiative to do the work and create a PR only for maintainers to not like the outcome. So it might be a bad experience for a contributor.

sondavidb:
Moved all binary compiling logic to a separate script and created a Makefile target for `release`, per previous discussions. This should allow for a slightly smoother process for automating releases on ARM64, which currently has to manually test and set up everything.

scripts/create-releases.sh
```diff
@@ -0,0 +1,72 @@
...
+release_version=${1/v/} # Remove v from tag name
+dynamic_binary_name=soci-snapshotter-${release_version}-linux-${ARCH}.tar.gz
+static_binary_name=soci-snapshotter-${release_version}-linux-${ARCH}-static.tar.gz
+
+make build
+cp $NOTICE_FILE $LICENSE_FILE $OUT_DIR
+tar -czvf $RELEASE_DIR/$dynamic_binary_name $OUT_DIR
+rm -rf $OUT_DIR/*
+
+make build
```
Kern--, line 66: This needs `STATIC=1` doesn't it?

scripts/create-releases.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
```
Kern--, line 1: Can you run this through shellcheck? I think we're going to be linting files that way soon.

Kern-- (approved):

Kern-- (dismissed):

austinvazquez (approved):

sondavidb:
Poorly-handled merge conflicts, sorry about that x.x

Kern-- (approved):

austinvazquez (approved):

Issue #832, opened 2023-09-12
[Bug] SOCI Removes Snapshots on SIGINT


sondavidb:
### Description

When stopping the daemon via a SIGINT, SOCI makes an effort to [clean up](https://github.com/awslabs/soci-snapshotter/blob/1e0057135d22a2bd91567200be8a73ead99bb63e/cmd/soci-snapshotter-grpc/main.go#L275) after itself. This includes both unmounting the requisite directories and [removing them entirely](https://github.com/awslabs/soci-snapshotter/blob/1e0057135d22a2bd91567200be8a73ead99bb63e/snapshot/snapshot.go#L567).

This leads to a fuse mount error upon restarting the snapshotter:

```shell
/usr/bin/fusermount: bad mount point /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/2/fs: No such file or directory
```

The snapshotter then only successfully start up when setting `allow_invalid_mounts_on_restart=true` in the config. After this, pulling any other images results in an error similar to the one below:

```shell
failed to prepare extraction snapshot "extract-xxxxx": failed to stat parent: stat /var/lib/soci-snapshotter-grpc/snapshotter/snapshots/1/fs: no such file or directory: unknown
```

As I ended the task gracefully, I would expect the snapshotter to start successfully, as I still possess the original image and index, but this is not the case, and the snapshotter fails to start. I must enable `allow_invalid_mounts_on_restart=true` in the `config.toml` file to get it to start up again, and then have to manually remove the previously loaded snapshot to pull any other images. This is seemingly unintuitive, so if this is expected behavior, I would like an explanation as to why it behaves this way.

Upon inspection, the following behaviors have been observed on cleanup:
- The uncompressed snapshots on disk are cleaned
- The snapshots metadata seems to be unmodified

The combination of the two seems to clash with each other, as on startup, the snapshotter expects the previously loaded snapshots to be in the snapshots directory per its metadata, yet a graceful cleanup in this case means the snapshots are not there anymore, whereas a nongraceful kill will retain the snapshots, and thus the snapshotter starts up just fine — I can pull a separate image and run either snapshot just fine.

If removing the snapshot is intended behavior, that's fine, but we should then clean the metadata to reflect this on a graceful cleanup. I think we should mimic containerd's default behavior if we do not already have a clear direction in mind for this.

### Steps to reproduce the bug

1. Pull any image
[...]

Issue #275, opened 2023-01-03
[BUG] killing soci-snapshotter-grpc while a container is running requires manual cleanup


sparr:
**Describe the bug**
When the soci process is killed while a container is running, some of the mounts and metadata are left in a problematic state.

**Steps To Reproduce**
1. `soci-snapshotter-grpc`
2. `ctr run --snapshotter soci [...]`
3. `killall soci-snapshotter-grpc`

**Expected behavior**
All state (running processes, files, container and image metadata in the data store and registry, etc) is left in a state allowing the snapshotter and container to be started again without manual intervention.

**Additional context**
Problem statement is non-specific because I did not take sufficient notes when encountering this problem while troubleshooting a more pressing issue. Investigating will be a necessary step of resolving this issue.

rdpsin:
The problem is that FUSE mounts' lifetimes are tied to the snapshotter. If the snapshotter crashes, the FUSE mounts will die too and we have no way to re-mount them. A couple of options are:

1) Separate out the FUSE implementation from the snapshotter, so that the mounts still exist even if the snapshotter crashes.

2) Persist some kind of state on disk that will allow the snapshotter to reconstruct the FUSE mount whenever it comes back online.

Kern--:
Is this the same problem?
https://github.com/awslabs/soci-snapshotter/issues/93

Kern--:
I think the "solution" to this is the config to ignore the broken data: https://github.com/awslabs/soci-snapshotter/blob/main/config/service.go#L79

Maybe we could also have the SOCI snapshotter call to containerd to remove the broken snapshots? That seems a bit weird, though.

PR #881, opened 2023-10-20
Keep directories when SIGINT sent to daemon

sondavidb:
**Issue #, if available:**
#275, #832

**Description of changes:**
On SIGINT, the daemon will perform some cleanup functions, including removing the directories for each of the snapshot layers. However, the metadata does not reflect this change. This meant that, upon restarting the snapshot, the snapshot would not start unless `allow_invalid_mounts_on_restart=true` was set in the config. Even with the variable set, one had to manually remove the snapshot to get the daemon in a working state.

This was pretty frustrating behavior. What should be a graceful kill ends up being more work than using a non-graceful kill (i.e. SIGKILL/SIGTERM, which allow the snapshots to instantly start working again, no questions asked). Additionally, it was unintuitive (to me) that killing the daemon also wipes all the snapshotter data, forcing one to re-pull an image every time the daemon is killed.

This change removes the cleanup step of wiping the snapshot directories. It will still unmount the FUSE mounts, but the snapshotter [already tries to restore snapshots](https://github.com/awslabs/soci-snapshotter/blob/1e0057135d22a2bd91567200be8a73ead99bb63e/snapshot/snapshot.go#L199) on startup, which just unmounts any mounts leftover and remounts existing mounts in the metadata, so it ties in very well to snapshot startup

**Testing performed:**
I performed some basic testing, largely including pulling an image + running a container with it after a SIGINT and confirmed normal behavior.

### DIFF
--- a/integration/run_test.go
+++ b/integration/run_test.go
@@ -424,6 +424,51 @@ func TestRootFolderPermission(t *testing.T) {
 	}
 }
 
+func TestRestartAfterSigint(t *testing.T) {
+	const containerImage = alpineImage
+	const killTimeout = 5
+	const startTimeout = "5s"
+
+	regConfig := newRegistryConfig()
+	sh, done := newShellWithRegistry(t, regConfig)
+	defer done()
+
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, false, tcpMetricsConfig))
+	copyImage(sh, dockerhub(containerImage), regConfig.mirror(containerImage))
+	indexDigest := buildIndex(sh, regConfig.mirror(containerImage), withMinLayerSize(0), withSpanSize(100*1024))
+	sh.X("soci", "image", "rpull", "--user", regConfig.creds(), "--soci-index-digest", indexDigest, regConfig.mirror(containerImage).ref)
+	sh.X("pkill", "-SIGINT", "soci-snapshotte") // pkill can only take up to 15 chars
+
+	var buffer []byte
+	timedOut := true
+	for i := 0; i < killTimeout*4; i++ {
+		buffer = sh.O("ps")
+		if !strings.Contains(string(buffer), "soci-snapshotte") {
+			timedOut = false
+			break
+		}
+		sh.X("sleep", "0.25")
+	}
+
+	if timedOut {
+		t.Fatalf("failed to kill snapshotter daemon")
+	}
+
+	timeoutCmd := []string{"timeout", "--preserve-status", startTimeout}
+	cmd := shell.C("/usr/local/bin/soci-snapshotter-grpc", "--log-level", sociLogLevel,
+		"--address", "/run/soci-snapshotter-grpc/soci-snapshotter-grpc.sock")
+	cmd = addConfig(t, sh, getSnapshotterConfigToml(t, false, tcpMetricsConfig), cmd...)
+	cmd = append(timeoutCmd, cmd...)
+
+	// exit status 0 — command timed out, so server is running — PASS
+	// exit status 1 — server crashed before time limit — FAIL
+	if _, err := sh.OLog(cmd...); err != nil {
+		if err.Error() != "exit status 0" {
+			t.Fatalf("error starting snapshotter daemon: %v", err)
+		}
+	}
+}
+
 func TestRunInContentStore(t *testing.T) {
 	imageName := helloImage
 	sh, done := newSnapshotterBaseShell(t)
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -556,18 +556,21 @@ func (o *snapshotter) getCleanupDirectories(ctx context.Context, t storage.Trans
 }
 
 func (o *snapshotter) cleanupSnapshotDirectory(ctx context.Context, dir string) error {
-
-	// On a remote snapshot, the layer is mounted on the "fs" directory.
-	// We use Filesystem's Unmount API so that it can do necessary finalization
-	// before/after the unmount.
-	mp := filepath.Join(dir, "fs")
-	if err := o.fs.Unmount(ctx, mp); err != nil {
-		log.G(ctx).WithError(err).WithField("dir", mp).Debug("failed to unmount")
+	if err := o.unmountSnapshotDirectory(ctx, dir); err != nil {
+		return fmt.Errorf("failed to unmount directory %q: %w", dir, err)
 	}
 	if err := os.RemoveAll(dir); err != nil {
 		return fmt.Errorf("failed to remove directory %q: %w", dir, err)
 	}
 	return nil
+}
+
+func (o *snapshotter) unmountSnapshotDirectory(ctx context.Context, dir string) error {
+	// On a remote snapshot, the layer is mounted on the "fs" directory.
+	// We use Filesystem's Unmount API so that it can do necessary finalization
+	// before/after the unmount.
+	mp := filepath.Join(dir, "fs")
+	return o.fs.Unmount(ctx, mp)
 }
 
 func (o *snapshotter) createSnapshot(ctx context.Context, kind snapshots.Kind, key, parent string, opts []snapshots.Opt) (_ storage.Snapshot, err error) {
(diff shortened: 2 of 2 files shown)


snapshot/snapshot.go
```diff
@@ -564,9 +564,6 @@ func (o *snapshotter) cleanupSnapshotDirectory(ctx context.Context, dir string)
 	if err := o.fs.Unmount(ctx, mp); err != nil {
 		log.G(ctx).WithError(err).WithField("dir", mp).Debug("failed to unmount")
 	}
-	if err := os.RemoveAll(dir); err != nil {
-		return fmt.Errorf("failed to remove directory %q: %w", dir, err)
-	}
```
austinvazquez, line 569: Without the `os.RemoveAll` this function may need a name change. It is now more of a `unmountFilesystem`. But before that other than `cleanup` this function is also has call sites from `createSnapshot` [on error] and `Remove` which makes sense for the directory to be cleaned up. My question is if graceful termination of snapshotter should take a seperate call path or expand cleanup to have cleanup options.

austinvazquez (commented):
Should we have some of these manual tests as automated tests with this change?

snapshot/snapshot.go
sondavidb: Your first point is correct, but the second point is hard to make a call on. I initially added to function headers but ultimately decided to make a new struct to store this information. It's a bit extra IMO, but also would make future changes easier to make (and really, I feel like a lot of our functions could be refactored to take data like this).

Could also add a `defaultCleanupConfig` if that proves useful in any form.

sondavidb:
> Should we have some of these manual tests as automated tests with this change?

I tried, but I don't actually know how to kill the snapshotter with a SIGINT. I see that we have [KillMatchingProcess](https://github.com/awslabs/soci-snapshotter/blob/114d5ddb8b8db27e5bd588afd30044fa65ed5e00/util/testutil/shell.go#L330), which is called by [rebootContainerd](https://github.com/awslabs/soci-snapshotter/blob/114d5ddb8b8db27e5bd588afd30044fa65ed5e00/integration/util_test.go#L632), but this sends a SIGKILL, not a SIGINT. Even if I could pass it in, `rebootContainerd` also wipes the content stores so I don't believe calling that is a viable option regardless.

I tried to manually kill but `killall` isn't a binary in the container created. I can technically fudge this by banking on the grpc being a specific pid every time the container starts, but from past experiences relying on an arbitrary value far out of the scope of what I'm trying to achieve is...not ideal, to say the least.

snapshot/snapshot.go
austinvazquez: When a boolean input modifies the behavior of the function, then to me that begins to indicate the function might be doing too much. Here the problem is amplified with two boolean inputs. It seems like we have a clear logical separation here in 1) unmount filesystem and 2) cleanup directory. Maybe this leans more into having two seperate implementations with explicit reponsibilities.

austinvazquez:
> > Should we have some of these manual tests as automated tests with this change?
> 
> I tried, but I don't actually know how to kill the snapshotter with a SIGINT. I see that we have [KillMatchingProcess](https://github.com/awslabs/soci-snapshotter/blob/114d5ddb8b8db27e5bd588afd30044fa65ed5e00/util/testutil/shell.go#L330), which is called by [rebootContainerd](https://github.com/awslabs/soci-snapshotter/blob/114d5ddb8b8db27e5bd588afd30044fa65ed5e00/integration/util_test.go#L632), but this sends a SIGKILL, not a SIGINT. Even if I could pass it in, `rebootContainerd` also wipes the content stores so I don't believe calling that is a viable option regardless.
> 
> I tried to manually kill but `killall` isn't a binary in the container created. I can technically fudge this by banking on the grpc being a specific pid every time the container starts, but from past experiences relying on an arbitrary value far out of the scope of what I'm trying to achieve is...not ideal, to say the least.

Give https://man7.org/linux/man-pages/man1/kill.1.html a try and if still stuck lmk and we can look together.

snapshot/snapshot.go
sondavidb: Ah, ok, that makes sense to me. The previously passed in boolean (cleanupCommitted) was used to route to `getCleanupDirectories`, so I guess we'll keep that function name for now.

I had already split the function in two — unmounting, and cleaning, of which the latter calls the former. This is desirable behavior in most cases, including in `cleanup()`, which is called in both `o.Cleanup()` and `o.Close()`. Per the comment in the code, "Cleanup cleans up disk resources from removed or abandoned snapshots", so I created a new function instead of reusing the existing cleanup() function, which hopefully makes the separation of ideas more apparent.

sondavidb:
I ended up using `pkill` instead of `kill` since for some reason `kill` doesn't work with a name even though the man page seems to imply that.

I added the test and ensured that it crashed with the previous version and passed with this one.

integration/run_test.go
```diff
@@ -424,6 +424,36 @@ func TestRootFolderPermission(t *testing.T) {
...
+	sh, done := newShellWithRegistry(t, regConfig)
+	defer done()
+
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, false, tcpMetricsConfig))
+
+	copyImage(sh, dockerhub(containerImage), regConfig.mirror(containerImage))
+	indexDigest := buildIndex(sh, regConfig.mirror(containerImage), withMinLayerSize(0), withSpanSize(100*1024))
+	sh.X("soci", "image", "rpull", "--user", regConfig.creds(), "--soci-index-digest", indexDigest, regConfig.mirror(containerImage).ref)
+	sh.X("pkill", "-2", "soci-snapshotte") // pkill can only take up to 15 chars
+
```
sondavidb, line 440: Would like some particular attention on the second half of this testing function. Mainly want to guarantee that a. the implementation would give deterministic output, b. that 5 seconds is long enough to wait for the server to crash (or not crash), and c. there are no potential race conditions (i.e. everything is guaranteed to execute before the function finishes).

sondavidb: Realistically I know b and c are unlikely to be issues, but I've wrote tests in other projects that ended up being flaky due to checking for strange conditions and implemented weird workarounds, and I really don't want to do that again :v

Kern--: Is it possible that the previous snapshotter isn't fully cleaned up before the next one tries to start? Should we put a short delay in?

integration/run_test.go
```diff
@@ -424,6 +424,36 @@ func TestRootFolderPermission(t *testing.T) {
...
+	regConfig := newRegistryConfig()
+	sh, done := newShellWithRegistry(t, regConfig)
+	defer done()
+
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, false, tcpMetricsConfig))
+
+	copyImage(sh, dockerhub(containerImage), regConfig.mirror(containerImage))
+	indexDigest := buildIndex(sh, regConfig.mirror(containerImage), withMinLayerSize(0), withSpanSize(100*1024))
+	sh.X("soci", "image", "rpull", "--user", regConfig.creds(), "--soci-index-digest", indexDigest, regConfig.mirror(containerImage).ref)
+	sh.X("pkill", "-2", "soci-snapshotte") // pkill can only take up to 15 chars
```
Kern--, line 439: nit: can we use `"-SIGINT"` instead of `"-2"` for clarity?

integration/run_test.go
sondavidb: I like that idea. I wonder if it's worth also doing a `ps -aux | grep soci` and printing that to log? It seems pretty unnecessary though.

Kern--: Speaking of which, maybe we could use something like `ps | grep soci` in a loop with a few hundred ms delay to determine when the process is gone rather than sleeping a full 5s. I'm not going to block on that, though.

Kern-- (approved):

integration/run_test.go
sondavidb: I like the idea, made said changes. Any chance we can apply this to the command that starts the server as well? I don't think so since lowering it would give a higher chance of a false positive, but wondering if you agree.

turan18 (dismissed):

Kern-- (dismissed):

snapshot/snapshot.go
```diff
@@ -743,12 +749,28 @@ func (o *snapshotter) Close() error {
...
+	}
+
+	log.G(ctx).Debugf("unmount: dirs=%v", cleanup)
+	for _, dir := range cleanup {
+		if err := o.unmountSnapshotDirectory(ctx, dir); err != nil {
+			log.G(ctx).WithError(err).WithField("path", dir).Warn("failed to unmount directory")
+		}
+	}
+
+	return nil
```
austinvazquez, line 771: Should we return first unmount error?

snapshot/snapshot.go
```diff
@@ -556,17 +556,23 @@ func (o *snapshotter) getCleanupDirectories(ctx context.Context, t storage.Trans
...
 	// We use Filesystem's Unmount API so that it can do necessary finalization
 	// before/after the unmount.
 	mp := filepath.Join(dir, "fs")
 	if err := o.fs.Unmount(ctx, mp); err != nil {
 		log.G(ctx).WithError(err).WithField("dir", mp).Debug("failed to unmount")
 	}
-	if err := os.RemoveAll(dir); err != nil {
-		return fmt.Errorf("failed to remove directory %q: %w", dir, err)
-	}
 	return nil
```
austinvazquez, line 576: Just noticed after the changes, this function always returns nil. Perhaps we can simplify the function now to:

```suggestion
    return o.fs.Unmount(ctx, mp)
```

snapshot/snapshot.go
sondavidb: I don't think so. We get passed every directory, including ones without a SOCI index, so we're basically guaranteed an error. We should definitely try to unmount every directory given to us.

What we really should be doing is only attempting to unmount directories that have a SOCI index attached to them, but that's probably another change for another PR :P

snapshot/snapshot.go
sondavidb: Sure, was gonna mention the fact that we'd lose logs but it looks like we double log anyway for every call to this.

austinvazquez: Yeah generally it is good to either log or return an error but not both. Although remote snapshotters sometimes need to break that mold because the error may be returned to containerd without be logged in the snapshotter's logs. /shrug

austinvazquez (approved):

turan18 (approved):

PR #886, opened 2023-10-24
Add global FUSE failure metric

turan18:
**Issue #, if available:**

**Description of changes:**

Add a FUSE failure metric at a global/process level. 
Added telemetry helper functions around FUSE operations.

**Testing performed:**

### DIFF
--- a/fs/layer/node.go
+++ b/fs/layer/node.go
@@ -187,14 +187,6 @@ func (f *FuseOperationCounter) Run(ctx context.Context) {
 	}
 }
 
-func incFuseOpFailureMetric(operationName string, layer digest.Digest) {
-	metric, ok := fuseOpFailureMetrics[operationName]
-	if !ok {
-		metric = commonmetrics.FuseUnknownFailureCount
-	}
-	commonmetrics.IncOperationCount(metric, layer)
-}
-
 // logFSOperations may cause sensitive information to be emitted to logs
 // e.g. filenames and paths within an image
 func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseInode uint32, opaque OverlayOpaqueType, logFSOperations bool, opCounter *FuseOperationCounter) (fusefs.InodeEmbedder, error) {
@@ -244,6 +236,31 @@ func (fs *fs) inodeOfStatFile() uint64 {
 	return (uint64(fs.baseInode) << 32) | 2 // reserved
 }
 
+// logAndIncrementOpCounter handles logging as well as incrementing
+// the FuseOperationCounter.
+func (fs *fs) logAndIncrementOpCounter(ctx context.Context, operationName string, path string) {
+	if fs.logFSOperations {
+		log.G(ctx).WithFields(logrus.Fields{
+			"operation": operationName,
+			"path":      path,
+		}).Debug("FUSE operation")
+	}
+	if fs.operationCounter != nil {
+		fs.operationCounter.Inc(operationName)
+	}
+}
+
+// reportFailure handles telemetry operations pertaining to FUSE failures
+// as well as writing an error to the state file.
+func (fs *fs) reportFailure(operationName string, stateError error) {
+	metric, ok := fuseOpFailureMetrics[operationName]
+	if !ok {
+		metric = commonmetrics.FuseUnknownFailureCount
+	}
+	commonmetrics.ReportFuseFailure(metric, fs.layerDigest)
+	fs.s.report(stateError)
+}
+
 func (fs *fs) inodeOfID(id uint32) (uint64, error) {
 	// 0 is reserved by go-fuse 1 and 2 are reserved by the state dir
 	if id > ^uint32(0)-3 {
@@ -264,15 +281,6 @@ type node struct {
 	entsMu     sync.Mutex
 }
 
-func (n *node) logOperation(ctx context.Context, operationName string) {
-	if n.fs.logFSOperations {
-		log.G(ctx).WithFields(logrus.Fields{
-			"operation": operationName,
-			"path":      n.Path(nil),
-		}).Debug("FUSE operation")
-	}
-}
-
 func (n *node) isRootNode() bool {
 	return n.id == n.fs.rootID
 }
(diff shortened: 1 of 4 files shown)


fs/layer/node.go
```diff
@@ -244,6 +244,26 @@ func (fs *fs) inodeOfStatFile() uint64 {
...
 
+// basicTelemetry wraps informational FUSE telemetry operations.
+func (fs *fs) basicTelemetry(ctx context.Context, operationName string, path string) {
+	if fs.logFSOperations {
+		log.G(ctx).WithFields(logrus.Fields{
+			"operation": operationName,
+			"path":      path,
+		}).Debug("FUSE operation")
+	}
+	if fs.operationCounter != nil {
```
sondavidb, line 255: Is there a case where this would not already be set? If not might be worth removing, as it could create some confusion.

sondavidb (approved):

fs/layer/node.go
turan18: We create a FUSE operation counter for every image, so it is always set. I think having a `nil` here is a good preventative measure against unexpected runtime errors, if we ever accidentally change any behavior.

fs/metrics/common/metrics.go
```diff
@@ -226,3 +232,36 @@ func AddBytesCount(operation string, layer digest.Digest, bytes int64) {
...
+// in a separate routine so that it does not block the main
+// thread.
+func ListenForFuseFailure(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-FuseFailureSignal:
+			IncOperationCount(FuseFailureState, digest.Digest(""))
+			select {
```
austinvazquez, line 248: Perhaps consider using an anonymous function for the nested select block.

e.g. 
```go
incrementWithBackoff := func(ctx context.Context) {
    IncOperationCount(FuseFailureState, digest.Digest(""))
    
    select {
    case <-ctx.Done():
        return
    case <-time.After(...):
        return
    }
}

for {
    select {
    case <-ctx.Done():
        return
    case <-FuseFailureSignal:
        incrementWithBackoff(...)
    }
}
```

Functionally the same, but gives a name to the inner select. (minor nit/not required)

fs/metrics/common/metrics.go
austinvazquez, line 1: Should we consider adding some unit tests to validate that reoccurring failures will not trigger a failure state alarm outside the backoff window?

fs/layer/node.go
```diff
@@ -244,6 +244,26 @@ func (fs *fs) inodeOfStatFile() uint64 {
 	return (uint64(fs.baseInode) << 32) | 2 // reserved
 }
 
+// basicTelemetry wraps informational FUSE telemetry operations.
+func (fs *fs) basicTelemetry(ctx context.Context, operationName string, path string) {
```
austinvazquez, line 248: Hmm, can we brainstorm this a little more. This function is logging and incrementing a counter. Maybe `logAndIncrementOperation`?

fs/layer/node.go
```diff
@@ -244,6 +244,26 @@ func (fs *fs) inodeOfStatFile() uint64 {
...
+	}
+	if fs.operationCounter != nil {
+		fs.operationCounter.Inc(operationName)
+	}
+}
+
+// reportTelemetry wraps FUSE failure telemetry operations.
+func (fs *fs) reportTelemetry(operationName string, stateError error) {
+	commonmetrics.ReportFuseFailure()
+	incFuseOpFailureMetric(operationName, fs.layerDigest)
```
austinvazquez, line 263: This to me feels like the filesystem is having to report the error twice? Maybe could be combined into a single responsibility.

fs/layer/node.go
```diff
@@ -244,6 +244,26 @@ func (fs *fs) inodeOfStatFile() uint64 {
...
+			"path":      path,
+		}).Debug("FUSE operation")
+	}
+	if fs.operationCounter != nil {
+		fs.operationCounter.Inc(operationName)
+	}
+}
+
+// reportTelemetry wraps FUSE failure telemetry operations.
+func (fs *fs) reportTelemetry(operationName string, stateError error) {
```
austinvazquez, line 261: nit: name does not communicate to me the failure state requirement for this function.

Kern-- (approved):
Agree with the small comments about naming. Overall approach looks good.

fs/metrics/common/metrics.go
```diff
@@ -226,3 +232,36 @@ func AddBytesCount(operation string, layer digest.Digest, bytes int64) {
...
+			case <-ctx.Done():
+				return
+			case <-time.After(5 * time.Minute):
```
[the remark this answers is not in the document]
turan18: Added a global `timeBlock` var. We can work on making this configurable in subsequent PR's.

fs/metrics/common/metrics.go
turan18: Added a unit test that checks that the counter value with concurrent sends.

fs/layer/node.go
turan18: `ReportFuseFailure` now handles incrementing and notifying.

fs/metrics/common/metrics.go
turan18: The integration test I added consistently produces 2 `file.Read` failures, so we are also testing this implicitly.

fs/layer/node.go
```diff
@@ -187,15 +187,7 @@ func (f *FuseOperationCounter) Run(ctx context.Context) {
...
-func incFuseOpFailureMetric(operationName string, layer digest.Digest) {
-	metric, ok := fuseOpFailureMetrics[operationName]
-	if !ok {
-		metric = commonmetrics.FuseUnknownFailureCount
-	}
-	commonmetrics.IncOperationCount(metric, layer)
-}
-
-// logFSOperations may cause sensitive information to be emitted to logs
+// logAndIncrementOpCounter may cause sensitive information to be emitted to logs
```
austinvazquez, line 190: Can you take a look here? The comment change does not appear to be intentional.

austinvazquez (approved):
Overall LGTM. Thanks for pushing this through!

fs/layer/node.go
turan18: Oops, fixed.

fs/metrics/common/metrics.go
austinvazquez: Thanks for looking into that @turan18. Yes agreed, ideally it would be a good candidate for unit testing but I understand the current design does not easily enable that. Something we should consider in a refactor but I am aligned we can take that as a seperate issue.

austinvazquez (approved):

sondavidb (approved):

PR #896, opened 2023-10-27
Improve benchmark make rules

Kern--:
**Issue #, if available:**

**Description of changes:**
This change improves benchmark make rules to reduce duplication, add dependencies between targets, and properly mark phony targets.

It also adds a new `BENCHMARK_FLAGS` option so that we can control how the benchmarks run from make, e.g.

```
BENCHMARK_FLAGS='-count 10' make benchmarks
```

**Testing performed:**
```
make benchmarks
make build-benchmarks
BENCHMARK_FLAGS='-count 1' make benchmarks-perf-test
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ CMD=soci-snapshotter-grpc soci
 
 CMD_BINARIES=$(addprefix $(OUTDIR)/,$(CMD))
 
-.PHONY: all build check add-ltag install uninstall clean test integration
+.PHONY: all build check add-ltag install uninstall clean test integration benchmarks build-benchmarks benchmarks-perf-test benchmarks-comparison-test
 
 all: build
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -44,7 +44,7 @@ CMD=soci-snapshotter-grpc soci
 
 CMD_BINARIES=$(addprefix $(OUTDIR)/,$(CMD))
 
-.PHONY: all build check add-ltag install uninstall clean test integration
+.PHONY: all build check add-ltag install uninstall clean test integration benchmarks build-benchmarks benchmarks-pert-test benchmarks-comparison-test
```
austinvazquez, line 47: typo: `benchmarks-perf-test`*

austinvazquez (approved):
LGTM

Makefile
```diff
@@ -92,19 +92,23 @@ integration: build
...
 
-benchmarks:
-	@echo "$@"
-	@cd benchmark/performanceTest ; GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/PerfTests . && sudo ../bin/PerfTests
-	@cd benchmark/comparisonTest ;  GO111MODULE=$(GO111MODULE_VALUE) go build -o ../bin/CompTests . && sudo ../bin/CompTests
+benchmarks: benchmarks-perf-test benchmarks-comparison-test
+
+build-benchmarks: benchmark/bin/PerfTests benchmark/bin/CompTests
+
+benchmark/bin/PerfTests: FORCE
```
sondavidb, line 99: nit: Any reason not to add this to phony targets? I presume it's because it's never intended to be directly called and thus helps create a separation of ideas in that sense?

sondavidb (approved):

Makefile
Kern--: It's not _technically_ a phony target - it really does produce this file. I think there's a question on whether we should consider our build targets to be phone and let go handle incremental builds (which is effectively what we're doing now), or actually have the targets depend on the go source files. 

I'm fine with either, I just want it all to be consistent and right now that means non-phony using `FORCE`.

PR #908, opened 2023-11-02
Allow benchmarks config as json

Kern--:
**Issue #, if available:**
N/A

**Description of changes:**
Benchmarks can currently be passed as a csv file. This additionally adds json as an option to make it not order dependent and to allow us to add more complex options in the future (e.g. mounts).

**Testing performed:**

### DIFF
--- a/benchmark/utils.go
+++ b/benchmark/utils.go
@@ -18,24 +18,49 @@ package benchmark
 
 import (
 	"encoding/csv"
+	"encoding/json"
 	"errors"
+	"io"
 	"os"
 	"os/exec"
 )
 
 type ImageDescriptor struct {
-	ShortName            string
-	ImageRef             string
-	SociIndexManifestRef string
-	ReadyLine            string
+	ShortName            string `json:"short_name"`
+	ImageRef             string `json:"image_ref"`
+	SociIndexManifestRef string `json:"soci_index_manifest_ref"`
+	ReadyLine            string `json:"ready_line"`
 }
 
-func GetImageListFromCsv(csvLoc string) ([]ImageDescriptor, error) {
-	csvFile, err := os.Open(csvLoc)
+func GetImageList(file string) ([]ImageDescriptor, error) {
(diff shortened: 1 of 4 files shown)


benchmark/utils.go
```diff
@@ -18,16 +18,43 @@ package benchmark
...
-	ImageRef             string
-	SociIndexManifestRef string
-	ReadyLine            string
+	ShortName            string `json:"short_name"`
+	ImageRef             string `json:"image_ref"`
+	SociIndexManifestRef string `json:"soci_index_manifest_ref"`
+	ReadyLine            string `json:"ready_line"`
+}
+
+func GetImageList(file string) ([]ImageDescriptor, error) {
```
sondavidb, line 34: nit: Might be slightly faster to open the file here and then parse it in their respective functions? Or would attempting to read in csv.NewReader/json.Unmarshal mutate the file descriptor even on failure?

sondavidb (dismissed):
Just food for thought — should we have a function explicitly to detect a file type before sending them to their respective functions? The logical flow here makes sense but I wonder if it would be a bit clearer to separate out the two functions. But maybe not, since as far as I'm aware, the only way to detect if a file is a json is unmarshaling it, which if you do it just to check, you would have to do it again to store the data anyway. But just something to consider.

austinvazquez (approved):

turan18 (dismissed):

benchmark/utils.go
Kern--: Done.

Kern--:
I don't think it makes sense to try to determine the file path without parsing the file. We could alternatively add a new flag for json files (although `-f` for csv makes just about anything for json feel like a second-class option), or my preference would be to eventually deprecate csv parsing for being too error prone.

sondavidb (approved):

austinvazquez (approved):

turan18 (approved):

PR #910, opened 2023-11-03
Pin GitHub Actions packages using commit SHA

austinvazquez:
**Issue #, if available:**
N/A

**Description of changes:**
This pull request pins third-party GitHub Actions packages by commit SHA based on recommended best practices in [Security Hardening for GitHub Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

Note [dependabot](https://github.com/dependabot/dependabot-core/issues/4691) will keep the human readable version tag up-to-date.

**Testing performed:**

### DIFF
--- a/.github/workflows/benchmark_visualization.yml
+++ b/.github/workflows/benchmark_visualization.yml
@@ -17,18 +17,18 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout main branch
-        uses: actions/checkout@v4
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
           fetch-depth: 0
           submodules: true
-      - uses: actions/setup-go@v4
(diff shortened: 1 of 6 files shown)


Kern--:
Is dependabot going to automatically update these anyway? Do we really gain anything with this?

austinvazquez:
> Is dependabot going to automatically update these anyway? Do we really gain anything with this?

Dependabot will keep these packages updated, but the approve and merge step is still manual. The benefit is partially build reproducibility but there also is a security aspect.

sondavidb:
This practice seems to only apply for third-party actions based on the link you provided. I think we should keep most of the actions as is and only pin the non-GH Actions ones (aka create-pull-request and action-gh-release) to a specific commit.

austinvazquez:
Going to close this one for now and we can revisit it if needed. The original intent was to pin CI dependencies to for security concerns but the likelihood of backdoors being introduced into actions packages might not be likely.

PR #912, opened 2023-11-03
Remove "soci run" from CLI

sondavidb:
With the effort to remove ctr code from CLI, we removed `soci run`, as it ran `ctr run` under the hood with no additional functionality.

In the testing suite, `soci run` has been replaced in favor of `nerdctl run` for similar reasons to above. `nerdctl run`, while it attempts to mimic ctr, does not function exactly the same as ctr out of the box. In particular default behavior in `nerdctl run` deviates from `ctr run` in the following ways:

- Auto generates an ID. `ctr run` requires the container have an ID after the image ref. Circumvented with `--name [id]` (not exactly the same, but good enough for tasks to find the container.
- Uses the [bridge CNI plugin](https://www.cni.dev/plugins/current/main/bridge/) for networking, which would add an unnecessary dependency. Circumvented this with `--net none` (maybe `--net host` might be more suitable for some tests? Unsure though)
-  Automatically pulls an image if the name is not found, which would cause some of our tests to unexpectedly succeed. Circumvented this with `--pull never`.
- Automatically closes after a task is finished, which would cause some of our tests to fail, as a container would start without a task but would need to stay running for later tasks. Circumvented this with `sleep infinity` for those tasks.

These are behaviors that I noticed but did not fully understand.
- `cannot remove containerd task: Device or resource busy` in `TestRunInNamespace`. Maybe resources between namespaces are shared strangely? In any case, adding `--rm` seemed to fix this
- `stdout` would output an empty string on task failure. This was noted in `TestNetworkRetry`, which is why the additional test condition was given. Really not sure as to why that works this way, since the tests should only return an error and thus stdout should be empty. Maybe because it actually prints an error message instead of nothing like ctr does?

It should also be noted that I had to replace `ctr task exec` with `nerdctl exec` because otherwise ctr could not find the containers to run the tasks in.

A lot of the replacement was done automatically via a custom shell script (which I can share if it helps) but proofread manually, I tried to be extremely thorough to ensure no funny replacements slipped through the cracks but do bear that in mind.

**Issue #, if available:**
#841

**Description of changes:**

**Testing performed:**
`make test` and `make integration`

### DIFF
--- a/integration/metrics_test.go
+++ b/integration/metrics_test.go
@@ -218,7 +218,7 @@ log_fuse_operations = true
 
 			sh.X("soci", "image", "rpull", "--soci-index-digest", indexDigest, imgInfo.ref)
 			// this command may fail due to fuse operation failure, use XLog to avoid crashing shell
-			sh.XLog("ctr", "run", "--rm", "--snapshotter=soci", imgInfo.ref, "test", "echo", "hi")
+			sh.XLog(append(runSociCmd, "--name", "test", "--rm", imgInfo.ref, "echo", "hi")...)
 
 			curlOutput := string(sh.O("curl", tcpMetricsAddress+metricsPath))
 			checkFuseOperationFailureMetrics(t, curlOutput, tc.metricToCheck, tc.expectFuseOperationFailure, tc.expectedCount)
@@ -253,7 +253,7 @@ fuse_metrics_emit_wait_duration_sec = 10
 			indexDigest := buildIndex(sh, imgInfo)
 
 			sh.X("soci", "image", "rpull", "--soci-index-digest", indexDigest, imgInfo.ref)
-			sh.XLog("ctr", "run", "-d", "--snapshotter=soci", imgInfo.ref, "test", "echo", "hi")
+			sh.XLog(append(runSociCmd, "--name", "test", "-d", imgInfo.ref, "echo", "hi")...)
 
 			curlOutput := string(sh.O("curl", tcpMetricsAddress+metricsPath))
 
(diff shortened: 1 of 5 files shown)


integration/metrics_test.go
```diff
@@ -218,7 +218,7 @@ log_fuse_operations = true
 
 			sh.X("soci", "image", "rpull", "--soci-index-digest", indexDigest, imgInfo.ref)
 			// this command may fail due to fuse operation failure, use XLog to avoid crashing shell
-			sh.XLog("ctr", "run", "--rm", "--snapshotter=soci", imgInfo.ref, "test", "echo", "hi")
+			sh.XLog("nerdctl", "run", "--name", "test", "--pull", "never", "--net", "none", "--rm", "--snapshotter", "soci", imgInfo.ref, "echo", "hi")
```
turan18, line 221: Should we extract parts of these common commands into a shared location?

eg:

```go
// Commands
runCmd := []string{"nerdctl"", "run", "--pull", "never", "--net", "none", "--snapshotter", "soci"}
pullCmd := []string{"soci", "image", "rpull"}

// pull an image
sh.X(append(pullCmd, "testImage")...)
// run a container
sh.X(append(runCmd, "testImage", "echo", "hello")...)

```

sondavidb: I like the idea — trying to refactor this was a massive pain for that reason. Could be something shared around the package in like utils.go or something. Should that be a separate PR, though?

Kern--: I don’t think that refactor makes this PR any larger or more complex, so it’s probably fine to do it now.

sondavidb: In that case I'll just stick to changing the run commands. I can do a separate PR after this is merged consolidating all of the other commands.

Kern-- (approved):

turan18 (approved):

PR #930, opened 2023-11-09
Redact HTTP query values from logs

austinvazquez:
**Issue #, if available:**
N/A

**Description of changes:**
HTTP client logs are mostly disabled with the exception of a request retry log. The issue observed is that error messages may contain the full HTTP request including the query component which can contain sensitive information like credentials or session tokens. To prevent leaking sensitive information, this change will redact HTTP query values from log messages.

**Testing performed:**
Added unit tests and benchmarks to measure overhead for redacting HTTP query values from errors in logs.

```
=== RUN   TestRedactHTTPQueryValuesFromError
=== RUN   TestRedactHTTPQueryValuesFromError/NilError
=== RUN   TestRedactHTTPQueryValuesFromError/NonURLError
=== RUN   TestRedactHTTPQueryValuesFromError/ErrorWithNoHTTPQuery
=== RUN   TestRedactHTTPQueryValuesFromError/ErrorWithHTTPQuery
--- PASS: TestRedactHTTPQueryValuesFromError (0.00s)
    --- PASS: TestRedactHTTPQueryValuesFromError/NilError (0.00s)
    --- PASS: TestRedactHTTPQueryValuesFromError/NonURLError (0.00s)
    --- PASS: TestRedactHTTPQueryValuesFromError/ErrorWithNoHTTPQuery (0.00s)
    --- PASS: TestRedactHTTPQueryValuesFromError/ErrorWithHTTPQuery (0.00s)
=== RUN   TestRedactHTTPQueryValuesFromURL
=== RUN   TestRedactHTTPQueryValuesFromURL/NilURL
=== RUN   TestRedactHTTPQueryValuesFromURL/URLWithNoQueries
=== RUN   TestRedactHTTPQueryValuesFromURL/URLWithQueries
--- PASS: TestRedactHTTPQueryValuesFromURL (0.00s)
    --- PASS: TestRedactHTTPQueryValuesFromURL/NilURL (0.00s)
    --- PASS: TestRedactHTTPQueryValuesFromURL/URLWithNoQueries (0.00s)
    --- PASS: TestRedactHTTPQueryValuesFromURL/URLWithQueries (0.00s)
PASS
coverage: 92.9% of statements
ok      github.com/awslabs/soci-snapshotter/util/http/log       1.017s  coverage: 92.9% of statements
```

```
goos: linux
goarch: amd64
pkg: github.com/awslabs/soci-snapshotter/util/http/log
cpu: Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz
BenchmarkRedactHTTPQueryValuesOverhead
BenchmarkRedactHTTPQueryValuesOverhead/BaselineLogging
BenchmarkRedactHTTPQueryValuesOverhead/BaselineLogging-2         	   12108	    104393 ns/op	    6220 B/op	      35 allocs/op
BenchmarkRedactHTTPQueryValuesOverhead/WithoutReplacement
BenchmarkRedactHTTPQueryValuesOverhead/WithoutReplacement-2      	   10000	    108078 ns/op	    6374 B/op	      42 allocs/op
BenchmarkRedactHTTPQueryValuesOverhead/WithErrorReplacement
BenchmarkRedactHTTPQueryValuesOverhead/WithErrorReplacement-2    	   10000	    114341 ns/op	    6799 B/op	      53 allocs/op
PASS
ok  	github.com/awslabs/soci-snapshotter/util/http/log	5.549s
```

### DIFF
--- a/util/http/log/redact_http_query_values.go
+++ b/util/http/log/redact_http_query_values.go
@@ -1,0 +1,51 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package log
+
+import (
+	"errors"
+	"net/url"
+)
+
+// RedactHTTPQueryValuesFromError is a log utility to parse an error as a URL error and redact
+// HTTP query values to prevent leaking sensitive information like encoded credentials or tokens.
+func RedactHTTPQueryValuesFromError(err error) error {
+	var urlErr *url.Error
+
+	if err != nil && errors.As(err, &urlErr) {
+		url, urlParseErr := url.Parse(urlErr.URL)
+		if urlParseErr == nil {
+			RedactHTTPQueryValuesFromURL(url)
+			urlErr.URL = url.Redacted()
+			return urlErr
+		}
+	}
+
+	return err
+}
+
+// RedactHTTPQueryValuesFromURL redacts HTTP query values from a URL.
+func RedactHTTPQueryValuesFromURL(url *url.URL) {
+	if url != nil {
+		if query := url.Query(); len(query) > 0 {
+			for k := range query {
+				query.Set(k, "redacted")
+			}
+			url.RawQuery = query.Encode()
+		}
+	}
+}
--- a/util/http/log/redact_http_query_values_test.go
+++ b/util/http/log/redact_http_query_values_test.go
@@ -1,0 +1,223 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package log
+
+import (
+	"bytes"
+	"errors"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	// mockURL is a fake URL modeling soci-snapshotter fetching content from S3.
+	mockURL = "https://s3.us-east-1.amazonaws.com/981ebdad55863b3631dce86a228a3ea230dc87673a06a7d216b1275d4dd707c9/12d7153d7eee2fd595a25e5378384f1ae4b6a1658298a54c5bd3f951ec50b7cb"
+
+	// mockQuery is a fake HTTP query with sensitive information which should be redacted.
+	mockQuery = "?username=admin&password=admin"
+
+	// redactedQuery is the expected result of redacting mockQuery.
+	// The query values will be sorted by key as a side-effect of encoding the URL query string back into the URL.
+	// See https://pkg.go.dev/net/url#Values.Encode
+	redactedQuery = "?password=redacted&username=redacted"
+)
+
+func TestRedactHTTPQueryValuesFromError(t *testing.T) {
+	testcases := []struct {
+		Name        string
+		Description string
+		Err         error
+		Assert      func(*testing.T, error)
+	}{
+		{
+			Name:        "NilError",
+			Description: "Utility should handle nil error gracefully",
+			Err:         nil,
+			Assert: func(t *testing.T, actual error) {
+				if actual != nil {
+					t.Fatalf("Expected nil error, got '%v'", actual)
+				}
+			},
+		},
+		{
+			Name:        "NonURLError",
+			Description: "Utility should not modify an error if error is not a URL error",
+			Err:         errors.New("this error is not a URL error"),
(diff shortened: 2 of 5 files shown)


turan18 (approved):

util/http/log/redact_http_query_values.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+var (
+	// httpQueryRegex is a compiled regular expression of httpQueryRegexPattern.
+	httpQueryRegex = regexp.MustCompile(httpQueryRegexPattern)
+)
+
+// RedactHTTPQueryValues parses an error and replaces any occurrences of an HTTP Query
+// value with '<redacted>' to prevent leakage of sensitive information like credentials.
+func RedactHTTPQueryValues(err error) error {
```
Kern--, line 51: I think we should do this in a more structured way by parsing the error as a `url.Error`

e.g. https://go.dev/play/p/LuQg7-Kl8mK

util/http/retry.go
```diff
@@ -83,7 +84,7 @@ func RetryStrategy(ctx context.Context, resp *http.Response, err error) (bool, e
 	retry, err2 := rhttp.DefaultRetryPolicy(ctx, resp, err)
 	if retry {
 		log.G(ctx).WithFields(logrus.Fields{
-			"error":    err,
+			"error":    logutil.RedactHTTPQueryValues(err),
```
Kern--, line 87: This feels error prone (what if we add a new log somewhere else?)

Similar to my other comment, I think we should do this in a structured way and I think it should be embedded in our http.Client. We might need to defer the embedding it until after #759.

Kern-- (changes requested):

util/http/retry.go
austinvazquez: Makes sense. I can create a task on the tracking issue for having this within the unified client.

Just for completeness, another approach considered was to utilize `logrus.Hook`s to have the redaction run on each log message. This might alleviate the concern for any added logs. The trade-off is taking a direct dependency on `logrus` (We already do this in other parts of the snapshotter). LMK if you have thoughts on that approach or if the current approach is fine.

util/http/log/redact_http_query_values_test.go
```diff
@@ -0,0 +1,163 @@
...
+					t.Fatalf("Expected '%s', got '%v'", expected, actual)
+				}
+			},
+		},
+		{
+			Name:        "ErrorWithNoHTTPQuery",
+			Description: "Utility should not modify an error if no HTTP queries are present.",
+			Err: &url.Error{
+				Op:  "GET",
+				URL: "https://s3.us-east-1.amazonaws.com/981ebdad55863b3631dce86a228a3ea230dc87673a06a7d216b1275d4dd707c9/12d7153d7eee2fd595a25e5378384f1ae4b6a1658298a54c5bd3f951ec50b7cb",
```
Kern--, line 62: Is this a real URL? and if so, what is it?

util/http/log/redact_http_query_values.go
Kern--: LGTM

util/http/retry.go
Kern--: Some of these errors escape over grpc. I think we should redact as close to the source as possible. In fact, we should probably be redacting `err2` before we return it at the end of this function.

Kern-- (changes requested):

util/http/log/redact_http_query_values_test.go
austinvazquez: The URL is fake itself is fake, but modeled after a real error.

Kern--: I think it would be helpful to either add that as a comment or make it look more obviously fake.

util/http/retry.go
austinvazquez: `err2` should only be context deadline exceeded or context cancelled error, but overhead is minimal so no harm in checking either.

util/http/log/redact_http_query_values_test.go
austinvazquez: Sounds good. Done in latest changeset.

util/http/retry.go
turan18: Just realized that we would still be implicitly logging the error if we fail on the last attempt: https://github.com/hashicorp/go-retryablehttp/blob/main/client.go#L759-L768. 

We log the error on every retry. If we have expired our retries, the rhttp client will just return the error and that will end up bubbling up to the call site. The `Client` has an [`ErrorHandler`](https://github.com/hashicorp/go-retryablehttp/blob/main/client.go#L422) attached, that will be called when we have exhausted our retries. We should set create an `ErrorHandler` that will redact parts of the error before returning.

austinvazquez: @turan18, good catch. I'll look into covering that one as well with this PR.

austinvazquez: @turan18, added error handler to latest commit. PTAL.

util/http/log/redact_http_query_values.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// RedactHTTPQueryValuesFromError is a log utility to parse an error as a URL error and redact
+// HTTP query values to prevent leaking sensitive information like encoded credentials or tokens.
+func RedactHTTPQueryValuesFromError(err error) error {
+	var urlErr *url.Error
+
+	if err != nil && errors.As(err, &urlErr) {
+		url, urlParseErr := url.Parse(urlErr.URL)
+		if urlParseErr != nil {
```
turan18, line 31: nit: can we leave the url as is and invert the condition

```go

url, urlParseErr := url.Parse(urlErr.URL)
if urlParseErr == nil { 
  RedactHTTPQueryValuesFromURL(url)
  urlErr.URL = url.Redacted()
  return urlErr
}
return err
```

austinvazquez: It's difficult. It seems to be a trade-off between security and maintainability. Could a malformed URL contain sensitive information? Maybe. Is it likely for SOCI to have malformed URLs? Probably not.

I guess it is fine to take the suggestion as it is simpler and continue to monitor.

Kern-- (approved):

turan18 (approved):

util/http/retry.go
turan18: LGTM thanks!

PR #938, opened 2023-11-13
Benchmark options

Kern--:
**Issue #, if available:**
N/A

**Description of changes:**

All of our existing benchmark images are run with the same set of containerd
options which are only configurable at the test level to control things
like which snapshotter is used.

This is a problem for benchmarking GPU workloads, for example, where we need to
pass additional options to mount the GPU in the container which don't
apply to all images in the test.

Additionally, our benchmarker assumes that the benchmarked images
require no configuration, however this can make experimentation hard in
cases where a single base-image can be used for multiple use cases
depending on environment variables, confiration mounts, etc.

This change adds the ability to configure image-specific options when
loading benchmarks from json. The options are not required and if not
passed, the benchmarker will behave as it did before this change.

The set of options available in this change are those that were
necessary for benchmarking the LLM workloads that I was trying to test.
They are not comprehensive, but can be built upon as use cases arise.

---

I kept 2 commits to separate some refactoring that I needed to make these change from the changes themselves. I'm happy to squash or pull out the refactoring into a separate commit if that's preferred.

**Testing performed:**
~/benchmark.json
```
[{
  "short_name": "djl-serving",
  "image_ref": "<ACCOUNT>.dkr.ecr.us-west-2.amazonaws.com/djl-serving:0.25.0-deepspeed",
  "soci_index_manifest_ref": "<INDEX DIGEST>",
  "ready_line": "ModelServer BOTH API bind to: http://0.0.0.0:8080/",
  "timeout_sec": 1000,
  "options": {
    "net": "host",
    "gpu": true,
    "shm_size": 1000000,
    "mounts": [
      {
        "source": "/home/ubuntu/model",
        "destination": "/opt/ml/model",
        "type": "bind",
        "options": ["rbind", "ro"]
      },
      {
        "source": "/home/ubuntu/model_logs",
        "destination": "/opt/djl/logs",
        "type": "bind",
        "options": ["rbind", "rw"]
      }
    ]
  }
}]
```

```
BENCHMARK_FLAGS="-count 1 -f ~benchmark-tgi.json" make benchmarks-perf-test
BENCHMARK_FLAGS="-count 1 -f ~/benchmark-tgi.json" make benchmarks-perf-test
```

### DIFF
--- a/benchmark/framework/containerd_utils.go
+++ b/benchmark/framework/containerd_utils.go
@@ -32,7 +32,6 @@ import (
 	"github.com/containerd/containerd/cio"
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/namespaces"
-	"github.com/containerd/containerd/oci"
 	"github.com/containerd/log"
 	"github.com/sirupsen/logrus"
 )
@@ -138,11 +137,8 @@ func (proc *ContainerdProcess) DeleteImage(ctx context.Context, imageRef string)
 
 func (proc *ContainerdProcess) CreateContainer(
 	ctx context.Context,
-	image containerd.Image,
 	opts ...containerd.NewContainerOpts) (containerd.Container, func(), error) {
 	id := fmt.Sprintf("%s-%d", testContainerID, time.Now().UnixNano())
-	opts = append(opts, containerd.WithNewSnapshot(id, image))
-	opts = append(opts, containerd.WithNewSpec(oci.WithImageConfig(image)))
 	container, err := proc.Client.NewContainer(
 		ctx,
 		id,
@@ -205,7 +201,8 @@ func (proc *ContainerdProcess) CreateTask(
 func (proc *ContainerdProcess) RunContainerTaskForReadyLine(
 	ctx context.Context,
 	taskDetails *TaskDetails,
-	readyLine string) (func(), error) {
+	readyLine string,
+	timeout time.Duration) (func(), error) {
 	stdoutScanner := bufio.NewScanner(taskDetails.stdoutReader)
 	stderrScanner := bufio.NewScanner(taskDetails.stderrReader)
 
@@ -214,7 +211,7 @@ func (proc *ContainerdProcess) RunContainerTaskForReadyLine(
 		return nil, err
 	}
 	resultChannel := make(chan string, 1)
-	timeoutCtx, cancel := context.WithTimeout(context.Background(), 180*time.Second)
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 	go func() {
 		select {
--- a/benchmark/performanceTest/main.go
+++ b/benchmark/performanceTest/main.go
@@ -94,15 +94,12 @@ func main() {
 	var drivers []framework.BenchmarkTestDriver
 	for _, image := range imageList {
 		shortName := image.ShortName
-		imageRef := image.ImageRef
-		sociIndexManifestRef := image.SociIndexManifestRef
-		readyLine := image.ReadyLine
 		testName := "SociFull" + shortName
 		driver := framework.BenchmarkTestDriver{
 			TestName:      testName,
 			NumberOfTests: numberOfTests,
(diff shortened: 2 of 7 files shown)


benchmark/utils.go
```diff
@@ -20,16 +20,87 @@ import (
...
+const testContainerID = "TEST_RUN_CONTAINER"
+
 type ImageDescriptor struct {
-	ShortName            string `json:"short_name"`
-	ImageRef             string `json:"image_ref"`
-	SociIndexManifestRef string `json:"soci_index_manifest_ref"`
-	ReadyLine            string `json:"ready_line"`
+	ShortName            string       `json:"short_name"`
+	ImageRef             string       `json:"image_ref"`
+	SociIndexManifestRef string       `json:"soci_index_manifest_ref"`
```
turan18, line 40: unrelated nit: Can we rename this to `SociIndexManifestDigest` or `SociIndexDigest`?

benchmark/utils.go
```diff
@@ -20,16 +20,87 @@ import (
...
+
+func (i *ImageDescriptor) Timeout() time.Duration {
+	if i.TimeoutSec <= 0 {
+		return 180 * time.Second
+	}
+	return time.Duration(i.TimeoutSec) * time.Second
+}
+
+// ImageOptions contains image-specific options needed to run the tests
+type ImageOptions struct {
```
turan18, line 54: Given image customization is tied to JSON, should we consider removing CSV entirely? We can track that in a separate issue.

turan18 (approved):

benchmark/performanceTest/main.go
```diff
@@ -94,15 +94,12 @@ func main() {
 	var drivers []framework.BenchmarkTestDriver
 	for _, image := range imageList {
```
austinvazquez, line 95: Do we need `image := image` here as well?

benchmark/stargzTest/main.go
```diff
@@ -61,13 +61,11 @@ func main() {
 	var drivers []framework.BenchmarkTestDriver
 	for _, image := range imageList {
```
austinvazquez, line 62: Also here?

benchmark/framework/containerd_utils.go
```diff
@@ -138,11 +137,8 @@ func (proc *ContainerdProcess) DeleteImage(ctx context.Context, imageRef string)
 
 func (proc *ContainerdProcess) CreateContainer(
 	ctx context.Context,
-	image containerd.Image,
-	opts ...containerd.NewContainerOpts) (containerd.Container, func(), error) {
+	opts []containerd.NewContainerOpts) (containerd.Container, func(), error) {
```
sondavidb, line 140: Any particular reason for changing the header this way?

sondavidb (commented):
Unrelated, but is there a reason benchmarkTests.go is in camelcase and everything else uses underscores?

benchmark/utils.go
Kern--: Yeah, that was generally the direction I wanted to go. I'll create an issue.

benchmark/framework/containerd_utils.go
Kern--: Whoops. At one point I was playing around with this function taking both `[]containerd.NewContainerOpts` and `[]oci.SpecOpts` so they were both slices instead of varargs. That ended up being confusing and error prone, so I went back to this signature, but forgot to switch the type back. Fixed in the next commit.

benchmark/performanceTest/main.go
austinvazquez: Ah, I see thanks.

austinvazquez (approved):

turan18 (approved):

Kern--:
> Unrelated, but is there a reason benchmarkTests.go is in camelcase and everything else uses underscores?

No, I don't think so. Historical accident.

benchmark/utils.go
Kern--: https://github.com/awslabs/soci-snapshotter/issues/946

PR #940, opened 2023-11-15
Replace "soci image rpull" with common base command

sondavidb:
**Issue #, if available:**

**Description of changes:**
Similar to #912's change making "nerdctl run" consolidated into one common command, this has the same function.

**Testing performed:**
go fmt passed. Should be no functional changes so as long as the replacement was done properly it should be alright.

### DIFF
--- a/integration/metrics_test.go
+++ b/integration/metrics_test.go
@@ -131,7 +131,7 @@ func TestOverlayFallbackMetric(t *testing.T) {
 				sh.X("nerdctl", "pull", "-q", imgInfo.ref)
 				indexDigest := tc.indexDigestFn(sh, contentStoreType, imgInfo)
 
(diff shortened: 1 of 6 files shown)


sondavidb:
That's awkward...turns out flags don't pass down the hierarchy in our CLI (which is strange, because [it seems they're supported](https://github.com/urfave/cli/issues/142)).

Kern--:
Oh, right. This is a slightly different issue than that upstream issue. urfave flags are tightly scoped to the command hierarchy.
```
soci --content-store soci image rpull
```
is not the same as

```
soci image rpull --content-store soci
```

The former is expecting the flag to be defined on the root command while the latter is expecting the flag to be defined on the `rpull` command.

sondavidb:
Ah, I see. But I would assume that kind of behavior is supported? I recall someone mentioning that it is supported in v3, so maybe we can hold off on this until then (unless this is already supported).

sondavidb:
Closing this for now since this can't work with current version of urfave/cli CLI.

Issue #920, opened 2023-11-06
[Bug] Refresh creds on forbidden


turan18:
### Description

Right now, we only [refresh](https://github.com/awslabs/soci-snapshotter/blob/ca55a07e2811d7fc0319b349a81990b2343d2bc6/fs/remote/resolver.go#L257-L258) credentials on `Unauthorized`/`401` when trying to fetch spans. Some registries return `403 Forbidden` on credential expiry. We should also consider `403` responses worthy of cred refreshing.

### Describe the results you expected

 We should consider `403` responses worthy of cred refreshing.

### Host information

1. OS: 
2. Snapshotter Version: 
3. Containerd Version:

PR #945, opened 2023-11-15
Support re-authentication on 403

turan18:
**Issue #, if available:**

Resolves: #920 

**Description of changes:**
Although in most cases 403 responses represent authorization issues that generally cannot be resolved by re-authentication, some registries like ECR, will return a 403 on credential expiration. We will attempt to re-authenticate only if the response body indicates credential expiration.

Ref: https://docs.aws.amazon.com/AmazonECR/latest/userguide/common-errors-docker.html#error-403

**Testing performed:**

### DIFF
--- a/util/http/errors.go
+++ b/util/http/errors.go
@@ -1,0 +1,19 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package http
+
+const ECRTokenExpiredResponse = "Your authorization token has expired. Reauthenticate and try again."
--- a/util/http/retry.go
+++ b/util/http/retry.go
@@ -17,7 +17,9 @@
 package http
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"math/rand"
@@ -28,6 +30,7 @@ import (
 	"github.com/awslabs/soci-snapshotter/config"
 	logutil "github.com/awslabs/soci-snapshotter/util/http/log"
 	"github.com/awslabs/soci-snapshotter/version"
+	"github.com/containerd/containerd/remotes/docker"
 	"github.com/containerd/log"
 	rhttp "github.com/hashicorp/go-retryablehttp"
 	"github.com/sirupsen/logrus"
@@ -102,7 +105,7 @@ func HandleHTTPError(resp *http.Response, err error, attempts int) (*http.Respon
 	)
 
 	if resp != nil {
-		drain(resp.Body)
+		Drain(resp.Body)
 
 		if resp.Request != nil {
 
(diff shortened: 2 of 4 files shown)


util/http/retry.go
```diff
@@ -89,3 +92,39 @@ func RetryStrategy(ctx context.Context, resp *http.Response, err error) (bool, e
...
+		// credential expiration. (ref https://docs.aws.amazon.com/AmazonECR/latest/userguide/common-errors-docker.html#error-403)
+		// In the case of ECR, the response body is structured according to the error format defined in the
+		// Docker v2 API spec. (ref https://distribution.github.io/distribution/spec/api/#errors).
+
+		// We will attempt to parse the response body as a `docker.Errors`. If it can be parsed
+		// we will ensure that the `Message` represents token expiration.
+
+		var errs docker.Errors
+		respBody := bytes.NewBuffer([]byte{})
+		_, err := io.Copy(respBody, resp.Body)
```
turan18, line 116: Not sure if we should employ some retry policy here?

util/http/errors.go
```diff
@@ -0,0 +1,19 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package http
+
+const ECRTokenExpiredResponse = "Your authorization token has expired. Reauthenticate and try again."
```
turan18, line 19: Wasn't sure where to put this. 

I do think this is a good place for storing general errors and it should enforce us to standardize some of our errors in resolver.go.

util/http/retry.go
```diff
@@ -89,3 +92,40 @@ func RetryStrategy(ctx context.Context, resp *http.Response, err error) (bool, e
 	}
 	return retry, err2
 }
+
+// ShouldReauthenticate takes an HTTP response and determines whether or not
+// it warrants re-authentication.
+func ShouldReauthenticate(resp *http.Response) bool {
+	defer resp.Body.Close()
```
Kern--, line 99: Hmm, we shouldn't consume the response body unless it actually is a 403. I think right now, this function eats the response whether or not there's actually an error, right?

We should also document that this will consume the response body on 403. In the future we might try to log the response body on failure or something which would not work.

util/http/retry.go
Kern--: Maybe an easier way is to do this?
```
json.NewDecoder(resp.Body).Decode(&errs)
```

Kern-- (changes requested):

util/http/retry.go
turan18: Oh right, we should close the body if the response warrants re-authentication.  I'll remove this from here and leave it up to the caller to decide what to do with the response.

turan18: In terms of re-reading the response, we can copy the body to buffer and re-assign in case the caller wants read from it.

fs/remote/resolver_test.go
```diff
@@ -369,3 +369,90 @@ func (u *userAgentRoundTripper) RoundTrip(req *http.Request) (*http.Response, er
...
+func (d *dummyAuthorizer) AddResponses(ctx context.Context, resp []*http.Response) error {
+	return nil
+}
+
+func (d *dummyAuthorizer) Authorize(ctx context.Context, req *http.Request) error {
+	d.authCount = d.authCount + 1
+	return nil
+}
+
+func TestReauthentication(t *testing.T) {
```
austinvazquez, line 393: +1, nice unit tests.

util/http/retry.go
```diff
@@ -89,3 +93,50 @@ func RetryStrategy(ctx context.Context, resp *http.Response, err error) (bool, e
...
+		// Docker v2 API spec. (ref https://distribution.github.io/distribution/spec/api/#errors).
+		// We will attempt to decode the response body as a `docker.Errors`. If it can be decoded,
+		// we will ensure that the `Message` represents token expiration.
+
+		// Since we drain the response body, we will copy it to a
+		// buffer and re-assign it so that callers can still read
+		// from it.
+		body, err := io.ReadAll(resp.Body)
+		defer func() {
+			resp.Body = io.NopCloser(bytes.NewReader(body))
```
austinvazquez, line 119: Should we close the old instance of `resp.Body` before reinstantiating with a new `ReadCloser`? I'm wondering if no doing so would result in the TCP connection not being reused.

Kern-- (dismissed):

util/http/retry.go
turan18: Yes, we should, good call.

austinvazquez (approved):
LGTM

Kern-- (approved):

Issue #687, opened 2023-06-23
[FEATURE] [CLI] Support rootless containerd


Kern--:
### Description

[Containerd/nerdctl have instructions](https://rootlesscontaine.rs/getting-started/containerd/) for setting up a rootless containerd with [rootlesskit](https://github.com/rootless-containers/rootlesskit). What this does is it launches containerd in a user namespace with some additional tools so that it can setup containers without root. 

[Nerdctl has a special code path](https://github.com/containerd/nerdctl/blob/main/pkg/rootlessutil/parent_linux.go#L86-L99) to enter the containerd namespace when not running as root by re-execing itself.

The SOCI CLI doesn't have this, so in order to work with rootless containerd, users need to manually enter the namespace before running SOCI commands:
```
$ pid=$(cat $XDG_RUNTIME_DIR/containerd-rootless/child_pid)
$ nsenter --user --mount --target $pid
```
After these commands, the new shell that the user finds themselves in works as expected:
```
# soci create docker.io/library/rabbitmq:latest
layer sha256:685bcbcad0b80b451a2415d07660eee735255aebf24055f0373ad754e8a43ec7 -> ztoc skipped
layer sha256:db610c556cbba7c80925a3c63541e67b83115ce413433a9ee10c814f81755df2 -> ztoc skipped
layer sha256:046b2545f90296ab8ebbfe6f83bc8a6cb48d3b5687c63e0b876ba3e32a6198d7 -> ztoc skipped
layer sha256:be93a5592724b5b988c99fd89f1de826660e326542ac0d8c5349134449d4c73c -> ztoc skipped
layer sha256:8733cec7e1ec21393fac1f2e08a0378b235ef2a96042926bfaa1c12441aba22c -> ztoc skipped
layer sha256:9d03529088ec86fd9d0df9ff4a95508f3df5372db5ea3c9c14d24b796dada8e1 -> ztoc skipped
layer sha256:c37a463dc99cf3733550357349eaeb27d4ce00e2e873b9c1bd3a0f7e8d4b8094 -> ztoc skipped
layer sha256:c414cb9e1350361b7afd48be1ff0382065a228d71fdd7d7d7af8ff6fee5643a5 -> ztoc sha256:8fa3584379f7df8141c979722cab39e2115b25331fd4062b48293fba9ef9781a
[...]

PR #950, opened 2023-11-19
feat: re-exec if rootless using nsenter

tuananh:
**Issue #, if available:**
Fixes: #687 

**Description of changes:**

**Testing performed:**

I setup containerd rootless mode, nsenter and then run soci-snapshoter in there.

Tested `index list` command

Need to create another PR for snapshotter too if this works

image


tuananh:
@Kern-- is this the correct way to go?

tuananh:
@Kern-- is this the correct way to go?

cmd/soci/main.go
```diff
@@ -97,8 +103,67 @@ func main() {
...
+		return nil
+	}
+
 	if err := app.Run(os.Args); err != nil {
 		fmt.Fprintf(os.Stderr, "soci: %v\n", err)
 		os.Exit(1)
 	}
 }
+
+func parentMain() error {
```
Kern--, line 120: It looks like a lot of this code came directly from nerdctl. That's fine since it's apache 2.0, but we need to add a comment that indicates where this came from and a note about the license/copywrite. 

If we can pull this out into another file, it's pretty easy to mark it with a header like: https://github.com/awslabs/soci-snapshotter/blob/6dcfca7bc0c70b00a5f23813d468379d6fbc8de5/fs/fs.go#L1-L31

cmd/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/awslabs/soci-snapshotter/cmd
 go 1.20
 
 require (
-	github.com/awslabs/soci-snapshotter v0.0.0-local
+	github.com/awslabs/soci-snapshotter v0.4.0
```
Kern--, line 6: This doesn't look right. Can we reduce the dependency changes down to just the ones needed for rootless?

It looks like there was a larger update here.

Kern-- (changes requested):
The overall direction looks right. A few comments on dependency updates and license requirements.

I'd also like to see that this works for building indexes too (I think it should, but just to be sure)

sondavidb:
Is the intention of this to be able to use rootless containerd with soci? If so, would we also have to make the daemon usable without root? Just curious on the general direction of this PR.

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/containerd/containerd v1.7.8
 	github.com/containerd/continuity v0.4.3
 	github.com/containerd/log v0.1.0
+	github.com/containerd/nerdctl v1.7.0
```
Kern--, line 9: I think this creates a circular dependency in nerdctl. Maybe we should see if we can get the rootless util package in a separate module.

Kern-- (commented):
Sorry for the delayed response. I think we will need to work with the nerdctl maintainers on this.

Issue #884, opened 2023-10-23
[FEATURE] Document TOML variables


sondavidb:
### Description

With #871 being merged as a good first step, we should document what each of the variables do.

### Describe the solution you'd like

Documentation of the example config.toml.

PR #954, opened 2023-11-22
Add documentation for most TOML variables

sondavidb:
**Issue #, if available:**
#884

**Description of changes:**
Added documentation for the example config.toml added from #871. Many variables were either unused or uncertain in their usage, so I put a best effort to describe each variable's impact on the application's behavior. Things related to k8s/CRI authentication were ignored as I do not have a good way to know how they work.

**Testing performed:**

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1,0 +1,97 @@
+# Config file options
+
+SOCI has a multitude of variables that can customize SOCI behavior to best suit users' needs. This doc will break down the possible values for each variable, where the variable can be found, and what it can do.
+
+This doc assumes the usage of a TOML file. Per TOML specs, Anything in [brackets] is the required header for the set of variables.
+
+An example configuration file can be found in [config/config.toml](../config/config.toml).
+
+The following format will be used throughout this doc:
+
+## File path
+### [toml_header]
+- `toml_var_name` (type) — Brief description. Default: some value.
+
+#
+
+# Uncategorized
+
+This set of variables must be at the top of your TOML file due to not belonging to any TOML header.
+
+## config/fs.go
+### FSConfig
+- `resolve_result_entry` (int) — Max amount of entries allowed in the cache. Default: 30.
+- `debug` (bool) — Enables debugging for go-fuse in logs. This often emits sensitive data, so this should be false in production. Default: false.
+- `disable_verification` (bool) — Allows skipping TOC validation, which can give slight performance improvements if files have already been verified elsewhere. Default: false.
+- `no_prometheus` (bool) — Toggle prometheus metrics. Default: false.
+- `mount_timeout_sec` (int) — Timeout for mount if a layer can't be resolved. Default: 30.
+- `fuse_metrics_emit_wait_duration_sec` (int) — The wait time before the snaphotter emits FUSE operation counts for an image. Default: 60.
+
+## config/config.go
+### Config
+- `metrics_address` (string) — If empty, no metrics will be polled. Default: "".
+- `metrics_network` (string) — Chooses protocol to send metrics over (e.g. tcp, unix, etc). Default: "tcp".
+- `no_prometheus` — Defined [above](#configfsgofsconfig), cannot be redeclared.
+- `debug_address` (string) — Address where [go pprof](https://pkg.go.dev/net/http/pprof) server will listen. If empty, no logs will be emitted. Default: "".
+- `metadata_store` (string) — Metadata storage type. Only "db" is valid. Default: "db".
+
+#
+
+# Categorized
+
+## config/fs.go
+
+### [http]
+- `MaxRetries` (int) — Max retries before giving up on a network request. Default: 8.
+- `MinWaitMsec` (int) — Min time between network request attempts. Default: 30.
+- `MaxWaitMsec` (int) — Max time between network request attempts. Default: 300000.
+- `DialTimeoutMsec` (int) — Max time for a connection before timeout. Default: 3000.
+- `ResponseHeaderTimeoutMsec` (int) — Maximum duration waiting for response headers before timeout. Default: 3000.
+- `RequestTimeoutMsec` (int) — Maximum duration waiting for entire request before timeout. Default: 30000.
+
+### [blob]
+- `valid_interval` (int) — Checks blob regularly at this interval in seconds. Default: 60.
+- `check_always` (bool) — Always check blobs. Default: false.
+- `fetching_timeout_sec` (int) — Sets maximum amount of seconds to wait before failing to fetch. Default: 300.
+- `max_retries` — Blob level MaxRetries. Will override the global MaxRetries set in [[http]](#http).
+- `min_wait_msec` — Blob level MinWaitMsec. Will override the global MinWaitMsec set in [[http]](#http).
+- `max_wait_msec` — Blob level MaxWaitMsec. Will override the global MaxWaitMsec set in in [[http]](#http).
+- `max_span_verification_retries` (int) — Defines number of retries if blob fetch fails. Default: 0.
+
+### [directory_cache]
+- `max_lru_cache_entry` (int) — Max items in Least Recently Used (LRU) Cache. Default: 10.
+- `max_cache_fds`  (int) — Max file descriptors in Least Recently Used (LRU) Cache. Default: 10.
+- `sync_add` (bool) — When true, synchronously adds data to cache. Default: false. 
+
+### [fuse]
+- `attr_timeout` (int) — Max timeout for a file system in seconds. Default: 1.
+- `entry_timeout` (int) — TTL for a directory name lookup in seconds. Default: 1.
+- `negative_timeout` (int) — Defines overall entry timeout for failed lookups in seconds. Default: 1.
+- `log_fuse_operations` (bool) — Similar to `debug`, enables debugging for FUSE FS in logs. This often emits sensitive data, so this should be false in production. Default: false.
+
+### [background_fetch]
+- `disable` (bool) — Disables the background fetcher. Default: false.
+- `silence_period_msec` (int) — Time that the background fetcher will be paused when a new image is mounted. Default: 30000.
+- `fetch_period_msec` (int) — How often spans will be fetched. Default: 500.
+- `max_queue_size` (int) — Max span managers that can be queued. Default: 100.
+- `emit_metric_period_sec` (int) — Interval of background fetcher metric emission. Default: 10.
+
+### [content_store]
+- `type` (string) — Sets content store (e.g. "soci", "containerd"). Default: "soci".
+- `namespace` (string) — Default: "default".
+
+## config/resolver.go
+
+### [resolver]
+#### [resolver.host]
+#### [resolver.host.examplehost]
+#### [[resolver.host.examplehost.mirrors]]
+- `host` (string) — hostname. Default: "".
+- `insecure` (bool) — Allows usage of http instead of https only. Default: true.
+- `request_timeout_sec` (int) — Timeout in seconds of each request to the registry. Default: infinity.
+
+## config/service.go
+
+### [snapshotter]
+- `min_layer_size` (int) — Sets the minimum threshold for lazy loading a layer. Any layer smaller than this value will ignore the zTOC for the layer and pull the entire layer ahead of time. We generally recommend setting it to 10MiB (10000000). Default: 0.
+- `allow_invalid_mounts_on_restart` (bool) — Allows the snapshotter to start even if preexisting snapshots cannot connect to their data source on startup. Useful on unexpected daemon crashes/corruption. Default: false.
--- a/config/config.toml
+++ b/config/config.toml
@@ -6,13 +6,14 @@
 # NOTE: Many variables set to zero are just an indicator
 # to use the built-in default. These values may change over time,
 # which is why the config uses zeroes.
+# Please see docs/config.md for more info on what these variables do.
 
 # TODO: Can we put these inside dedicated TOML vars to refer to them?
 
 # config/fs.go FSConfig
 http_cache_type=""
 filesystem_cache_type=""
-resolve_result_entry=0 # Actually zero
+resolve_result_entry=0
 debug=false
 allow_no_verification=true
 # disable_verification=false


docs/config.md
```diff
@@ -0,0 +1,135 @@
...
+- `max_wait_msec` — Overridden by MaxWaitMsec in [[http]](#http)
+- `max_span_verification_retries` — MaxSpanVerificationRetries (int). Defines the number of additional times to attempt to attempt to fetch in case of span verification failure. Defaults to 0.
+
+### [directory_cache] DirectoryCacheConfig
+- `max_lru_cache_entry` — MaxLRUCacheEntry (int). Max number of items in Least Recently Used (LRU) Cache. Defaults to 0, which sets it to the value of defaultMaxLRUCacheEntry (10).
+- `max_cache_fds` — MaxCacheFds (int). Max number of file descriptors in Least Recently Used (LRU) Cache. Defaults to 0, which sets it to the value of defaultMaxCacheFds (10).
+- `sync_add` — SyncAdd (bool). When true, wait until the data is fully written to the cache directory before proceeding with operations. Defaults to false. 
+- `direct` — Direct (bool). Unused variable. Defaults to true.
+
+### [fuse] FuseConfig
```
sondavidb, line 70: This section I didn't really understand super well.

docs/config.md
```diff
@@ -0,0 +1,135 @@
...
+- `insecure` — Insecure (bool). Allows usage of http instead of https only. Defaults to true.
+- `request_timeout_sec` — RequestTimeoutSec (int). Timeout, in seconds, of each request to the registry. Defaults to 0, which creates no timeout.
+
+## config/service.go
+
+### [snapshotter] SnapshotterConfig
+- `min_layer_size` — MinLayerSize (int). Sets the minimum threshold for creating a SOCI index for a layer. Defaults to 0, which sets it to defaultMinLayerSize (10MiB).
+- `allow_invalid_mounts_on_restart` — AllowInvalidMountsOnRestart (bool). Allows the snapshotter to start even if preexisting snapshots cannot connect to their data source on startup. Defaults to false.
+
+TODO: Study and explain behavior of below variables.
```
austinvazquez, line 102: Let's just exclude these for now until these are better understood.

docs/config.md
```diff
@@ -0,0 +1,135 @@
...
+- `allow_no_verification` — AllowNoVerification (bool). Unused variable. Defaults to false.
+- `disable_verification` — DisableVerification (bool). Allows skipping TOC validation, which can give slight performance improvements if files have already been verified elsewhere. Defaults to false.
+- `max_concurrency` — MaxConcurrency (int). Unused variable. Defaults to 0.
+- `no_prometheus` — NoPrometheus (bool). Allows users to not use prometheus metrics. Defaults to false.
+- `mount_timeout_sec` — MountTimeoutSec (int). Amount of time a mount will take to time out if a layer can't be resolved. Defaults to 0, which sets it to the value of defaultMountTimeoutSec (30).
+- `fuse_metrics_emit_wait_duration_sec` — FuseMetricsEmitWaitDurationSec (int). Amount of time the snapshotter will wait before emitting the metrics for a FUSE operation. Defaults to 0, which sets it to the value of defaultMountTimeoutSec (60).
+
+## config/config.go
+### Config
+- `metrics_address` — MetricsAddress (string). Directory to emit Prometheus metrics to. If `no_prometheus=true`, this will have no effect. Defaults to empty string, which means no metrics will be emitted.
```
Kern--, line 36: nit: this is the address that the prometheus metrics server listens on. Metrics are polled, not emitted.

docs/config.md
```diff
@@ -0,0 +1,135 @@
...
+- `metrics_network` — MetricsNetwork (string). Chooses protocol to send metrics over (e.g. tcp, unix, etc). Defaults to empty string, which will send metrics via TCP.
+- `no_prometheus` — Defined [above](#configfsgofsconfig), cannot be redeclared
+- `debug_address` — DebugAddress (string). Address to send debug logs to. Defaults to empty string, which means no debugging metrics will be emitted.
```
Kern--, line 39: nit: similar to above, this is the address on which the go pprof servier will listen: https://pkg.go.dev/net/http/pprof

docs/config.md
```diff
@@ -0,0 +1,135 @@
...
+- `ResponseHeaderTimeoutMsec` — ResponseHeaderTimeoutMsec (int). Maximum duration waiting for response headers before a request attempt is timed out. This does not include reading the body. Defaults to 0, which sets it to the value of defaultResponseHeaderTimeoutMsec (3000).
+- `RequestTimeoutMsec` — RequestTimeoutMsec (int). Maximum duration before the entire request attempt is timed out. Defaults to 0, which sets it to the value of defaultRequestTimeoutMsec (30000).
+
+### [blob] BlobConfig
+- `valid_interval` — ValidInterval (int). Checks blob regularly at this interval in seconds. Defaults to 0, which sets it to the value of defaultValidIntervalSec (60).
+- `check_always` — CheckAlways (bool). Overrides `valid_interval` to be properly 0. Defaults to false.
+- `fetching_timeout_sec` — FetchTimeoutSec (int). Sets maximum amount of secodns to wait before failing to fetch. Defaults to 0, which sets it to the value of defaultFetchTimeoutSec (300).
+- `force_single_range_mode` — ForceSingleRnageMode (bool). Unused variable. Defaults to false.
+- `max_retries` — Overridden by MaxRetries in [[http]](#http)
```
Kern--, line 59: This is backwards right? This overrides the MaxRetries in http?

docs/config.md
```diff
@@ -0,0 +1,135 @@
...
+### [resolver.host] HostConfig
+### [resolver.host.yourmirrorconfigname1.mirrors] MirrorHost
+- `host` — Host (string). Hostname to apply this config to. Defaults to empty string.
+- `insecure` — Insecure (bool). Allows usage of http instead of https only. Defaults to true.
+- `request_timeout_sec` — RequestTimeoutSec (int). Timeout, in seconds, of each request to the registry. Defaults to 0, which creates no timeout.
+
+## config/service.go
+
+### [snapshotter] SnapshotterConfig
+- `min_layer_size` — MinLayerSize (int). Sets the minimum threshold for creating a SOCI index for a layer. Defaults to 0, which sets it to defaultMinLayerSize (10MiB).
```
Kern--, line 99: This is the threshold for lazily loading a layer. If the layer is smaller than this value, then the snapshotter will ignore a ztoc for the layer and pull it ahead of time anyway. It also actually defaults to 0 meaning it will never ignore ztocs.

docs/config.md
```diff
@@ -0,0 +1,135 @@
...
+
+#
+
+# Uncategorized
+
+This set of variables must be at the top of your TOML file due to not belonging to any TOML header.
+
+## config/fs.go
+### FSConfig
+- `http_cache_type` — HTTPCacheType (string). When set to "memory" it will create an HTTP cache in memory, otherwise it will create a cache on the disk. Defaults to empty string. (Unused?)
```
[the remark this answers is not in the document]
sondavidb: My reasoning was that, regardless of if it was used or not, it should be covered to show that we are aware it exists. But it probably creates too much clutter so I can remove these

docs/config.md
```diff
@@ -0,0 +1,135 @@
...
+
+The following format will be used throughout this doc:
+
+## File path
+### StructName [toml_header]
+- `toml_var_name` — StructVarName (type). Brief description. Defaults to some value.
+
+#
+
+# Uncategorized
```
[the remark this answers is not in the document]
sondavidb: Not really, it's just here because there's no TOML variable corresponding to each of these variables despite residing inside their respective structs.

docs/config.md
sondavidb: Actually, is http even used? I recall looking into it and finding out that the http client is just legacy code from stargz and blob client is the actual client we use.

In that case, should we just remove the entire http section?

docs/config.md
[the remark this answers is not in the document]
sondavidb: > Adding some in place examples would help as well.

As in example values for the variables? I attempted to put them in a couple of places but unsure if more would be helpful.

docs/config.md
```diff
@@ -0,0 +1,94 @@
...
+
+### [http]
+- `MaxRetries` (int) — Max retries before giving up on a network request. Default: 8.
+- `MinWaitMsec` (int) — Min time between network request attempts. Default: 30.
+- `MaxWaitMsec` (int) — Max time between network request attempts. Default: 30.
+- `DialTimeoutMsec` (int) — Max time for a connection before timeout. Default: 3000.
+- `ResponseHeaderTimeoutMsec` (int) — Maximum duration waiting for response headers before timeout. Default: 3000.
+- `RequestTimeoutMsec` (int) — Maximum duration waiting for entire request before timeout. Default: 30000.
+
+### [blob]
```
sondavidb, line 50: Would it be helpful to explain exactly what a blob is? Feels like it would be difficult to elaborate on what they do without explaining some of the technical detail, but again I wouldn't want this to be too overwhelming.

docs/config.md
```diff
@@ -0,0 +1,94 @@
...
+- `max_cache_fds`  (int) — Max file descriptors in Least Recently Used (LRU) Cache. Default: 10.
+- `sync_add` (bool) — When true, synchronously adds data to cache. Default: false. 
+
+### [fuse]
+- `attr_timeout` (int) — Max timeout for a file system in seconds. Default: 1.
+- `entry_timeout` (int) — TTL for a directory name lookup in seconds. Default: 1.
+- `negative_timeout` (int) — Defines overall entry timeout for failed lookups in seconds. Default: 1.
+- `log_fuse_operations` (bool) — Similar to `debug`, enables debugging for FUSE FS in logs. This often emits sensitive data, so this should be false in production. Default: false.
+
+### [background_fetch]
```
sondavidb, line 70: Similarly, would it be helpful to explain exactly what a span is?

sondavidb (commented):
Hopefully this is a bit more concise and relevant? I'm still worried it's too verbose but I'm struggling to know when exactly I should stop explaining.

docs/config.md
```diff
@@ -0,0 +1,96 @@
...
+- `disable_verification` (bool) — Allows skipping TOC validation, which can give slight performance improvements if files have already been verified elsewhere. Default: false.
+- `no_prometheus` (bool) — Toggle prometheus metrics. Default: false.
+- `mount_timeout_sec` (int) — Timeout for mount if a layer can't be resolved. Default: 30.
+- `fuse_metrics_emit_wait_duration_sec` (int) — Time until snapshotter emits FUSE metrics in logs. Default: 60.
+
+## config/config.go
+### Config
+- `metrics_address` (string) — If empty, no metrics will be polled. Default: "".
+- `metrics_network` (string) — Chooses protocol to send metrics over (e.g. tcp, unix, etc). Default: "tcp".
+- `no_prometheus` — Defined [above](#configfsgofsconfig), cannot be redeclared.
```
turan18, line 34: Oh interesting, looks like this is duplicated at the FSConfig level. This should only be at the root level. We can consider removing it from FSConfig later.

docs/config.md
```diff
@@ -0,0 +1,96 @@
...
+#
+
+# Categorized
+
+## config/fs.go
+
+### [http]
+- `MaxRetries` (int) — Max retries before giving up on a network request. Default: 8.
+- `MinWaitMsec` (int) — Min time between network request attempts. Default: 30.
+- `MaxWaitMsec` (int) — Max time between network request attempts. Default: 30.
```
turan18, line 47: The default for this is 300,000.

docs/config.md
```diff
@@ -0,0 +1,96 @@
...
+- `DialTimeoutMsec` (int) — Max time for a connection before timeout. Default: 3000.
+- `ResponseHeaderTimeoutMsec` (int) — Maximum duration waiting for response headers before timeout. Default: 3000.
+- `RequestTimeoutMsec` (int) — Maximum duration waiting for entire request before timeout. Default: 30000.
+
+### [blob]
+- `valid_interval` (int) — Checks blob regularly at this interval in seconds. Default: 60.
+- `check_always` (bool) — Always check blobs. Default: false.
+- `fetching_timeout_sec` (int) — Sets maximum amount of secodns to wait before failing to fetch. Default: 300.
+- `max_retries` — Overrides MaxRetries in [[http]](#http)
```
turan18, line 56:
```suggestion
- `max_retries` — Blob level MaxRetries. Will override the global MaxRetries set in [[http]](#http)
```

docs/config.md
```diff
@@ -0,0 +1,96 @@
...
+- `min_wait_msec` — Overrides MinWaitMsec in [[http]](#http)
```
turan18, line 57: Same for these.

docs/config.md
```diff
@@ -0,0 +1,96 @@
...
+
+### [content_store]
+- `type` (string) — Sets content store (e.g. "soci", "containerd"). Default: "soci".
+- `namespace` (string) — Default: "default".
+
+## config/resolver.go
+
+### [resolver]
+### [resolver.host]
+### [resolver.host.yourmirrorconfigname1.mirrors]
```
turan18, line 87:
```suggestion
### [resolver]
#### [resolver.host]
#### [resolver.host.hostname]
#### [[resolver.host.hostname.mirrors]]
```

turan18 (changes requested):

docs/config.md
turan18: HTTP is the global level HTTP configuration that applies to fetching spans or SOCI artifacts. The blob config only applies to fetching spans, so it overrides the retry configuration in the global HTTP config. We may want to remove the retry stuff from `BlobConfig`, I think it's mainly there for historical reasons. I don't see why we want blob specific HTTP knobs.

docs/config.md
```diff
@@ -0,0 +1,96 @@
+# Config file options
+
+SOCI has a multitude of variables that can customize SOCI behavior to best suit users' needs. This doc will break down the possible values for each variable, where the variable can be found, and what it can do.
+
+This doc assumes the usage of a TOML file. Per TOML specs, Anything in [brackets] is the required header for the set of variables.
+
+An example configuration file can be found in config/config.toml.
```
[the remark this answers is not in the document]
sondavidb: I don't think that's possible, since actually navigating there would be up to however you're opening the project view. But I could link to the GitHub page?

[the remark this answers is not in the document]
Kern--: Github handles relative paths correctly (e.g.)
https://github.com/awslabs/soci-snapshotter/blob/eaf3231606368c675561281e572c275f295f1ba3/docs/build.md?plain=1#L140

Kern-- (dismissed):
I'm in favor of getting this merged and updating entries as needed.

turan18 (approved):

docs/config.md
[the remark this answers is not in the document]
sondavidb: Done.

turan18 (approved):

Kern-- (approved):

Issue #967, opened 2023-11-29
[Bug] Memory aliasing keeps entire ztoc file in memory


Kern--:
### Description

When looking at a heap profile of the soci snapshotter while a workload is running, the memory usage is dominated by `io.ReadAll` coming from `ztoc.Unmarshal`. The memory from `io.ReadAll` is the memory allocated to the ztoc when we read it from disk [here](https://github.com/awslabs/soci-snapshotter/blob/21ec5445ea5e0908861e60e92cbdcd70d3251c93/ztoc/ztoc_marshaler.go#L57). Normally, this buffer would be GC'd once the method returns, but `flatbufToZtoc` creates a `*Ztoc` which has a [direct reference to `compressionInfo.CheckpointBytes()`](https://github.com/awslabs/soci-snapshotter/blob/main/ztoc/ztoc_marshaler.go#L100) which forces the entire unparsed ztoc to remain in memory until that reference is released. Since the ztoc stays in memory for the lifetime of the container, the memory is used indefinitely.

I haven't looked too closely, so there might be other aliasing, but at the very least we should copy the checkpoint bytes so we can free the rest of the unparsed ztoc.

Example flamegraph (the image doesn't really matter, it seems to always be ztocs that use the memory):
<img width="1399" alt="heapprofile" src="https://github.com/awslabs/soci-snapshotter/assets/2304727/a0bec627-9233-47b6-8c55-3122e38ccd15">

### Steps to reproduce the bug

1) Launch a container with SOCI where the config contains `debug_address = ":6060"`
2) Wait for the ztocs to be downloaded and the fuse filesystems to be set up
3) curl for a heap profile: `curl -s localhost:6060/debug/pprof/heap > heap.pprof`
4) visualize the pprof: `go tool pprof -http :8080 ./heap.pprof`

### Describe the results you expected

`io.ReadAll` should not be present in the heap profile.

### Host information

1. OS: 
2. Snapshotter Version: 
3. Containerd Version:

sondavidb:
Fixed in #975

PR #975, opened 2023-12-01
Free zTOC from memory

sondavidb:
**Issue #, if available:**
#967 

**Description of changes:**
Before this change, the uncompressed zTOC would stay in memory. This was because when converting the full uncompressed bytes into a struct, we erreneously retained a reference to the original byte array in `ztoc.Checkpoints`, because `compressionInfo.CheckpointsBytes()` returns a slice of the uncompressed bytes. This change copies the bytes into a dedicated buffer to free up the full uncompressed byte array.

**Testing performed:**
`make test`. Also redid the steps in #967 and confirmed that `io.ReadAll` was no longer in the FlameGraph, and that it is smaller than before (73MB -> 65MB).

Using Tensorflow image:

Before:
before_flamegraph

After:
after_flamegraph

### DIFF
--- a/ztoc/ztoc_marshaler.go
+++ b/ztoc/ztoc_marshaler.go
@@ -97,7 +97,10 @@ func flatbufToZtoc(flatbuffer []byte) (z *Ztoc, err error) {
 		dgst, _ := digest.Parse(string(compressionInfo.SpanDigests(i)))
 		ztoc.SpanDigests[i] = dgst
 	}
-	ztoc.Checkpoints = compressionInfo.CheckpointsBytes()
+	// Since compressionInfo.CheckpointsBytes() returns a slice,
+	// we need to give it its own array so the GC can free compressionInfo.
+	ztoc.Checkpoints = make([]byte, len(compressionInfo.CheckpointsBytes()))
+	copy(ztoc.Checkpoints, compressionInfo.CheckpointsBytes())
 	ztoc.CompressionAlgorithm = strings.ToLower(compressionInfo.CompressionAlgorithm().String())
 	return ztoc, nil
 }


ztoc/ztoc_marshaler.go
```diff
@@ -97,7 +97,8 @@ func flatbufToZtoc(flatbuffer []byte) (z *Ztoc, err error) {
 		dgst, _ := digest.Parse(string(compressionInfo.SpanDigests(i)))
 		ztoc.SpanDigests[i] = dgst
 	}
-	ztoc.Checkpoints = compressionInfo.CheckpointsBytes()
+	ztoc.Checkpoints = make([]byte, len(compressionInfo.CheckpointsBytes()))
+	copy(ztoc.Checkpoints, compressionInfo.CheckpointsBytes())
```
austinvazquez, line 101: Maybe it be helpful to add a comment here that `CheckpointsBytes` returns a slice so need to manually copy for garbage collection to reclaim the flatbuffer memory.

Also any thoughts on validation? Are there any unit tests or integration tests that should be included with this change?

Kern-- (approved):

ztoc/ztoc_marshaler.go
sondavidb: It might be tough. Instinctively I would like to compare pointer addresses, but it would be tough to do so since directly referencing pointers is [considered unsafe](https://go101.org/article/unsafe.html).

It also is a somewhat weird thing to test, since it's specifically testing one field in one struct in one part of the code that's only done in the beginning of the code and never again. I think it would be cool to do a general test to try and make sure we're not holding pointers that we shouldn't be, but that seems like a much larger scope thing and probably needs to be implemented object by object or something like that.

I think it might be ok to not test this. Feel free to resolve if you agree, otherwise open to more ideas.

PR #993, opened 2023-12-11
Remove test dependency on docker-compose

austinvazquez:
**Issue #, if available:**
N/A

**Description of changes:**
This PR drops the dependency on Docker Compose and replaces it with Docker's builtin compose command. The benefit is one less dependency required for manual ARM release process.

**Testing performed:**

CI passes and ran `make test` and `make integration` on EC2 ARM instance.

### DIFF
--- a/util/dockershell/compose/compose.go
+++ b/util/dockershell/compose/compose.go
@@ -47,10 +47,7 @@ import (
 
 // Supported checks if this pkg can run on the current system.
 func Supported() error {
-	if err := exec.Command("docker", "version").Run(); err != nil {
-		return err
-	}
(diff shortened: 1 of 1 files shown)


sondavidb (approved):
Is there any direct call to Docker in our release automation? I thought I did and checked but couldn't find anything.

If so, we should definitely remove the docker-compose dependency there too. If not, LGTM

Kern-- (approved):

austinvazquez:
> Is there any direct call to Docker in our release automation? I thought I did and checked but couldn't find anything.
> 
> If so, we should definitely remove the docker-compose dependency there too. If not, LGTM

No direct calls on docker-compose from release automation.

Issue #677, opened 2023-06-20
[FEATURE] Solidify SOCI CLI intended use


Kern--:
### Description

The [SOCI CLI](https://github.com/awslabs/soci-snapshotter/tree/main/cmd/soci) is currently a mix of build time tooling, runtime tooling, and debugging tooling. Each of these components have a range of "production readiness" with the debug tooling being the least mature, then runtime tooling, then build time tooling. This makes it hard for the community to know whether they can expect stability if they deploy the CLI as part of their production pipelines.

[Our official statement is that the CLI is unstable](https://github.com/awslabs/soci-snapshotter/blob/main/RELEASES.md#api-stability), but it would be good to add some clarity around what our intentions are. 

### Describe the solution you'd like

A design document describing which parts of the CLI we intend to stabilize, which parts are intended to be replaced with other tooling, and which parts are intentionally unspecified as we see which way the community takes them.

turan18:
Related To: #841

sondavidb:
We've done a good amount of work around the CLI but have yet to document anything since the issue. We can probably consider this closed once we align on a decision for the CLI's intended usage

PR #998, opened 2023-12-13
Remove "image rpull" from CLI

sondavidb:
As an effort to streamline our CLI usage, `image rpull` has been removed. While it technically retains some functionality special to SOCI, it is ultimately up to the CLI to pass the requisite parameters to the remote snapshotter.

This also removes the last of our references to `ctr`.

**Issue #, if available:**
#677
#841 

**Description of changes:**

Removed `cmd/soci/commands/image/`, consequentially also removing the `rpull` command. Replaced all instances of `soci image rpull` with `nerdctl pull --snapshotter soci` in the testing suite.

**Testing performed:**
`make integration`


sondavidb:
Our docs should probably be updated to reflect this.

Would this be considered a breaking change? I wonder if we want to stall this for a later time. I don't really think so, but worth thinking about.

austinvazquez:
> Our docs should probably be updated to reflect this.

Can we update docs with this PR so the change is atomic?

> Would this be considered a breaking change? I wonder if we want to stall this for a later time. I don't really think so, but worth thinking about.

Yes, but that also is the nature of projects <1.0.0. Typically there isn't much expectation for API stability until 1.0 is released.

sondavidb:
Docs updated to reflect these changes.

Since we are now relegating implementation of remote snapshot pulling to other CLIs now, we should include an extra paragraph about how to connect it to our CLI. Once this is merged I can open an issue for this.

Kern-- (approved):

turan18:
> Since we are now relegating implementation of remote snapshot pulling to other CLIs now, we should include an extra paragraph about how to connect it to our CLI. Once this is merged I can open an issue for this.

I think this can be a part of #677. We should make sure to mention these changes in our next release. The tracking issue #841 has the v0.5 milestone tag so that should be enforcing mechanism for us to not forget to do that.

turan18 (approved):

PR #1001, opened 2023-12-14
Log successful startup of soci-snapshotter-grpc

Subzidion:
**Issue #, if available:**
Resolves: #515 

**Description of changes:** Currently, the beginning of the startup process is logged, but no line is logged when startup succeeds.

**Testing performed:**
Before:
```
[ec2-user@ip-10-0-1-240 soci-snapshotter]$ sudo soci-snapshotter-grpc
{"level":"info","msg":"starting soci-snapshotter-grpc","revision":"5d1022b0def534e457fd0f649b99e3874dc38eaa","time":"2023-12-14T22:01:22.686850482Z","version":"5d1022b"}
{"emitMetricPeriod":10000000000,"fetchPeriod":500000000,"level":"info","maxQueueSize":100,"msg":"constructing background fetcher","silencePeriod":30000000000,"time":"2023-12-14T22:01:22.905906907Z"}
^C{"level":"info","msg":"Got interrupt","time":"2023-12-14T22:01:29.001103791Z"}
{"error":"bucket does not exist: not found","level":"warning","msg":"failed to unmount snapshots on close","time":"2023-12-14T22:01:29.001244038Z"}
{"level":"info","msg":"Exiting","time":"2023-12-14T22:01:29.001289768Z"}
```

After:
```
[ec2-user@ip-10-0-1-240 soci-snapshotter]$ sudo soci-snapshotter-grpc
{"level":"info","msg":"starting soci-snapshotter-grpc","revision":"1f95399f909bdd1ba02529255b682165d3007949","time":"2023-12-14T22:01:58.597241960Z","version":"1f95399"}
{"emitMetricPeriod":10000000000,"fetchPeriod":500000000,"level":"info","maxQueueSize":100,"msg":"constructing background fetcher","silencePeriod":30000000000,"time":"2023-12-14T22:01:58.615551090Z"}
{"level":"info","msg":"soci-snapshotter-grpc successfully started, listening on \"/run/soci-snapshotter-grpc/soci-snapshotter-grpc.sock\"","time":"2023-12-14T22:01:58.618808388Z"}
^C{"level":"info","msg":"Got interrupt","time":"2023-12-14T22:02:37.162371412Z"}
{"error":"bucket does not exist: not found","level":"warning","msg":"failed to unmount snapshots on close","time":"2023-12-14T22:02:37.162574419Z"}
{"level":"info","msg":"Exiting","time":"2023-12-14T22:02:37.162618239Z"}
```

[...]

Issue #515, opened 2023-03-22
[FEATURE] Log successful startup of soci-snapshotter-grpc


sparr:
### Description

We recently added a "starting" line to the log output:

```json
{"level":"info","msg":"starting soci-snapshotter-grpc","revision":"0993b79bbecf5aabb1630b6cfc81cd40706bf720.m","time":"2023-03-22T15:58:58.380825434Z","version":"0993b79.m"}
```

However, this is just the start of the startup process. Startup can still fail, such as with invalid config:

```json
{"error":"(1, 17): was expecting token =, but got keys cannot contain new lines instead","level":"fatal","msg":"failed to load config file \"/etc/soci-snapshotter-grpc/config.toml\"","time":"2023-03-22T15:58:58.380929165Z"}
```

or with an unwritable /var/lib:

```json
{"error":"mkdir /var/lib/soci-snapshotter-grpc/snapshotter: permission denied","level":"fatal","msg":"snapshotter is not supported","time":"2023-03-22T15:49:15.787091799Z"}
```

### Describe the solution you'd like

Emit a log line on successful startup.

### Any additional context or information about the feature request

This would be useful for #376 and #371

PR #1002, opened 2023-12-15
Change LogMonitor to look for correct string

sondavidb:
**Issue #, if available:**
Related #515

**Description of changes:**
Per the comment in #515, after our changes in #1001, we should modify `monitorStartup` to look for this instead of just the background fetcher.

**Testing performed:**

### DIFF
--- a/util/testutil/shell.go
+++ b/util/testutil/shell.go
@@ -216,7 +216,8 @@ func monitorStartup(errs chan error) func(string) {
 				errs <- errors.New("fatal snapshotter log entry encountered, snapshotter failed to start")
 				return
 			}
-			if strings.Contains(logline.Msg, "background") {
+			// Looking for "soci-snapshotter-grpc successfully started"
+			if strings.Contains(logline.Msg, "successfully") {
 				errs <- nil
 				return
 			}


austinvazquez (approved):

util/testutil/shell.go
```diff
@@ -216,7 +216,8 @@ func monitorStartup(errs chan error) func(string) {
 				errs <- errors.New("fatal snapshotter log entry encountered, snapshotter failed to start")
 				return
 			}
-			if strings.Contains(logline.Msg, "background") {
+			// Looking for "soci-snapshotter-grpc successfully started"
+			if strings.Contains(logline.Msg, "successfully") {
```
Kern--, line 220: Can we use a constant that's used both here and in the log message itself?

sondavidb: We'd have to import the testutil into the CLI package or vice versa. The former is being pretty messy to figure out (and honestly not even sure if it's possible) while the latter doesn't make logical sense to me.

I think we should just keep it as is. The comment clarifies exactly what we're looking for in the logs.

Kern-- (approved):

Issue #1007, opened 2023-12-21
[Bug] The binary file path has changed in the release artifact starting from 0.4.1


debajyoti-truefoundry:
### Description

![image](https://github.com/awslabs/soci-snapshotter/assets/97010181/8136e40a-f323-454f-9111-ae2ca9ce56aa)

Please check the screenshot attached above. Was this an intended change? 

Feel free to close the issue if this will be the binary path in the release tar file from now on.

### Describe the results you expected

The binary path is the same as 0.4.0.

### Host information

Not required.

debajyoti-truefoundry:
@austinvazquez This issue may be relevant to you based on your changes [here](https://github.com/awslabs/soci-snapshotter/pull/1005).

Kern--:
Reopening to track that we still need to update the release artifacts

austinvazquez:
Thanks for reporting the issue. Just returning back from holiday and will look into this more.

austinvazquez:
Update on this issue. It looks like #1009 will resolve the issue for future releases. For the v0.4.1 release, I have put together a change management form to regenerate and rerelease the artifacts with the issue resolved. ECD is 1/4.

austinvazquez:
@debajyoti-truefoundry , the artifacts for v0.4.1 release have been updated with the pathing fix. Thanks for reporting the issue!

PR #1009, opened 2023-12-22
fix: strip path in release tar

tuananh:
**Issue #, if available:**

Fixes: #1007 #1010 

Update the create tar release from

```
tar -czvf "$RELEASE_DIR"/"$dynamic_binary_name" "$OUT_DIR"
```

to

```sh
tar -czvf "$RELEASE_DIR"/"$dynamic_binary_name" -C "$OUT_DIR" .
```

after the change, when I tried to do `make RELEASE_TAG=v0.4.1 release`, I will get the following

image

**Description of changes:**

**Testing performed:**

### DIFF
--- a/scripts/create-releases.sh
+++ b/scripts/create-releases.sh
@@ -60,13 +60,15 @@ static_binary_name=soci-snapshotter-${release_version}-linux-${ARCH}-static.tar.
 
 make build
 cp "$NOTICE_FILE" "$LICENSE_FILE" "${OUT_DIR}"
-tar -czvf "$RELEASE_DIR"/"$dynamic_binary_name" "$OUT_DIR"
+tar -czvf "$RELEASE_DIR"/"$dynamic_binary_name" -C "$OUT_DIR" .
 rm -rf "{$OUT_DIR:?}"/*
 
 STATIC=1 make build
 cp "$NOTICE_FILE" "$LICENSE_FILE" "$OUT_DIR"
-tar -czvf "$RELEASE_DIR"/"$static_binary_name" "$OUT_DIR"
+tar -czvf "$RELEASE_DIR"/"$static_binary_name" -C "$OUT_DIR" .
 rm -rf "{$OUT_DIR:?}"/*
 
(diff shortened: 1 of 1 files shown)


scripts/create-releases.sh
```diff
@@ -60,12 +60,12 @@ static_binary_name=soci-snapshotter-${release_version}-linux-${ARCH}-static.tar.
 
 make build
 cp "$NOTICE_FILE" "$LICENSE_FILE" "${OUT_DIR}"
-tar -czvf "$RELEASE_DIR"/"$dynamic_binary_name" "$OUT_DIR"
+tar -czvf "$RELEASE_DIR"/"$dynamic_binary_name" -C "$OUT_DIR" .
```
Kern--, line 63: I think it would be better to just tar the files instead of the directory (that's what we did in previous releases).

To fix #1010 as well, we should probably `pushd` into the release dir, create the archive with `tar *`, create the sha256, move both to the release dir, `popd`.

Kern-- (changes requested):
Thanks for this. I think we might want a slightly different approach so we can address #1010 as well.

scripts/create-releases.sh
tuananh: can we just `cd $RELEASE_DIR && shas256sum ...` ? I think that would work too.

Kern--: It does work. I don’t think there’s any reason we would have hidden files in release that would be picked up by archiving the whole folder so it should be equivalent.

Can we at least use `pushd` and `popd` so the script doesn’t leave us in the release folder?

tuananh: Sure. Lemme update the pr that way

Kern-- (approved):

tuananh:
> Can we squash the commits?

squashed.

sondavidb:
Thanks, will merge now. And thanks for the fix, it's much appreciated!

sondavidb (approved):

PR #1021, opened 2024-01-04
Disable xattrs if layer has no xattrs or opaque directories

Kern--:
**Issue #, if available:**

**Description of changes:**
This is a rebase + update to #718 

## Original change
The original insight is that GetXAttr and ListXAttr are the most expensive FUSE operations, accounting for significant SOCI overhead in some use cases, but most layers don't actually contain any xattrs. If we can disable xattrs in these cases, then we can get better runtime performance with SOCI. 

Since image layers are immutable, if the layer tarball doesn't contain any xattrs, then we can disable xattrs for the FUSE mount and the kernel will stop issuing the calls to the snapshotter. 

### Fix for opaque directories
The original insight is mostly correct, but it doesn't handle OveralyFS opaque directories. To delete a file in OverlayFS, you can either create a whiteout file in a high layer (`.wh.$ORIGINAL_FILE_NAME`) or you can implicitly delete all files in a directory by returning an opaque xattr (`trusted.overlay.opaque=y`). The image spec encodes opaque xattrs into the layer as a special file (`$DIR/.wh..wh..opq`) instead of directly putting the xattr into the tarball, and the snapshotter translates that back into an xattr at runtime. With the original change xattrs would be disabled so the opaque file wouldn't be converted back into an xattr which means that the files would be visible.

This change updates the logic for determining whether to disable xattrs to include checking  for opaque directories

### `--xattr-optimization` -> `--optimizations xattr`

Following a suggestion on the original PR, I renamed the flag to `--optimizations` so that we can experiment with optimizations in the future. I'm on the fence on whether there should be a concept of "experimental optimizations" and "stable optimizations". I can also imagine that this xattrs optimization eventually just becomes on-by-default and disappears from the CLI entirely. Thoughts on this are welcome.

### Everything is now "Disable Xattrs"
In the original PR, the discussions were back and forth on whether we should use variations of "Has xattrs", "No xattrs", or "Disable Xattrs". Since now it also includes whether the layer has whiteouts, I opted to stick to "Disable Xattrs" everywhere since it's not directly related to whether xattrs exist in the layer, but whether xattrs should be enabled for OverlayFS. This also has the nice property that the default false ignores the optimization as pointed out in the original PR.

**Testing performed:**
```
```

### DIFF
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -18,6 +18,7 @@ package commands
 
 import (
 	"errors"
+	"fmt"
 	"os"
 
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/internal"
@@ -31,7 +32,7 @@ const (
 	buildToolIdentifier = "AWS SOCI CLI v0.1"
 	spanSizeFlag        = "span-size"
 	minLayerSizeFlag    = "min-layer-size"
-	xattrFlag           = "xattr-optimization"
+	optimizationFlag    = "optimizations"
 )
 
 // CreateCommand creates SOCI index for an image
@@ -54,15 +55,24 @@ var CreateCommand = cli.Command{
 			Usage: "Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10 MiB.",
 			Value: 10 << 20,
 		},
-		cli.BoolFlag{
-			Name:  xattrFlag,
-			Usage: "If set to true layers without xattr will have xattr disabled, yielding a performance benefit. Does not effect layers where is xattrs are present. Default is False.",
+		cli.StringSliceFlag{
+			Name:  optimizationFlag,
+			Usage: fmt.Sprintf("(Experimental) Enable optional optimizations. Valid values are %v", soci.Optimizations),
 		},
 	),
 	Action: func(cliContext *cli.Context) error {
 		srcRef := cliContext.Args().Get(0)
 		if srcRef == "" {
 			return errors.New("source image needs to be specified")
+		}
+
+		var optimizations []soci.Optimization
+		for _, o := range cliContext.StringSlice(optimizationFlag) {
+			optimization, err := soci.ParseOptimization(o)
+			if err != nil {
+				return err
+			}
+			optimizations = append(optimizations, optimization)
 		}
 
 		client, ctx, cancel, err := internal.NewClient(cliContext)
@@ -79,7 +89,6 @@ var CreateCommand = cli.Command{
 		}
 		spanSize := cliContext.Int64(spanSizeFlag)
 		minLayerSize := cliContext.Int64(minLayerSizeFlag)
-		xattr := cliContext.Bool(xattrFlag)
 		// Creating the snapshotter's root path first if it does not exist, since this ensures, that
 		// it has the limited permission set as drwx--x--x.
 		// The subsequent oci.New creates a root path dir with too broad permission set.
@@ -110,7 +119,7 @@ var CreateCommand = cli.Command{
 			soci.WithMinLayerSize(minLayerSize),
 			soci.WithSpanSize(spanSize),
 			soci.WithBuildToolIdentifier(buildToolIdentifier),
-			soci.WithXAttrOptimization(xattr),
+			soci.WithOptimizations(optimizations),
 		}
 
 		for _, plat := range ps {
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -31,6 +31,7 @@ const (
 	buildToolIdentifier = "AWS SOCI CLI v0.1"
 	spanSizeFlag        = "span-size"
 	minLayerSizeFlag    = "min-layer-size"
+	xattrFlag           = "xattr-optimization"
 )
 
 // CreateCommand creates SOCI index for an image
@@ -53,6 +54,10 @@ var CreateCommand = cli.Command{
 			Usage: "Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10 MiB.",
 			Value: 10 << 20,
 		},
+		cli.BoolFlag{
+			Name:  xattrFlag,
+			Usage: "If set to true layers without xattr will have xattr disabled, yielding a performance benefit. Does not effect layers where is xattrs are present. Default is False.",
+		},
 	),
 	Action: func(cliContext *cli.Context) error {
 		srcRef := cliContext.Args().Get(0)
@@ -74,6 +79,7 @@ var CreateCommand = cli.Command{
 		}
 		spanSize := cliContext.Int64(spanSizeFlag)
 		minLayerSize := cliContext.Int64(minLayerSizeFlag)
+		xattr := cliContext.Bool(xattrFlag)
 		// Creating the snapshotter's root path first if it does not exist, since this ensures, that
 		// it has the limited permission set as drwx--x--x.
 		// The subsequent oci.New creates a root path dir with too broad permission set.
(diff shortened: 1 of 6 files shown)


austinvazquez (dismissed):
Thanks for the detailed description and connecting the all the dots. Change LGTM. I am pretty aligned with the `--optimizations` approach but it would be nice if you could add `docs/optimizations.md` with a short writeup on the disable xattr optimization and the state of the optimization. i.e. experimental and disabled by default.

cmd/soci/commands/create.go
```diff
@@ -53,13 +55,26 @@ var CreateCommand = cli.Command{
 			Usage: "Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10 MiB.",
 			Value: 10 << 20,
 		},
+		cli.StringSliceFlag{
+			Name:  optimizationFlag,
+			Usage: fmt.Sprintf("(Experimental) Enable optional optimizations. Valid values are %v", soci.Optimizations),
```
sondavidb, line 60: nit: In the future if soci.Optimizations is ever empty, do we want to have it display a different message?

sondavidb (commented):
Current changes LGTM, really like the new structure to allow us to do more experimental optimizations in the future.

Seconding Austin's comment on the documentation so that users have clarity over what exactly these optimizations do.

> I can also imagine that this xattrs optimization eventually just becomes on-by-default and disappears from the CLI entirely.

I think we should have discussions/guidelines on how we decide if features are experimental or not. To me, this change seems like it should be a default behavior, since I view it as a raw performance increase for the majority of users, and inconsequential for the others. I'd like to know if you had a particularly strong reason for making this experimental/opt-out by default, and from there assess how we want to evaluate future features.

For now though, we should definitely keep it as is, since it would justify the new Optimizations type.

cmd/soci/commands/create.go
```diff
@@ -53,13 +55,26 @@ var CreateCommand = cli.Command{
...
+		},
 	),
 	Action: func(cliContext *cli.Context) error {
 		srcRef := cliContext.Args().Get(0)
 		if srcRef == "" {
 			return errors.New("source image needs to be specified")
 		}
 
+		var optimizations []soci.Optimization
+		for _, o := range cliContext.StringSlice(optimizationFlag) {
```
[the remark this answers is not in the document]
austinvazquez: So the [v1](https://cli.urfave.org/v1/examples/flags/) documentation does not provide any guidance, but the [v2](https://cli.urfave.org/v2/examples/flags/#multiple-values-per-single-flag) documentation suggests it would be `soci --optimizations xattr --optimizations other`

soci/soci_index.go
```diff
@@ -595,3 +646,22 @@ func WriteSociIndex(ctx context.Context, indexWithMetadata *IndexWithMetadata, c
...
+		if ztocDesc.Annotations == nil {
+			ztocDesc.Annotations = make(map[string]string, 1)
+		}
+		ztocDesc.Annotations[IndexAnnotationDisableXAttrs] = "true"
+	}
+}
+
+func shouldDisableXattrs(ztoc *ztoc.Ztoc) bool {
+	for _, md := range ztoc.TOC.FileMetadata {
+		if len(md.Xattrs()) > 0 || strings.HasSuffix(md.Name, whiteoutOpaqueDir) {
```
turan18, line 661: Don't we also need to check for single white out files with the `.wh` prefix (eg: `.wh.somefile`)?

Kern--: No, regular file whiteouts show up in the list of files, not xattrs:

```
A whiteout is created as a character device with 0/0 device number.
When a whiteout is found in the upper level of a merged directory, any
matching name in the lower level is ignored, and the whiteout itself
is also hidden.
```
https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt

turan18: For layer tar's aren't all whiteout file represented as files like `.wh` or `.wh..wh..opq`? Meaning wouldn't we also need to check if `md.Name` has the prefix `.wh`, since we're operating on the TAR file and not the uncompressed file. 

Containerd does this when decompressing/applying a diff, where they check if the TAR file is a whiteout before writing it to disk:

https://github.com/containerd/containerd/blob/a2f167e836437393187da7147f2c0e7983202fdd/archive/tar.go#L279

Kern--: `.wh` and `.wh..wh..opq` are container concepts (based on AUFS if understand correctly) and need to be translated into hidden files/opaque directories by each snapshotter. What you linked is for the naive case where we are unpacking just into a directory. OverlayFS has a different whiteout translator:
https://github.com/containerd/containerd/blob/a2f167e836437393187da7147f2c0e7983202fdd/archive/tar_opts_linux.go#L29-L52

The logic is:
- `$BASEDIR/.wh..wh..opq` becomes an xattr `trusted.overlay.opaque=y` on `$BASEDIR` and the file is not returned as part of listing the directory (i.e. not written to disk)
- `$BASEDIR/.wh.$FILE` becomes a character device with the same name with device id 0/0 and does show up as part of the directory listing (i.e. there is a character device in the filesystem). OverlayFS handles hiding this from the overlay directory listing.

We have the same logic in `node.readdir`

https://github.com/awslabs/soci-snapshotter/blob/21181edae6b06ce5511bf72eb8204c853fee3ae1/fs/layer/node.go#L329-L336

https://github.com/awslabs/soci-snapshotter/blob/21181edae6b06ce5511bf72eb8204c853fee3ae1/fs/layer/node.go#L357-L371

and in `node.GetXattr`
https://github.com/awslabs/soci-snapshotter/blob/21181edae6b06ce5511bf72eb8204c853fee3ae1/fs/layer/node.go#L508-L517

Kern--: `$DIR/.wh.$FILE` whiteouts don't interact with OverlayFS with xattrs, so we don't need to check for them here.

`$DIR/.wh..wh..opq` does interact with OverlayFS with xattrs on `$DIR`, so we shouldn't disable xattrs if any file in the layer ends wit h`.wh..wh..opq`.

turan18: > $DIR/.wh.$FILE whiteouts don't interact with OverlayFS with xattrs, so we don't need to check for them here.
Ah, gotcha. Cool, thanks.

sondavidb:
Did some analysis on this and wanted to report my findings:

TL;DR is that this seems to deliver some very promising performance increases.

The images were tested by first starting up the SOCI snapshotter, waiting 15 seconds, pulling and running the container, then exiting 40 seconds after initially starting up the snapshotter, then waiting another 15 seconds before killing the process.

The prometheus metrics pictures are slightly edited just to show the exact values at each time frame, instead of having to send 4 separate pictures.

The first image that I tested was an alpine image with the fio package, which makes many kernel calls. The layer itself only has two lazy-loaded layers. One had xattrs while the other didnt.

Prometheus metrics (first one before change, second one after change)
image

Flamegraphs (first one before change, second one after change)
image
image

We can see that in the prometheus metrics, total CPU usage was cut by over 33%. In the FlameGraphs, we can see the server loop uses also has over 33% less CPU usage.

The other test was run with a rabbitmq image. Same was applied. This one had three lazy-loaded layers, and none had xattrs.

Prometheus metrics (first one before change, second one after change)
image

Flamegraphs (first one before change, second one after change)
image
image

Here we see total CPU usage down by about 45% in the Flamegraphs, and nearly 50% in the Prometheus metrics.

sondavidb (approved):

sondavidb:
One final note (for another PR, not this one) — I think that this should be default behavior. As I mentioned above, I don't see a use case where this degrades performance, but also would like to have this structure in place for future experimental optimizations we decide to implement.

For DisableXAttrs, we should probably add a flag/config value to enable/disable this behavior during runtime, as well as log whenever DisableXAttrs is enabled for each layer. This is to ensure we have visibility on whether or not this functionality is enabled.

Unsure if I should open a new PR or just ship this and add this as a followup, though.

henry118 (approved):

PR #1026, opened 2024-01-11
Update Go version in workflows

sondavidb:
**Issue #, if available:**

**Description of changes:**
Updated Go version from v1.20.12 -> v1.20.13 in workflows

**Testing performed:**

### DIFF
--- a/.github/workflows/benchmark_visualization.yml
+++ b/.github/workflows/benchmark_visualization.yml
@@ -13,7 +13,7 @@ permissions:
   deployments: write
 
 env:
-  GO_VERSION: '1.20.12'
+  GO_VERSION: '1.20.13'
       
 jobs:
   benchmark:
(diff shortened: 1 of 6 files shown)


turan18 (dismissed):

austinvazquez (changes requested):
You need to also update the Go version in the top level Dockerfile which will trigger the workflows to run.

sondavidb:
Thanks for the catch @austinvazquez. Added a failsafe to the Dockerfile in case this happens again. This could also be useful if we ever want to start testing multiple go versions (e.g. 1.21).

austinvazquez:
Closing and reopening to pick up workflow changes.

sondavidb:
It looks like the new env section overwrites the old one, and I don't see a way to carry it over.

I think I'll just revert these changes for now. I think we'll want to do another matrix like we do with containerd anyway if we want to test multiple Go versions, and that would warrant a separate PR IMO.

austinvazquez (approved):

turan18 (approved):

Issue #911, opened 2023-11-03
[FEATURE] Run GitHub workflows when the workflows change


austinvazquez:
### Description

As part of #910, I noticed the build workflow did not run due to conditional skips. As a maintainer, it would be helpful for the workflows to run when the workflows themselves change for validation.

### Describe the solution you'd like

The following conditions to be added to the respective workflows:

Benchmark workflow:
- SHOULD run when benchmarks when the benchmark workflow is changed.
- SHOULD run when benchmarks when benchmarks are added or changed.
- SHOULD NOT post results to benchmark website on pull request runs.

Build workflow:
- SHOULD run when build workflow is changed.

Comparison test workflow:
- SHOULD run when comparison test workflow is changed.
- SHOULD run when comparison test benchmark is changed.

Release workflow
- SHOULD run when release workflow is changed. (using a test release tag)
- SHOULD NOT create a release on pull request runs.

```[tasklist]
### Tasks
- [x] Run build workflows on change
- [x] Run release workflow on change
- [ ] Run benchmark workflow on change
```

sondavidb:
#1186 runs release workflows when they're changed. I think we still need something to run benchmarks when those workflows change, since it looks like they currently only run on commit [push](https://github.com/awslabs/soci-snapshotter/blob/739c04f167b63b334e6cdae4486cabe8b17f69a7/.github/workflows/benchmark_visualization.yml#L4)

PR #1027, opened 2024-01-11
Update containerd versions in workflows

sondavidb:
**Issue #, if available:**

**Description of changes:**
Updated containerd versions in matrix.
v1.6.19 -> v1.6.26
v1.7.0 -> v1.7.11

**Testing performed:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        containerd: ["1.6.19", "1.7.0"]
+        containerd: ["1.6.26", "1.7.11"]
     env:
       DOCKER_BUILD_ARGS: "CONTAINERD_VERSION=${{ matrix.containerd }}"
     steps:
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        containerd: ["1.6.19", "1.7.0"]
+        containerd: ["1.6.26", "1.7.11"]
     env:
       DOCKER_BUILD_ARGS: "CONTAINERD_VERSION=${{ matrix.containerd }}"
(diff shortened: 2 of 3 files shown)


turan18 (dismissed):

austinvazquez (changes requested):
The changes themselves LGTM, but can you run the build workflow in your fork and paste the link under testing performed in the PR description just to validate the integration tests pass with the new containerd versions.

For context, our workflows do not currently run when modified. We have #911 to track that work.

austinvazquez:
Opened #1028 to help with this.

sondavidb:
Rebased after changes in #1028 so now it should auto-run the integration tests.

While we're here, I wanted a sanity check — our Dockerfile has an argument `CONTAINERD_VERSION` that gets overwritten in our GH Actions workflow. Is there any particular reason we keep this variable at 1.6.17? I wonder if we should change it to reflect the latest version(s?) in our workflow. It wouldn't affect us, but might help clear confusion for future users looking to run the testing suite locally.

turan18:
> Rebased after changes in #1028 so now it should auto-run the integration tests.
> 
> While we're here, I wanted a sanity check — our Dockerfile has an argument `CONTAINERD_VERSION` that gets overwritten in our GH Actions workflow. Is there any particular reason we keep this variable at 1.6.17? I wonder if we should change it to reflect the latest version(s?) in our workflow. It wouldn't affect us, but might help clear confusion for future users looking to run the testing suite locally.

We pass the containerd version from GH actions as a build argument, 1.6.17 is just the default, meaning if don't explicitly set one it's used. I do think it's a good idea to update the Dockerfile whenever we make relevant changes to our GA actions, as it can cause a discrepancy between our GA tests and local tests.

sondavidb:
Thanks. Change has been made to reflect @turan18's suggestion.

austinvazquez (approved):

turan18 (approved):

PR #1028, opened 2024-01-11
Run build workflow on change

austinvazquez:
**Issue #, if available:**
Partial #911 

**Description of changes:**
Runs build workflow on changes.

**Testing performed:**
Build workflow ran on this PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,7 @@ on:
   pull_request:
     branches: ['main', 'release/**']
     paths:
+      - '.github/workflows/build.yml'
       - '**.go'
       - 'go.*'
       - 'cmd/go.*'


.github/workflows/build.yml
```diff
@@ -10,6 +10,7 @@ on:
   pull_request:
     branches: ['main', 'release/**']
     paths:
+      - '.github/workflows/build.yml'
```
Kern--, line 13: we should include `prebuild.yml` as well.

sondavidb: Wouldn't doing this also make it so changes in `prebuild.yml` would make `build.yml` also run?

Kern--: Oh, right. I guess I was thinking prebuild should rerun if prebuild changes, but that's not what this is. Whoops.

Kern-- (approved):

sondavidb (approved):
LGTM, we'll need to be a bit more thorough to fully fix #911 but this is a simple quick fix for now.

Thanks for this. I'll merge it immediately so that I can get #1027 to auto-build.

PR #1031, opened 2024-01-16
Call fs.Unmount only if needed

dims:
I see a bunch of logs like this. Could we avoid calling Unmount if it is not really needed?

```
Jan 16 02:37:53 ip-172-31-45-161.ec2.internal soci-snapshotter-grpc[27054]: {"error":"failed to unmount directory \"/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/2\": specified path \"/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/2/fs\" isn't a mountpoint","level":"warning","msg":"failed to remove directory","path":"/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/2","time":"2024-01-16T02:37:53.665272849Z"}
```


Kern-- (approved):
Thanks!

turan18 (approved):

Issue #1034, opened 2024-01-16
[FEATURE] Fallback to `containerd` if we cannot lazy load image


turan18:
### Description

If we are unable to lazy load a single layer in an image we fallback to our internal unpacker that fetches and unpacks the layer to disk. Our local unpacker only operates on the current layer, unlike in `containerd` [where all neighboring layers are fetched concurrently](https://github.com/containerd/containerd/blob/a83449cf1181b622a3f1c799e58257f8860e4db3/pkg/unpack/unpacker.go#L352-L367). This is mostly fine, since we wouldn't want to accidentally pre-fetch neighboring layers that _should_ be lazy loaded. However, if we know that we cannot lazy load any layers in an image, we could in theory concurrently pull all layers.

### Describe the solution you'd like

If we cannot fetch SOCI artifacts for an image, we should throw some sort of error that indicates that the image (in it's entirety) cannot be lazy loaded. We can check for that error in `Prepare` and if it is present we can fallback to containerd instead of attempting to locally unpack each layer.  

### Describe any alternative solutions/features you've considered

Have our local/internal unpacker support concurrent fetches. With this approach, we would need to maintain some state per image, that indicates whether or not an image can be lazy loaded. 

### Any additional context or information about the feature request

Related to: https://github.com/awslabs/soci-snapshotter/discussions/988

_No response_
```[tasklist]
### Tasks
```

PR #1035, opened 2024-01-16
Fallback to containerd if we are unable to fetch SOCI artifacts

turan18:
If we know that we cannot lazy load an image (eg: SOCI index does not exist for an image), we should fallback to the underlying runtime to do the fetching/unpacking of all layers.

**Issue #, if available:**

Fixes: #1034 

**Description of changes:**

**Testing performed:**

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -409,7 +409,7 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	client := src[0].Hosts[0].Client
 	c, err := fs.getSociContext(ctx, imageRef, sociIndexDigest, imgDigest, client)
 	if err != nil {
-		return fmt.Errorf("unable to fetch SOCI artifacts: %w", err)
+		return fmt.Errorf("%w: unable to fetch SOCI artifacts for image %q: %w", snapshot.ErrUnableToLazyLoadImage, imageRef, err)
 	}
 
 	// Resolve the target layer
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -77,7 +77,10 @@ const (
 )
 
 var (
-	// Error returned by `fs.Mount` when there is no ztoc for a particular layer.
+	// ErrUnableToLazyLoadImage is returned by `fs.Mount` when the image cannot/should not be
+	// lazy loaded. (eg: If a SOCI index does not exist for an image)
+	ErrUnableToLazyLoadImage = errors.New("unable to lazy load image")
+	// ErrNoZtoc is returned by `fs.Mount` when there is no zTOC for a particular layer.
 	ErrNoZtoc = errors.New("no ztoc for layer")
 )
 
@@ -302,6 +305,8 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 	//       log is used by tests in this project.
 	lCtx := log.WithLogger(ctx, log.G(ctx).WithField("key", key).WithField("parent", parent))
 
+	var skipLazyLoadingImage bool
+
 	// remote snapshot prepare
 	if !o.skipRemoteSnapshotPrepare(lCtx, base.Labels) {
 		err := o.prepareRemoteSnapshot(lCtx, key, base.Labels)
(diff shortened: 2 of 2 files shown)


fs/fs.go
```diff
@@ -409,7 +409,7 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	client := src[0].Hosts[0].Client
 	c, err := fs.getSociContext(ctx, imageRef, sociIndexDigest, imgDigest, client)
```
turan18, line 410: I suppose this could be it's own issue/discussion, but right now we fallback either to overlayfs/the default snapshotter even on partial artifact fetch success, meaning we were able to fetch the index and only _some_ zTOC's. Should that be the case? Should we still try to remote mount layers that correspond to zTOC's that we _were_ able to fetch?

sondavidb: To be clear, this is when the index specifies a number of zTOCs, but we get less than that expected number, correct? What scenarios might cause this to happen?

I think this behavior is probably fine since I mostly see it as a network issue if this fails (e.g. auth revoked, internet conenction lost, etc), and so it's somewhat reasonable to assume that the user no longer has access to the required repo at all.

sondavidb (commented):
LGTM, from what I understand since we currently just use all local mounts it makes sense to defer to container runtime.

Definitely not in the scope of this PR, but wonder if we'd ever want behavior where we would want a toggle to attempt to auto-generate a SOCI index if one isn't found. We need a SOCI index to do any work so it might be something nice to have. Just a thought for now though.

sondavidb (approved):

dims:
This is probably useful even when an image has been already downloaded *before* soci-snapshotter was enabled, right?

austinvazquez (commented):
Overall the changes LGTM. A few questions I had:
1. Are we able to mock the ErrUnableToLazyLoad returned from the filesystem mount in snapshotter unit tests to validate a fallback occurs?
2. Wondering if we should emit a metric on this case where SOCI artifacts exist but we failed to pull them. My thought is it might not be worth tracking because we have the log. Curious if you had some thoughts on it.
3. Just want to validate an integration test here is difficult because we'd need a method for pulling an image with SOCI artifacts but an error occurs while pulling. e.g. network error.

sondavidb:
> Just want to validate an integration test here is difficult because we'd need a method for pulling an image with SOCI artifacts but an error occurs while pulling. e.g. network error.

We do a similar thing in [TestNetworkRetry](https://github.com/awslabs/soci-snapshotter/blob/db7df3ab2a840bc927417a84448263885b3e21ff/integration/run_test.go#L223), where we pull an image from a repo, run the image with SOCI, attempt to do run a command, then block network access to the registry and try again. TestNetworkRetry specifically is not implemented correctly, though, and I'm not sure how we would simulate this, but the idea is there.

Does this need to be an integration test though? Is there any way we can make it a unit test, e.g. by calling Mount with a bunch of local mounts?

Kern-- (approved):
I think `Prepare` is in need of some rethinking/refactoring to make it more clear what the logic actually is and how that logic appears in logs.


For example, we now have both `skipLazyLoadingImage` and `o.skipRemoteSnapshotPrepare`. If we skip lazy loading an image, we will log "failed to prepare remote snapshot" and "failed to prepare snapshot; deferring to container runtime" for every layer which doesn't really convey what's actually going on.

sondavidb:
Should we also fall back if a bad SOCI index digest is passed in? Right now, pulling while specifying a bad index digest ends up having a similar issue, so unless we want to switch between SOCI indexes for different layers of the same image (which I really don't think we do), it might be a nice thing to add.

turan18:
> This is probably useful even when an image has been already downloaded _before_ soci-snapshotter was enabled, right?

@dims  Sorry for the delay. All this change will do is ensure that we defer back to containerd earlier when we know we can't lazy load the image. If the image has already been downloaded and the blob content exists in the containerd content store, this will ensure that SOCI won't blindly attempt to sequentially fetch each layer again.

fs/fs.go
turan18: > To be clear, this is when the index specifies a number of zTOCs, but we get less than that expected number, correct?

Yeah.

>  What scenarios might cause this to happen?

Not exactly sure, but I'm imagining network failures.

turan18:
TODO:

1. Do not emit mount failure metric if the `err` is no referrers (eg: the index does not exist)
2. Add integration test

Kern--:
Fixed as part of #1302

PR #1036, opened 2024-01-16
Leave Debug breadcrumbs for snapshotter functions called

dims:
When debugging issues with snapshotter, it was hard to figure out if anything was even called or not especially with `--log-level debug` should show some traffic in the logs

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -209,6 +209,7 @@ func NewSnapshotter(ctx context.Context, root string, targetFs FileSystem, opts 
 // Should be used for parent resolution, existence checks and to discern
 // the kind of snapshot.
 func (o *snapshotter) Stat(ctx context.Context, key string) (snapshots.Info, error) {
+	log.G(ctx).WithField("key", key).Debug("stat")
 	ctx, t, err := o.ms.TransactionContext(ctx, false)
 	if err != nil {
 		return snapshots.Info{}, err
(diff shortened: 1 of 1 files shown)


Kern-- (approved):

austinvazquez (approved):

snapshot/snapshot.go
```diff
@@ -301,6 +305,7 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 	//       or not, using the key `remoteSnapshotLogKey` defined in the above. This
 	//       log is used by tests in this project.
 	lCtx := log.WithLogger(ctx, log.G(ctx).WithField("key", key).WithField("parent", parent))
+	log.G(lCtx).WithField(remoteSnapshotLogKey, prepareSucceeded).Debug("preparing snapshot")
```
turan18, line 308: small change: Could we omit `WithField(remoteSnapshotLogKey, prepareSucceeded)` here. It should only be logged after we were able to perform a "remote" mount: https://github.com/awslabs/soci-snapshotter/pull/1036/files#diff-5c6bcd0ff5f76d3d6e311122b52e890cff76233ee29e88a138f074bcc74518e1R318

turan18, line 308: small change: Could we omit `WithField(remoteSnapshotLogKey, prepareSucceeded)` here. It should only be logged after we were able to perform a "remote" mount: https://github.com/awslabs/soci-snapshotter/pull/1036/files#diff-5c6bcd0ff5f76d3d6e311122b52e890cff76233ee29e88a138f074bcc74518e1R318

turan18 (approved):

snapshot/snapshot.go
dims: Done!

turan18 (approved):
Thanks!

Kern-- (approved):

Issue #901, opened 2023-10-31
[Bug] File descriptor leak in unpacker when unpacking remote content


austinvazquez:
### Description

Potential file descriptor leak in filesystem unpacker after fetching remote content. ([Ref](https://github.com/awslabs/soci-snapshotter/blob/5af486cba306bf0b04da5e8b66072a13d259063a/fs/unpacker.go#L82-L91))

### Steps to reproduce the bug

Unpack content which is not yet local.

### Describe the results you expected

The remote reader ([Ref](https://github.com/awslabs/soci-snapshotter/blob/5af486cba306bf0b04da5e8b66072a13d259063a/fs/unpacker.go#L87)) should be closed after the unpack.

```go
		- rc.Close() // this close is the same as the defer rc.Close() above
		rc, _, err = lu.fetcher.Fetch(ctx, desc)
		if err != nil {
			return fmt.Errorf("cannot fetch layer: %w", err)
		}
		+ defer rc.Close()
```

### Host information

1. OS: Linux
2. Snapshotter Version: v0.4.0
3. Containerd Version: v1.6.24

austinvazquez:
As a proof-of-concept, https://go.dev/play/p/e7Dg7pO8y94 simulates the current implementation and shows a reassignment of c would not affect the original defer close.

Output:
```
2009/11/10 23:00:00 Closing 0
2009/11/10 23:00:00 Closing 0

Program exited.
```

Kern--:
It's a relatively simple fix, but personally, I think we should just rip out the local mount stuff entirely and always defer to the client to do the unpack when we pull ahead of time. It adds complexity and costs features (e.g. zstd) for little benefit.

turan18:
> should just rip out the local mount stuff entirely

I don't know if we can. If we don't always return an `ErrAlreadyExists` containerd will start [pre-fetching](https://github.com/containerd/containerd/blob/main/pkg/unpack/unpacker.go#L359-L365) all subsequent layers. If there was some `sn.Stat` check before each async fetch, that could work.

PR #1037, opened 2024-01-18
Fix file descriptor leak

sondavidb:
**Issue #, if available:**
#901 

**Description of changes:**
Moved `defer rc.Close()` outside of the local check, so that it will always close the correct file descriptor

**Testing performed:**

### DIFF
--- a/fs/unpacker.go
+++ b/fs/unpacker.go
@@ -77,18 +77,19 @@ func (lu *layerUnpacker) Unpack(ctx context.Context, desc ocispec.Descriptor, mo
 	if err != nil {
 		return fmt.Errorf("cannot fetch layer: %w", err)
 	}
-	defer rc.Close()
 
 	if !local {
-		if err = lu.fetcher.Store(ctx, desc, rc); err != nil {
+		err := lu.fetcher.Store(ctx, desc, rc)
+		rc.Close()
+		if err != nil {
 			return fmt.Errorf("cannot store layer: %w", err)
 		}
-		rc.Close()
 		rc, _, err = lu.fetcher.Fetch(ctx, desc)
 		if err != nil {
 			return fmt.Errorf("cannot fetch layer: %w", err)
 		}
 	}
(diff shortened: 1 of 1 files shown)


fs/unpacker.go
```diff
@@ -77,7 +77,6 @@ func (lu *layerUnpacker) Unpack(ctx context.Context, desc ocispec.Descriptor, mo
 	if err != nil {
 		return fmt.Errorf("cannot fetch layer: %w", err)
 	}
-	defer rc.Close()
 
 	if !local {
 		if err = lu.fetcher.Store(ctx, desc, rc); err != nil {
```
Kern--, line 82: nit: we need to close `rc` on error here too.

sondavidb: I don't think so, since it looks like all the possible return values in [Fetch](https://github.com/awslabs/soci-snapshotter/blob/335515f746f50c964ed48159257e1aeba04805b6/fs/artifact_fetcher.go#L88) only return nil upon returning an error, so this would actually segfault.

We could be extra safe, though, and just do something like

```golang
if rc != nil {
    rc.Close()
}
```

I don't think it's super necessary but if you think it's better we can.

austinvazquez: I think the weakness with that approach is it assumes the current implementation for a interface. My suggestion would be to keep the close on line 80 and add another close on line 90 after the `rc` variable is reinstantiated.

austinvazquez (approved):
LGTM.

Kern-- (commented):
This double-closes the original `io.ReadCloser` (once in the defer and once in the explicit call to `Close`). This is undefined behavior (https://pkg.go.dev/io#Closer).

I think we have two choices: either we let the defer close the original when the function exits or we manually close it, being careful to handle error cases. We shouldn't do both.

austinvazquez (approved):

Kern-- (approved):

PR #1054, opened 2024-01-29
Modify artifact verification script

sondavidb:
**Issue #, if available:**
N/A

**Description of changes:**
Allowed artifact verification to be called without having to cd into the release directory

**Testing performed:**

### DIFF
--- a/scripts/verify-release-artifacts.sh
+++ b/scripts/verify-release-artifacts.sh
@@ -18,6 +18,10 @@
 
 set -o pipefail
 
+cur_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+soci_snapshotter_project_root="$(cd -- "$cur_dir"/.. && pwd)"
+release_dir="${soci_snapshotter_project_root}/release"
+
 arch=""
 case $(uname -m) in
     x86_64) arch="amd64" ;;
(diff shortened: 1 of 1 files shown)


scripts/verify-release-artifacts.sh
```diff
@@ -18,6 +18,10 @@
 
 set -o pipefail
 
+cur_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+soci_snapshotter_project_root="$(cd -- "$CUR_DIR"/.. && pwd)"
```
Kern--, line 22: nit: can we just use `"$CUR_DIR"/..` like we do in other files? e.g. https://github.com/awslabs/soci-snapshotter/blob/db7df3ab2a840bc927417a84448263885b3e21ff/scripts/add-ltag.sh#L20

Also, I think you have a bug because `CUR_DIR` is capitalized here, but not in the declaration above. 

I think most of our scripts are using the wrong convention and we should be using lower case letters. https://stackoverflow.com/questions/673055/correct-bash-and-shell-script-variable-capitalization

scripts/verify-release-artifacts.sh
```diff
@@ -35,10 +39,16 @@ if [ $# -eq 0 ]; then
...
+if [ ! -d "$release_dir" ]; then
+    echo "$0: Release directory not found in $release_dir"
+    exit 1
+fi
+
 release_tag=$1
 # Strip 'v' from release tag.
 release_version=${release_tag/v/} 
 
+cd $release_dir
```
Kern--, line 51: nit: we should use `pushd` here and `popd` at the end of the file so we don't mess with a user's working directory after executing this script.

scripts/verify-release-artifacts.sh
sondavidb: I'll try and fix that for all of our scripts at some point, but for now fixed for this

sondavidb: Oops, I remember why I did this. It makes the path funny because it'll end up making the release directory something like `~/soci-snapshotter/scripts/../release`. Fixed capitalization bug though.

austinvazquez (approved):

Kern-- (approved):

PR #1055, opened 2024-01-29
Add metadata benchmarks

turan18:
**Issue #, if available:**

TODO: 

1. Add concurrent reader benchmark 
2. Add more complex options to `GenerateFileTarEntries` like depth level.

**Description of changes:**

**Testing performed:**


util/testutil/util.go
```diff
@@ -181,3 +186,12 @@ func RandomDigest() string {
 	d := digest.FromBytes(RandomByteData(10))
 	return d.String()
 }
+
+// RandString returns a random string of length n
+func RandString(n int) string {
```
turan18, line 191: File names must be random since we use them to traverse the fs tree when creating the DB and so RandString does not use our seeded random. We still use a fixed size, so there shouldn't be any variance between runs.

metadata/reader_test.go
```diff
@@ -33,47 +33,260 @@
...
+			entries: 50_000,
+		},
+		{
+			name:    "Create metadata.Reader with an enormous amount of TOC entries",
+			entries: 100_000,
+		},
 	}
-	defer os.Remove(f.Name())
-	db, err := bolt.Open(f.Name(), 0600, nil)
+	cwdPath, err := os.Getwd()
```
turan18, line 224: Note: Since we want to measure write performance to disk, we have to write to non tmpfs location

metadata/util_test.go
```diff
@@ -46,168 +46,62 @@ import (
...
-				numOfNodes(6), // root dir + 1 file + 1 dir + 1 cdev + 1 bdev + 1 fifo
-				hasFile("bar/foo", 0),
-				hasChardev("bar/cdev", 10, 11),
-				hasBlockdev("bar/bdev", 100, 101),
-				hasFifo("bar/fifo"),
-			},
-		},
+// newTestableReader creates a test bbolt db as-well as a metadata.Reader given a TOC.
+func newTestableReader(sr *io.SectionReader, toc ztoc.TOC, opts ...Option) (testableReader, error) {
+	db, clean, err := newTempDB("")
```
sondavidb, line 61: Will this just create the temp db at root? Maybe we should allow it to be configurable, e.g. make a tempdir and pass it into the TestableReader or something.

metadata/reader_test.go
```diff
@@ -33,47 +33,260 @@
...
+			}
+		}
+	}
 }
 
-func newTestableReader(sr *io.SectionReader, toc ztoc.TOC, opts ...Option) (testableReader, error) {
-	f, err := os.CreateTemp("", "readertestdb")
-	if err != nil {
-		return nil, err
+func BenchmarkMetadataReader(b *testing.B) {
```
sondavidb, line 202: Does this ever get called?

sondavidb (commented):
Generally LGTM, a lot of minor changes that I just want a bit of attention on before approving. Overall the functionality looks great and I think it's a pretty cool addition to our testing suite.

metadata/reader_test.go
```diff
@@ -33,47 +33,260 @@
...
+	"gzip-nocompression":      gzip.NoCompression,
+	"gzip-bestspeed":          gzip.BestSpeed,
+	"gzip-bestcompression":    gzip.BestCompression,
+	"gzip-defaultcompression": gzip.DefaultCompression,
+	"gzip-huffmanonly":        gzip.HuffmanOnly,
+}
+
 func TestMetadataReader(t *testing.T) {
-	testReader(t, newTestableReader)
+	sampleTime := time.Now().Truncate(time.Second)
```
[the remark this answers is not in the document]
turan18: The time is used to as the mod time for certain test files. It doesn't have any intrinsic meaning outside of validation.

metadata/reader_test.go
```diff
@@ -33,47 +33,260 @@
...
+				hasFile("bar/foo", 0),
+				hasChardev("bar/cdev", 10, 11),
+				hasBlockdev("bar/bdev", 100, 101),
+				hasFifo("bar/fifo"),
+			},
+		},
+	}
+	for _, tt := range tests {
+		for _, prefix := range allowedPrefix {
+			prefix := prefix
```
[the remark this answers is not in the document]
turan18: I think it's generally good practice to include `_ := _` in all of our test cases, in case we ever attempt to parallelize them. I'll add it to `srcCompression`.

metadata/reader_test.go
```diff
@@ -33,47 +33,260 @@
...
+	"gzip-huffmanonly":        gzip.HuffmanOnly,
+}
+
 func TestMetadataReader(t *testing.T) {
-	testReader(t, newTestableReader)
+	sampleTime := time.Now().Truncate(time.Second)
+	tests := []struct {
+		name string
+		in   []testutil.TarEntry
+		want []check
```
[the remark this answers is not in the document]
turan18: This test is specifically testing whether our TOC->FS translation is accurate and so we are explicit when creating our TOC entries. There may be more elegant ways to express the structure of a filesystem, but I think that is out of scope for this.

metadata/util_test.go
turan18: If no dir is provided (""), It will create the temp DB under`/tmp`. If you provide a dir it will create the temp DB file under that dir.

metadata/reader_test.go
turan18: No, this is a benchmark function (similar to a TestXXX), so it will never get invoked outside of `go test -run=TestXXX` or `go test -bench=BenchmarkXXX`.

metadata/reader_test.go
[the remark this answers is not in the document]
sondavidb: I get the logic, I just generally don't see the call made outside of functions that end up being run in parallel.

It's ultimately still a nit, though, so feel free to resolve.

metadata/reader_test.go
```diff
@@ -33,47 +33,262 @@
...
+			b.Fatalf("failed to initialize temp db: %v", err)
+		}
+		toc, err := generateTOC(tc.entries)
+		if err != nil {
+			b.Fatalf("failed to generate TOC: %v", err)
+		}
+		b.ResetTimer()
+		b.Run(tc.name, func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				if _, err := NewReader(tempDB, nil, toc); err != nil {
```
Kern--, line 243: Does this rely on the fact that `NewReader` will write new entries to the `tempDB` even if this TOC has already been used?

util/testutil/util.go
Kern--: I don't understand this statement.

util/testutil/tar.go
```diff
@@ -493,3 +493,132 @@ func permAndExtraMode2TarMode(fm os.FileMode) (tm int64) {
...
+
+	return tarEntries, nil
+}
+
+// randRangeInc returns a random range between [min,max] inclusive.
+func randRangeInc(min int, max int) int {
+	if min == max {
+		return min
+	}
+	return r.Intn((max+1)-min) + min
```
Kern--, line 603: I probably should have documented this better, but the intention with having this unexported, global `ThreadSafeRandom` was just to make it so that we didn't have to go back and update every existing test immediately.

I don't think any new code should use it since it makes reproducibility dependent on test execution order. The idea was that we want "randomness" in the sense that we want various input object structures, but we actually want comparability by making the test deterministic.

turan18: > I don't think any new code should use it since it makes reproducibility dependent on test execution order. The idea was that we want "randomness" in the sense that we want various input object structures, but we actually want comparability by making the test deterministic.

Sorry, does the seeded random not return the same value regardless of execution order and whether or not we use the global or a local instance?

metadata/reader_test.go
turan18: It does, yeah. I suppose I could initialize a new DB per-run with `b.Start` and `b.Stop`.

util/testutil/util.go
turan18: If we used the seeded random all our filenames would be the same and the only thing differentiating them would be depth level (eg: `file` vs `file/file` vs `file/file/file`). This is OK when writing to our nodes bucket since we don't care about relationships between nodes. This doesn't hold true when writing to our metadata bucket, though.

Every node is identified by some ID and has a corresponding metadata entry. The metadata entry has a map of it's children with the child name as the key. When we are adding children to the metadata entry of a node we will end up overwriting any existing children if they share the same name, which never happens in practice since you cannot multiple children nodes/files with the same name under a single parent node/directory. What this means in practice, is that our metadata bucket will not be fully populated.

Instead, we use rand so we can get an actual pseudo random string. We still use a fixed length of 10 for the filename, to avoid any variance between runs.

util/testutil/tar.go
Kern--: The TSR will generate the same random sequence, but if multiple tests use the same TSR, then the sequence they see depends on the order they run.


If you create a new TSR in a test and only use that, then the input data will be exactly the same every single time you run the test, not matter what other tests are running. If you use the global TSR, then the input depends on what other tests are running and in what order.

sondavidb (approved):

util/testutil/util.go
turan18: If we used the seeded random all our filenames would be the same and the only thing differentiating them would be depth level (eg: `file` vs `file/file` vs `file/file/file`).

When looping through the TOC we maintain a map of node ID to metadata entry. The metadata entry has a map of the nodes children where the child name is the key. When we are adding children to the metadata entry of a node we will end up overwriting any existing children if they share the same name, which never happens in practice since you cannot multiple children nodes/files with the same name under a single parent node/directory. This means that our metadata/nodes bucket will not be fully populated, since a parent can really only have 1 child. 

To avoid this, we use rand so we can get an actual pseudo random string. We still use a fixed length of 10 for the filename, to ensure their isn't any variance in bbolt write performance between benchmark runs. (bbolt doesn't care about the content of a KV pair since they are just interpreted as byte slices; the length, however, does matter since it controls how nodes/pages are split before writing to disk).

PR #1061, opened 2024-01-31
Add concurrency limits

### LINKED ISSUE #970: [FEATURE] Support concurrency limits when setting up layers
### Description

When pulling an image with Docker, Nerdctl, or Kubernetes, there is a per-image concurrency limiter which limits the number of inflight layer pulls. 

When SOCI handles a layer, it unconditionally spawns a goroutine to handle all adjacent layers:
https://github.com/awslabs/soci-snapshotter/blob/21ec5445ea5e0908861e60e92cbdcd70d3251c93/fs/fs.go#L427-L447

For images with a large number of layers or a large number of concurrent image pulls, the SOCI snapshotter has no limit to the number of layers it will try to resolve. This could put pressure on the network, but more likely would put pressure on the metadata bboltdb and could cause pull times to grow indefinitely. 

There should be an option to limit the number of in-flight resolutions. It is an open question whether this should be a global limit or a per-image limit.

_No response_

An additional question is whether there is a race where if we're slow to set up an adjacent layer, can we end up resolving it again when handling the next layer?

sondavidb:
**Issue #, if available:**
Fixes #970 

**Description of changes:**
Implemented a limit on number of layers that can be pulled at once, reusing an old variable in the TOML for FSConfig, `max_concurrency`. This ended up slightly refactoring our pre-resolve calls to be slightly more efficient on larger images.

With the changes, instead of spawning a bunch of goroutines in order and having many layers compete with each other, every layer to be preresolved is instead sent to the queue. From there, the main queue function will first ensure that the amount of Goroutines running do not exceed the max concurrency limit, and if not, it will read from the channel and resolve the layers in the order they came in.

We initially wanted this to only make unique calls, but this PR doesn't have it yet. I feel like the queue system already does a lot to ensure that the most important work is done first. Keeping track of what layers it has already resolved has two caveats:
1. This requires new data structure besides a channel to store all of what is being queued, which will probably be less efficient than a channel. But, it's possible.
2. If a layer is pulled and then for some reason content is removed from the filesystem, it will never add the layer to the queue again until the service is restarted. This can be circumvented by having another data structure (e.g. a hashset) and just removing from the hashset after the layer has been resolved.

I don't think either of the two caveats are a huge deal, and thus I'll probably end up implementing # 2, but figured I'd write them down in case we feel it's unnecessary/unneeded. Also just want to make sure the PR is going in the right direction :p

**Testing performed:**

### DIFF
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -74,6 +74,9 @@ const (
 	// defaultFuseMetricsEmitWaitDurationSec is the amount of time the snapshotter will wait before emitting the metrics for FUSE operation.
 	defaultFuseMetricsEmitWaitDurationSec = 60
 
+	// defaultMaxConcurrency is the maximum number of layers allowed to be pulled at once
+	defaultMaxConcurrency = 100
+
 	defaultValidIntervalSec = 60
 
 	defaultFetchTimeoutSec = 300
--- a/config/fs.go
+++ b/config/fs.go
@@ -187,6 +187,13 @@ func parseFSConfig(cfg *Config) {
 	if cfg.FuseMetricsEmitWaitDurationSec == 0 {
 		cfg.FuseMetricsEmitWaitDurationSec = defaultFuseMetricsEmitWaitDurationSec
 	}
+	if cfg.MaxConcurrency == 0 {
+		cfg.MaxConcurrency = defaultMaxConcurrency
+	}
+	// If MaxConcurrency is negative, disable concurrency limits entirely.
+	if cfg.MaxConcurrency < 0 {
+		cfg.MaxConcurrency = 0
+	}
 	// Parse nested fs configs
 	parsers := []configParser{parseFuseConfig, parseBackgroundFetchConfig, parseRetryableHTTPClientConfig, parseBlobConfig, parseContentStoreConfig}
 	for _, p := range parsers {
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -74,12 +74,73 @@ import (
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/semaphore"
 )
 
 var (
 	defaultIndexSelectionPolicy = SelectFirstPolicy
 	fusermountBin               = "fusermount"
+	preresolverQueueBufferSize  = 1024 // arbitrarily chosen buffer size
 )
+
+// Preresolver will resolve a number of layers in parallel,
+// up to the amount specified by MaxConcurrency.
+type preresolver struct {
+	queue chan func(context.Context) string
+	cache *sync.Map
+	smp   *semaphore.Weighted
+}
+
+func newPreresolver(maxConcurrency int64) *preresolver {
+	pr := &preresolver{}
+	pr.queue = make(chan func(context.Context) string, preresolverQueueBufferSize)
+	pr.cache = &sync.Map{}
+	if maxConcurrency > 0 {
+		pr.smp = semaphore.NewWeighted(maxConcurrency)
+	}
+	return pr
+}
+
+func (pr *preresolver) Start(ctx context.Context) error {
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				log.G(ctx).Info("exiting preresolver")
+				return
+			default:
+				resolveFn := <-pr.queue
+				// If concurrency limits are disabled,
+				// we don't need to wait for a semaphore
+				if pr.smp != nil {
+					pr.smp.Acquire(ctx, 1)
+				}
+
+				go func() {
+					digest := resolveFn(ctx)
+					pr.cache.Delete(digest)
+					if pr.smp != nil {
+						pr.smp.Release(1)
+					}
+				}()
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (pr *preresolver) Enqueue(imgNameAndDigest string, fn func(context.Context) string) {
+	if _, ok := pr.cache.Load(imgNameAndDigest); !ok {
+		select {
+		case pr.queue <- fn:
+			pr.cache.Store(imgNameAndDigest, struct{}{})
+		default:
+			return
+		}
+	}
+
+}
 
 type Option func(*options)
 
@@ -88,6 +149,7 @@ type options struct {
 	resolveHandlers   map[string]remote.Handler
 	metadataStore     metadata.Store
 	overlayOpaqueType layer.OverlayOpaqueType
+	maxConcurrency    int64
 }
 
 func WithGetSources(s source.GetSources) Option {
@@ -114,6 +176,12 @@ func WithMetadataStore(metadataStore metadata.Store) Option {
 func WithOverlayOpaqueType(overlayOpaqueType layer.OverlayOpaqueType) Option {
 	return func(opts *options) {
 		opts.overlayOpaqueType = overlayOpaqueType
+	}
+}
+
+func WithMaxConcurrency(maxConcurrency int64) Option {
+	return func(opts *options) {
+		opts.maxConcurrency = maxConcurrency
 	}
 }
 
@@ -180,6 +248,9 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 		return nil, fmt.Errorf("failed to setup resolver: %w", err)
 	}
 
+	pr := newPreresolver(fsOpts.maxConcurrency)
+	pr.Start(ctx)
+
 	var ns *metrics.Namespace
 	if !cfg.NoPrometheus {
 		ns = metrics.NewNamespace("soci", "fs", nil)
@@ -215,6 +286,7 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 		bgFetcher:                   bgFetcher,
 		mountTimeout:                mountTimeout,
 		fuseMetricsEmitWaitDuration: fuseMetricsEmitWaitDuration,
+		pr:                          pr,
 	}, nil
 }
 
@@ -318,6 +390,7 @@ type filesystem struct {
 	bgFetcher                   *bf.BackgroundFetcher
 	mountTimeout                time.Duration
 	fuseMetricsEmitWaitDuration time.Duration
+	pr                          *preresolver
 }
 
 func (fs *filesystem) MountLocal(ctx context.Context, mountpoint string, labels map[string]string, mounts []mount.Mount) error {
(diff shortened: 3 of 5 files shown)


fs/fs.go
```diff
@@ -81,13 +82,66 @@ var (
...
+	maxSize  int
+	currSize int
+}
+
+func (lq *layerQueue) Start() error {
+	if lq.queue != nil {
+		return fmt.Errorf("layer queue already instantiated")
+	}
+
+	lq.queue = make(chan func() string)
```
Kern--, line 98: This channel needs to be buffered or else all enqueue operations will block when we're at our concurrency limit.

fs/fs.go
```diff
@@ -81,13 +82,66 @@ var (
 	fusermountBin               = "fusermount"
 )
 
+type layerQueue struct {
+	mu       sync.Mutex
```
Kern--, line 86: I think we can simplify this structure if we replace the mutex + hashset with a `sync.Map` and max size + cur size with a `semaphore.Weighted`.

That way the structures themselves handle concurrent access and we don't need to pull in a whole data structures library.

fs/fs.go
```diff
@@ -81,13 +82,66 @@ var (
...
+	queue    chan func() string
+	cache    *hashset.Set
+	maxSize  int
+	currSize int
+}
+
+func (lq *layerQueue) Start() error {
+	if lq.queue != nil {
```
Kern--, line 94: nit: A common way to solve this is to have a `newLayerQueue` function which initializes the struture appropriately.

fs/fs.go
```diff
@@ -81,13 +82,66 @@ var (
...
+		return fmt.Errorf("layer queue already instantiated")
+	}
+
+	lq.queue = make(chan func() string)
+	lq.cache = hashset.New()
+
+	go func() {
```
Kern--, line 101: nit: we should probably associate this go-routine with an errgroup or waitgroup so that it can be closed later. It should probably also select over a context so it can be canceled.

fs/fs.go
```diff
@@ -81,13 +82,66 @@ var (
...
+			}()
+		}
+	}()
+
+	return nil
+}
+
+func (lq *layerQueue) Queue(digest string, fn func() string) {
+	if !lq.cache.Contains(digest) {
+		lq.queue <- fn
```
Kern--, line 128: nit: we should probably do a non-blocking write here so that we don't get stuck waiting to setup pre-resolving if the snapshotter falls behind.

fs/fs.go
```diff
@@ -81,13 +82,66 @@ var (
...
+				lq.cache.Remove(digest)
+				lq.mu.Unlock()
+			}()
+		}
+	}()
+
+	return nil
+}
+
+func (lq *layerQueue) Queue(digest string, fn func() string) {
```
Kern--, line 126: nit: Enqueue?

Kern-- (changes requested):

fs/fs.go
sondavidb: Not super familiar with how exactly this works, does the new version this look about right?

config/defaults.go
```diff
@@ -74,6 +74,9 @@ const (
 	// defaultFuseMetricsEmitWaitDurationSec is the amount of time the snapshotter will wait before emitting the metrics for FUSE operation.
 	defaultFuseMetricsEmitWaitDurationSec = 60
 
+	// defaultMaxConcurrency is the maximum number of layers allowed to be pulled at once
+	defaultMaxConcurrency = 100
```
sondavidb, line 78: Chosen arbitrarily, lmk if this needs to be adjusted.

config/fs.go
```diff
@@ -187,6 +187,9 @@ func parseFSConfig(cfg *Config) {
 	if cfg.FuseMetricsEmitWaitDurationSec == 0 {
 		cfg.FuseMetricsEmitWaitDurationSec = defaultFuseMetricsEmitWaitDurationSec
 	}
+	if cfg.MaxConcurrency <= 0 {
```
sondavidb, line 190: The process segfaults on startup if this value is negative, since we attempt to set a negative value to the weighted semaphore. I feel like it might be better to fix the problem for the user instead of an ungraceful crash, but if we think I should change `<=` to `==` to follow convention in the rest of the file, feel free to lmk.

fs/fs.go
```diff
@@ -74,20 +74,67 @@ import (
...
+
+func newLayerQueue(maxConcurrency int64) *layerQueue {
+	lq := &layerQueue{}
+	lq.queue = make(chan func() string) // arbitrarily chosen buffer size
+	lq.cache = &sync.Map{}
+	lq.smp = semaphore.NewWeighted(maxConcurrency)
+	return nil
+}
+
+func (lq *layerQueue) Start(ctx context.Context) error {
```
sondavidb, line 101: I think this errgroup and context stuff is done correctly, just want to double check. Using the same context as the filesystem ties it to the lifecycle of the filesystem, which I think makes sense since we only use this in fs.Mount.

sondavidb (commented):
Functionality is done, I just want to add an integration test. Mentioned a few stuff I wasn't super familiar with in my review below so would like a bit more attention on those bits.

sondavidb:
Added integration tests. Good to review now.

fs/fs.go
```diff
@@ -74,20 +74,68 @@ import (
...
 )
 
 var (
 	defaultIndexSelectionPolicy = SelectFirstPolicy
 	fusermountBin               = "fusermount"
+	layerQueueBufferSize        = 1024
 )
 
+type layerQueue struct {
+	queue chan func() string
```
turan18, line 88: I think the func here should take a context so we pass in the FileSystem context. This way we can cancel in-flight pre-resolves as-well.

fs/fs.go
```diff
@@ -74,20 +74,68 @@ import (
...
+				lq.cache.Delete(digest)
+				lq.smp.Release(1)
+			}()
+		}
+	})
+
+	return nil
+}
+
+func (lq *layerQueue) Enqueue(digest string, fn func() string) {
```
turan18, line 119: We currently reference a `layer` by it's [img ref + digest](https://github.com/awslabs/soci-snapshotter/blob/04a83aaa56243152a0c14828a12cb810997ea6d8/fs/layer/layer.go#L234), and so a `layer` is heavily associated with the image it's a part of. If we only map by digest, we'll only concurrently resolve one layer if two images share the same layer. We may want to make the key ref + digest, so that we can concurrently pre-resolve shared layers as-well.

fs/fs.go
turan18: Do we want a wait/err group? If the current queue is finished resolving, than the `layerQueue` will exit as-well. Meaning, if you were to pull an image some time after an initial pull, than the `layerQueue` will be inactive and so all layer resolves will be sequential, right?

I think the `layerQueue` should be tied to the lifecycle of the process/filesystem.

config/fs.go
sondavidb: Ah, so then if one wants to disable it they'd have to do so explicitly. I'm not positive that they should ever want to, but I suppose it could help to cover the use case. I'd have to refactor the function a bit to make this work, but it shouldn't be anything too difficult probably.

fs/fs.go
```diff
@@ -74,20 +74,68 @@ import (
...
+	lq := &layerQueue{}
+	lq.queue = make(chan func() string, layerQueueBufferSize) // arbitrarily chosen buffer size
+	lq.cache = &sync.Map{}
+	lq.smp = semaphore.NewWeighted(maxConcurrency)
+	return lq
+}
+
+func (lq *layerQueue) Start(ctx context.Context) error {
+	eg, egCtx := errgroup.WithContext(ctx)
+	eg.Go(func() error {
```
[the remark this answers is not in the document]
sondavidb: Sure, this is a bit simpler.

> I don't think we want an errgroup here

Is this because this is the only one in the group and thus is doing more than what we need?

fs/fs.go
```diff
@@ -74,20 +74,68 @@ import (
...
+		for {
+			lq.smp.Acquire(egCtx, 1)
```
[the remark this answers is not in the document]
sondavidb: Done, would just like to know if it's to not have to hold onto memory for a little longer than needed, or if there's another reason.

sondavidb:
More changes pushed. Big notable change is that I changed the struct name from `layerQueue` to `preresolver`, since it felt much clearer of a name to explain what it did. Everything else was just incorporating the feedback.

fs/fs.go
```diff
@@ -74,20 +74,79 @@ import (
...
 	fusermountBin               = "fusermount"
+	preresolverQueueBufferSize  = 1024 // arbitrarily chosen buffer size
 )
 
+// Preresolver will resolve a number of layers in parallel,
+// up to the amount specified by MaxConcurrency.
+type preresolver struct {
+	queue chan func(context.Context) string
+	cache *sync.Map
+	smp   *semaphore.Weighted
```
sondavidb, line 91: Is this a clear enough name for a semaphore? Not sure if there's generally a convention for what to call them in a struct

config/fs.go
monirul: Can we also add a comment in the code to explain what are different meaning of the values eg. 0 means default, >0 means exact value, -1 means disabled.

fs/fs.go
monirul: `smp   *semaphore.Weighted` what does the weighted means here?

fs/fs.go
```diff
@@ -74,20 +74,79 @@ import (
...
+
+func (pr *preresolver) Start(ctx context.Context) error {
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				resolveFn := <-pr.queue
+				if pr.smp != nil {
```
monirul, line 112: Can you pls add a comment at this line something like
`when the concurrency limit is enabled then wait for the turn`

monirul (commented):
Added some comments.

fs/fs.go
sondavidb: It's an implementation of the semaphore in [golang's library](https://pkg.go.dev/golang.org/x/sync/semaphore#Weighted). It works effectively like a semaphore, but instead of only allowing one thread per `Acquire()`, it allows processes to take multiple threads instead.

config/defaults.go
sondavidb: I'm not sure how benchmarking would work in cases like these.

The preresolver has two tasks — deduplication and resource limiting. The former should be a strict performance increase and requires no tuning, but the resource limiting is what we are tuning with this variable. Finding a "best" value is a bit difficult here, since ultimately, we want the majority of customers to be unaware of this change, since they are not pulling too many layers at once. But, ideally, we also want to set a somewhat reasonable value for those who are doing such tasks.

I don't believe it's impossible, of course, I'm just unsure of how I'd go about finding a good value. Would love to hear more thoughts.

monirul: can we run benchmark for a set of images for different values and check if the overall execution time is better or worse for different value?

sondavidb: We actually want execution time to be slower whenever this value kicks in. We might instead want to see what memory usage looks like during a heavy image pull and fine-tune the value until it no longer crosses some threshold. I can discuss and see what that threshold might be.

Kern-- (approved):

fs/fs.go
```diff
@@ -190,6 +257,11 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 		metrics.Register(ns) // Register layer metrics.
 	}
 
+	pr := newPreresolver(fsOpts.maxConcurrency)
+	if err := pr.Start(ctx); err != nil {
```
turan18, line 261: nit: Since the preresolver doesn't return an error, I think we can just run pr.Start in a goroutine here:

```go
pr := newPreresolver(fsOpts.maxConcurrency)
go pr.Start(ctx)
```

fs/fs.go
```diff
@@ -74,20 +74,81 @@ import (
...
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				resolveFn := <-pr.queue
+				// If concurrency limits are disabled,
+				// we don't need to wait for a semaphore
+				if pr.smp != nil {
```
turan18, line 114: I wonder if instead of checking if the semaphore is nil here, we only create a `preresolver` if the max concurrency is > 0 in fs.go. We can check if `preresolver != nil` before getting into the pre-resolve codeblock [here](https://github.com/awslabs/soci-snapshotter/pull/1061/files#diff-c7b54bb28e1654d0ca47364a2c97456c124ee0d3a65854db18d116c509d89136L444). This way we don't have an unnecessary struct and go routine in memory.

sondavidb: The preresolver still handles deduplication in [Enqueue()](https://github.com/awslabs/soci-snapshotter/blob/8b303bdb3b1d214d2671f2b43320a616f5960ee3/fs/fs.go#L132), so I think we should keep this even with concurrency limits disabled.

turan18 (approved):

fs/fs.go
sondavidb: We can, I just wrapped it in that format in case we decide that Start ever needs to return an error. The only context I can think of now that would make sense, though, is if we pass in a bad context and want to be louder on the preresolver failing. But I think it might be better to add logging to show this, instead.

monirul (approved):
LGTM

turan18 (approved):

sondavidb:
Will override and merge to get the change out. Kern previously approved but it got dismissed after changing a nit, and he is unable to review it anytime soon.

Issue #969, opened 2023-11-29
[Bug] Ztocs with a lot of files use a lot of memory/cpu to build metadata db


Kern--:
### Description

When parsing a TOC with many files (>100k) into a fuse metadata db, the soci snapshotter uses a ton of CPU/Memory. From heap/cpu profiles, it looks like most of that time/memory is shuffling around memory related to writing to bbolt.

Heap:
![peak_mem](https://github.com/awslabs/soci-snapshotter/assets/2304727/dd3548af-996a-4ada-b023-6a1ad83fb7b2)

CPU:
![cpu_setup](https://github.com/awslabs/soci-snapshotter/assets/2304727/24c6aa35-3228-48b4-8ffa-e906efc67ea2)

The bbolt documentation explicitly has this limitation that we're probably hitting:

>Bulk loading a lot of random writes into a new bucket can be slow as the
  page will not split until the transaction is committed. Randomly inserting
  more than 100,000 key/value pairs into a single new bucket in a single
  transaction is not advised.

-[bbolt readme](https://github.com/etcd-io/bbolt/blob/d630253fcd673a8613b5a6c836d34ed24dfae620/README.md?plain=1#L846-L849)

We should see if we can change the way we setup the metadata db to mitigate this and if not consider if bbolt is the right tool for what we're doing.

### Steps to reproduce the bug

1) create an image using the following dockerfile:
```
FROM ubuntu@sha256:c9cf959fd83770dfdefd8fb42cfef0761432af36a764c077aed54bbc5bb25368
RUN mkdir /layer1 && for i in $(seq 20000); do dd if=/dev/random bs=1024 count=1 of=/layer1/${i}; done
RUN mkdir /layer2 && for i in $(seq 50000); do dd if=/dev/random bs=1024 count=1 of=/layer2/${i}; done
RUN mkdir /layer3 && for i in $(seq 100000); do dd if=/dev/random bs=1024 count=1 of=/layer3/${i}; done
```
2) Push the image and create a soci index
3) Launch the snapshotter with `debug_address=":6060"`
4) Pull the image with the soci snapshotter
5) Pull a heap profile `curl -s localhost:6060/debug/pprof/heap > heap.pprof` 
6) Pull a CPU profile `curl -s localhost:6060/debug/pprof/cpu?seconds=10`
7) Observe that the pull takes much longer than usual

### Describe the results you expected

Images with many files shouldn't see significantly more memory/cpu utilization than other images

### Host information

[...]

PR #1066, opened 2024-02-05
[WIP | Benchmarking] Optimize bbolt memory usage

turan18:
**Issue #, if available:**

Resolves: #969 

**Description of changes:**

**Testing performed:**


austinvazquez:
@turan18 can you document if there is any performance trade-offs to chunked insertion?

metadata/reader.go
```diff
@@ -671,3 +686,13 @@ func (r *reader) NumOfNodes() (i int, _ error) {
 	}
 	return
 }
+
+// partition partitions a slice of type T into chunks of size chunkSize
+func partition[T any](data []T, chunkSize int) [][]T {
```
austinvazquez, line 691: Can we add some unit tests that show this is chunking correctly? Also we should either add a unit test or documentation that chunkSize never be zero or this function will segfault.

metadata/reader.go
```diff
@@ -53,7 +53,13 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-// reader stores filesystem metadata parsed from ztoc to metadata DB
+const (
+	// bboltInsertionChunkSize roughly determines the maximum number of insertions to
+	// bbolt per transaction.
+	bboltInsertionChunkSize = 5000
```
austinvazquez, line 59: Can we document why 5000? Even a line stating "manually tuned and gave the best performance" could give context in a few years time.

metadata/db.go
```diff
@@ -325,16 +332,18 @@ func writeMetadataEntry(md *bolt.Bucket, m *metadataEntry) error {
...
-		}
-		if err := md.Put(bucketKeyChildName, []byte(firstChildName)); err != nil {
-			return fmt.Errorf("failed to put name first child %q: %w", firstChildName, err)
-		}
 		if len(m.children) > 1 {
+			// Sort children by base name.
+			keys := make([]string, len(m.children))
+			for k := range m.children {
+				keys = append(keys, k)
+			}
```
austinvazquez, line 340: If we want to preallocate, then it should be `keys := make([]string, 0, len(m.children)`. Right now we are using twice the memory.

The current form will create the container with size len(m.children) and fill with zero values. i.e. empty strings. This results in the empty keys check below. That can be removed with this fix.

metadata/db.go
```diff
@@ -325,16 +332,18 @@ func writeMetadataEntry(md *bolt.Bucket, m *metadataEntry) error {
...
+			sort.Slice(keys, func(i, j int) bool {
+				return bytes.Compare([]byte(m.children[keys[i]].base), []byte(m.children[keys[j]].base)) < 0
+			})
 			var cbkt *bolt.Bucket
-			for k, c := range m.children {
-				if k == firstChildName {
+			for _, key := range keys {
+				if key == firstChildName || key == "" {
```
austinvazquez, line 346: `key == ""` check can be removed with above fix.

austinvazquez (changes requested):

metadata/db.go
turan18: Ah, right. I forgot the capacity is doubled once we hit the length.

metadata/reader.go
turan18: Sure, I can add unit tests. This was taken from the Go Wiki: https://go.dev/wiki/SliceTricks#batching-with-minimal-allocation.

turan18:
> @turan18 can you document if there is any performance trade-offs to chunked insertion?

Added the pull performance benchmarks to the description.

metadata/db.go
```diff
@@ -162,36 +170,16 @@ func getMetadataBucketByID(md *bolt.Bucket, id uint32) (*bolt.Bucket, error) {
...
-		if v.val != 0 {
-			val, err := dbutil.EncodeInt(v.val)
-			if err != nil {
-				return err
-			}
-			if err := b.Put(v.key, val); err != nil {
-				return err
-			}
-		}
+// writeMetadataEntry writes node metadata to the appropriate node sub-bucket inside the top-level nodes bucket.
```
austinvazquez, line 173: nit: writeNodeEntry*

austinvazquez (dismissed):
Minor nit, but changes look good to me. Thanks for all the detail in the PR description.

austinvazquez (approved):

sondavidb (approved):

PR #1080, opened 2024-02-16
Remove unparsed references from zTOC

### LINKED ISSUE #968: [FEATURE] Don't store parsed ztoc in memory
### Description

Related to #967

When we set up a fuse filesystem for a layer that will be lazily loaded by soci, we first unmarshal a ztoc:
https://github.com/awslabs/soci-snapshotter/blob/21ec5445ea5e0908861e60e92cbdcd70d3251c93/fs/layer/layer.go#L291

We then give the TOC portion to the metadata store to convert into bbolt db on disk:
https://github.com/awslabs/soci-snapshotter/blob/21ec5445ea5e0908861e60e92cbdcd70d3251c93/fs/layer/layer.go#L326

And we hold a reference to the ztoc and a zinfo in the span manager:
https://github.com/awslabs/soci-snapshotter/blob/21ec5445ea5e0908861e60e92cbdcd70d3251c93/fs/span-manager/span_manager.go#L92-L105

The problem with this is that we end up holding onto a lot of memory that we don't need.

Once we convert the TOC to the metadata DB, we never reference the `ztoc.TOC` again which means this memory could be freed, but it can't because we hold a reference to the ztoc in the span manager. Similarly, when we create the zinfo, we `malloc` space in the C code and copy all the checkpoint data into a parsed form:

https://github.com/awslabs/soci-snapshotter/blob/21ec5445ea5e0908861e60e92cbdcd70d3251c93/ztoc/compression/gzip_zinfo.c#L567

This means that we're holding on to both a TOC that we will never use and 2 copies of the zInfo. 

Ideally we wouldn't hold any memory that we don't need so we wouldn't have the checkpoint bytes nor TOC in memory after the initial set up. We can't just remove the reference to the ztoc because we still need the span digests.

I think we should reconsider the APIs here since we kind of explode the ztoc into 3 parts (TOC on disk, zinfo in a separate struct, and span digests in the ztoc itself) and we don't have a clear notion of ownership/responsibility for those components. 

It would be good if we can come up with some API that allows us to have a nice wrapper to make programming against the zTOC easy if you don't really care about memory, but simple to destructively decompose into parts so we don't have duplication. 

Alternatively we could do the simple fix and just set TOC = nil when we're done with it and do #81 so that we never store an unparsed zinfo so we only ever have 1 copy.

I think this is likely to bite us for not thinking about what the ztoc APIs should look like.

_No response_

sondavidb:
**Issue #, if available:**
Fixes #968 

**Description of changes:**
Our current zTOC structure uses more memory than needed. Specifically, the Checkpoints and FileMetadata arrays only get called once, yet they cannot be freed because the SpanManager retains a reference to them, despite never needing either past their initial calls. This is an inherent design flaw with our zTOC APIs, and this fix is a temporary workaround to increase performance. Once we can flesh out the zTOC API, this solution can be much more elegant, or even unneeded.

**Testing performed:**

I additionally ran a Prometheus server to show memory usage on heap over time. On my branch, I started the snapshotter, pulled the tensorflow image, then waited for it to stabilize. Then I did the same with main. The first metrics are with my change, the second ones are without.

image

We can see that we now have a baseline of roughly 18-30MB being used, instead of the 50-95MB previously used. The initial pulls remain unchanged (and can be somewhat volatile), but now while the snapshotter is running, it holds onto much less memory than before.

### DIFF
--- a/fs/layer/layer.go
+++ b/fs/layer/layer.go
@@ -327,6 +327,7 @@ func (r *Resolver) Resolve(ctx context.Context, hosts []docker.RegistryHost, ref
 	if err != nil {
 		return nil, err
 	}
+	ztoc.TOC.FileMetadata = nil
 	log.G(ctx).Debugf("[Resolver.Resolve]Initialized metadata store for layer sha=%v", desc.Digest)
 
 	spanManager := spanmanager.New(ztoc, sr, spanCache, r.config.BlobConfig.MaxSpanVerificationRetries, cache.Direct())
--- a/ztoc/ztoc.go
+++ b/ztoc/ztoc.go
@@ -180,6 +180,7 @@ func (zt Ztoc) ExtractFile(r *io.SectionReader, filename string) ([]byte, error)
 		return nil, nil
 	}
 	defer zinfo.Close()
+	zt.Checkpoints = nil
 
 	spanStart := zinfo.UncompressedOffsetToSpanID(entry.UncompressedOffset)
 	spanEnd := zinfo.UncompressedOffsetToSpanID(entry.UncompressedOffset + entry.UncompressedSize)
@@ -244,6 +245,7 @@ func (zt Ztoc) ExtractFromTarGz(gz string, filename string) (string, error) {
 		return "", err
 	}
 	defer zinfo.Close()
+	zt.Checkpoints = nil
 
 	bytes, err := zinfo.ExtractDataFromFile(gz, entry.UncompressedSize, entry.UncompressedOffset)
 	if err != nil {
--- a/cmd/soci/commands/ztoc/info.go
+++ b/cmd/soci/commands/ztoc/info.go
@@ -91,6 +91,7 @@ var infoCommand = cli.Command{
 			return err
 		}
 		defer gzInfo.Close()
+		ztoc.Checkpoints = nil
 
 		multiSpanFiles := 0
 		zinfo := Info{
--- a/cmd/soci/commands/ztoc/info.go
+++ b/cmd/soci/commands/ztoc/info.go
@@ -90,6 +90,7 @@ var infoCommand = cli.Command{
 		if err != nil {
 			return err
 		}
+		defer gzInfo.Close()
 
 		multiSpanFiles := 0
 		zinfo := Info{
--- a/fs/span-manager/span_manager.go
+++ b/fs/span-manager/span_manager.go
@@ -93,6 +93,7 @@ func New(ztoc *ztoc.Ztoc, r *io.SectionReader, cache cache.BlobCache, retries in
 	if err != nil {
 		return nil
 	}
+	ztoc.Checkpoints = nil
 	spans := make([]*span, ztoc.MaxSpanID+1)
 	m := &SpanManager{
 		cache:                             cache,


fs/layer/layer.go
```diff
@@ -327,6 +327,7 @@ func (r *Resolver) Resolve(ctx context.Context, hosts []docker.RegistryHost, ref
 	if err != nil {
 		return nil, err
 	}
+	ztoc.TOC.FileMetadata = nil // We no longer need to hold onto the TOC
```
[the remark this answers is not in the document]
sondavidb: Done, I would prefer this gets passed into the metadataStore function, but since it's a function that must be implemented by whatever store uses it, I figured it was better to just consume it after using it.

ztoc/ztoc.go
```diff
@@ -253,8 +253,23 @@ func (zt Ztoc) ExtractFromTarGz(gz string, filename string) (string, error) {
...
+// algorithm in the ztoc. This process nils the byte array used in the process, so
+// it can only be called once.
+func (zt Ztoc) ConsumeZinfo() (compression.Zinfo, error) {
+	if zt.Checkpoints != nil {
+		zinfo, err := compression.NewZinfo(zt.CompressionAlgorithm, zt.Checkpoints)
+		if err != nil {
+			return nil, err
+		}
+		// Zinfo holds all of the checkpoint data, so we can free the raw Checkpoints array.
+		zt.Checkpoints = nil
```
turan18, line 270: Does this acutally work? The methods exposed by Ztoc take non-pointer receivers, so we are modifying the deferenced _copy_  of the Ztoc. 

eg: https://go.dev/play/p/Ng5JA8n3m06

I think while we work on stabilizing the Ztoc API, it should be up to consumers to manually free the checkpoint bytes. 

```go

type SpanManager struct {
	zinfo                             compression.Zinfo
	ztoc                              *ztoc.Ztoc
}

func New(ztoc *ztoc.Ztoc, r *io.SectionReader, cache cache.BlobCache, retries int, cacheOpt ...cache.Option) *SpanManager {
       index, err := ztoc.Zinfo()
	if err != nil {
		return nil
	}
        ztoc.Checkpoints = nil
	m := &SpanManager{
		zinfo:                             index,
		ztoc:                              ztoc,
                ...
	}
}
```

ztoc/ztoc.go
```diff
@@ -253,8 +253,23 @@ func (zt Ztoc) ExtractFromTarGz(gz string, filename string) (string, error) {
 	return string(bytes), nil
 }
 
-// Zinfo deserilizes and returns a Zinfo based on the zinfo bytes and compression
-// algorithm in the ztoc.
-func (zt Ztoc) Zinfo() (compression.Zinfo, error) {
-	return compression.NewZinfo(zt.CompressionAlgorithm, zt.Checkpoints)
+func (zt Ztoc) ConsumeTOC() {
+	zt.TOC.FileMetadata = nil
```
turan18, line 257: Same issue as below.

turan18 (changes requested):

ztoc/ztoc.go
sondavidb: > Does this acutally work? The methods exposed by Ztoc take non-pointer receivers, so we are modifying the deferenced copy of the Ztoc.

Ah, good catch. Would it be considered poor practice then to have some methods refer to pointers, while others do not? 

> it should be up to consumers to manually free the checkpoint bytes

Should it, though? The checkpoint bytes are essentially garbage for anything outside of creating the Zinfo, so I don't think it's a bad thing to have the Zinfo auto-consume. Though, if it's bad practice to refer to some methods by reference and others by value, then I don't think we have a choice outside of what you mentioned.

sondavidb:
Pushed some new changes to address @turan18's comments. Also added unit tests to ensure that these entries were freed.

I think I'll close this PR if this method is ultimately considered to be a bad approach to this problem, but I think this premise should still be used, regardless.

ztoc/ztoc.go
turan18: Yeah, mixed receivers are generally considered bad practice. I suppose we could change all Ztoc methods to take pointer receivers, but that's _technically_ considered an API change and so we _may_ want to think about what a stable access API looks like before that, although I'm not strongly convinced by that either.

sondavidb: > I suppose we could change all Ztoc methods to take pointer receivers

I was going to pitch changing everything to be a pointer, but then functions such as ExtractFile would consume the zinfo, and doesn't pass the zinfo back, which really doesn't make sense.

I think I'll just revert a bunch of these changes and set the Checkpoints/TOC to nil manually. But I want to stress our need to come back to this soon.

sondavidb:
With the new changes per previous discussions I removed both of the unit tests. Removing the metadata store function might be worth testing to ensure it doesn't re-write with a nil TOC, but it felt a little out-of-place to test. (Not to mention it's specific to the implementation of each store.)

sondavidb:
Flamegraphs have been updated. It's very noteworthy that we can no longer see the zTOC data structure at all in the flamegraphs. The SpanDigests that are stored are probably negligible compared to the giant checkpoints and FileMetadata arrays that we were keeping around, hence why we can't see it in the graph. (But that's just my guess.)

turan18 (approved):

PR #1119, opened 2024-03-19
Fix max concurrent uploads on SOCI push

### LINKED ISSUE #1118: [Bug] Max concurrent uploads flag is unused
### Description

Follow-up to #1113, where it was found the `max-concurrent-uploads` flag can be set but is never read. The result is the default concurrent limit of 3 is used (defined by oras-go library [ref](https://github.com/oras-project/oras-go/blob/8d139f0be9190ee4e48fe00cda08e0185ccd2494/copy.go#L93))

Push SOCI index to registry with `--max-concurrent-uploads` != 3.

As a user, I expect this flag to be honored by the soci push command.

### Host information

1. OS: Amazon Linux 2023 x86_64
2. Snapshotter Version: soci version 7af8577 7af8577ac627f991d561f94d476ea4529a02626a
3. Containerd Version: containerd github.com/containerd/containerd v1.6.28 ae07eda36dd25f8a1b98dfbf587313b99c0190bb

_No response_

austinvazquez:
**Issue #, if available:**
Fixes #1118 

**Description of changes:**
This change fixes the SOCI push command to set the maximum number of concurrent copy tasks to the value passed via the CLI flag. Before this change, the push command would not use the concurrency limit value set via CLI and default to 3 copy tasks. [Ref](https://github.com/oras-project/oras-go/blob/8d139f0be9190ee4e48fe00cda08e0185ccd2494/copy.go#L93)

**Testing performed:**

Added some logs to validate concurrency is now being set by the flag.
```
$ mysudo soci push docker.io/library/hello-world:latest
The concurrent limits are 3
soci: could not find any soci indices to push
$ 
```

```
$ mysudo soci push --max-concurrent-uploads=20 docker.io/library/hello-world:latest
The concurrent limits are 20
soci: could not find any soci indices to push
$ 
```

Before the change:
```
$ mysudo soci push docker.io/library/hello-world:latest
The concurrent limits are 0
soci: could not find any soci indices to push
$
```

```
$ mysudo soci push --max-concurrent-uploads=20 docker.io/library/hello-world:latest
The concurrent limits are 0
soci: could not find any soci indices to push
$ 
```

### DIFF
--- a/cmd/soci/commands/push.go
+++ b/cmd/soci/commands/push.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"sort"
 	"strings"
@@ -36,6 +37,15 @@ import (
 	oraslib "oras.land/oras-go/v2"
 	"oras.land/oras-go/v2/registry/remote"
 	"oras.land/oras-go/v2/registry/remote/auth"
+)
+
+const (
+	maxConcurrentUploadsFlag = "max-concurrent-uploads"
+
+	// defaultMaxConcurrentUploads is the default number of copy tasks used to upload the SOCI artifacts.
+	// This aligns with the ORAS Go library default value.
+	// Reference: https://github.com/oras-project/oras-go/blob/850a24737c15927603d1cecddc74c87f5f3377f4/copy.go#L38
+	defaultMaxConcurrentUploads = 3
 )
 
 // PushCommand is a command to push an image artifacts from local content store to the remote repository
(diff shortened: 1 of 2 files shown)


cmd/soci/commands/push.go
```diff
@@ -38,6 +39,11 @@ import (
 	"oras.land/oras-go/v2/registry/remote/auth"
 )
 
+const (
+	maxConcurrentUploadsFlag    = "max-concurrent-uploads"
+	defaultMaxConcurrentUploads = 10
```
sondavidb, line 44: ORAS already has a default value of 3, I think we should be more in-line with it. A value of 3 or 0 are both ok. I think the latter may be better with a comment explaining that this will set the value to the default one set in ORAS.

austinvazquez: Sure pushed a default value of 3. 0 is a little strange IMO for default behavior to rely on third party library. Also updated docs with the default value change.

coderbirju (approved):

cmd/soci/commands/push.go
sondavidb: Sure, I suggested zero to ensure we stay consistent but deviating in this case is probably not a problem.

sondavidb (approved):

PR #1129, opened 2024-03-21
Update go version in CI to v1.21.8

coderbirju:
**Issue #, if available:**

**Description of changes:**

**Testing performed:**


.github/workflows/bump-deps.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     - uses: actions/checkout@v4
     - uses: actions/setup-go@v5
       with:
-        go-version: '1.21.3'
+        go-version: ${{ env.GO_VERSION }}
```
pendo324, line 26: Seems like GO_VERSION isn't set in the `env` for this workflow?

coderbirju: My bad, added it.

austinvazquez:
Slight nit but can we call out the Go version in the commit subject? e.g. Update Golang version in CI to v1.21.8

This just helps to be more explicit so at a glance we can differentiate it from other update commits which will eventually happen. :)

pendo324 (approved):

sondavidb (approved):

PR #1133, opened 2024-03-23
Disable xattrs by default

sondavidb:
**Issue #, if available:**
Followup to #1021

**Description of changes:**
#1021 introduced a check to not call `GetXAttrs()` and related calls, as well as building an index with these annotations. This behavior was opt-out by default. As it generally increases performance, this PR changes this to be default behavior.

This also removes the need for the Optimizations structure in the CLI, thus it was removed.

**Testing performed:**
`make test`. Additionally made some indexes and confirmed the flag created/didn't create the annotations based on expectation.

### DIFF
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -18,7 +18,6 @@ package commands
 
 import (
 	"errors"
-	"fmt"
 	"os"
 
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/internal"
@@ -32,7 +31,7 @@ const (
 	buildToolIdentifier = "AWS SOCI CLI v0.1"
 	spanSizeFlag        = "span-size"
 	minLayerSizeFlag    = "min-layer-size"
-	optimizationFlag    = "optimizations"
+	disableXAttrsFlag   = "disable-xattrs"
 )
 
 // CreateCommand creates SOCI index for an image
@@ -55,24 +54,15 @@ var CreateCommand = cli.Command{
 			Usage: "Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10 MiB.",
 			Value: 10 << 20,
 		},
-		cli.StringSliceFlag{
-			Name:  optimizationFlag,
-			Usage: fmt.Sprintf("(Experimental) Enable optional optimizations. Valid values are %v", soci.Optimizations),
+		cli.BoolTFlag{
+			Name:  disableXAttrsFlag,
+			Usage: "When true, adds DisableXAttrs annotation to SOCI index. This annotation often helps performance at pull time. Default is true",
 		},
 	),
 	Action: func(cliContext *cli.Context) error {
 		srcRef := cliContext.Args().Get(0)
 		if srcRef == "" {
 			return errors.New("source image needs to be specified")
-		}
-
-		var optimizations []soci.Optimization
-		for _, o := range cliContext.StringSlice(optimizationFlag) {
-			optimization, err := soci.ParseOptimization(o)
-			if err != nil {
-				return err
-			}
-			optimizations = append(optimizations, optimization)
 		}
 
 		client, ctx, cancel, err := internal.NewClient(cliContext)
@@ -119,7 +109,10 @@ var CreateCommand = cli.Command{
 			soci.WithMinLayerSize(minLayerSize),
 			soci.WithSpanSize(spanSize),
 			soci.WithBuildToolIdentifier(buildToolIdentifier),
-			soci.WithOptimizations(optimizations),
+		}
+
+		if !cliContext.Bool(disableXAttrsFlag) {
+			builderOpts = append(builderOpts, soci.WithNoDisableXAttrs())
 		}
 
 		for _, plat := range ps {
--- a/config/config.toml
+++ b/config/config.toml
@@ -19,7 +19,7 @@ allow_no_verification=true
 # disable_verification=false
 # Causes TestRunWithDefaultConfig to break, but
 # fine to use in /etc/soci-snapshotter-grpc-config.toml
-max_concurrency=0  # Actually zero
+max_concurrency=0
 no_prometheus=false
 mount_timeout_sec=0
 fuse_metrics_emit_wait_duration_sec=0
--- a/docs/cli-usage.md
+++ b/docs/cli-usage.md
@@ -33,6 +33,7 @@ Flags:
 
  - ```--span-size``` : Span size that soci index uses to segment layer data. Default is 4MiB 
  - ```--min-layer-size``` : Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10MiB
+ - ```--disable-xattrs``` : When true, adds DisableXAttrs annotation to SOCI index. This annotation often helps performance at pull time. Default is true.
 
 **Example:** 
 ```
--- a/fs/layer/layer.go
+++ b/fs/layer/layer.go
@@ -345,14 +345,14 @@ func (r *Resolver) Resolve(ctx context.Context, hosts []docker.RegistryHost, ref
 	if err != nil {
 		return nil, fmt.Errorf("failed to read layer: %w", err)
 	}
-	disableXattrs := getDisableXAttrAnnotation(sociDesc)
+	disableXAttrs := getDisableXAttrAnnotation(sociDesc)
 	// Combine layer information together and cache it.
-	l := newLayer(r, desc, blobR, vr, bgLayerResolver, opCounter, disableXattrs)
+	l := newLayer(r, desc, blobR, vr, bgLayerResolver, opCounter, disableXAttrs)
 	r.layerCacheMu.Lock()
 	cachedL, done2, added := r.layerCache.Add(name, l)
 	r.layerCacheMu.Unlock()
 	if !added {
-		l.close() // layer already exists in the cache. discrad this.
+		l.close() // layer already exists in the cache. discard this.
 	}
 
 	log.G(ctx).Debugf("resolved layer")
--- a/soci/soci_index.go
+++ b/soci/soci_index.go
@@ -68,6 +68,7 @@ const (
 	// [soci-snapshotter] NOTE: this is a duplicate of fs/layer/layer.go so that the SOCI package and the snapshotter can
 	// be independent (and SOCI could be split out into it's own module in the future).
 	whiteoutOpaqueDir = ".wh..wh..opq"
+	disableXAttrsTrue = "true"
 )
 
 var (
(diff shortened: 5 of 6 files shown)


fs/layer/layer.go
```diff
@@ -345,7 +345,11 @@ func (r *Resolver) Resolve(ctx context.Context, hosts []docker.RegistryHost, ref
 	if err != nil {
 		return nil, fmt.Errorf("failed to read layer: %w", err)
 	}
-	disableXattrs := getDisableXAttrAnnotation(sociDesc)
+	disableXattrs := false
+	if !r.config.ForceGetXAttrs {
```
haddscot, line 349: nitpick but this reads a little strange to me, as opposed to

```
disableXattrs := true
if r.config.ForceGetXAttrs {
```

feels a bit odd that the "default behavior" is to enter the if statement

fs/layer/layer.go
```diff
@@ -345,7 +345,11 @@ func (r *Resolver) Resolve(ctx context.Context, hosts []docker.RegistryHost, ref
...
+		fmt.Println("xattrs checking")
```
haddscot, line 350: I don't see any other instances of `fmt.Println` in this file – should this be a log statement? can it be a little more verbose?

sondavidb: Debugging log for myself that I forgot to remove, whoops

fs/layer/layer.go
sondavidb: The way I see it is, by default, we're checking for xattrs. But, if our config does not tell us that we are required to do this, check the annotation to see if we can maybe skip this step.

It's kind of confusing, though, I agree. I just pushed a change to this, LMK if it's better. (It's technically a tiny bit less optimal but probably inconsequential)

TBH I think the var name is just bad :p If it were up to me, it would be something like "checkXAttrs" and pass that down the call chain. But since FUSE names this variable [DisableXAttrs](https://github.com/hanwen/go-fuse/blob/36b359112698c1ccc5d5139693b92f62eb02477c/fuse/api.go#L219) it would probably be more confusing to deviate from the name.

fs/layer/layer.go
```diff
@@ -345,9 +345,12 @@ func (r *Resolver) Resolve(ctx context.Context, hosts []docker.RegistryHost, ref
 	if err != nil {
 		return nil, fmt.Errorf("failed to read layer: %w", err)
 	}
-	disableXattrs := getDisableXAttrAnnotation(sociDesc)
+	disableXAttrs := getDisableXAttrAnnotation(sociDesc)
+	if r.config.ForceGetXAttrs {
+		disableXAttrs = false
+	}
```
austinvazquez, line 351: Slight optimization we can use short circuiting to avoid parsing the string value set in layer annotations when the configuration is to force get xattrs.

```suggestion
        disableXAttrs := r.config.ForceGetXAttrs && getDisableXAttrAnnotation(sociDesc)
```

soci/soci_index.go
```diff
@@ -511,6 +525,9 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 		IndexAnnotationImageLayerDigest:    desc.Digest.String(),
 	}
 	b.maybeAddDisableXattrAnnotation(&ztocDesc, toc)
+	if desc.Annotations[IndexAnnotationDisableXAttrs] == "true" {
```
austinvazquez, line 528: Let's just add a constant to ensure this always matches the set value in `maybeAddDisableXattrAnnotation` function.

soci/soci_index.go
```diff
@@ -216,8 +216,15 @@ type buildConfig struct {
...
 	platform            ocispec.Platform
 	optimizations       []Optimization
+	forceGetXAttrs      bool
 }
 
+// Optimization represents an optional optimization to be applied when building the SOCI index
+type Optimization string
+
+// Uncomment the below block when adding optimizations.
+// This is commented out for now to pass CI lint checker.
```
austinvazquez, line 226: Maybe it is best to either remove the code for now or exception the lint checks here if we want to keep this implementation.

austinvazquez (changes requested):
Overall LGTM, thanks for carrying this forward. Let's add some more for the documentation of XAttrs in CLI and configuration.

soci/soci_index.go
sondavidb: I think the question to ask is, how likely is it that we find some other form of experimental change? I think it's pretty unlikely, but if it ever happens, creating a new structure for it instead of appending to what's already existing sounds like it could end up being more convoluted than needed. But, having dead code for months, if not indefinitely, sounds like poor coding practice. But it's also not like it takes up a lot of space.

I created a placeholder and excepted the function for now, but very open to just removing it entirely.

fs/layer/layer.go
sondavidb: Good idea, though the first variable should be inverted, since the only condition where it is true is if we do not want to force GetXAttrs and the annotation is also true.

This probably goes to show how confusing this bit is, though. I added an extra comment to hopefully clarify for future readers.

soci/soci_index.go
```diff
@@ -68,6 +68,7 @@ const (
 	// [soci-snapshotter] NOTE: this is a duplicate of fs/layer/layer.go so that the SOCI package and the snapshotter can
 	// be independent (and SOCI could be split out into it's own module in the future).
 	whiteoutOpaqueDir = ".wh..wh..opq"
+	DisableXAttrsTrue = "true"
```
austinvazquez, line 71: Small nit. Let's reduce the scope of this until it needs to be exported (if ever).

soci/soci_index.go
austinvazquez: Yeah I agree with you. I don't have a strong opinion either way.

austinvazquez (dismissed):
+1, happy with how this turned out. Great work.

config/config.toml
```diff
@@ -19,10 +19,11 @@ allow_no_verification=true
 # disable_verification=false
 # Causes TestRunWithDefaultConfig to break, but
 # fine to use in /etc/soci-snapshotter-grpc-config.toml
-max_concurrency=0  # Actually zero
+max_concurrency=0
 no_prometheus=false
 mount_timeout_sec=0
 fuse_metrics_emit_wait_duration_sec=0
+force_get_xattrs=false
```
henry118, line 26: this config name is a bit weird. why just call it "disable_xattrs" and default to "true"?

cmd/soci/commands/create.go
```diff
@@ -29,10 +29,11 @@ import (
...
 
 const (
-	buildToolIdentifier = "AWS SOCI CLI v0.1"
-	spanSizeFlag        = "span-size"
-	minLayerSizeFlag    = "min-layer-size"
-	optimizationFlag    = "optimizations"
+	buildToolIdentifier    = "AWS SOCI CLI v0.1"
+	spanSizeFlag           = "span-size"
+	minLayerSizeFlag       = "min-layer-size"
+	forceXAttrsEnabledFlag = "force-xattrs-enabled"
```
henry118, line 35: Can we just name this option "disable-xattrs" and default to true?

soci/soci_index.go
```diff
@@ -227,17 +233,15 @@ func (b *buildConfig) hasOptimization(o Optimization) bool {
...
 
-// Optimization represents an optional optimization to be applied when building the SOCI index
-type Optimization string
-
 const (
-	// XAttrOptimization optimizes xattrs by disabling them for layers where there are no xattrs or opaque directories
-	XAttrOptimization Optimization = "xattr"
+	// NoneOptimization is a placeholder for any future optimizations.
+	// If future optimizations are added, the below line can be commented/removed.
+	NoneOptimization = "none"
```
henry118, line 239: why do we need this? can we not just leave Optimizations as an empty array?

sondavidb: The previous iteration had this entire code block commented out to be example code. We could just make it empty, though.

config/config.toml
sondavidb: I think `disable_xattrs` is a bit of a misleading name. The ultimate goal of the flag is to decide whether or not the runtime honors the DisableXAttrs annotation. The variable names were intentionally chosen to avoid making users believe this disables xattrs for files that have xattrs, when it reality it just skips the check for layers that don't have them anyway.

With that in mind, I think the variable name is fine, but I'm open for alternative variable names.

cmd/soci/commands/create.go
sondavidb: urfave/cli doesn't allow bool flags to default to true, but I agree that it's a bit of a wordy flag name. Unsure of an alternative variable name that would not confuse users. Maybe something like `--skip-xattr-check` or something? Though it's not much better either...

config/config.toml
henry118: The DisableXAttrs annotation was set internally based on "optimizations" list. This PR changed the logic based on a new config/flag. Either way it's transparent to end users, who seem only care about if [DisableXAttrs](https://github.com/awslabs/soci-snapshotter/blob/main/fs/fs.go#L604C3-L604C16) FUSE mount option is turned on.

Another option here is enable all optimizations and allow users to opt-out individual ones, and change the optimizations flag to something like "disable-optimizations".

sondavidb: > Either way it's transparent to end users, who seem only care about if [DisableXAttrs](https://github.com/awslabs/soci-snapshotter/blob/main/fs/fs.go#L604C3-L604C16) FUSE mount option is turned on.

It doesn't inherently turn it on, though, since it only means we'll read the annotation.

With this comment + previous comments on when I plumbed this variable into the code, it's becoming pretty clear that my verbosity in clarifying variable behavior is probably doing the opposite, so I'll just keep it simple here and use `disable_xattrs`.

cmd/soci/commands/create.go
henry118: I believe you can do it with [BoolTFlag](https://github.com/urfave/cli/blob/v1.22.14/flag_bool_t.go#L10) :)

sondavidb: Oh that's pretty neat. Just pushed changes.

docs/cli-usage.md
```diff
@@ -33,6 +33,7 @@ Flags:
 
  - ```--span-size``` : Span size that soci index uses to segment layer data. Default is 4MiB 
  - ```--min-layer-size``` : Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10MiB
+ - ```--force-xattrs-enabled``` : When true, do not add DisableXAttrs annotation. This annotation often helps performance at pull time. Default is false.
```
henry118, line 36: this requires update

henry118 (dismissed):
LGTM. Thanks.

cmd/soci/commands/create.go
```diff
@@ -55,6 +56,10 @@ var CreateCommand = cli.Command{
 			Usage: "Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10 MiB.",
 			Value: 10 << 20,
 		},
+		cli.BoolTFlag{
+			Name:  disableXAttrsFlag,
+			Usage: "Specify this flag to skip adding DisableXAttrs annotation for every layer.",
```
austinvazquez, line 61: We need to update this to specify it adds the disable xattrs optimization and default value of true. To disable, users will need to `--disable-xattrs=false`.

docs/cli-usage.md
```diff
@@ -33,6 +33,7 @@ Flags:
 
  - ```--span-size``` : Span size that soci index uses to segment layer data. Default is 4MiB 
  - ```--min-layer-size``` : Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10MiB
+ - ```--disable-xattrs``` : Specify this flag to skip adding DisableXAttrs annotation. This annotation often helps performance at pull time.
```
austinvazquez, line 36: Same here.

config/fs.go
```diff
@@ -56,6 +56,10 @@ type FSConfig struct {
 	NoPrometheus                   bool   `toml:"no_prometheus"`
 	MountTimeoutSec                int64  `toml:"mount_timeout_sec"`
 	FuseMetricsEmitWaitDurationSec int64  `toml:"fuse_metrics_emit_wait_duration_sec"`
+	// DisableXAttrs will read the DisableXAttrs annotation at pull time.
+	// Disabling this often results in degraded performance, as GetXAttrs
+	// is an expensive call, but it might be useful for debugging purposes.
```
austinvazquez, line 61: Slightly confusing with the new configuration naming.

henry118: We seem to all have slightly different interpretations of this configuration. Let's have a quick sync up offline tomorrow. @sondavidb @austinvazquez

sondavidb:
After some discussion, we decided to remove the TOML variable to ignore this at runtime. Instead, should users wish to disable this behavior, they must recreate a new SOCI index with the flag disabled. This is done for clarity on the developer side, and users are very unlikely to want to turn off this behavior anyway.

henry118 (approved):

austinvazquez (approved):

PR #1135, opened 2024-03-25
Fix scripts/check-dco.sh CI failure

sondavidb:
**Issue #, if available:**

**Description of changes:**
CI started failing after our last commit because our [fetch depth](https://github.com/awslabs/soci-snapshotter/blob/9d6a4202b4831daa193ccd08ae125ba216b09780/.github/workflows/prebuild.yml#L22) in checkout is only 21. We only changed `scripts/check-dco.sh` because we erroneously allowed a non-DCO signed commit into mainline, so we can revert the change from then.

Also added some newlines at the end of our scripts.

**Testing performed:**


austinvazquez (approved):
LGTM

pendo324 (approved):

PR #1141, opened 2024-03-27
Fine-tune shell scripts

sondavidb:
**Issue #, if available:**

**Description of changes:**
Added some modularity in our installation scripts by moving versions into their own variable. Also added an integrity check for cmake, as well as TODO comments for the other scripts.

**Testing performed:**
Ran both scripts and ensured they both ran fine

### DIFF
--- a/scripts/install-dep.sh
+++ b/scripts/install-dep.sh
@@ -22,10 +22,29 @@ set -eux -o pipefail
 TMPDIR=$(mktemp -d)
 pushd "${TMPDIR}"
 
+arch="$(uname -m)"
+
 # install cmake
+cmake_ver="3.24.1"
 if ! command -v cmake &> /dev/null
 then
-    wget https://github.com/Kitware/CMake/releases/download/v3.24.1/cmake-3.24.1-Linux-x86_64.sh -O cmake.sh
+    wget https://github.com/Kitware/CMake/releases/download/v"${cmake_ver}"/cmake-"${cmake_ver}"-Linux-"${arch}".sh -O cmake.sh
+    wget https://github.com/Kitware/CMake/releases/download/v"${cmake_ver}"/cmake-"${cmake_ver}"-SHA-256.txt -O shasums.txt
+
+    # Verify integrity
+
+    # The sha256sum will come in the format like the following:
+    # sha256hash filename
+    # Since we rename the script for convenience purposes,
+    # we only want to compare hashes to verify file integrity.
+    cmake_expected_shasum="$(grep "${arch}".sh < shasums.txt | awk '{print $1}')"
+    cmake_actual_shasum="$(sha256sum cmake.sh | awk '{print $1}')"
+    if [ "${cmake_expected_shasum}" != "${cmake_actual_shasum}" ]
+    then
+        echo "error: cmake sha256sum did not match"
+        exit 1
+    fi
+    
     sh cmake.sh --prefix=/usr/local/ --exclude-subdir
     rm -rf cmake.sh
 else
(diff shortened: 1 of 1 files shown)


scripts/install-check-tools.sh
```diff
@@ -17,11 +17,14 @@
...
+golangci_lint_ver="1.56.2"
+curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$(go env GOPATH)/bin" v"${golangci_lint_ver}"
+# TODO: Verify script integrity via checksums
 go install github.com/kunalkushwaha/ltag@v0.2.4
 go install github.com/vbatts/git-validation@v1.2.0
 
 scversion="v0.10.0"
 arch=$(uname -m)
 wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${scversion?}/shellcheck-${scversion?}.linux.${arch}.tar.xz" | tar -xJv
+# TODO: Verify script integrity via checksums
```
austinvazquez, line 29: Hmm, maybe another idea to circumvent this is to run the Docker [container](https://hub.docker.com/r/koalaman/shellcheck/tags) instead. I'll try this in another thread we don't have to get it here.

pendo324: Can we generate per-architecture shasums ourselves and hardcode them in this file? At least we will know that the files don't change from what we expect

scripts/install-check-tools.sh
```diff
@@ -17,11 +17,14 @@
 echo "Install soci check tools"
 set -eux -o pipefail
 
-curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$(go env GOPATH)/bin" v1.56.2
+golangci_lint_ver="1.56.2"
+curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$(go env GOPATH)/bin" v"${golangci_lint_ver}"
+# TODO: Verify script integrity via checksums
```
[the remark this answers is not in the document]
sondavidb: Right, but we can't verify the contents of the script itself.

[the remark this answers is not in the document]
austinvazquez: Another candidate to run as a container?

scripts/install-check-tools.sh
sondavidb: Sure, did it for both of them.

scripts/install-check-tools.sh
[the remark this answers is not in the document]
sondavidb: Moved this to a GH Action and removed changes here to avoid issues w/ rebasing in the future.
https://github.com/awslabs/soci-snapshotter/pull/1145

scripts/install-dep.sh
```diff
@@ -22,33 +22,73 @@ set -eux -o pipefail
...
+    wget https://github.com/google/flatbuffers/archive/refs/tags/v"${flatc_ver}".tar.gz -O flatbuffers.tar.gz
+
+    flatc_actual_shasum="$(sha256sum flatbuffers.tar.gz | awk '{print $1}')"
+    if [ "${flatc_expected_shasum}" != "${flatc_actual_shasum}" ]
+    then
+        echo "error: flatc sha256sum did not match"
+        exit 1
+    fi
+
+    # TODO: Verify script integrity via checksums
```
austinvazquez, line 68: We can remove the TODO now.

pendo324 (approved):

austinvazquez (approved):

Issue #1084, opened 2024-02-22
[Bug] ImagePullBackoff when pulling an image from private ECR with SOCI Index being present.


debajyoti-truefoundry:
### Description

I have set up CRI-based authentication to make Soci Snapshotter work with our private ECR Registry. We use EKS.

I noticed that one of the containers was stuck in the ImagePullBackoff state. This was the event.

```
Failed to pull image "111111111.dkr.ecr.eu-west-1.amazonaws.com/tfy-truefoundry-billing-server-1egxs:75159d954d0b3aed9dfc9ca7df2654b5220f174f-77": [failed to pull and unpack image "111111111.dkr.ecr.eu-west-1.amazonaws.com/tfy-truefoundry-billing-server-1egxs:75159d954d0b3aed9dfc9ca7df2654b5220f174f-77": failed to resolve reference "111111111.dkr.ecr.eu-west-1.amazonaws.com/tfy-truefoundry-billing-server-1egxs:75159d954d0b3aed9dfc9ca7df2654b5220f174f-77": unexpected status from HEAD request to https://111111111.dkr.ecr.eu-west-1.amazonaws.com/v2/tfy-truefoundry-billing-server-1egxs/manifests/75159d954d0b3aed9dfc9ca7df2654b5220f174f-77: 403 Forbidden, rpc error: code = Unavailable desc = failed to pull and unpack image "111111111.dkr.ecr.eu-west-1.amazonaws.com/tfy-truefoundry-billing-server-1egxs:75159d954d0b3aed9dfc9ca7df2654b5220f174f-77": failed to prepare extraction snapshot "extract-639718280-mVNW sha256:e3d7a120de01b04d536f8f8479b4f942dc3a4c6c52a4070e73c39e70f5a298b0": layer "551" unavailable: unavailable]
```

These are the logs from the Containerd, Soci snapshotter and Kubelet.

```
[root@ip-10-2-45-158 /]# journalctl -u containerd | grep "extract-639718280-mVNW"
[...]

PR #1147, opened 2024-03-29
Check connection only when image isn't fully cached

### LINKED ISSUE #1116: [FEATURE] Don't make registry calls when content has completed fetching
### Description

SOCI doesn't use any sort of credential rotation system, so in systems such as Kubernetes, we can encounter issues where a user can attempt to run tasks after the credential has expired. stargz gets around this by only calling the registry if the fetched size is under the expected layer size: https://github.com/containerd/stargz-snapshotter/pull/1584. We can do the same here.

Ultimately, this is a workaround, and we need a better credentials mechanism, but this should be a simple fix that should help for now.

_No response_

_No response_

There's a separate issue here which is that we don't know exactly when the background fetcher has concluded fetching all of its data. I think if we can implement that, this fix can be a bit cleaner.

sondavidb:
Taken from stargz:
https://github.com/containerd/stargz-snapshotter/pull/1584/files

**Issue #, if available:**
Related to #1084
Fixes #1116

**Description of changes:**
SOCI does not use any form of credential rotation. Thus, it can refuse to run on pods with stale credentials unless a new image is pulled. This fix should mitigate a pain point where a user will want to use a cached image on a pod with expired credentials.

**Testing performed:**

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -641,10 +641,13 @@ func (fs *filesystem) Check(ctx context.Context, mountpoint string, labels map[s
 		return fmt.Errorf("layer not registered")
 	}
 
-	// Check the blob connectivity and try to refresh the connection on failure
-	if err := fs.check(ctx, l, labels); err != nil {
-		log.G(ctx).WithError(err).Warn("check failed")
-		return err
+	if l.Info().FetchedSize < l.Info().Size {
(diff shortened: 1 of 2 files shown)


henry118 (approved):
LGTM

haddscot (approved):

PR #1153, opened 2024-04-02
Pull container images from ECR Public instead of Docker Hub

austinvazquez:
**Issue #, if available:**
As an Amazon best practice, CI should pull container images from ECR Public instead of Docker Hub.

**Description of changes:**
This change pulls the base container images from ECR Public.

**Testing performed:**
CI is successful

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,8 +16,9 @@ ARG CONTAINERD_VERSION=1.6.30
 ARG RUNC_VERSION=1.1.12
 ARG NERDCTL_VERSION=1.7.1
 ARG GO_VERSION=1.21.8
+ARG REGISTRY_VERSION=3.0.0-alpha.1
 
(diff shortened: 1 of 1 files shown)


sondavidb (approved):
LGTM

nit: Can we also self-host the [zot image](https://github.com/project-zot/zot/releases/tag/v2.0.3)? This should remove all of our non-public ECR image dependencies. Not a blocker, though, just might be nice to lump it in this PR.

henry118 (approved):

austinvazquez:
> LGTM
> 
> nit: Can we also self-host the [zot image](https://github.com/project-zot/zot/releases/tag/v2.0.3)? This should remove all of our non-public ECR image dependencies. Not a blocker, though, just might be nice to lump it in this PR.

Looked into this. Unfortunately there does not appear to be a ECR Public image for ZOT so there might be more effort involved. Let's split this and follow-up to not block these changes.

PR #1175, opened 2024-04-11
Fix network retry integration test

### LINKED ISSUE #1039: [Bug] TestNetworkRetry is broken
### Description

Our `TestNetworkRetry` is both flaky and has some confusing logic. Most notably, it seems that iptables doesn't always go into effect before testing. We should make them more consistent

`make GO_TEST_FLAGS='-run TestNetworkRetry -count 1' integration`

Network retries are properly tested

### Host information

1. OS: AL2
2. Snapshotter Version: main
3. Containerd Version: v1.6.26

_No response_

sondavidb:
**Issue #, if available:**
Fixes #1039

**Description of changes:**
Taken from @turan18's branch, this PR fixes the TestNetworkRetry logic.

The previous test worked strangely with iptables, and for some reason did not always process them. Additionally, the checks were not robust. So, even though it was always passing, in reality the errors were not being properly handled and thus always fell through. It also operated under the assumption that certain spans (specifically the ones associated with `times` binary) were fetched with background fetcher disabled, which we found to be incorrect. The setup also seemed to behave strangely on ARM machines, not catching the error and causing it to hang indefinitely. Still haven't figured out why yet.

This change fixes the logic by streamlining the read channels and only running one command, as no binaries should be expected to be fetched on container startup. This should also stop it from hanging as all buffered channel reads are now in select statements or goroutines.

This initially had issues and stayed hanging on ARM, but changing the timeout from 5 to 6 seconds seems to make it very consistent on ARM machines.

Thanks for the work done on this @turan18 

**Testing performed:**
`GO_TEST_FLAGS='-run TestNetworkRetry -count 1' make integration` on ARM64 and AMD64

### DIFF
--- a/integration/run_test.go
+++ b/integration/run_test.go
@@ -35,20 +35,19 @@ package integration
 import (
 	"bufio"
 	"bytes"
-	"encoding/json"
 	"fmt"
-	"io"
 	"math"
 	"os"
 	"regexp"
-	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/awslabs/soci-snapshotter/soci/store"
 	shell "github.com/awslabs/soci-snapshotter/util/dockershell"
+	"github.com/awslabs/soci-snapshotter/util/testutil"
 )
 
 func TestRunWithDefaultConfig(t *testing.T) {
@@ -221,15 +220,7 @@ type retryConfig struct {
 // TestNetworkRetry runs a container, disables network access to the remote image, asks the container
 // to do something requiring the remote image, waits for some/all requests to fail, enables the network,
 // confirms retries and success/failure
-
-// TODO: Refactor test
 func TestNetworkRetry(t *testing.T) {
-	// This test makes the testing suite hang on ARM.
-	// Disabling for now till we can refactor this.
-	if runtime.GOARCH == "arm64" {
-		return
-	}
-
 	const containerImage = alpineImage
 
 	tests := []struct {
@@ -247,12 +238,12 @@ func TestNetworkRetry(t *testing.T) {
 			},
 		},
 		{
-			name: "1s network interruption, no retries allowed, failure",
+			name: "6s network interruption, no retries allowed, failure",
 			config: retryConfig{
 				maxRetries:         -1,
 				minWaitMsec:        0,
 				maxWaitMsec:        0,
-				networkDisableMsec: 1000,
+				networkDisableMsec: 6000,
 				expectedSuccess:    false,
 			},
 		},
@@ -267,12 +258,12 @@ func TestNetworkRetry(t *testing.T) {
 			},
 		},
 		{
-			name: "10s network interruption, ~6-7s retries allowed, failure",
+			name: "12s network interruption, ~6-7s retries allowed, failure",
 			config: retryConfig{
 				maxRetries:         1,
 				minWaitMsec:        100,
 				maxWaitMsec:        1600,
-				networkDisableMsec: 10000,
+				networkDisableMsec: 12000,
 				expectedSuccess:    false,
 			},
 		},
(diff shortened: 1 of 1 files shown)


sondavidb:
TL;DR I used `testutil.LogMonitor` to tie the stdout and stderr streams from `sh.R` to a new testreporter. This uses the same stdout and stderr that the logreporter uses, thus allowing `sh.R`'s open pipes to be closed and the goroutines returned.

Ok, so I went back to the drawing board a bit and I think I found a solution I'm somewhat happy with. Instead of using our own goroutines that are not tied to any context or other structure, I opted to do what we previously did and just add a LogMonitor. This LogMonitor scans the given stdout and stderr readers and matches them with the writers for stdout and stderr, respectively. This fixes the issue we had previously with `sh.R` not properly exiting due to the pipes being left open.

I still kept the logic of not processing any of the data and just getting the raw string, though, i.e. any data from stdout = success, any data from stderr = failure. It felt a lot more elegant and removed the need to unmarshal into a json and whatnot.

This change was a comparatively significant refactor, and thus I opted to put it in its own commit so we can easily revert if needed, but I plan on squashing before we merge this.

Kern-- (approved):
I think this is a good simplification of the test.

The previous test attempted to check that both 1) reading files that are already downloaded works with no network and 2) reading files that are not already downloaded may or may not work depending on the retry policy/network interruption duration.

I think trying to do those at the same time was the reason we ended up with such a flaky test, but is there still value in a follow up test that verifies that the network can be interrupted after the image is downloaded?

sondavidb:
> The previous test attempted to check that both 1) reading files that are already downloaded works with no network and 2) reading files that are not already downloaded may or may not work depending on the retry policy/network interruption duration.

It's worth noting that, before changing to `nerdctl run`, I recall this test being fine. I was confused on why `sh -c times` worked before, though. Is it because with the entrypoint it would already have `times` loaded? But what's the guarantee there? Since as noted on ARM it did hang, which makes me wonder why that would be the case. Perhaps the image difference between architectures somehow made this a problem?

> is there still value in a follow up test that verifies that the network can be interrupted after the image is downloaded?

Is this not what we're doing here though? Or do you mean fully download the image, wait for the background fetcher to finish, then run the snapshotter?

Kern--:
> It's worth noting that, before changing to nerdctl run, I recall this test being fine. I was confused on why sh -c times worked before, though. Is it because with the entrypoint it would already have times loaded? But what's the guarantee there? Since as noted on ARM it did hang, which makes me wonder why that would be the case. Perhaps the image difference between architectures somehow made this a problem?

Yeah, the old logic was that by running the container before disabling the network, we force some amount of the image to be downloaded for `sleep inf` (probably span 0 which probably includes `/bin/sh`). We never had strong guarantees that would keep working, though.

>> is there still value in a follow up test that verifies that the network can be interrupted after the image is downloaded?

> Is this not what we're doing here though? Or do you mean fully download the image, wait for the background fetcher to finish, then run the snapshotter?

The test now verifies that if we __don't__ have the data we need and we have a network interruption, there will be retries until the network comes back and we can get the data.

The test that we lost is if __do__ have the data we need, there is no network dependency and the container will work whether the network comes back or not (this is what we were trying to test with `sh -c times` although it didn't work reliably).

It doesn't have to be fully downloaded, but I think that makes it easy to make it more reliable. Assuming we had a way to determine that an image is fully downloaded, the test I'm imagining is:
1) Pull the image with the bg fetcher enabled
2) Wait for the image to be fully downloaded
3) Disable the network
4) run the container
5) Verify that the container runs successfully without requiring the network

sondavidb:
I think it's possible. We could possibly just add something like this:

```golang
// Pull image
entrypointWithNetwork := ["sh", "-c", "times"]
sh.X(append(runSociCmd, "-d", "--rm", image, entrypointWithNetwork...)...)

// Disable network access here, or return if first test case
stdout, stderr, out := sh.R(append(runSociCmd, "-d", "--rm", image, entrypointWithNetwork...)...)) // In case any regression causes this to hang
// Setup listeners for this command

// Proceed with rest of test
```

There's no reasonable case for the first test to hang IMO, and the first run prefetches the required spans so the second command can then test our theory. WDYT?

Kern--:
LGTM. I don't think that blocks this PR, so we can defer that to a follow up issue.

sondavidb:
Sure, will open that once this gets merged.

integration/run_test.go
```diff
@@ -296,135 +286,80 @@ max_wait_msec = ` + strconv.FormatInt(tt.config.maxWaitMsec, 10) + `
...
-				if err := json.Unmarshal([]byte(rawL), &logLine); err == nil {
-					if logLine.Msg == "statFile error" && strings.Contains(logLine.Error, "giving up after") {
-						gaveUpChannel <- true
+			listener := func(c chan string) func(string) {
+				return func(rawLog string) {
+					select { // Make sure this isn't blocking
+					case c <- rawLog:
+					default:
 						return
 					}
```
austinvazquez, line 320: Let's include a comment here to explain more of why we don't want to block. i.e. the channel is unbuffered and we only really care about the first message so write the message to the channel and drop everything else.

Alternatively it might be more straightforward to have a `sync.Once` for writing to the channel and close the channel when you have written the first message.

sondavidb: WDYT about this?

```golang
			listener := func(c chan string) func(string) {
				return func(rawLog string) {
					var once sync.Once
					tempFunc := func() {
						c <- rawLog
					}
					once.Do(tempFunc)
				}
			}
```

sondavidb:
Squashed commits

integration/run_test.go
```diff
@@ -296,135 +287,77 @@ max_wait_msec = ` + strconv.FormatInt(tt.config.maxWaitMsec, 10) + `
...
-				}
-				var logLine ErrorLogLine
-				if err := json.Unmarshal([]byte(rawL), &logLine); err == nil {
-					if logLine.Msg == "statFile error" && strings.Contains(logLine.Error, "giving up after") {
-						gaveUpChannel <- true
-						return
-					}
+			listener := func(c chan string) func(string) {
+				return func(rawLog string) {
+					var once sync.Once
```
austinvazquez, line 317: You would need to move this higher in scope else a new variable will be initialized when the listener return function is called. i.e.
```suggestion
listener := func(c chan string) func(string) {
        var once sync.Once
        return func(rawLog string) {
            once.Do(func() { c <- rawLog })
        }
}
```

integration/run_test.go
```diff
@@ -296,135 +287,77 @@ max_wait_msec = ` + strconv.FormatInt(tt.config.maxWaitMsec, 10) + `
...
+					once.Do(func() { c <- rawLog })
 				}
 			}
```
austinvazquez, line 320: You would need to move this higher in scope else a new variable will be initialized when the listener return function is called. i.e.
```suggestion
            listener := func(c chan string) func(string) {
                    var once sync.Once
                    return func(rawLog string) {
                        once.Do(func() { c <- rawLog })
                    }
            }
```

sondavidb: The intention here is to allow writing to both stdout and stderr in a single test case. Then, in the [select block](https://github.com/awslabs/soci-snapshotter/pull/1175/files#diff-8a6db7c58ab40add58f9046cee9ffab883d23e66c3dbc4a315dba43314c06d0bR350) further down, if one is written to unintentionally, it would get caught here and return an error.

sondavidb: Just pushed a change to make the latter checks more rigorous. Hopefully it clarifies the intention better.

Kern--: If I understand this correctly, Austin's concern is that the `sync.Once` is a new instance for every log line instead of being a single instance across all log lines.

sondavidb: In that case, would the solution still work? Under the assumption this is pass-by-value, then it definitely should, but I'm not sure of a way to confirm this.

sondavidb: idk why I said that, I just added a few print statements to check.

Anything before the return statement always runs. Anything within it is run every time we get a new message. So Austin was correct. Made the changes Austin recommended.

Kern-- (approved):

austinvazquez (approved):

sondavidb:
> I think it's possible. We could possibly just add something like this:
> 
> 
> 
> ```golang
> 
> // Pull image
> 
> entrypointWithNetwork := ["sh", "-c", "times"]
> 
> sh.X(append(runSociCmd, "-d", "--rm", image, entrypointWithNetwork...)...)
> 
> 
> 
> // Disable network access here, or return if first test case
> 
> stdout, stderr, out := sh.R(append(runSociCmd, "-d", "--rm", image, entrypointWithNetwork...)...)) // In case any regression causes this to hang
> 
> // Setup listeners for this command
> 
> 
> 
> // Proceed with rest of test
> 
> ```
> 
> 
> 
> There's no reasonable case for the first test to hang IMO, and the first run prefetches the required spans so the second command can then test our theory. WDYT?

Opened #1231 to track this

PR #1181, opened 2024-04-13
Add timeouts to testing in CI

austinvazquez:
**Issue #, if available:**
When making the test container to be Alpine based, the GitHub Action workflow ran for 6+ hours due to tests getting stuck.
https://github.com/awslabs/soci-snapshotter/actions/runs/8671778098

**Description of changes:**
This change adds a 15 and 30 minute timeout to unit and integration tests respectively to prevent wasting compute resources.

**Testing performed:**
CI is successful

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,6 +25,7 @@ env:
 jobs:
   test:
     runs-on: ubuntu-20.04
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
@@ -34,6 +35,7 @@ jobs:
       - run: make test
   integration:
     runs-on: ubuntu-20.04
+    timeout-minutes: 40
     strategy:
       fail-fast: false
       matrix:


.github/workflows/build.yml
```diff
@@ -34,6 +35,7 @@ jobs:
       - run: make test
   integration:
     runs-on: ubuntu-20.04
+    timeout-minutes: 30
```
sondavidb, line 38: I think I've seen integration tests take 35 minutes occasionally. Can we do something like 40?

sondavidb:
Do we want all of the other workflows to have a timeout as well? I can't see a reason for them to ever hang, but may not be a bad idea

> the GitHub Action workflow ran for 6+ hours due to tests getting stuck

FWIW, #1175 should fix the hanging issue.

austinvazquez:
@sondavidb, updated. PTAL.

sondavidb (approved):
LGTM, maybe we can consider unit tests to be shorter (~10min)?

pendo324 (approved):

PR #1196, opened 2024-04-24
Trigger Docker auth on ECR token expiry

### LINKED ISSUE #1148: [Bug] ECR reauthentication on 403 does not seem to work
### Description

I'm having trouble reproducing since the expiry time is 12h for an ECR token, but even with the [amazon-ecr-credential-helper](https://github.com/awslabs/amazon-ecr-credential-helper) we're seeing that after the authorization expires it's never refreshed.

These are the logs in question.
```
Mar 28 17:20:52 ip-10-0-1-2.us-west-2.compute.internal soci-snapshotter-grpc[690]: {"level":"info","msg":"Received status code: 403 Forbidden. Authorizing...","time":"2024-03-28T17:20:52.858737822Z"}
Mar 28 17:20:52 ip-10-0-1-2.us-west-2.compute.internal soci-snapshotter-grpc[690]: {"error":"failed to refresh URL: status 403 Forbidden","key":"sha256:[sha256]","level":"warning","mountpoint":"/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/51/fs","msg":"failed to connect to blob","time":"2024-03-28T17:20:52.859196457Z"}
Mar 28 17:20:52 ip-10-0-1-2.us-west-2.compute.internal soci-snapshotter-grpc[690]: {"error":"unable to create remote fetcher: failed to redirect: request to registry failed: unknown \"unknown\": giving up request after 1 attempt(s): context canceled (host \"0123456789012.dkr.ecr.us-west-2.amazonaws.com\", ref:\"0123456789012.dkr.ecr.us-west-2.amazonaws.com/repo/image:tag\", digest:\"sha256:[sha256]\")","key":"sha256:[sha256]","level":"warning","mountpoint":"/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/52/fs","msg":"failed to refresh the layer \"sha256:[sha256]\" from \"0123456789012.dkr.ecr.us-west-2.amazonaws.com/repo/image:tag\"","time":"2024-03-28T17:20:52.858771180Z"}
```
(same instance, different occurrence)
```
Mar 28 15:41:12 ip-10-0-1-2.us-west-2.compute.internal soci-snapshotter-grpc[690]: {"error":"unable to create remote fetcher: failed to redirect: request to registry failed: failed to handle challenge: failed to find supported auth scheme: not implemented (host \"0123456789012.dkr.ecr.us-west-2.amazonaws.com\", ref:\"0123456789012.dkr.ecr.us-west-2.amazonaws.com/repo/image:tag\", digest:\"sha256:[sha256]\")","key":"sha256:[sha256]","level":"warning","mountpoint":"/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/79/fs","msg":"failed to refresh the layer \"sha256:[sha256]\" from \"0123456789012.dkr.ecr.us-west-2.amazonaws.com/repo/image:tag\"","time":"2024-03-28T15:41:12.206163948Z"}
```

Restarting `soci-snapshotter.service` fixes the issue. Pods are running with `pullPolicy: Always`.

_No response_

After 12 hours, soci-snapshotter is still able to pull layers (images?) from ECR.

### Host information

```
# uname -a
Linux [snip] 6.5.0-1016-aws #16~22.04.1-Ubuntu SMP Wed Mar 13 18:54:49 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
# containerd --version
containerd github.com/containerd/containerd v1.7.13 7c3aca7a610df76212171d200ca3811ff6096eb8
# soci-snapshotter-grpc --version
soci-snapshotter-grpc version v0.5.0 77f218d0b11c88d17a832ad7456dbd2165d19c86
# kubelet --version
Kubernetes v1.28.5-eks-5e0fdde
```

https://github.com/awslabs/soci-snapshotter/pull/945 appears to be an attempt to fix this, but I don't think AWS responds with the `WWW-Authenticate` header necessary to trigger re-authorization [in containerd](https://github.com/containerd/containerd/blob/release/1.7/remotes/docker/auth/parse.go#L100).

sondavidb:
**Issue #, if available:**
Fixes #1148 

**Description of changes:**
An expired token passed to ECR will return a 403, which lacks a Www-Authenticate header required to trigger the docker authorizer. This meant that credential helpers like amazon-ecr-credential-helper would not refresh the token. This change adds the proper header to fix this behavior.

Thanks @steved for starting work on the fix!

**Testing performed:**
Tested on an EC2 instance with SOCI and amazon-ecr-credential-helper and confirmed that images could still be pulled after 12 hours.

### DIFF
--- a/service/resolver/client.go
+++ b/service/resolver/client.go
@@ -250,6 +250,26 @@ func shouldAuthenticate(resp *http.Response) bool {
 		for _, e := range errs {
 			if err, ok := e.(docker.Error); ok {
 				if err.Message == ecrTokenExpiredResponseMessage {
+					// ECR's 403 doesn't return a Www-Authenticate and so doesn't trigger the
+					// basic re-authentication in containerd's docker authorizer.
+					// Ideally ECR would return the Www-Authenticate for expired tokens,
+					// but until then we'll have to use this workaround.
+					if resp.Header == nil {
+						resp.Header = map[string][]string{}
+					}
+
+					authenticateHeader := http.CanonicalHeaderKey("Www-Authenticate")
+					if _, exists := resp.Header[authenticateHeader]; !exists {
+						realm := ""
+						if resp.Request != nil {
+							realm = "https://" + resp.Request.URL.Host + "/"
+						}
+						service := "ecr.amazonaws.com"
+
+						authHeaderContent := fmt.Sprintf("Basic realm=\"%s\",service=\"%s\"", realm, service)
+						resp.Header[authenticateHeader] = []string{authHeaderContent}
+					}
+
 					return true
 				}
 			}
--- a/service/resolver/client_test.go
+++ b/service/resolver/client_test.go
@@ -167,9 +167,10 @@ func TestAuthentication(t *testing.T) {
 	normalForbiddenResponse, _ := docker.Errors([]error{docker.ErrorCodeDenied}).MarshalJSON()
 	unauthorizedResponse, _ := docker.Errors([]error{docker.ErrorCodeUnauthorized}).MarshalJSON()
 	testCases := []struct {
-		name        string
-		performAuth bool
-		response    *http.Response
+		name               string
+		performAuth        bool
+		response           *http.Response
+		expectedAuthHeader string
 	}{
 		{
 			name:        "Authenticate on 403 with token expiry.",
@@ -178,6 +179,7 @@ func TestAuthentication(t *testing.T) {
 				StatusCode: http.StatusForbidden,
 				Body:       io.NopCloser(bytes.NewReader(ecrForbiddenResponse)),
 			},
+			expectedAuthHeader: "Basic realm=\"\",service=\"ecr.amazonaws.com\"",
 		},
 		{
 			name:        "Do not authenticate on 403 without token expiry.",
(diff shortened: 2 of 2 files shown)


sondavidb:
Rebased

service/resolver/client.go
```diff
@@ -250,6 +250,21 @@ func shouldAuthenticate(resp *http.Response) bool {
...
 				if err.Message == ecrTokenExpiredResponseMessage {
+					// ECR's 403 doesn't return a Www-Authenticate and so doesn't trigger the
+					// basic re-authentication in containerd's docker authorizer
+					authenticateHeader := http.CanonicalHeaderKey("Www-Authenticate")
+					if resp.Header == nil {
+						resp.Header = map[string][]string{}
+					}
+					if _, exists := resp.Header[authenticateHeader]; !exists {
+						realm := ""
+						if resp.Request != nil {
```
sondavidb, line 264: This check was added due to our [TestAuthentication](https://github.com/awslabs/soci-snapshotter/blob/dfcd5e5fd22bf0e78fb6b666b522613964a3f933/service/resolver/client_test.go#L165) unit test [failing](https://github.com/awslabs/soci-snapshotter/actions/runs/8824348550/job/24226633471) without it. Should we make sure that `resp.Request.URL` is nil as well? I would imagine a request would always have URL populated but curious if we want to be a little more vigilant.

If the above situation is possible, this means that realm is sometimes empty, but it seems that [containerd docker authorizer](https://github.com/containerd/containerd/blob/01ed3ff1233f4179bbfecac3c81d9e6c6555a44b/core/remotes/docker/authorizer.go#L186-L200) only checks the scheme, and some digging seems to show that service and realm are only used in the context of bearer auth, not basic auth. Which is to say, we could theoretically change the header to be `"Basic realm=\"\",service=\"\""` and it should still work fine. But, it's probably better practice to craft the header properly to cover most, if not all cases.

Kern-- (approved):

austinvazquez (commented):
Can you add a unit test to validate the authentication header is added properly?

austinvazquez (approved):

coderbirju (approved):

PR #1233, opened 2024-05-13
Add integration test for offline content

### LINKED ISSUE #1231: [FEATURE] Add test for container starting without internet access
Followup to #1175, which removed this test, we want to make sure we can run a container without internet access if we have the required files.

sondavidb:
**Issue #, if available:**
Fixes #1231

**Description of changes:**
This basically copies a lot of the new TestNetworkRetry's code to pull an image, run a container to access a certain span, disable network requests, then re-run the command to ensure it passes.

We can probably put this inside of TestNetworkRetry, but the test is already pretty bloated with actions, but then this code ends up being a lot of copy-paste from that, so maybe it's better to do the former for better code maintenance? Unsure what the right move here is TBH. 

**Testing performed:**
`make integration`

### DIFF
--- a/integration/run_test.go
+++ b/integration/run_test.go
@@ -36,7 +36,6 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
-	"math"
 	"os"
 	"regexp"
 	"strconv"
@@ -267,6 +266,16 @@ func TestNetworkRetry(t *testing.T) {
 				expectedSuccess:    false,
 			},
 		},
+		{
+			name: "Permanent network interruption after loading content, success",
+			config: retryConfig{
+				maxRetries:         -1,
+				minWaitMsec:        0,
+				maxWaitMsec:        0,
+				networkDisableMsec: -1,
+				expectedSuccess:    true,
+			},
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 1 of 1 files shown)


sondavidb:
Rebased

sondavidb:
Ended up merging with TestNetworkRetry as an extra test case and added a couple extra conditions. A bit clunky but prevents a lot of code copy-paste.

integration/run_test.go
```diff
@@ -333,12 +348,11 @@ disable = true
...
-			// retrying.
-			time.Sleep(time.Duration(tt.config.networkDisableMsec) * time.Millisecond)
-
 			// Restore network access
 			if tt.config.networkDisableMsec > 0 {
+				// Wait with network disabled. At this point the snapshotter should be
+				// retrying.
+				time.Sleep(time.Duration(tt.config.networkDisableMsec) * time.Millisecond)
 				// Restore access to the registry and image
 				sh.X("iptables", "-D", "OUTPUT", "-d", registryHostIP, "-j", "DROP")
```
Kern--, line 357: Does skipping this make our tests order dependent? If this test doesn't run last, do the next tests not have access to the registry?

sondavidb: No, since every test runs with a new container. But, I think this is a valid concern anyway, so I did a bit of refactoring to deterministically restore network access at the end of every test.

Kern-- (approved):

integration/run_test.go
```diff
@@ -267,6 +267,16 @@ func TestNetworkRetry(t *testing.T) {
 				expectedSuccess:    false,
 			},
 		},
+		{
+			name: "Permanent network interruption after loading content, success",
```
austinvazquez, line 271: Does this depend on `No network interruption` test case to be ran first? Generally I would prefer test cases to be order agnostic.

sondavidb: These tests are each run in their [own container](https://github.com/awslabs/soci-snapshotter/pull/1233/files#diff-8a6db7c58ab40add58f9046cee9ffab883d23e66c3dbc4a315dba43314c06d0bR285), so order doesn't matter.

austinvazquez: Oh , I understand the permanent network interruption test runs the container to fetch the required spans.

sondavidb: Yeah, the container gets run [here](https://github.com/awslabs/soci-snapshotter/pull/1233/files#diff-8a6db7c58ab40add58f9046cee9ffab883d23e66c3dbc4a315dba43314c06d0bR315) for this use case so that when it goes offline it will have the content loaded.

Kern-- (approved):

austinvazquez (approved):

PR #1238, opened 2024-05-16
Add release automation to update getting started guide

austinvazquez:
**Issue #, if available:**
Updating the version in the getting started guide is currently a manual step for every release. It would be nice to have automation to generate the PR so maintainers only have to approve and merge the pull request.

**Description of changes:**
This change adds a workflow to open a pull request to update the version in the getting started guide when a release is made.

**Testing performed:**
Pushed workflow to fork and created a release tag to show workflow opened a PR to update the getting started version.

Automated version PR - https://github.com/austinvazquez/soci-snapshotter/pull/29
Release used to trigger automation - https://github.com/austinvazquez/soci-snapshotter/releases/tag/v0.6.1-alpha.0
GitHub Action workflow run - https://github.com/austinvazquez/soci-snapshotter/actions/runs/9208357812

Mock workflow runs in pull request on changes to GitHub Actions workflow file and asserts the changes are made to update the version.

**Alternatives considered:**
1. An approach which creates the PR once the release tag is pushed. i.e. the existing release workflow
a. The trade-off is the release artifacts will not yet be published. Although this is mitigated because the PR still requires manual approval and publishing a tag is a one-way door, to avoid confusion it would be good to only update the getting started version once the artifact is publicly available.

**Limitations:**
1. Similar to the existing update dependencies automation, the PR must be closed and reopened for workflow checks to run.


.github/workflows/update-getting-started-guide.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Create PR
+        if: github.event_name == 'release'
+        uses: peter-evans/create-pull-request@v6
+        with:
+          title: 'Update getting started to ${{ github.event.release.tag_name }}'
+          commit-message: 'Update getting started to ${{ github.event.release.tag_name }}'
+          body: |
+            This PR must be closed and reopened manually to trigger automated checks.
+
+            Auto-generated by [create-pull-request][1].
```
pendo324, line 47: Can't this just be markdown? Or is this going to be added as a commit message upon merge as well?

.github/workflows/update-getting-started-guide.yml
```diff
@@ -0,0 +1,56 @@
...
+jobs:
+  update-version:
+    runs-on: ubuntu-20.04
+    env:
+      RELEASE_TAG: ''
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Mock release tag on pull request
+        if: github.event_name == 'pull_request'
```
pendo324, line 24: Why do we need to run this on pull requests?

austinvazquez: Enables spot checking of any changes to the workflow in pull request. It won't catch every issue, but it can give some confidence.

.github/workflows/update-getting-started-guide.yml
austinvazquez: Yeah, this can be markdown. It is not included in the commit message.

sondavidb:
IMO it wouldn't be a bad idea to also include automation to update third-party licenses in this workflow. This will get run on patch versions and patch versions can potentially have different versions which could potentially have different licenses. Definitely not likely at all but a possibility that could be a problem if not caught IMO.

.github/workflows/update-getting-started-guide.yml
sondavidb: My understanding is that, on PR, it will create a dummy tag, replace the tag with the dummy tag, and diff check it, but ultimately not fail for any particular reason. What errors are we expecting this to catch?

austinvazquez: e.g. changes to the getting started guide layout which would break the update workflow.

austinvazquez: Added run on pull request to docs/getting-started.md in latest commit. PTAL.

austinvazquez:
@pendo324, @sondavidb, I have added an assertion which is ran on pull request to the workflow and docs/getting-started.md to validate the update behaves correctly. Please take a look and lmk your thoughts.

pendo324 (approved):

.github/workflows/update-getting-started-guide.yml
```diff
@@ -0,0 +1,74 @@
...
+          release_tag=${{ env.RELEASE_TAG }}
+          release_version=${release_tag/v/}
+          sed -i -E "s/version=\"([0-9]+\.){2}[0-9]+\"/version=\"${release_version}\"/" docs/getting-started.md
+
+      - name: Assert version was updated correctly
+        if: github.event_name == 'pull_request'
+        run: |
+          release_tag=${{ env.RELEASE_TAG }}
+          release_version=${release_tag/v/}
+          output=$(git diff --exit-code) && {
```
Kern--, line 47: Can we move this to a script?

I think it's easier to discover, tests, and understand if the github action yaml only interacts with github env and then calls a script.

austinvazquez: Sure. Normally we through scripts into `scripts/`; in this case the script is an assertion to validate GitHub Actions changes so it might not be useful into many more contexts. Would `.github/scripts` be more appropriate here?

sondavidb (dismissed):
LGTM, cool to automate another part of our release process

austinvazquez:
@sondavidb, apologies I was doing some permissions testing in another branch and somehow mixed up the changes in this branch. Added permission `contents: write` back to this PR.

pendo324 (dismissed):

.github/actions/update-getting-started/action.yml
```diff
@@ -0,0 +1,17 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: "Update getting started version"
+      shell: bash
+      run: |
+        version=${{ inputs.version }}
+        version=${version/v/} # strip prefix v if not already stripped
+        sed -i -E "s/version=\"([0-9]+\.){2}[0-9]+\"/version=\"${version}\"/" docs/getting-started.md
```
Kern--, line 16: Sorry, I don't think I was clear before. This is the part that I think would be useful in a script.

For example, if need to update this script for some reason (e.g. we add another reference to the version, say a link to the specific release) then right now we can't test the change without pushing to a fork.

IMO the actions yaml should just move inputs from the github env into variables and then call scripts that do the actual work.

austinvazquez: @Kern-- , oh yeah I misunderstood. Pushed a new version to address this. PTAL. Good callout though, this simplifies what I wanted. Thanks!

Kern-- (approved):

sondavidb (approved):

PR #1232, opened 2024-05-13
update containerd version to 1.7.16

coderbirju:
**Issue #, if available:**

**Description of changes:**

**Testing performed:**

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1244, opened 2024-05-21
Enforce lease when building SOCI index

sondavidb:
**Issue #, if available:**

**Description of changes:**
Previously, in our CLI create command, we would create zTOCs, push them to the content store, then label the index to refer to the zTOCs. This created a problem where a user using the containerd content store might have their zTOCs deleted by the containerd garbage collector before it could be labeled. While a lease was used, it was only done during the latter step, so it did not prevent zTOCs from being deleted before being labeled.

This commit fixes this by using a lease on the entire process via a new command, BuildAndWriteIndex, which is called by "soci create".

This should unblock #1232 as well.

**Testing performed:**

### DIFF
--- a/integration/create_test.go
+++ b/integration/create_test.go
@@ -19,6 +19,7 @@ package integration
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"path/filepath"
 	"testing"
 
@@ -200,5 +201,34 @@ func TestSociCreate(t *testing.T) {
 				t.Fatalf("failed to validate soci index: %v", err)
 			}
 		})
+	}
+}
+
+// TestSociCreateGarbageCollection ensures that SOCI index artifacts are not garbage collected after creation.
+func TestSociCreateGarbageCollection(t *testing.T) {
+	image := rabbitmqImage
+	smallImage := alpineImage
+	sh, done := newSnapshotterBaseShell(t)
+	defer done()
+
+	extraContainerdConfig := `
+[plugins."io.containerd.gc.v1.scheduler"]
+	deletion_threshold = 1`
+
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false, extraContainerdConfig), getSnapshotterConfigToml(t, false, tcpMetricsConfig, GetContentStoreConfigToml(store.WithType(config.ContainerdContentStoreType))))
+
+	imgInfo := dockerhub(image)
+	sh.X("nerdctl", "pull", "-q", imgInfo.ref)
+	indexDigest := buildIndex(sh, imgInfo, withMinLayerSize(0), withContentStoreType(config.ContainerdContentStoreType))
+
+	// Pull and remove a small image to automatically trigger GC.
+	smallImageInfo := dockerhub(smallImage)
+	sh.X("nerdctl", "pull", "-q", smallImageInfo.ref)
+	sh.X("nerdctl", "rmi", smallImageInfo.ref)
+
+	sh.X(append(imagePullCmd, "--soci-index-digest", indexDigest, imgInfo.ref)...)
+	curlOutput := string(sh.O("curl", tcpMetricsAddress+metricsPath))
+	if err := checkOverlayFallbackCount(curlOutput, 0); err != nil {
+		t.Fatal(fmt.Errorf("resources unexpectedly garbage collected: %v", err))
 	}
 }
--- a/soci/soci_index.go
+++ b/soci/soci_index.go
@@ -39,7 +39,6 @@ import (
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 
-	orascontent "oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/errdef"
 )
 
@@ -273,14 +272,14 @@ func WithArtifactsDb(db *ArtifactsDb) BuildOption {
 // IndexBuilder creates soci indices.
 type IndexBuilder struct {
 	contentStore content.Store
-	blobStore    orascontent.Storage
+	blobStore    store.Store
 	ArtifactsDb  *ArtifactsDb
 	config       *buildConfig
 	ztocBuilder  *ztoc.Builder
 }
 
 // NewIndexBuilder returns an `IndexBuilder` that is used to create soci indices.
-func NewIndexBuilder(contentStore content.Store, blobStore orascontent.Storage, artifactsDb *ArtifactsDb, opts ...BuildOption) (*IndexBuilder, error) {
+func NewIndexBuilder(contentStore content.Store, blobStore store.Store, artifactsDb *ArtifactsDb, opts ...BuildOption) (*IndexBuilder, error) {
 	defaultPlatform := platforms.DefaultSpec()
 	config := &buildConfig{
 		spanSize:            defaultSpanSize,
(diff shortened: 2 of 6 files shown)


sondavidb:
Worth noting here — in my testing, I tried to have the second test case ("image with some layers not having ztoc and no fs.Mount results in 0 overlay fallback with containerd content store") fail, but could never get it to fail, but when I lowered the min layer size so that 4 zTOCs were created instead of 3, it would fail intermittently.

I also thought that the initial 3 layers used in the default index config were never being garbage collected during testing, but I haven't been able to prove that, as I could never reproduce it after resetting my test environment. Potentially worth looking into.

soci/soci_index.go
```diff
@@ -305,18 +305,42 @@ func NewIndexBuilder(contentStore content.Store, blobStore orascontent.Storage,
 	}, nil
 }
 
+// BuildAndPush builds a SOCI index, pushes and labels the artifacts, and returns the SOCI index.
+func (b *IndexBuilder) BuildAndPush(ctx context.Context, blobStore store.Store, img images.Image) (*IndexWithMetadata, error) {
```
austinvazquez, line 309: It is a little odd to me the Builder interface is changing to accommodate a behavior specific to one of the blob store implementations. Can the handling of the lease be pushed into the containerd blob store?

sondavidb: Honestly the entire building interface is super weird IMO. Why are we building an index and pushing zTOCs before labeling them? The flow naturally leads to race conditions and so this was the best way I could think of without more refactoring work. But, I could put in more time to refactor it deeper, it just seemed unnecessary.

> Can the handling of the lease be pushed into the containerd blob store?

IMO it shouldn't since, while SOCI content store does not garbage collect, we could theoretically support it in the future, and in fact it might be something we want to pursue, since leftover artifacts not being cleaned up on some operations (e.g. image removal with a SOCI image only removes image manifests) can lead to unexpected behaviors.

soci/soci_index.go
```diff
@@ -305,18 +305,42 @@ func NewIndexBuilder(contentStore content.Store, blobStore orascontent.Storage,
...
+	return sociIndexWithMetadata, nil
+}
+
 // Build builds a soci index for `img` and return the index with metadata.
-func (b *IndexBuilder) Build(ctx context.Context, img images.Image) (*IndexWithMetadata, error) {
+// To prevent garbage collection of the zTOCs pushed by this command,
+// we obtain a lease from containerd and also return it.
+// It is the responsibility of the caller to terminate the lease.
+// Ideally the lease ends after labelling and pushing the index and related zTOCs.
+func (b *IndexBuilder) Build(ctx context.Context, blobStore store.Store, img images.Image) (*IndexWithMetadata, store.CleanupFunc, error) {
```
Kern--, line 328: This feels like a leaky abstraction. We create the bach here, but pass up the cleanup to the caller because there's more work to be done. Shouldn't the caller be the one to create the batch?

sondavidb:
Went back to the drawing board a bit, TL;DR agree with above that the API changes were strange. We essentially leave users with the option of "the right way" and "the wrong way" which doesn't make sense, so we changed the API back to its original state, but now it both builds, pushes, and labels in one go, with a lease.

I also added an integration test, but TBH I'm not a huge fan of it because it passes in v1.7.15 without my changes and fails in v1.7.16 without my changes. Ideally, it should fail both. The previous issue stemmed from the garbage collector and time between pushing and labeling being racy, and I can't think of a good way to simulate that other than injecting a sleep between pushing and labeling, then forcing garbage collection, but then of course that's going to fail.

This is all to say, this integration test is kind of our best bet — ensure that labels are done properly, so that the containerd content store will not garbage collect us after the operation is entirely done. It doesn't really do what I'd like it to do, but it's the best I could think of.

sondavidb:
Well, that was an annoying edit. Had to fix the unit tests (forgot to run before updating branch, whoops...) and that was a lot of tedious copy-pasting, but CI should pass now fingers crossed.

soci/soci_index.go
```diff
@@ -305,18 +305,42 @@ func NewIndexBuilder(contentStore content.Store, blobStore orascontent.Storage,
...
 
+// BuildAndPush builds a SOCI index, pushes and labels the artifacts, and returns the SOCI index.
+func (b *IndexBuilder) BuildAndPush(ctx context.Context, blobStore store.Store, img images.Image) (*IndexWithMetadata, error) {
+	sociIndexWithMetadata, done, err := b.Build(ctx, blobStore, img)
+	if err != nil {
+		return nil, err
+	}
+	defer done(ctx)
+
+	err = WriteSociIndex(ctx, sociIndexWithMetadata, blobStore, b.ArtifactsDb)
```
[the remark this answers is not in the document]
sondavidb: Changed so the function does not use a batch

soci/soci_index.go
sondavidb: Batch no longer gets returned

soci/soci_index.go
```diff
@@ -305,8 +304,31 @@ func NewIndexBuilder(contentStore content.Store, blobStore orascontent.Storage,
 	}, nil
 }
 
-// Build builds a soci index for `img` and return the index with metadata.
+// Build builds a soci index for `img`, pushes it, and returns the index with metadata.
 func (b *IndexBuilder) Build(ctx context.Context, img images.Image) (*IndexWithMetadata, error) {
```
austinvazquez, line 308: Let's be explicit here where the index is pushed. i.e. ... for `img`, stores the index in the blob store, and returns ...

Also the index return value is no longer needed by the command line with this change. Should we go ahead and refactor to remove it?

soci/soci_index.go
```diff
@@ -305,8 +304,31 @@ func NewIndexBuilder(contentStore content.Store, blobStore orascontent.Storage,
...
+	// batch will prevent content from being garbage collected in the middle of the following operations
+	ctx, done, err := b.blobStore.BatchOpen(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer done(ctx)
+
+	// Create and push zTOCs to content store
```
austinvazquez, line 316: blob store*

soci/soci_index.go
```diff
@@ -542,14 +563,7 @@ func GetImageManifestDescriptor(ctx context.Context, cs content.Store, imageTarg
 }
 
 // WriteSociIndex writes the SociIndex manifest to oras `store`.
```
austinvazquez, line 565: Let's go ahead and update this comment while the hood is open on this bit of code. It should match the function name and also state the write is to the blob store.

austinvazquez (commented):
The fix looks good. Just had a few comments on cleaning up some comments.

soci/soci_index.go
sondavidb: There's no other way to get the digest/info for a SOCI index if we don't return it here. I can imagine it might be useful for those using our library, and they can always choose to ignore it.

austinvazquez (approved):
Changes LGTM. Thanks for pushing this through. +1 for adding the integration test!

coderbirju (dismissed):

soci/soci_index.go
```diff
@@ -305,8 +304,31 @@ func NewIndexBuilder(contentStore content.Store, blobStore orascontent.Storage,
...
+	// Label zTOCs and push SOCI index
+	err = b.writeSociIndex(ctx, index)
+	if err != nil {
+		return nil, err
+	}
+
+	return index, nil
+}
+
+func (b *IndexBuilder) build(ctx context.Context, img images.Image) (*IndexWithMetadata, error) {
```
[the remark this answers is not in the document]
sondavidb: Added doc comments around the board to warn people to call funcs w/ a batch

coderbirju (approved):

integration/create_test.go
```diff
@@ -202,3 +203,32 @@ func TestSociCreate(t *testing.T) {
 		})
 	}
 }
+
+// TestSociCreateGarbageCollection ensures that SOCI index artifacts are not garbage collected after creation.
+func TestSociCreateGarbageCollection(t *testing.T) {
```
Kern--, line 208: Does this fail consistently without the changes? I wonder if this test passing isn't a strong indicator that the fix is working.

Kern-- (approved):

integration/create_test.go
sondavidb: It doesn't, and I really don't like that it doesn't.

This test more for a failsafe in case we move around some function calls and accidentally forget to label some artifacts and never notice because we don't test much in the containerd content store. I don't think there's a good way to test this fix works, though. The only way we can make sure it doesn't fail is if we call private commands in a unit test, but then of course if you call garbage collection on unlabeled zTOCs it will get deleted, so it seemed like a silly idea. This was just the best I could think of that was adjacently related.

PR #1253, opened 2024-05-29
Add workflow to label pull requests

austinvazquez:
**Issue #, if available:**
Offline discussion from #1238. Risk for malicious actors contributing code to run on AWS's GitHub Action compute is mitigated by requiring maintainer approval to run workflows; however, this requires changes to GitHub Actions to be monitored by maintainers for malicious changes to workflows with write permissions. It would be helpful to auto label pull requests with `github_actions` for maintainers to more identify the changes need to be thoroughly examined for malicious contributions.

**Description of changes:**
This change adds a GitHub Actions workflow to label new pull requests with the following labels:

- `documentation` - updates to `**/*.md`
- `go` - updates to `**/*.go`
- `testing` - updates to `**/*_test.go`
- `github_actions` - updates to `.github/`
- `benchmarking` - updates to `benchmark/`
- `dependencies` - updates to `go.(mod|sum)`

**Testing performed:**

**Limitations:**
1. Because the new pull requests workflow depends on mainline configuration for labels to avoid checking out code from the pull request branch, a risk mitigation for pwn pull requests, the workflow will not run on the contributing pull request which adds the workflow.

**Alternatives considered:**
1. Variations to the label logic:
a. Requiring all files in a glob to match a label instead of any file.

### DIFF
--- a/.github/new-pull-request-labels.yml
+++ b/.github/new-pull-request-labels.yml
@@ -1,0 +1,22 @@
+documentation:
+  - '**/*.md'
+
+github_actions:
+  - .github/**
+  - scripts/**
+
+go:
+  - '**/*.go'
+
+testing:
+  - integration/**
+  - '**/*_test.go'
+
+benchmarking:
+  - benchmark/**
+
+dependencies:
+  - cmd/go.mod
+  - cmd.go.sum
+  - go.mod
+  - go.sum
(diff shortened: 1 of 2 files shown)


sondavidb:
> `github_actions` - updates to `.github/`

I think the label for this should be a bit more explicit in ensuring that we are more careful with these PRs. Perhaps an extra label like `needs-additional-review` for external contributors? (Is that even possible?)

I guess my only concern is that we might forget what the label means down the line, so we should just be a little more explicit on which parts need some scrutiny.

sondavidb (approved):

PR #1254, opened 2024-05-29
Add codebuild integrations for our workflows

coderbirju:
Requires #1255 
This PR adds the capability to run our workloads on codebuild instances. This allows us to manage the instance size/performance in a consistent manner. It also allows the use of ARM runners which will enable us to automate the release process for arm binaries in the future.

~~Running the Integration tests on both GH runners and codebuild to make sure we are able to identify any potential kernel specific issues.~~

_Don't see any compelling reasons to run the integration tests on both github and codebuild and have kept the fork runs exclusively on github. If there are any kernel issues then we will see it when we make updates and run these tests in the fork._

We now use Ubuntu 22.04 as the underlying OS for all our workloads

**Issue #, if available:**

**Description of changes:**

**Testing performed:**

### DIFF
--- a/.github/workflows/benchmark_visualization.yml
+++ b/.github/workflows/benchmark_visualization.yml
@@ -17,7 +17,9 @@ env:
 
 jobs:
   benchmark:
-    runs-on: codebuild-soci-visualization-workflow-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-7.0-xlarge
+    # skip running benchmarks in forks
+    if: github.repository == 'awslabs/soci-snapshotter'
+    runs-on: codebuild-soci-snapshotter-benchmark-visualization-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-7.0-xlarge
     steps:
       - name: Checkout main branch
         uses: actions/checkout@v4
@@ -37,7 +39,8 @@ jobs:
           path: ${{github.workspace}}/benchmark/performanceTest/output/results.json
 
   download-and-convert-benchmark-result-to-visualization-data:
-    runs-on: codebuild-soci-visualization-workflow-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-7.0-xlarge
+    if: github.repository == 'awslabs/soci-snapshotter'
+    runs-on: codebuild-soci-snapshotter-benchmark-visualization-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-7.0-xlarge
     needs: benchmark
     steps:
       - name: Checkout main branch
@@ -66,8 +69,9 @@ jobs:
       matrix: ${{ steps.set-matrix.outputs.files }}
 
   push-benchmark-result-gh-pages:
+    if: github.repository == 'awslabs/soci-snapshotter'
     name: Push benchmark result to Github-pages
-    runs-on: codebuild-soci-visualization-workflow-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-7.0-xlarge
+    runs-on: codebuild-soci-snapshotter-benchmark-visualization-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-7.0-xlarge
     needs: download-and-convert-benchmark-result-to-visualization-data
     strategy:
       matrix:
--- a/.github/workflows/benchmark_visualization.yml
+++ b/.github/workflows/benchmark_visualization.yml
@@ -17,7 +17,7 @@ env:
 
 jobs:
   benchmark:
-    runs-on: ubuntu-20.04
+    runs-on: codebuild-soci-visualization-workflow-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-7.0-xlarge
     steps:
       - name: Checkout main branch
         uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


.github/workflows/bump-deps.yml
austinvazquez, line 1: Is there benefit to running this workflow on a CodeBuild runner?

.github/workflows/prebuild.yml
austinvazquez, line 1: Do any of these jobs benefit from running on CodeBuild runners?

austinvazquez:
Does using the CodeBuild runner for a workflow prevent the workflow from running in developer forks? If so, we should document that as a limitation for this approach.

.github/workflows/releases.yml
```diff
@@ -20,14 +20,17 @@ permissions:
...
     env:
       # Set during setup.
       RELEASE_TAG: ''
       DYNAMIC_BINARY_NAME: ''
       STATIC_BINARY_NAME: ''
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
```
sondavidb, line 33: Do we still need Go here?

.github/workflows/benchmark_visualization.yml
```diff
@@ -17,7 +17,7 @@ env:
 
 jobs:
   benchmark:
-    runs-on: ubuntu-20.04
+    runs-on: codebuild-soci-visualization-workflow-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-7.0-xlarge
```
austinvazquez, line 20: Jobs which run on CodeBuild runner would not be able to run in developers forks. It might be good to add a conditional for repository is `awslabs/soci-snapshotter`

sondavidb (commented):
+1 on both of Austin's comments, IMO the default should be to run on GH Actions runners, and only run tests on CodeBuild if we have a reason.

With that reasoning, I think we should only run benchmarking (for consistency) + integration tests (for ARM support) on CodeBuild. Perhaps unit tests as well, just so we're running all of our SOCI-related tests in a consistent environment. But that's just my thoughts, feel free to disagree.

.github/workflows/build.yml
austinvazquez, line 1: It is pretty useful to run the test jobs in developer forks to test changes which are not ready to open a pull request or even a draft pull request. CodeBuild runners prevent this use case. Have we considered using CodeBuild runners only for ARM unit/integration test jobs to still support running x86 on GitHub runners?

.github/workflows/bump-deps.yml
austinvazquez, line 1: Would this workflow benefit from the CodeBuild runners?

.github/workflows/prebuild.yml
austinvazquez, line 1: Similar question here, these jobs are useful to run in developer forks. Would any of them really benefit from the CodeBuild runners?

.github/workflows/releases.yml
austinvazquez, line 1: For the release workflow, is there a plan to follow-up with changes for onboarding the ARM support? At this time, I do not really see any benefits to running this workflow on CodeBuild runners without ARM support.

austinvazquez (changes requested):

.github/workflows/benchmark_visualization.yml
sondavidb: +1, looks like this isn't natively supported but there's workarounds to make this work

.github/workflows/releases.yml
coderbirju: It was required for the codebuild env, now since we are not including this workflow it is no longer required.

.github/workflows/benchmark_visualization.yml
coderbirju: This is a good call out, I have added a conditional statement to run the workflows on github runners if the trigger is on a fork.

.github/workflows/bump-deps.yml
coderbirju: No benefits, falling back to running on github runners.

.github/workflows/releases.yml
coderbirju: The next step would be to integrate building arm binaries through codebuild, will hold back on updating this till we have a set plan for that.

.github/workflows/benchmark_visualization.yml
```diff
@@ -17,7 +17,7 @@ env:
 
 jobs:
   benchmark:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ 'github.event.pull_request.head.repo.fork '  && 'ubuntu-22.04' || 'codebuild-soci-visualization-workflow-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-7.0-xlarge' }}
```
sondavidb, line 20: Based on the tests run in this PR it seems this will just always run 22.04.

Also, I assume we're switching to 22.04 because that's what CodeBuild uses?

coderbirju: Updated the workflow, there is not really a simpler way to add conditions to the runner type. One of those times where WET code makes more sense to reduce complicated conditions.

Kern-- (dismissed):

sondavidb (approved):

PR #1302, opened 2024-07-03
Fallback to containerd if we are unable to fetch SOCI artifacts

### LINKED ISSUE #1034: [FEATURE] Fallback to `containerd` if we cannot lazy load image

sondavidb:
**Issue #, if available:**
Continuation on #1035
Fixes #1034 

**Description of changes:**
Following up on the last [comment](https://github.com/awslabs/soci-snapshotter/pull/1035#issuecomment-1967131892) on #1035, I made the change to not send a FUSE failure for this new error, and added an integration test.

From original PR:

If we know that we cannot lazy load an image (eg: SOCI index does not exist for an image), we should fallback to the underlying runtime to do the fetching/unpacking of all layers.

**Testing performed:**

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -337,7 +337,7 @@ func (c *sociContext) Init(fsCtx context.Context, ctx context.Context, imageRef,
 
 			desc, err := client.SelectReferrer(ctx, ocispec.Descriptor{Digest: imgDigest}, defaultIndexSelectionPolicy)
 			if err != nil {
-				retErr = fmt.Errorf("cannot fetch list of referrers: %w", err)
+				retErr = fmt.Errorf("%w: cannot fetch list of referrers: %w", snapshot.ErrNoIndex, err)
 				return
 			}
 			indexDesc = desc
@@ -347,7 +347,7 @@ func (c *sociContext) Init(fsCtx context.Context, ctx context.Context, imageRef,
 
 		index, err := FetchSociArtifacts(fsCtx, refspec, indexDesc, store, remoteStore)
 		if err != nil {
-			retErr = fmt.Errorf("error trying to fetch SOCI artifacts: %w", err)
+			retErr = fmt.Errorf("%w: error trying to fetch SOCI artifacts: %w", snapshot.ErrNoIndex, err)
 			return
 		}
 		c.sociIndex = index
@@ -482,7 +482,7 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	client := src[0].Hosts[0].Client
 	c, err := fs.getSociContext(ctx, imageRef, sociIndexDigest, imgDigest, client)
 	if err != nil {
-		return fmt.Errorf("%w: unable to fetch SOCI artifacts for image %q: %w", snapshot.ErrUnableToLazyLoadImage, imageRef, err)
+		return fmt.Errorf("unable to fetch SOCI artifacts for image %q: %w", imageRef, err)
 	}
 
 	// Resolve the target layer
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -482,7 +482,7 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	client := src[0].Hosts[0].Client
 	c, err := fs.getSociContext(ctx, imageRef, sociIndexDigest, imgDigest, client)
 	if err != nil {
-		return fmt.Errorf("unable to fetch SOCI artifacts: %w", err)
+		return fmt.Errorf("%w: unable to fetch SOCI artifacts for image %q: %w", snapshot.ErrUnableToLazyLoadImage, imageRef, err)
 	}
 
 	// Resolve the target layer
--- a/integration/metrics_test.go
+++ b/integration/metrics_test.go
@@ -114,10 +114,36 @@ func TestOverlayFallbackMetric(t *testing.T) {
 		{
 			name:  "image with fs.Mount errors results in non-zero overlay fallback",
 			image: rabbitmqImage,
+			indexDigestFn: func(sh *shell.Shell, contentStoreType store.ContentStoreType, image imageInfo) string {
+				indexDigest := buildIndex(sh, image, withMinLayerSize(defaultSpanSize), withContentStoreType(contentStoreType))
+				contentStorePath := store.DefaultSociContentStorePath
+				if contentStoreType == "containerd" {
+					contentStorePath = store.DefaultContainerdContentStorePath
+				}
+
+				output := strings.Trim(string(sh.O("soci", "ztoc", "list")), "\n")
+				outputLines := strings.Split(output, "\n")
+				if len(outputLines) < 2 {
+					t.Fatalf("soci ztoc list output has no ztocs, actual output: %s", output)
+				}
+
+				// Choose a random ztoc to corrupt
+				ztocInfo := strings.Fields(outputLines[1])
+				corruptZtocDigest := strings.Split(ztocInfo[0], ":")[1]
+				// Do a random substitution to corrupt the specific ztoc
+				sh.X("sed", "-i", "s/a/abc/g", fmt.Sprintf("%s/blobs/sha256/%s", contentStorePath, corruptZtocDigest))
+
+				return indexDigest
+			},
+			expectedFallbackCount: 1,
+		},
+		{
+			name:  "image with no soci index results in no overlay fallback",
+			image: rabbitmqImage,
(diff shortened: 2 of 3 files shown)


sondavidb:
Huh, interesting. During testing I'm finding out that every other mount will message "deferring to container runtime", instead of just the first or just every mount. Wonder if this could be pointing to an issue.

sondavidb:
Seeing that TestOverlayFallbackMetric (specifically [this](https://github.com/awslabs/soci-snapshotter/blob/9aaf67e8cca2fd24e91dffea3b8e1b9c09842c84/integration/metrics_test.go#L115) case) is now failing due to this behavior. The problem is, with this behavior I'm not sure we can capture how many layers are falling back to overlayfs, as only every other layer seems to traverse this. Though, perhaps the answer is that we shouldn't update the metric? But that doesn't seem right.

sondavidb:
Fixed up the tests to pass. We decided that not sending an OverlayFallback metric made sense due to this being intended behavior, and the metric in question tracks unexpected overlay fallback metrics. This does have the small caveat that an invalid specified SOCI index digest will not emit this metric, but realistically very few people, if any, are actually manually specifying SOCI index digests, so it's probably fine as-is.

For the failing test, it was expecting overlay fallbacks on a bad zTOC, which resulted in the above edge case happening. I refactored the test to corrupt a zTOC in a somewhat hacky way, but it seems to work :P

I previously reported that performance didn't seem to increase, but as I was testing with an image with only one large layer, the lack of parallelization would make a minimal difference in that case. In images with multiple large layers this will heavily increase performance.

There's still one more caveat here, which is that, per the logs, it seems to use double the number of mounts actually required to set up the snapshot. But, upon inspecting the mounts (`/var/lib/soci-snapshotter-grpc/snapshotter/snapshots`), and checking the content store, it does not store double the number of blobs, so I don't believe this to be an issue, unless I'm looking in the wrong place.

integration/metrics_test.go
```diff
@@ -114,10 +114,28 @@ func TestOverlayFallbackMetric(t *testing.T) {
 		{
 			name:  "image with fs.Mount errors results in non-zero overlay fallback",
 			image: rabbitmqImage,
+			indexDigestFn: func(sh *shell.Shell, contentStoreType store.ContentStoreType, image imageInfo) string {
+				corruptZtocDigest := "b6f44b39963e56996b1273cf1c686089e77787f9cebe191b6c8eb7ef9361e625"
```
Kern--, line 118: Where did this digest come from? Can we choose a random ztoc digest from the index instead?

snapshot/snapshot.go
```diff
@@ -77,7 +77,10 @@ const (
 )
 
 var (
-	// Error returned by `fs.Mount` when there is no ztoc for a particular layer.
+	// ErrUnableToLazyLoadImage is returned by `fs.Mount` when the image cannot/should not be
+	// lazy loaded. (eg: If a SOCI index does not exist for an image)
+	ErrUnableToLazyLoadImage = errors.New("unable to lazy load image")
```
Kern--, line 82: What about `ErrNoIndex`? I think that `ErrUnableToLazyLoadImage` could be useful somewhere else, but right now you can't distinguish between "there is no index" and "there's an issue with the registry" that might be useful down the line.

fs/fs.go
```diff
@@ -482,7 +482,7 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	client := src[0].Hosts[0].Client
 	c, err := fs.getSociContext(ctx, imageRef, sociIndexDigest, imgDigest, client)
 	if err != nil {
-		return fmt.Errorf("unable to fetch SOCI artifacts: %w", err)
+		return fmt.Errorf("%w: unable to fetch SOCI artifacts for image %q: %w", snapshot.ErrUnableToLazyLoadImage, imageRef, err)
```
Kern--, line 485: If we go with the suggestion to use `ErrNoIndex`, I think this should get pushed down into `SociContext.Init`

integration/metrics_test.go
sondavidb: It was from the rabbitmq image. Yeah, we can just generate it dynamically.

snapshot/snapshot.go
sondavidb: Made these changes, but this then separates the bad index case from the no index found case. Perhaps we should make a separate error for the former and test that as well?

snapshot/snapshot.go
```diff
@@ -335,24 +344,27 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
...
-	if err == nil {
-		err := o.commit(ctx, false, target, key, append(opts, snapshots.WithLabels(base.Labels))...)
-		if err == nil || errdefs.IsAlreadyExists(err) {
-			// count also AlreadyExists as "success"
-			// there's no need to provide any details on []mount.Mount because mounting is already taken care of
-			// by snapshotter
-			log.G(lCtx).Info("local snapshot successfully prepared")
-			return nil, fmt.Errorf("target snapshot %q: %w", target, errdefs.ErrAlreadyExists)
+	// If the underlying FileSystem deems that the image is unable to be lazy loaded, then we
+	// should completely fallback to the underlying runtime (containerd) to handle pulling and
```
austinvazquez, line 348: nit: The wording underlying runtime (containerd) can be simplified to just containerd here.

snapshot/snapshot.go
```diff
@@ -325,7 +330,11 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 		}
 		log.G(lCtx).WithField(remoteSnapshotLogKey, prepareFailed).WithError(err).Warn("failed to prepare remote snapshot")
 		if !errors.Is(err, ErrNoZtoc) {
-			commonmetrics.IncOperationCount(commonmetrics.FuseMountFailureCount, digest.Digest(""))
+			if errors.Is(err, ErrNoIndex) {
+				skipLazyLoadingImage = true
+			} else {
+				commonmetrics.IncOperationCount(commonmetrics.FuseMountFailureCount, digest.Digest(""))
+			}
```
austinvazquez, line 337: It might be good to refactor this error handling into a switch case with more clear logging.

snapshot/snapshot.go
```diff
@@ -335,24 +344,27 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
...
+				// count also AlreadyExists as "success"
+				// there's no need to provide any details on []mount.Mount because mounting is already taken care of
+				// by snapshotter
+				log.G(lCtx).Info("local snapshot successfully prepared")
+				return nil, fmt.Errorf("target snapshot %q: %w", target, errdefs.ErrAlreadyExists)
+			}
+			log.G(lCtx).WithError(err).Warn("failed to internally commit local snapshot")
+			// Don't fallback here (= prohibit to use this key again) because the FileSystem
+			// possible has done some work on this "upper" directory.
+			return nil, err
```
austinvazquez, line 365: Might be worth refactoring this section to reduce cyclomatic complexity. For example, 

```
if skipLazyLoading {
    return mounts, nil
}

// Continue to handle edge cases
...
```

snapshot/snapshot.go
sondavidb: Given the wording of other errors (e.g. "deferring to container runtime" I'll just reword to "fallback to the container runtime"

Kern-- (approved):
I think the whole `Prepare` method could use a more structured approach to refactoring to make it less complex and to fix all the log noise it produces.

I don't want to block getting this in though, so I'll create an issue for it.

austinvazquez (approved):

PR #1310, opened 2024-07-13
add codebuild integration to build workflow

coderbirju:
**Issue #, if available:**

**Description of changes:**

**Testing performed:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,26 +20,67 @@ on:
       - 'scripts/**'
 
 env:
-  GO_VERSION: '1.21.12'
+  GO_VERSION: '1.21.10'
 
 jobs:
+  setup:
+    # This job sets up the runners to be used in the matrix for the build workflow.
+    # It provides a list of available runners with stable, human-friendly names and a mapping
+    # from those names to the actual `runs-on` value for each runner type. This allows us to
+    # use codebuild-hosted runners for awslabs/soci-snapshotter without requiring forks to also
+    # have codebuild-hosted runners.
+    #
+    # If you want to use codebuild runners for your personal fork, follow the instructions to set
+    # up a codebuild project. https://docs.aws.amazon.com/codebuild/latest/userguide/action-runner.html
+    # Then, replace 'soci-snapshotter-instance' with the name of the project you created.
+    name: Setup Build Matrix
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != 'awslabs/soci-snapshotter'
+    strategy:
+      matrix:
+        # We're using a matrix with a single entry so that we can define some config as YAML rather than
+        # having to write escaped json in a string.
+        include:
+          - use-codebuild: ${{ github.repository_owner == 'awslabs' }}
+            runs-on-names-cb: [ubuntu-x86, al2-arm]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -23,28 +23,73 @@ env:
...
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != 'awslabs/soci-snapshotter'
+    strategy:
+      matrix:
+        # We're using a matrix with a single entry so that we can define some config as YAML rather than
+        # having to write escaped json in a string.
+        include:
+          - use-codebuild: ${{ github.repository_owner == 'awslabs' }}
+            runs-on-names-cb: [ubuntu-x86, al2-arm]
+            runs-on-names: [ubuntu]
```
pendo324, line 46: nit: maybe `runs-on-names-gh` is better in contrast to `runs-on-names-cb`

pendo324 (dismissed):
Left a minor nit. LGTM after merging and rebasing on https://github.com/awslabs/soci-snapshotter/pull/1311

pendo324 (approved):

.github/workflows/build.yml
```diff
@@ -23,28 +23,72 @@ env:
...
+      matrix:
+        # We're using a matrix with a single entry so that we can define some config as YAML rather than
+        # having to write escaped json in a string.
+        include:
+          - use-codebuild: ${{ github.repository_owner == 'awslabs' }}
+            runs-on-names-cb: [ubuntu-x86, al2-arm]
+            runs-on-names-gh: [ubuntu]
+            runner-labels:
+              ubuntu: ubuntu-22.04
+              ubuntu-x86: "codebuild-soci-snapshotter-instance-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-7.0-xlarge"
```
sondavidb, line 48: GitHub atm usually runs in ~16-18 minutes, so we should probably choose instance sizes that will be roughly in line with that. xlarge seems overkill to me but if large takes too long it's probably fine

.github/workflows/build.yml
```diff
@@ -23,28 +23,73 @@ env:
...
-    runs-on: ubuntu-20.04
+    needs: setup
+    runs-on: ${{ fromJSON(needs.setup.outputs.runner-labels)[matrix.os] }}
     timeout-minutes: 40
     strategy:
       fail-fast: false
       matrix:
-        containerd: ["1.6.33", "1.7.18", "2.0.0-rc.3"]
+        os: ${{ fromJSON(needs.setup.outputs.available-runners) }}
+        containerd: ["1.6.30", "1.7.14"]
```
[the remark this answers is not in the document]
coderbirju: I don't see a need to test against release candidates, open to change it if there's a compelling argument for it.
Agreed on the containerd versions, updated

.github/workflows/build.yml
coderbirju: Agreed, changed it to large instance type, seems to be most similar to the one [Github provides](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) for public repos

pendo324 (approved):

Kern-- (approved):

Issue #1314, opened 2024-07-17
[Bug] xattrs cannot be disabled on layers with volume mounts


sondavidb:
### Description

There's a bug in https://github.com/containerd/continuity/issues/244 where CopyDir() fails on filesystems without xattrs. The syscall to list xattrs will fail with ENOTSUP, and continuity will treat it as a regular error and say it failed. When creating a container, if there are volumes to mount, nerdctl will make a call to this [CopyDir()](https://github.com/containerd/nerdctl/blob/0dfa3bb8de82f281f21a25e9e6eeacbb12caeeb1/pkg/cmd/container/run_mount.go#L361) function. This is a problem for us, since by default we disable xattrs unless whiteout directories or file xattrs are detected.

For now we can have xattrs be on by default and allow this change to be opted into, but we should either wait for the upstream bug fix (which could take a long time) or find a somewhat elegant workaround in SOCI.

### Steps to reproduce the bug

1. Pull image with a layer that has a volume mount without whiteout dirs or xattrs. (ARM64 for `rabbitmq@sha256:19e69a7a65fa6b1d0a5c658bad8ec03d2c9900a98ebbc744c34d49179ff517bf` worked for this.)
2. Create SOCI index with xattrs optimization enabled.
3. Push to registry
4. Pull from registry with SOCI and attempt to run container

### Describe the results you expected

Container should run instead of erroring with "operation not supported"

### Host information

1. OS: Linux
2. Snapshotter Version: v0.6.1
3. Containerd Version: v1.7.18

PR #1315, opened 2024-07-17
Revert "Disable xattrs by default"

sondavidb:
**Issue #, if available:**
Related: #1314 

**Description of changes:**

This reverts commit 2a70d12d833f8e93f62dea30d400bac1e2d7810d and some extra reverts related to it. This is due to a bug we have where containers fail to start if the image has a layer with xattrs disabled and a volume mount is present in said layer.

This ideally gets fixed upstream, but as that might take a while to propagate into nerdctl, for now we should revert this commit so that this is an opt-in behavior instead of a default one. 

**Testing performed:**

### DIFF
--- a/docs/cli-usage.md
+++ b/docs/cli-usage.md
@@ -33,7 +33,9 @@ Flags:
 
  - ```--span-size``` : Span size that soci index uses to segment layer data. Default is 4MiB 
  - ```--min-layer-size``` : Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10MiB
- - ```--disable-xattrs``` : When true, adds DisableXAttrs annotation to SOCI index. This annotation often helps performance at pull time. Default is true.
+ - ```--optimizations``` : Enable experimental features by name. Usage is `--optimizations opt_name`.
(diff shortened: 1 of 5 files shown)


pendo324 (approved):

docs/cli-usage.md
```diff
@@ -33,7 +33,9 @@ Flags:
 
  - ```--span-size``` : Span size that soci index uses to segment layer data. Default is 4MiB 
  - ```--min-layer-size``` : Minimum layer size to build zTOC for. Smaller layers won't have zTOC and not lazy pulled. Default is 10MiB
- - ```--disable-xattrs``` : When true, adds DisableXAttrs annotation to SOCI index. This annotation often helps performance at pull time. Default is true.
+ - ```--optimizations``` : Enable experimental features by name. Usage is `--optimizations opt_name`.
```
swagatbora90, line 36: These are not part of the initial PR, but still coming up in the revert. intentional?

austinvazquez (approved):

docs/cli-usage.md
sondavidb: The original PR never added documentation, so I just added it here.

swagatbora90 (approved):

PR #1329, opened 2024-07-31
Update to containerd v2

apostasie:
Follow-on #1305


apostasie:
Ping @sondavidb here is the adapted PR to merge into the custom branch.

Given the way the CI is configured, it is not running on a PR against a branch.

Do we want to change that or was that done on purpose for some reason?

sondavidb:
Hey, thanks for this again!

> Given the way the CI is configured, it is not running on a PR against a branch.

Not a particularly good reason, but we don't really do feature branches usually, so we can change this. I have a PR out for this at #1330.

However, we did some discussion internally and realized that, should nerdctl release v2 with this branch's commit, we will have to keep up this branch forever, so it would not be a temporary branch. This is fine if necessary, but we would prefer to not do so.

As nerdctl's only dependency is [this function](https://github.com/awslabs/soci-snapshotter/blob/main/fs/source/source.go#L175-L220) in fs/source, I'd like to see if nerdctl maintainers are interested in just keeping a local copy in nerdctl instead, so I've opened a [PR](https://github.com/containerd/nerdctl/pull/3266) in nerdctl. We can always revert this once we update to containerd v2, but as we don't want to be forced to keep this branch unless necessary, the nerdctl change is preferred.

apostasie:
Hey @sondavidb 

Your PR got merged into nerdctl - which does solve my problem. Thanks a lot!

This PR here still has value of course - would you like me to close it and re-open it against master, and we go back to the previously agreed state (wait until containerd v2 is out, rebase)?

sondavidb:
> Your PR got merged into nerdctl

🎉

> would you like me to close it and re-open it against master, and we go back to the previously agreed state (wait until containerd v2 is out, rebase)?

Sure, we can do that. I don't think you have to close this, though. I can edit your PR to main instead of the feature branch, so I've gone ahead and did that.

LMK if there's anything else you need. Thanks a bunch!

apostasie:
Hey @sondavidb !

Hope this finds you well.
Is the team here interested in reviving this?

Lmk and I can dust it off.

sondavidb:
Hey @apostasie !

Given that we require compatibility with older client versions, this isn't something that we believe we can do anytime soon, as some of our users will require support for containerd 1.7 for a while. Currently the plan is to wait for an LTS version to be announced before moving SOCI to v2.0.0.

apostasie:
> Hey @apostasie !
> 
> Given that we require compatibility with older client versions, this isn't something that we believe we can do anytime soon, as some of our users will require support for containerd 1.7 for a while. Currently the plan is to wait for an LTS version to be announced before moving SOCI to v2.0.0.

Totally understandable.
I need this for myself anyhow - so, will maintain it on a private fork.
Ping me if you ever need this to land (or take it and run with it).

Cheers.

Issue #1081, opened 2024-02-16
[FEATURE] Add test coverage metrics


sondavidb:
### Description

We know that our testing suite, while somewhat comprehensive, lacks in a few areas. It would be great future-proofing to ensure that we have good testing coverage that anyone can easily run.

### Describe the solution you'd like

A Makefile target that will test for code coverage. Ideally this gets called in our workflow as well.

### Describe any alternative solutions/features you've considered

None shy of manual debugging

### Any additional context or information about the feature request

Once this is implemented, we can be more rigorous on tackling certain issues related to testing coverage. #290, #569, #570, #571 are all currently open as of opening this issue.

PR #1366, opened 2024-09-05
Add test coverage

sondavidb:
**Issue #, if available:**
#1081 

**Description of changes:**
Added necessary code to add code coverage for our tests. This should help us to get a better idea of where our test suite stands as a whole.

This is really only a preliminary step, the next step would probably be finding a way to combine these metrics, but for now this should help us ensure we're on the right path.

**Testing performed:**
`make test-with-coverage` and `make integration-with-coverage`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,9 @@ jobs:
         if: matrix.os == 'al2-arm'
         run: dnf install zlib-static.aarch64 -y
       - run: make
-      - run: make test
+      - run: make test-with-coverage
+      - name: Show test coverage
+        run: make show-test-coverage
 
   integration:
     needs: setup
@@ -91,4 +93,6 @@ jobs:
       - name: Install zlib static on AL2 ARM instances
         if: matrix.os == 'al2-arm'
         run: dnf install zlib-static.aarch64 -y
-      - run: make integration
+      - run: make integration-with-coverage
+      - name: Show test coverage
+        run: make show-integration-coverage
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@
 CMD_DESTDIR ?= /usr/local
 GO111MODULE_VALUE=auto
 OUTDIR ?= $(CURDIR)/out
+COVDIR ?= $(CURDIR)/cov
 PKG=github.com/awslabs/soci-snapshotter
 VERSION=$(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
 REVISION=$(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -99,6 +106,7 @@ clean: clean-integration
 	@echo "🧹 ... 🗑️"
 	@rm -rf $(OUTDIR)
 	@rm -rf $(CURDIR)/release/
+	@rm -rf $(COVDIR)
```
sondavidb, line 109: This fails when running integration tests because the binary runs with elevated permissions. Is there a way to fix this?

util/testutil/shell.go
```diff
@@ -367,7 +367,7 @@ func KillMatchingProcess(sh *shell.Shell, psLinePattern string) error {
 
 	var allErr error
 	for _, pid := range targets {
-		if err := sh.Command("kill", "-9", fmt.Sprintf("%d", pid)).Run(); err != nil {
+		if err := sh.Command("kill", "-2", fmt.Sprintf("%d", pid)).Run(); err != nil {
```
austinvazquez, line 370: It might be worth a comment to explain why this needs to be SIGTERM instead of SIGKILL.

```suggestion
// Send SIGTERM so the unit under test correctly writes integration coverage reports to Go coverage directory.
```

.github/workflows/build.yml
```diff
@@ -91,4 +93,6 @@ jobs:
       - name: Install zlib static on AL2 ARM instances
         if: matrix.os == 'al2-arm'
         run: dnf install zlib-static.aarch64 -y
-      - run: make integration
+      - run: make integration-with-coverage
+      - name: Re-run to show test coverage
```
austinvazquez, line 97: Maybe just `Show test coverage`?

Makefile
austinvazquez: The _way_ would be to map the user running integration tests to the root user in the container for integration tests, but that might have its own set of challenges.

sondavidb: Oh, I forgot to resolve this, the reason we had this issue was that the folder itself was being created by root instead of by the user. I think somewhere in the testing suite setup or when setting up the containers it saw the folder wasn't created and did it on its own. My newest change just adds the `mkdir` comment.

austinvazquez (dismissed):
Nice, excited to see this come through!

.github/workflows/build.yml
```diff
@@ -70,7 +70,9 @@ jobs:
         if: matrix.os == 'al2-arm'
         run: dnf install zlib-static.aarch64 -y
       - run: make
-      - run: make test
+      - run: make test-with-coverage
+      - name: Show test coverage
+        run: make test-with-coverage
```
Kern--, line 75: Why do we need to run this twice?

sondavidb:
Discussed offline, new changes do the following
- Separate cleaning coverage data from the regular `clean` Makefile directive (though it still runs it, now it is its own target)
- Separate running testing suites with coverage from showing the results of said coverage
- Add option to generate HTML line-by-line reports for developers to use locally

austinvazquez (approved):

Kern-- (approved):

PR #1378, opened 2024-09-18
Support systemd socket activation

### LINKED ISSUE #1088: [FEATURE]  Support systemd socket activation for soci-snapshotter-grpc
### Description

I would like to use systemd socket activation with the `soci-snapshotter-grpc` helper, so that system administrators can decide how they expose this endpoint to services like prometheus

I would like `soci-snapshotter-grpc` to expose a flag or configuration to allow it to use the file descriptors provided by systemd when the `soci-snapshotter-grpc` service is activated through a socket. For more details refer to [system.socket](https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html). This feature is already supported by other services like docker, see [this file](https://github.com/moby/moby/blob/c70d7905fbd9e872bd39bbc0b9d59afb82f771d7/daemon/listeners/listeners_linux.go#L19) for reference. The docker implementation is a bit overkill for SOCI, so the functionality I would like is:

- Extend the `--address` flag to understand the "protocol", e.g.: `file://`, `unix://`, `fd://`, similar to what Docker does
- Parse the given address, and use that to start the server. Unlike the docker implementation, only one server should be started instead of one per "host" provided.

I could use something like `socat` and redirect traffic with it, however this adds an additional dependency, and that wouldn't be great for container-optimized distributions like Bottlerocket.

_No response_

Kern--:
**Issue #, if available:**
Closes #1088 

**Description of changes:**
This change adds 2 features:
1) the --address flag of the soci snapshotter supports a network prefix 
2) The soci snapshotter can listen on a file descriptor passed by
   systemd socket activation.

If the user tells soci to listen on a file descriptor, but doesn't use systemd socket activation to launch the process, the snapshotter will fall back to the default address as a unix socket.

The example systemd service file is updated to use systemd activation if possible. There is an example systemd socket file, but no strict dependency from the service file so that users who are always starting SOCI on boot won't get a warning from systemd.

**Testing performed:**
Tested manually and verified that:
1) socket activation works
2) the example systemd unit file works without socket activation
3) setting a custom socket address still works both with and without socket activation.

### DIFF
--- a/cmd/soci-snapshotter-grpc/main.go
+++ b/cmd/soci-snapshotter-grpc/main.go
@@ -34,6 +34,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -43,6 +44,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"time"
 
 	_ "net/http/pprof"
@@ -66,6 +68,7 @@ import (
 	"github.com/containerd/containerd/snapshots"
 	runtime_alpha "github.com/containerd/containerd/third_party/k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 	"github.com/containerd/log"
+	"github.com/coreos/go-systemd/v22/activation"
 	sddaemon "github.com/coreos/go-systemd/v22/daemon"
 	metrics "github.com/docker/go-metrics"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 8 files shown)


cmd/soci-snapshotter-grpc/main.go
```diff
@@ -337,3 +340,37 @@ func getCriConn(criAddr string) (*grpc.ClientConn, error) {
...
+	}
+}
+
+func listenFd(ctx context.Context) (net.Listener, error) {
+	listeners, err := activation.Listeners()
+	if err != nil {
+		return nil, err
+	}
+	if len(listeners) == 0 {
+		log.G(ctx).Info("Address was set to listen on a file descriptor, but no file descriptors were passed. Perhaps soci was launched directly without using systemd socket activation? Using the default address.")
```
sondavidb, line 366: nit: Can we separate the last sentence into a separate log message for clarity?

sondavidb (commented):
Do we want to add an integration test to this? My understanding is that this doesn't change anything in the integration tests so maybe a quick one using socket activation. (Though actually, alpine doesn't support systemd, so maybe not something very feasible?)

austinvazquez (approved):

austinvazquez (approved):
Love it. Thanks for the added tests.

pendo324 (approved):

sondavidb (approved):

PR #1383, opened 2024-10-03
Use AL2023 for integ test container

Kern--:
**Issue #, if available:**
Part of #1378 

**Description of changes:**
This change moves our integ test container from Alpine to AL2023. The motivation is to be able to test features of a bigger OS, such as testing systemd socket activation.

This also fixes a few bugs I found in the integ test setup, mostly around not handling errors which caused things like missing packages to fail silently. 

**Testing performed:**
`make integration && golangci-lint run`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ ARG NERDCTL_VERSION=1.7.1
 
 FROM public.ecr.aws/docker/library/registry:3.0.0-alpha.1 AS registry
 
-FROM public.ecr.aws/docker/library/golang:1.22.7-alpine AS containerd-snapshotter-base
+FROM public.ecr.aws/amazonlinux/amazonlinux:2023 AS containerd-snapshotter-base
 
 ARG CONTAINERD_VERSION
 ARG RUNC_VERSION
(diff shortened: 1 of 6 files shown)


integration/util_test.go
```diff
@@ -722,8 +723,14 @@ func rebootContainerd(t *testing.T, sh *shell.Shell, customContainerdConfig, cus
 	)
 
 	// cleanup directories
-	testutil.KillMatchingProcess(sh, "containerd")
+	err := testutil.KillMatchingProcess(sh, "containerd")
+	if err != nil {
+		sh.Fatal("failed to kill containerd: %v", err)
+	}
 	testutil.KillMatchingProcess(sh, "soci-snapshotter-grpc")
```
pendo324, line 730: Think we need to reassign `err` here. May also consider refactoring this error check and the preceding error check to check the errors inline since there's no other output from the `KillMatchingProcess` function

```suggestion
	err = testutil.KillMatchingProcess(sh, "soci-snapshotter-grpc")
```

Dockerfile
```diff
@@ -29,16 +29,15 @@ ENV GOCOVERDIR /test_coverage
...
 COPY . $GOPATH/src/github.com/awslabs/soci-snapshotter
-RUN apk update && apk upgrade
-RUN apk add --no-cache \
-    btrfs-progs-libs \
-    curl \
-    fuse \
-    gcc \
-    libc6-compat \
-    libseccomp-dev \
+RUN dnf update && dnf upgrade && dnf install -y \
```
pendo324, line 32: Does go come by default on `amazonlinux:2023` images? Just noticed we're switching from a `golang-alpine` image to a standard AL2023 image, but we're not `dnf` installing golang

integration/util_test.go
Kern--: Whoops, will fix.

Dockerfile
Kern--: We don't actually use go inside the container. We build the snapshotter on the host and copy the binary into the container and the integ tests run on the host with execs into the container to run nerdctl commands.

austinvazquez (approved):
LGTM, thanks for getting the issues with errors on process kill.

coderbirju (approved):

PR #1392, opened 2024-10-09
Add ID-mapping capabilities

sondavidb:
**Issue #, if available:**

**Description of changes:**
Working off the POC at [Kern--/idmap-v2](https://github.com/Kern--/soci-snapshotter/tree/idmap-v2), added id-mapping capabilities to the snapshotter.

The PR additionally has a commit to add a separate ~~ctr binary~~ nerdctl binary (EDIT: later iterations changed this to a nerdctl binary) to the container. This is because containerd doesn't support id-mapping in ~~ctr~~ nerdctl in an official release yet, so we need a specific binary to pass the proper labels. This is done by adding a new Makefile target and copying it in the Dockerfile, but I'm open to different approaches here.

**Testing performed:**
`make test` and `make integration`

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -45,9 +45,13 @@ package fs
 import (
 	"context"
 	"fmt"
+	"io"
 	golog "log"
 	"net/http"
+	"os"
 	"os/exec"
+	"path/filepath"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -59,6 +63,7 @@ import (
 	layermetrics "github.com/awslabs/soci-snapshotter/fs/metrics/layer"
 	"github.com/awslabs/soci-snapshotter/fs/remote"
 	"github.com/awslabs/soci-snapshotter/fs/source"
+	"github.com/awslabs/soci-snapshotter/idtools"
 	"github.com/awslabs/soci-snapshotter/metadata"
 	"github.com/awslabs/soci-snapshotter/snapshot"
 	"github.com/awslabs/soci-snapshotter/soci"
@@ -67,6 +72,7 @@ import (
 	ctdsnapshotters "github.com/containerd/containerd/pkg/snapshotters"
 	"github.com/containerd/containerd/reference"
 	"github.com/containerd/containerd/remotes/docker"
+	"github.com/containerd/errdefs"
 	"github.com/containerd/log"
 	metrics "github.com/docker/go-metrics"
 	fusefs "github.com/hanwen/go-fuse/v2/fs"
@@ -455,6 +461,58 @@ func (fs *filesystem) getSociContext(ctx context.Context, imageRef, indexDigest,
 	return c, err
 }
 
+func getIDMappedMountpoint(mountpoint, activeLayerID string) string {
+	d := filepath.Dir(mountpoint)
+	return filepath.Join(fmt.Sprintf("%s_%s", d, activeLayerID), "fs")
+}
+
+func (fs *filesystem) IDMapMount(ctx context.Context, mountpoint, activeLayerID string, idmapper idtools.IDMap) (string, error) {
+	newMountpoint := getIDMappedMountpoint(mountpoint, activeLayerID)
+	logger := log.G(ctx).WithField("mountpoint", newMountpoint)
+
+	logger.Debug("creating remote id-mapped mount")
+	if err := os.Mkdir(filepath.Dir(newMountpoint), 0700); err != nil {
+		return "", err
+	}
+	if err := os.Mkdir(newMountpoint, 0755); err != nil {
+		return "", err
+	}
+
+	fs.layerMu.Lock()
+	l := fs.layer[mountpoint]
+	if l == nil {
+		fs.layerMu.Unlock()
+		logger.Error("failed to create remote id-mapped mount")
+		return "", errdefs.ErrNotFound
+	}
+	fs.layer[newMountpoint] = l
+	fs.layerMu.Unlock()
+	node, err := l.RootNode(0, idmapper)
+	if err != nil {
+		return "", err
+	}
+
+	fuseLogger := log.L.
+		WithField("mountpoint", mountpoint).
+		WriterLevel(logrus.TraceLevel)
+
+	return newMountpoint, fs.setupFuseServer(ctx, newMountpoint, node, l, fuseLogger, nil)
+}
+
+func (fs *filesystem) IDMapMountLocal(ctx context.Context, mountpoint, activeLayerID string, idmapper idtools.IDMap) (string, error) {
+	newMountpoint := getIDMappedMountpoint(mountpoint, activeLayerID)
+	logger := log.G(ctx).WithField("mountpoint", newMountpoint)
+
+	logger.Debug("creating local id-mapped mount")
+	if err := idtools.RemapDir(ctx, mountpoint, newMountpoint, idmapper); err != nil {
+		logger.WithError(err).Error("failed to create local mount")
+		return "", err
+	}
+
+	logger.Debug("successfully created local mountpoint")
+	return newMountpoint, nil
+}
+
 func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[string]string) (retErr error) {
 	// Setting the start time to measure the Mount operation duration.
 	start := time.Now()
@@ -560,7 +618,7 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 		}
 	}()
 
-	node, err := l.RootNode(0)
+	node, err := l.RootNode(0, idtools.IDMap{})
 	if err != nil {
 		log.G(ctx).WithError(err).Warnf("Failed to get root node")
 		retErr = fmt.Errorf("failed to get root node: %w", err)
@@ -577,6 +635,17 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	fs.layerMu.Unlock()
 	fs.metricsController.Add(mountpoint, l)
 
+	// Pass in a logger to go-fuse with the layer digest
+	// The go-fuse logs are useful for tracing exactly what's happening at the fuse level.
+	fuseLogger := log.L.
+		WithField("layerDigest", labels[ctdsnapshotters.TargetLayerDigestLabel]).
+		WriterLevel(logrus.TraceLevel)
+
+	retErr = fs.setupFuseServer(ctx, mountpoint, node, l, fuseLogger, c)
+	return
+}
+
+func (fs *filesystem) setupFuseServer(ctx context.Context, mountpoint string, node fusefs.InodeEmbedder, l layer.Layer, logger *io.PipeWriter, c *sociContext) error {
 	// mount the node to the specified mountpoint
 	// TODO: bind mount the state directory as a read-only fs on snapshotter's side
 	rawFS := fusefs.NewNodeFS(node, &fusefs.Options{
@@ -585,11 +654,6 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 		NegativeTimeout: &fs.negativeTimeout,
 		NullPermissions: true,
 	})
-	// Pass in a logger to go-fuse with the layer digest
-	// The go-fuse logs are useful for tracing exactly what's happening at the fuse level.
-	logger := log.L.
-		WithField("layerDigest", labels[ctdsnapshotters.TargetLayerDigestLabel]).
-		WriterLevel(logrus.TraceLevel)
 	mountOpts := &fuse.MountOptions{
 		AllowOther:    true,   // allow users other than root&mounter to access fs
 		FsName:        "soci", // name this filesystem as "soci"
@@ -600,25 +664,26 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	if _, err := exec.LookPath(fusermountBin); err == nil {
 		mountOpts.Options = []string{"suid"} // option for fusermount; allow setuid inside container
 	} else {
-		log.G(ctx).WithError(err).Infof("%s not installed; trying direct mount", fusermountBin)
+		log.G(ctx).WithField("binary", fusermountBin).WithError(err).Info("fusermount binary not installed; trying direct mount")
 		mountOpts.DirectMount = true
 	}
 	server, err := fuse.NewServer(rawFS, mountpoint, mountOpts)
 	if err != nil {
-		log.G(ctx).WithError(err).Debug("failed to make filesystem server")
-		retErr = err
-		return
+		log.G(ctx).WithError(err).Error("failed to make filesystem server")
+		return err
 	}
 
 	go server.Serve()
 
-	// Send a signal to the background fetcher that a new image is being mounted
-	// and to pause all background fetches.
-	c.bgFetchPauseOnce.Do(func() {
-		if fs.bgFetcher != nil {
-			fs.bgFetcher.Pause()
-		}
-	})
+	if c != nil {
+		// Send a signal to the background fetcher that a new image is being mounted
+		// and to pause all background fetches.
+		c.bgFetchPauseOnce.Do(func() {
+			if fs.bgFetcher != nil {
+				fs.bgFetcher.Pause()
+			}
+		})
+	}
 
 	return server.WaitMount()
 }
@@ -681,6 +746,11 @@ func (fs *filesystem) check(ctx context.Context, l layer.Layer, labels map[strin
 	return rErr
 }
 
+func isIDMappedDir(mountpoint string) bool {
+	dirName := filepath.Base(mountpoint)
+	return len(strings.Split(dirName, "_")) > 1
+}
+
 func (fs *filesystem) Unmount(ctx context.Context, mountpoint string) error {
 	fs.layerMu.Lock()
 	l, ok := fs.layer[mountpoint]
@@ -688,8 +758,13 @@ func (fs *filesystem) Unmount(ctx context.Context, mountpoint string) error {
 		fs.layerMu.Unlock()
 		return fmt.Errorf("specified path %q isn't a mountpoint", mountpoint)
 	}
-	delete(fs.layer, mountpoint) // unregisters the corresponding layer
-	l.Done()
+
+	delete(fs.layer, mountpoint)
+	// If the mountpoint is an id-mapped layer, it is pointing to the
+	// underlying layer, so we cannot call done on it.
+	if !isIDMappedDir(mountpoint) {
+		l.Done()
+	}
 	fs.layerMu.Unlock()
 	fs.metricsController.Remove(mountpoint)
 	// The goroutine which serving the mountpoint possibly becomes not responding.
--- a/fs/layer/node.go
+++ b/fs/layer/node.go
@@ -56,6 +56,7 @@ import (
 	commonmetrics "github.com/awslabs/soci-snapshotter/fs/metrics/common"
 	"github.com/awslabs/soci-snapshotter/fs/reader"
 	"github.com/awslabs/soci-snapshotter/fs/remote"
+	"github.com/awslabs/soci-snapshotter/idtools"
 	"github.com/awslabs/soci-snapshotter/metadata"
 	"github.com/containerd/log"
 	fusefs "github.com/hanwen/go-fuse/v2/fs"
@@ -189,7 +190,7 @@ func (f *FuseOperationCounter) Run(ctx context.Context) {
 
 // logFSOperations may cause sensitive information to be emitted to logs
 // e.g. filenames and paths within an image
-func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseInode uint32, opaque OverlayOpaqueType, logFSOperations bool, opCounter *FuseOperationCounter) (fusefs.InodeEmbedder, error) {
+func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseInode uint32, opaque OverlayOpaqueType, logFSOperations bool, opCounter *FuseOperationCounter, idMapper idtools.IDMap) (fusefs.InodeEmbedder, error) {
 	rootID := r.Metadata().RootID()
 	rootAttr, err := r.Metadata().GetAttr(rootID)
 	if err != nil {
@@ -210,9 +211,10 @@ func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseIno
 	}
 	ffs.s = ffs.newState(layerDgst, blob)
 	return &node{
-		id:   rootID,
-		attr: rootAttr,
-		fs:   ffs,
+		id:       rootID,
+		attr:     rootAttr,
+		fs:       ffs,
+		idMapper: idMapper,
 	}, nil
 }
 
@@ -272,9 +274,10 @@ func (fs *fs) inodeOfID(id uint32) (uint64, error) {
 // node is a filesystem inode abstraction.
 type node struct {
 	fusefs.Inode
-	fs   *fs
-	id   uint32
-	attr metadata.Attr
+	fs       *fs
+	id       uint32
+	attr     metadata.Attr
+	idMapper idtools.IDMap
 
 	ents       []fuse.DirEntry
 	entsCached bool
@@ -407,14 +410,14 @@ func (n *node) Lookup(ctx context.Context, name string, out *fuse.EntryOut) (*fu
 				n.fs.reportFailure(fuseOpLookup, fmt.Errorf("%s: %v", fuseOpLookup, err))
 				return nil, syscall.EIO
 			}
-			entryToAttr(ino, tn.attr, &out.Attr)
+			n.entryToAttr(ino, tn.attr, &out.Attr)
 		case *whiteout:
 			ino, err := n.fs.inodeOfID(tn.id)
 			if err != nil {
 				n.fs.reportFailure(fuseOpLookup, fmt.Errorf("%s: %v", fuseOpLookup, err))
 				return nil, syscall.EIO
 			}
-			entryToAttr(ino, tn.attr, &out.Attr)
+			n.entryToAttr(ino, tn.attr, &out.Attr)
 		default:
 			n.fs.reportFailure(fuseOpLookup, fmt.Errorf("%s: unknown node type detected", fuseOpLookup))
 			return nil, syscall.EIO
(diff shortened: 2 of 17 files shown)


idtools/idmap.go
```diff
@@ -0,0 +1,186 @@
...
+
+// toHost takes an id mapping and a remapped ID, and translates the
+// ID to the mapped host ID. If no map is provided, then the translation
+// assumes a 1-to-1 mapping and returns the passed in id #
+func toHost(contID uint32, idMap []specs.LinuxIDMapping) (uint32, error) {
+	if idMap == nil {
+		return contID, nil
+	}
+	for _, m := range idMap {
+		high, err := safeSum(m.ContainerID, m.Size)
```
austinvazquez, line 119: Should use builtin https://pkg.go.dev/math/big#Int instead?

idtools/idmap.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func RemapDir(ctx context.Context, originalMountpoint, newMountpoint string, idMap IDMap) error {
+	idmappedSnapshotBase := filepath.Dir(newMountpoint)
+	if err := os.Mkdir(idmappedSnapshotBase, 0755); err != nil {
+		return err
+	}
+
+	if err := exec.Command("cp", "-R", originalMountpoint, idmappedSnapshotBase).Run(); err != nil {
+		// if err := continuity.CopyDir(newMountpoint, originalMountpoint); err != nil {
```
austinvazquez, line 150: To be removed?

snapshot/snapshot.go
```diff
@@ -200,6 +212,8 @@ func NewSnapshotter(ctx context.Context, root string, targetFs FileSystem, opts
 		userxattr:                   userxattr,
 		minLayerSize:                config.minLayerSize,
 		allowInvalidMountsOnRestart: config.allowInvalidMountsOnRestart,
+		allowIDMap:                  config.allowIDMap,
+		idmapped:                    make(map[string]interface{}),
```
austinvazquez, line 216: As an optimization, should this only allocate memory if allow ID map is configured?

snapshot/snapshot.go
```diff
@@ -757,23 +832,67 @@ func (o *snapshotter) mounts(ctx context.Context, s storage.Snapshot, checkKey s
...
 			Options: options,
 		},
 	}, nil
+}
+
+func (o *snapshotter) getParentPaths(ctx context.Context, s storage.Snapshot) ([]string, error) {
+	parentPaths := make([]string, len(s.ParentIDs))
+
+	for i := range s.ParentIDs {
+		id := s.ParentIDs[i]
```
austinvazquez, line 858:
```suggestion
        for i, id := range s.ParentIDs
```

snapshot/snapshot.go
```diff
@@ -757,23 +832,67 @@ func (o *snapshotter) mounts(ctx context.Context, s storage.Snapshot, checkKey s
...
+		if _, ok := o.idmapped[s.ID]; ok {
+			id = fmt.Sprintf("%s_%s", s.ParentIDs[i], s.ID)
```
austinvazquez, line 860: Can use `id` here instead of reading again.

snapshot/snapshot.go
```diff
@@ -757,23 +832,67 @@ func (o *snapshotter) mounts(ctx context.Context, s storage.Snapshot, checkKey s
...
+		}
+		parentPaths[i] = o.upperPath(id)
+	}
+
+	return parentPaths, nil
+}
+
+func (o *snapshotter) createIDMapMounts(ctx context.Context, s storage.Snapshot, idmap idtools.IDMap) error {
+	log.G(ctx).Infof("mapping ids")
```
austinvazquez, line 869: Formatting is not required, but also is this logging too verbose? Perhaps consider removing or reducing to trace.

snapshot/snapshot.go
```diff
@@ -757,23 +832,67 @@ func (o *snapshotter) mounts(ctx context.Context, s storage.Snapshot, checkKey s
...
+	}
 
+	return idtools.RemapRoot(ctx, o.upperPath(s.ID), idmap)
+}
+
+func (o *snapshotter) createIDMapMount(ctx context.Context, path, id string, idmap idtools.IDMap) error {
+	// s.ID is the shortest unique identifier for each new container,
+	// so append it to the end of the new mountpoint
+	_, err := o.fs.IDMapMount(ctx, path, id, idmap)
+	if err == errdefs.ErrNotFound {
```
austinvazquez, line 885: For readability, let's use `errdefs.IsNotFound(err)` here instead.

fs/fs.go
```diff
@@ -455,6 +459,79 @@ func (fs *filesystem) getSociContext(ctx context.Context, imageRef, indexDigest,
...
+		log.G(ctx).WithError(err).Infof("%s not installed; trying direct mount", fusermountBin)
+		mountOpts.DirectMount = true
+	}
+	server, err := fuse.NewServer(rawFS, newMountpoint, mountOpts)
+	if err != nil {
+		log.G(ctx).WithError(err).Debug("failed to make filesystem server")
+		return "", err
+	}
+
+	go server.Serve()
```
austinvazquez, line 516: Does this goroutine need to be tethered to the passed in context? If not, it looks like the passed in context is only being used for logging.

fs/layer/node.go
```diff
@@ -808,7 +812,9 @@ func entryToAttr(ino uint64, e metadata.Attr, out *fuse.Attr) fusefs.StableAttr
 	mtime := e.ModTime
 	out.SetTimes(nil, &mtime, nil)
 	out.Mode = fileModeToSystemMode(e.Mode)
-	out.Owner = fuse.Owner{Uid: uint32(e.UID), Gid: uint32(e.GID)}
+	// Potentially dangerous casting int -> uint32? But probably fine.
```
austinvazquez, line 815: ?

austinvazquez (changes requested):
Thanks for including the modified ctr as a individual commit. Makes it easy to revert if containerd accepts the upstream contribution later.

Overall nice work iterating on this. I would like to see the logging cleaned up a bit. Some considerations:
1. Quite a bit of info logging occurring. Should some of these be reduced to debug or trace logs? Users running in production operate a large scale, so cost of logs increases quickly. So make each log count.
2. Prefer structured logging over formatting. Think about the tech that gets paged and has a hard time finding where in code an error originated because it has been formatted.

fs/fs.go
```diff
@@ -455,6 +459,79 @@ func (fs *filesystem) getSociContext(ctx context.Context, imageRef, indexDigest,
...
+		return "", errdefs.ErrNotFound
+	}
+	fs.layer[newMountpoint] = l
+	fs.layerMu.Unlock()
+	node, err := l.RootNode(0, idmapper)
+	if err != nil {
+		return "", err
+	}
+
+	rawFS := fusefs.NewNodeFS(node, &fusefs.Options{
```
Kern--, line 491: Is this all the same as in `fs.Mount`? Can we have a helper function instead?

idtools/idmap.go
```diff
@@ -0,0 +1,186 @@
...
+
+// ToHost returns the host user ID pair for the container ID pair.
+func (i IDMap) ToHost(pair User) (User, error) {
+	var (
+		target User
+		err    error
+	)
+	target.Uid, err = toHost(pair.Uid, i.UidMap)
+	if err != nil {
+		return invalidUser, err
```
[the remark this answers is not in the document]
sondavidb: 0 can map to a user (root) so invalidUser makes it more clear that this is not the case

fs/fs.go
```diff
@@ -455,6 +459,79 @@ func (fs *filesystem) getSociContext(ctx context.Context, imageRef, indexDigest,
...
+		AttrTimeout:     &fs.attrTimeout,
+		EntryTimeout:    &fs.entryTimeout,
+		NegativeTimeout: &fs.negativeTimeout,
+		NullPermissions: true,
+	})
+	mountOpts := &fuse.MountOptions{
+		AllowOther:    true,   // allow users other than root&mounter to access fs
+		FsName:        "soci", // name this filesystem as "soci"
```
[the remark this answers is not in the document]
sondavidb: I don't really see a reason for this

fs/fs.go
sondavidb: Is it a bad thing to tether logging to the context?

fs/fs.go
```diff
@@ -706,6 +785,10 @@ func (fs *filesystem) Unmount(ctx context.Context, mountpoint string) error {
 	return syscall.Unmount(mountpoint, syscall.MNT_FORCE)
 }
 
+func (fs *filesystem) UnmountLocal(ctx context.Context, mountpoint string) error {
```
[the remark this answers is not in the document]
sondavidb: For this specific one, I think some filesystem abstractions might want to take advantage of this. LMK if you disagree

fs/layer/node.go
sondavidb: I'm generally cautious of casting from a signed int to an unsigned int, perhaps overly so 😅 I'd feel a bit more comfortable if we have some sort of bounds check before this?

idtools/idmap.go
sondavidb: Not sure what the benefit here is exactly

fs/fs.go
```diff
@@ -455,6 +459,79 @@ func (fs *filesystem) getSociContext(ctx context.Context, imageRef, indexDigest,
...
+	}
+
+	fs.layerMu.Lock()
+	l := fs.layer[mountpoint]
+	if l == nil {
+		fs.layerMu.Unlock()
+		log.G(ctx).WithField("mountpoint", newMountpoint).Info("failed to create remote id-mapped mount")
+		return "", errdefs.ErrNotFound
+	}
+	fs.layer[newMountpoint] = l
```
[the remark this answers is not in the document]
sondavidb: It just reuses the same layer and builds on top of it AFAIK. I don't believe we need to unassign this in any way?

sondavidb:
New changes address logging concerns as well as moved the FUSE server setup to a new function. I think this also made the tests pass properly, possibly because I previously didn't setup the fuse logging properly before?

Also added a new internal patch which will make the diff look even bigger but it's just so we can store the patch locally.

sondavidb:
FWIW, I tested these changes on the ubuntu runners and they all seemed to pass?? Not sure what it might be indicative of.

https://github.com/sondavidb/soci-snapshotter/pull/20

Kern-- (commented):
I'm still reviewing, but I wanted to get my comments in so far before they drift too far.

idtools/idmap.go
```diff
@@ -0,0 +1,187 @@
...
+
+func RemapDir(ctx context.Context, originalMountpoint, newMountpoint string, idMap IDMap) error {
+	idmappedSnapshotBase := filepath.Dir(newMountpoint)
+	if err := os.Mkdir(idmappedSnapshotBase, 0755); err != nil {
+		return err
+	}
+
+	// It would be preferred to use something like continuity.CopyDir, but we need to avoid copying hardlinks,
+	// unless there is a more efficient way to ensure we don't id-map the same inode multiple times
+	if err := exec.Command("cp", "-R", originalMountpoint, idmappedSnapshotBase).Run(); err != nil {
```
Shubhranshu153, line 151: copy option to be cp -a (archive option) otherwise the file permissions are not preserved.

idtools/idmap.go
```diff
@@ -0,0 +1,187 @@
...
+	UidMap []specs.LinuxIDMapping `json:"UidMap"`
+	GidMap []specs.LinuxIDMapping `json:"GidMap"`
+}
+
+func LoadIDMap(id string, labels map[string]string) (IDMap, error) {
+	var idmap IDMap
+	uidmapJSON, okUID := labels[snapshots.LabelSnapshotUIDMapping]
+	gidmapJSON, okGID := labels[snapshots.LabelSnapshotGIDMapping]
+	if okUID && okGID {
+		if err := json.Unmarshal([]byte(uidmapJSON), &idmap.UidMap); err != nil {
```
Shubhranshu153, line 79: the idmap are sent via marshalling using the idmap serializer, need to use the corresponding deserializer (imported from containerd pr)

Reference PR: https://github.com/containerd/containerd/pull/10722/files#diff-d4bc8cdbfad4af153a088d3594ef1c77ca112ef540c70a6089424bfbd359a98a

```
func (i *IDMap) Unmarshal(uidMap, gidMap string) error {
+       unmarshal := func(str string, fn func(m specs.LinuxIDMapping)) error {
+               if len(str) == 0 {
+                       return nil
+               }
+               for _, mapping := range strings.Split(str, ",") {
+                       m, err := deserializeLinuxIDMapping(mapping)
+                       if err != nil {
+                               return err
+                       }
+                       fn(m)
+               }
+               return nil
+       }
+       if err := unmarshal(uidMap, func(m specs.LinuxIDMapping) {
+               i.UidMap = append(i.UidMap, m)
+       }); err != nil {
+               return err
+       }
+
+       return unmarshal(gidMap, func(m specs.LinuxIDMapping) {
+               i.GidMap = append(i.GidMap, m)
+       })
+}
+
+func deserializeLinuxIDMapping(str string) (specs.LinuxIDMapping, error) {
+       var (
+               hostID, ctrID, length int64
+       )
+       _, err := fmt.Sscanf(str, "%d:%d:%d", &ctrID, &hostID, &length)
+       if err != nil {
+               return specs.LinuxIDMapping{}, fmt.Errorf("input value %s unparsable: %w", str, err)
+       }
+       if ctrID < 0 || ctrID >= invalidID || hostID < 0 || hostID >= invalidID || length < 0 || length >= invalidID {
+               return specs.LinuxIDMapping{}, fmt.Errorf("invalid mapping \"%s\"", str)
+       }
+       return specs.LinuxIDMapping{
+               ContainerID: uint32(ctrID),
+               HostID:      uint32(hostID),
+               Size:        uint32(length),
+       }, nil
+}
```

idtools/idmap.go
```diff
@@ -0,0 +1,187 @@
+/*
```
Shubhranshu153, line 1: idmap package will be from containerd and will be throw away.
license to be for containerd.

snapshot/snapshot.go
```diff
@@ -302,7 +361,15 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 	}
 
 	target, ok := base.Labels[targetSnapshotLabel]
+	// !ok means we are in an active snapshot
 	if !ok {
+		// Setup id-mapped mounts if config allows.
+		// Any error here needs to stop the container from starting.
+		if o.allowIDMap {
```
Shubhranshu153, line 368: lets say the allowIDMap config is not true then we should return false in case we find the labels?

snapshot/snapshot.go
```diff
@@ -150,6 +160,8 @@ type snapshotter struct {
 	userxattr                   bool  // whether to enable "userxattr" mount option
 	minLayerSize                int64 // minimum layer size for remote mounting
 	allowInvalidMountsOnRestart bool
+	allowIDMap                  bool
```
Shubhranshu153, line 163: whats the specific need for this flag? i am not very clear on this.

snapshot/snapshot.go
sondavidb: You're right, good catch

sondavidb: Rethinking this, though, is there ever a reason to disable id-mapping? It's a new feature sure but in theory it's entirely opt-in behavior that is only triggered by adding the labels.

Maybe we should get rid of this option in the config entirely.

snapshot/snapshot.go
```diff
@@ -578,7 +647,13 @@ func (o *snapshotter) getCleanupDirectories(ctx context.Context, t storage.Trans
 	cleanup := []string{}
 	for _, d := range dirs {
 		if !cleanupCommitted {
-			if _, ok := ids[d]; ok {
+			currDir := d
```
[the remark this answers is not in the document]
sondavidb: This is too short to make a function. I'll just leave a more descriptive comment/var name instead.

snapshot/snapshot.go
```diff
@@ -613,7 +688,10 @@ func (o *snapshotter) unmountSnapshotDirectory(ctx context.Context, dir string)
 		return err
 	}
 	if mounted {
-		return o.fs.Unmount(ctx, mp)
+		if err := o.fs.Unmount(ctx, mp); err == errdefs.ErrNotFound {
```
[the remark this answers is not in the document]
sondavidb: Discussed offline, not sure why I made this change honestly but reverted it as it seems to serve no real purpose.

sondavidb:
Looks like I undid a ton of changes and messed things up while rebasing 😓 Manually re-added everything I think but should double check.

Will double-check multi-uid/gid mapping shenanigans tomorrow as well as using nerdctl instead of ctr but I'll have to see.

snapshot/snapshot.go
Shubhranshu153: yes it should not be required.

snapshot/snapshot.go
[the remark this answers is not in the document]
Shubhranshu153: functions ideally should be short (even one line) if they serve a singular purpose. The overall prepare function is too long and has too many logic ingrained, comments help but having a function with description is more readable.

idtools/idmap.go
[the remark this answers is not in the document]
Shubhranshu153: this package is imported from containerd and will be used as a package when containerd code is merged by henry

fs/layer/node.go
Shubhranshu153: you can add a check if e.UID < 0 throw and exception error, the comment itself is not necessary.

fs/fs.go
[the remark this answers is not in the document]
Shubhranshu153: this looks a bit iffy, should do a check if the key doesnt exist?

integration/run_test.go
```diff
@@ -520,3 +521,152 @@ func TestRunInNamespace(t *testing.T) {
...
+				if err != nil {
+					t.Fatalf("error listing files in %s", baseSnapshotDir)
+				}
+
+				if len(filenames) == len(newFilenames) {
+					t.Fatalf("error: id-mapping failed")
+				}
+
+				for _, check := range tt.checkFiles {
+					fullCheckPath := filepath.Join(baseRuntimeDir, containerID, "rootfs", check.path)
```
Shubhranshu153, line 656: is this checking each file uid gid to match the expected uid gid?

integration/run_test.go
```diff
@@ -520,3 +521,152 @@ func TestRunInNamespace(t *testing.T) {
...
+				{
+					path:        "/usr/bin/sh",
+					expectedUID: "12345",
+					expectedGID: "12345",
+				},
+			},
+		},
+		{
+			name:           "with multiple substitutions",
+			subUIDContents: fmt.Sprintf("%s:12345:1000\n%s:22222:1", dummyuser, dummyuser),
```
Shubhranshu153, line 588: this test doesnt seem to be correct, as even if we dont do multiple uid and gid it will show the same result?

integration/run_test.go
sondavidb: In each of the specified files to check, yes

integration/run_test.go
sondavidb: What do you mean exactly?

sondavidb:
Added a check for containerd version to skip the new tests which means we should wait for #1401 to be merged before merging this if we are happy with the changes

snapshot/snapshot.go
sondavidb: Removed this

fs/fs.go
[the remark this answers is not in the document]
sondavidb: You're right, I removed this in the latest changes

fs/fs.go
```diff
@@ -692,7 +770,8 @@ func (fs *filesystem) Unmount(ctx context.Context, mountpoint string) error {
 	l, ok := fs.layer[mountpoint]
 	if !ok {
 		fs.layerMu.Unlock()
-		return fmt.Errorf("specified path %q isn't a mountpoint", mountpoint)
+		log.G(ctx).Errorf("specified path %q isn't a remote mount", mountpoint)
+		return errdefs.ErrNotFound
 	}
 	delete(fs.layer, mountpoint) // unregisters the corresponding layer
 	l.Done()
```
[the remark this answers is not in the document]
sondavidb: Added a check to not delete layers if they are id-mapped

idtools/idmap.go
sondavidb: Opted for `cp -pR` instead as `cp -a` kept hardlinks which proved to be problematic previously when doing chown

fs/fs.go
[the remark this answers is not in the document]
sondavidb: Theoretically we should never hit this but if you want I can add it as a failsafe.

fs/layer/node.go
sondavidb: The func itself doesn't return an error so I'll just remove the comment

sondavidb:
Added commits from #1401 preemptively just to get CI running and make sure we're all good, will rebase before merging but CI should be green now.

Shubhranshu153 (approved):

integration/run_test.go
```diff
@@ -520,3 +521,132 @@ func TestRunInNamespace(t *testing.T) {
 		}
 	}
 }
+
+func TestRunWithIdMap(t *testing.T) {
```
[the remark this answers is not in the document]
Shubhranshu153: discused offline to create separate pr for more robust release.

idtools/idmap.go
Shubhranshu153: i am not sure this is correct, can we test the hard link case with -a. 
AFAIK replica here should not impact once we create the overlay,
we would need to test with SM with this change first also to verify it doestn break anything else.

Shubhranshu153 (changes requested):
not sure about the -pR change, we need to test more to approve this
if with -a the concerns about hardlink is indeed true  then we can change, otherwise i will prefer not to change this option.

idtools/idmap.go
```diff
@@ -0,0 +1,235 @@
...
+	return z, nil
+}
+
+func RemapDir(ctx context.Context, originalMountpoint, newMountpoint string, idMap IDMap) error {
+	idmappedSnapshotBase := filepath.Dir(newMountpoint)
+	if err := os.Mkdir(idmappedSnapshotBase, 0755); err != nil {
+		return err
+	}
+
+	// opt for '-pR' instead of '-a' to not copy hardlinks.
```
Shubhranshu153, line 196: -p needs to take in attribute list it seems, i am not certain this works.

Shubhranshu153 (approved):

sondavidb:
Ended up just using `-a --no-preserve=links` and it seems to work

Shubhranshu153 (dismissed):

sondavidb:
Rebased with main

integration/run_test.go
sondavidb: Test has been updated to use a different image with multiple user permissions inside.

idtools/idmap.go
sondavidb: Ended up changing this to `cp -a --no-preserve=links`

fs/fs.go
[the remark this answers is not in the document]
sondavidb: This function was removed entirely in later iterations so can be resolved.

fs/fs.go
```diff
@@ -455,6 +461,58 @@ func (fs *filesystem) getSociContext(ctx context.Context, imageRef, indexDigest,
...
+	fs.layer[newMountpoint] = l
+	fs.layerMu.Unlock()
+	node, err := l.RootNode(0, idmapper)
+	if err != nil {
+		return "", err
+	}
+
+	fuseLogger := log.L.
+		WithField("mountpoint", mountpoint).
+		WriterLevel(logrus.TraceLevel)
```
austinvazquez, line 497: Interesting technique. I might have to try this with background goroutines. :)

idtools/idmap.go
```diff
@@ -0,0 +1,186 @@
...
+	Uid uint32
+	Gid uint32
+}
+
+// IDMap contains the mappings of Uids and Gids.
+//
+//nolint:revive
+type IDMap struct {
+	UidMap []specs.LinuxIDMapping `json:"UidMap"`
+	GidMap []specs.LinuxIDMapping `json:"GidMap"`
```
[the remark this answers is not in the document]
austinvazquez: Just realizing this is imported from containerd probably to be removed at a later date. Ignore convention here.

idtools/idmap.go
```diff
@@ -0,0 +1,186 @@
...
+			return IDMap{}, err
+		}
+		if err := json.Unmarshal([]byte(gidmapJSON), &idmap.GidMap); err != nil {
+			return IDMap{}, err
+		}
+	}
+	return idmap, nil
+}
+
+// ToHost returns the host user ID pair for the container ID pair.
```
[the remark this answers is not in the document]
austinvazquez: This one is on me not realizing this is imported from containerd.

austinvazquez (approved):
Thanks for tightening up the logging. It looks much better!

Change LGTM.

fs/fs.go
```diff
@@ -455,6 +461,58 @@ func (fs *filesystem) getSociContext(ctx context.Context, imageRef, indexDigest,
 	return c, err
 }
 
+func getIDMappedMountpoint(mountpoint, activeLayerKey string) string {
```
Kern--, line 464: nit: can we rename this to `activeLayerId`?

`key` has a specific meaning to containerd and that's not what we mean.

fs/fs.go
```diff
@@ -600,25 +664,26 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	if _, err := exec.LookPath(fusermountBin); err == nil {
 		mountOpts.Options = []string{"suid"} // option for fusermount; allow setuid inside container
 	} else {
-		log.G(ctx).WithError(err).Infof("%s not installed; trying direct mount", fusermountBin)
+		log.G(ctx).WithField("binary", fusermountBin).WithError(err).Info("fuse binary not installed; trying direct mount")
```
Kern--, line 667: nit:  `fuse binary` -> `fusermount binary`

fs/fs.go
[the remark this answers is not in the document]
Kern--: It still calls `l.Done()` and invalidates the layer, though. Isn't that shared across all fuse mounts?

integration/run_test.go
```diff
@@ -520,3 +521,198 @@ func TestRunInNamespace(t *testing.T) {
...
```
Kern--, line 525: Should this also test that the ids are mapped correctly inside the container?

snapshot/snapshot.go
Kern--: This is accessed concurrently, so it needs to be a `sync.Map`

integration/run_test.go
sondavidb: Yeah, good point, in some manual tests that I ran the id mapping would fail because the container wouldn't recognize the IDs inside. Added tests for this.

fs/fs.go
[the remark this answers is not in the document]
sondavidb: Oh whoops, yeah you're right, fixed it up.

sondavidb:
Newest changes should address most recent concerns with variable naming and also adds testing for files inside the container as well

sondavidb:
After this is merged I think documentation on how to actually use this feature would be nice. I can take it as a followup.

Kern-- (dismissed):

Kern-- (approved):

Shubhranshu153 (approved):

PR #1394, opened 2024-10-11
Remove VerifiableReader

Kern--:
**Issue #, if available:**
N/A

**Description of changes:**

VerifiableReader was code that SOCI inherited from Stargz, but didn't use. With estargz images, the TOC is embedded in the layer in the remote store. When setting up a reader, you can optionally pass a digest of the TOC which the stargz-snaphostter will verify against the actual TOC in the remote registry.

SOCI stores TOCs in zTOCs which have integrity rooted in the SOCI index. Therefore, you get equivalent verification by specifing a SOCI index digest when pulling an image. By the time we're setting up a reader, we've already verified the ztocs and their layer mapping which means there is nothing to verify here.

Initially we initially just made Verify a no-op and bypassed it. This change cleans up that unneeded code entirely.

**Testing performed:**

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -560,12 +560,6 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 		}
 	}()
 
-	// Verification is needed to instantiate reader
-	l.SkipVerify()
-	log.G(ctx).Infof("Verification forcefully skipped")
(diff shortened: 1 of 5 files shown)


austinvazquez (approved):
LGTM thanks

coderbirju (approved):

PR #1440, opened 2024-12-19
Fix regression in metrics startup order

Kern--:
**Issue #, if available:**
N/A

**Description of changes:**
Before this change, if the metrics socket was a unix socket, SOCI would
not create the parent dir before trying to bind. This exposed an
implicit dependency on bind order if you tried to put the metrics
address next to the main socket which we broke when we introduced
systemd socket activation.

This change ensures that we create the parent dir before binding the
metrics socket.

**Testing performed:**
`GO_TEST_FLAGS='-run TestSnapshotterStartup' make integration`

I verified that the updated test failed before the snapshotter change and succeeds after.


timj-hh (approved):

austinvazquez (approved):
LGTM.

timj-hh (approved):

coderbirju (approved):

PR #1455, opened 2025-01-15
Update golangci-lint to v1.63.4 and resolve warnings

austinvazquez:
**Issue #, if available:**
n/a

**Description of changes:**
This change updates golangci-lint to v1.63.4 and resolves new linter warnings for shadowing of built-in functions.

**Testing performed:**
CI

### DIFF
--- a/.github/workflows/prebuild.yml
+++ b/.github/workflows/prebuild.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   GO_VERSION: '1.22.11'
-  GOLANGCI_LINT_VERSION: '1.61.0'
+  GOLANGCI_LINT_VERSION: '1.63.4'
 
 jobs:
   check:
--- a/fs/remote/blob_test.go
+++ b/fs/remote/blob_test.go
@@ -463,17 +463,17 @@ func multiRoundTripper(t *testing.T, contents []byte, opts ...interface{}) Round
 		})
 		var sparse bool
 		if sorted[0].b == 0 {
-			var max int64
+			var maxRegionEnd int64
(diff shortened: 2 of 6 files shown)


sondavidb:
So I get the idea behind not shadowing builtin functions but it but it feels like it's overcorrecting a bit, e.g. min and max [here](https://github.com/awslabs/soci-snapshotter/pull/1455/files#diff-22826213cf1ef6947ce54041a96a609db644105152593d1726b7361532158679L126) are very clearly the minimum and maximum duration and having to restate that is a bit silly.

Is it worth disabling this? Or maybe we should just use better variable names :p

austinvazquez:
Open to other suggestions if other variable names? `minimum`, `maximum`?

Kern-- (approved):
`minBackoff` and `maxBackoff` would also be reasonable, but I don't think it's that important.

The likelihood that we have an issue with this is low, but it's also good hygiene to not shadow builtins.

sondavidb:
> Open to other suggestions if other variable names? `minimum`, `maximum`?

I'm less hung up on your specific replacements and more so hung up on the fact that it doesn't allow us to use these variable names at all, i.e. it seems that we can no longer call any variable "min" in any context anymore even if it makes sense. That's the part that kind of irks me.

I do agree with Kern that this is generally a good practice, though, and given the vastness of the codebase and how few false positives we have, I guess it's reasonable to assume this shouldn't really ever be a problem.

sondavidb (dismissed):

pendo324 (approved):

sondavidb (approved):

PR #1456, opened 2025-01-21
Move io abstractions to ioutils

Kern--:
**Issue #, if available:**

**Description of changes:**
SOCI has a couple of extensions to combine various `io` interfaces. This PR moves these under `ioutils` instead of defining them in the packages where they were first needed to encourage reuse.

**Testing performed:**
`make && make test`

### DIFF
--- a/util/ioutils/multireadcloser.go
+++ b/util/ioutils/multireadcloser.go
@@ -1,0 +1,48 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


util/ioutils/multireadcloser.go
```diff
@@ -0,0 +1,48 @@
...
+	"errors"
+	"io"
+)
+
+type MultiReaderCloser struct {
+	c []io.ReadCloser
+	io.Reader
+}
+
+func NewMultiReadCloser(rcs []io.ReadCloser) *MultiReaderCloser {
```
sondavidb, line 29: Any particular reason we don't just use the exact readclosers given to us?

Kern--: Because `io.MultiReader` takes an `[]io.Reader`. `[]io.ReadCloser` is not the same type as `[]io.Reader` and can't be used in its place without this sort of conversion.

sondavidb (approved):

pendo324 (approved):

austinvazquez (approved):

PR #1457, opened 2025-01-21
Add full reader verification

Kern--:
WIP: needs test

**Issue #, if available:**

**Description of changes:**
Before this change, the SOCI snapshotter would verify files on read. After this change, SOCI will verify all files after the background fetcher is done, whether their read or not.

**Testing performed:**
`make`


metadata/reader.go
```diff
@@ -617,10 +617,6 @@ func (r *reader) OpenFile(id uint32) (File, error) {
 			return fmt.Errorf("failed to get file bucket %d: %w", id, err)
 		}
 		size, _ = binary.Varint(b.Get(bucketKeySize))
-		m, _ := binary.Uvarint(b.Get(bucketKeyMode))
-		if !os.FileMode(uint32(m)).IsRegular() {
-			return fmt.Errorf("%q is not a regular file", id)
-		}
```
sondavidb, line 623: Why did this chunk of code get removed?

fs/span-manager/span_manager.go
```diff
@@ -314,6 +319,15 @@ func (m *SpanManager) getSpanContent(spanID compression.SpanID, offsetStart, off
 	return io.NopCloser(buf), nil
 }
 
+func (m *SpanManager) MarkDownloaded() {
+	m.downloadedMu.Lock()
+	if !m.downloaded {
+		m.downloaded = true
+		close(m.DownloadedC)
```
sondavidb, line 326: Just to make sure — this will send a signal in [this](https://github.com/awslabs/soci-snapshotter/pull/1457/files#diff-d3bff8da8e98692f2a18b8ca2e5b5e6c6e9e497a5231ab4f830b783917a3c4f0R64) switch case, correct?

sondavidb (commented):
Just wanted to leave some comments for now, overall seems good but I need to look a little closer at the verifier tomorrow.

fs/fs.go
```diff
@@ -246,10 +247,19 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 		}
 		go bgFetcher.Run(context.Background())
```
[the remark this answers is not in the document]
Kern--: Yes, we probably should.

metadata/reader.go
Kern--: This prevented calling `metadataReader.OpenFile` on a non-regular file. That's probably not the right thing to do because a `metadata.File` is the information about a file's uncompressed offset/size and tar header offset/size within the uncompressed layer. That applied to all files, not just regular files.

In a practical sense, the big issue is that `reader.Verify` can't verify directories without this change.

fs/span-manager/span_manager.go
Kern--: Yes, reading from a closed channel returns the zero value and doesn't block.

PR #1465, opened 2025-02-03
Add non-lazy-loaded-layer verification

sondavidb:
**Issue #, if available:**

**Description of changes:**
Though we attempt to mimic overlay in MountLocal, we do not do any verification once we receive the content. This change adds functionality for this, and the option to disable it as well.

**Testing performed:**
`make && make test`

### DIFF
--- a/config/fs.go
+++ b/config/fs.go
@@ -56,6 +56,10 @@ type FSConfig struct {
 	MountTimeoutSec                int64  `toml:"mount_timeout_sec"`
 	FuseMetricsEmitWaitDurationSec int64  `toml:"fuse_metrics_emit_wait_duration_sec"`
 
+	// VerifyLocalMounts will verify the shasum of compressed and uncompressed
+	// content for layers pulled ahead of time (i.e. non-FUSE mounts)
+	VerifyLocalMounts bool `toml:"verify_local_mounts" default:"true"`
+
 	RetryableHTTPClientConfig `toml:"http"`
 	BlobConfig                `toml:"blob"`
 
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -44,6 +44,7 @@ package fs
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	golog "log"
@@ -257,6 +258,11 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 	pr := newPreresolver(fsOpts.maxConcurrency)
 	pr.Start(ctx)
 
+	var manifests *manifestMap
+	if cfg.VerifyLocalMounts {
+		manifests = new(manifestMap)
+	}
+
 	var ns *metrics.Namespace
 	if !cfg.NoPrometheus {
 		ns = metrics.NewNamespace("soci", "fs", nil)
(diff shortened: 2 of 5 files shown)


config/fs.go
```diff
@@ -56,6 +56,10 @@ type FSConfig struct {
 	MountTimeoutSec                int64  `toml:"mount_timeout_sec"`
 	FuseMetricsEmitWaitDurationSec int64  `toml:"fuse_metrics_emit_wait_duration_sec"`
 
+	// VerifyLocalMounts will verify the shasum of compressed and uncompressed
+	// content for layers pulled ahead of time (i.e. non-FUSE mounts)
+	VerifyLocalMounts bool `toml:"verify_local_mounts" default:"true"`
```
Kern--, line 61: Newer versions of go-toml don't support `default`. Can we move this out into a function like we did for pull modes?

fs/fs.go
```diff
@@ -449,6 +486,63 @@ func (fs *filesystem) MountLocal(ctx context.Context, mountpoint string, labels
 	return nil
 }
 
+// getLayerDiffMap will return the map of a compressed layer shasum to its uncompressed shasum.
+// If VerifyLocalMounts is disabled, this will return nil with no error.
+func (fs *filesystem) getLayerDiffMap(ctx context.Context, labels map[string]string, remoteStore resolverStorage) (map[string]digest.Digest, error) {
```
Kern--, line 491: Should this use an `ArtifactFetcher` instead so that we can access the local store?

fs/unpacker.go
```diff
@@ -49,26 +50,63 @@ func NewLayerArchive() Archive {
...
 	defer decompressReader.Close()
-	return archive.Apply(ctx, root, decompressReader, opts...)
+
+	r = wrapReader(decompressReader, verifier, false)
+	n, err := archive.Apply(ctx, root, r, opts...)
+	if err != nil {
+		return 0, err
+	}
+
+	if verifier != nil {
```
Kern--, line 82: Instead of a bunch of nil checks around, can we add a concept of a no-op verifier that just always returns true? That way, we can guarantee it's never nil and reduce the number of special cases?

fs/fs.go
```diff
@@ -377,6 +384,30 @@ func (c *sociContext) populateImageLayerToSociMapping(sociIndex *soci.Index) {
 	}
 }
 
+// manifestMap is a wrapper for an atomic map
+// of a manifest digest to its appropriate list of diffIDs.
+//
+// key: manifest digest
+// val: map where
```
Kern--, line 391: This val map is only ever used to construct a `layerVerifier`, can we construct it eagerly instead?

fs/unpacker.go
```diff
@@ -49,26 +50,63 @@ func NewLayerArchive() Archive {
 	return &layerArchive{}
 }
 
-func (la *layerArchive) Apply(ctx context.Context, root string, r io.Reader, opts ...archive.ApplyOpt) (int64, error) {
+// wrapReader uses a TeeReader to feed content into verifier (compressed or uncompressed).
+// If verification is disabled, this is a no-op.
+func wrapReader(r io.Reader, verifier *layerVerifier, compressed bool) io.Reader {
```
Kern--, line 55: Combined with the suggested change to make `layerVerifier` always non-nil, can we replace this with:
```
verifier.WrapCompressedReader(r)
```
and 
```
verifier.WrapUncompressedReader(r)
```

sondavidb:
Superceded by the parallel pull changes in #1607, which also added layer verification for all layers pulled ahead of time.

PR #1474, opened 2025-02-07
Add ARM release workflow

sondavidb:
**Issue #, if available:**
Final part of #447

**Description of changes:**
While x86 release binaries were already automated, GitHub Actions does not have native ARM support so we have been manually creating those. This PR should allow for automatic ARM artifact release generation.

This was achieved by moving the `build.yml` setup into a new file `setup.yml` so that it can be reused between the Build and Release workflows.

With this, our release process should be as, creating a branch, adding the third-party licenses, and pushing the tag. From there the automation should do the rest.

**Testing performed:**
Created my own tag locally which created [this](https://github.com/sondavidb/soci-snapshotter/releases/tag/v9.9.9) dummy release (GitHub Actions run [here](https://github.com/sondavidb/soci-snapshotter/actions/runs/13190790625)).

I haven't tested for multiple architectures yet, but I believe we now have something in our CI to do this so hopefully CI will catch any errors.


sondavidb:
It's worth noting that with this artifact generation the glibc version used to generate the artifacts might differ. Probably not a big deal, but also probably worth noting.

Kern-- (approved):

austinvazquez (approved):
LGTM, sorry for the delay. @sondavidb , this is fine to bring in now?

sondavidb:
> this is fine to bring in now?

Should be 🤞 Thanks!

PR #1482, opened 2025-03-13
fix: fuse file seek

Shubhranshu153:
**Issue #, if available:**
Find file for some containers were failing while launching containers with SOCI

**Description of changes:**
We checked this was introduced via the go-fuse update from 2.5.1 to 2.7.2 in this commit https://github.com/hanwen/go-fuse/commit/e885cea8d4d40a5a9bb92bc3cef7193f2a316f59, due to removal of the file seek functionality DirStreams.

This was subsequently fixed via re introducing the seek functionality in the pr: https://github.com/hanwen/go-fuse/commit/d6170d09d743644ccf6099744e5bad1d2c3e552f

**Testing performed:**
1. Reproduce the error with a impacted container and soci-snapshotter
2. Verify the fix with the update go-fuse version.

### DIFF
--- a/cmd/go.mod
+++ b/cmd/go.mod
@@ -55,7 +55,7 @@ require (
 	github.com/google/go-cmp v0.6.0 // indirect
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/uuid v1.6.0 // indirect
-	github.com/hanwen/go-fuse/v2 v2.7.2 // indirect
+	github.com/hanwen/go-fuse/v2 v2.7.3-0.20250115180623-d6170d09d743 // indirect
 	github.com/hashicorp/errwrap v1.1.0 // indirect
 	github.com/hashicorp/go-cleanhttp v0.5.2 // indirect
 	github.com/hashicorp/go-multierror v1.1.1 // indirect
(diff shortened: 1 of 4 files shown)


Kern-- (changes requested):
Can you please add more context to the commit message?

henry118:
@Shubhranshu153 Can we have an integ test to cover this case?

Shubhranshu153:
@henry118 will add in a new pr the integ tests as it is hard to reproduce consistently.

Kern-- (approved):

sondavidb (approved):
Agreed with Shubh, think we should get the fix out first then focus on integration tests.

From looking at the go-fuse issue (https://github.com/hanwen/go-fuse/issues/549), it looks like only very specific Seek operations were failing, but the majority seem to not be impacted. Specifically a user found that `ls` commands worked but their file manager could not do the same. Attempting to narrow down the very specific syscalls used would be tough, not to mention a whole comprehensive testing suite.

I'll create a new issue to track this.

chews93319 (approved):

sondavidb:
Opened #1484 to track this

PR #1025, opened 2024-01-09
Eagerly resolve local blobs

sondavidb:
**Issue #, if available:**
Fixes #931

**Description of changes:**
For local snapshots, we previously encountered a bug where the size of a layer was resported as zero in its descriptor. This caused the Fetch function to attempt to resolve it as a manifest, which would then cause the local snapshot to defer to container runtime, causing containerd to fetch all remaining layers ahead of time. This change resolves the local blob earlier to populate the size field, avoiding this issue.

**Testing performed:**
Followed steps in #931 and ensured local snapshots were successfully prepared.

After setting up containerd/crictl configs I ran the following script:

```
# Empty content store beforehand

TOKEN=$(aws ecr get-authorization-token --region us-west-2 | jq -r '.authorizationData[0].authorizationToken' | base64 -d)
sudo crictl \
  --runtime-endpoint "unix:///var/run/containerd/containerd.sock" \
  --image-endpoint "unix:///var/run/containerd/containerd.sock" \
  run \
  --creds $TOKEN \
  container-config.json \
  pod-config.json
```

<details>
  <summary>Output</summary>

``` json
{"level":"info","msg":"starting soci-snapshotter-grpc","revision":"5217c3718d46a675fb964ff06c2399540b03794e","time":"2024-01-09T21:56:31.727674500Z","version":"5217c37"}
{"emitMetricPeriod":10000000000,"fetchPeriod":500000000,"level":"info","maxQueueSize":100,"msg":"constructing background fetcher","silencePeriod":30000000000,"time":"2024-01-09T21:56:31.763298289Z"}
{"address":"/run/soci-snapshotter-grpc/soci-snapshotter-grpc.sock","level":"info","msg":"soci-snapshotter-grpc successfully started","time":"2024-01-09T21:56:31.767434398Z"}
{"key":"k8s.io/1/extract-544072554-KtoC sha256:e3e5579ddd43c08e4b5c74dc12941a4ef656fab070b1087a1fd5a8a836b71e7d","level":"info","msg":"preparing filesystem mount at mountpoint=/var/lib/soci-snapshotter-grpc/snapshotter/snapshots/1/fs","parent":"","time":"2024-01-09T21:56:32.550227833Z"}
[...]

PR #1493, opened 2025-03-20
Remove ORAS Resolve calls in MountLocal

sondavidb:
**Issue #, if available:**

**Description of changes:**
This PR has two parts:

Previously in #1025, we fixed an issue where MountLocal would fail as the artifact fetcher would make a call to the remote store, but since we passed it an ORAS Repository, the Resolve call would assume it was given a ref to a manifest, which will never be the case in this call stack.

By passing the artifact fetcher a BlobStore instead in this callstack, we can remove redundant logic. However, the ORAS resolve call will always try to resolve with a HEAD request, which some repos do not support. This is not a problem for the remote snapshot codepath, as it resolves manually via the RoundTripper. These two codepaths should really be using the same logic but currently are not, so by removing the ORAS resolve codepath, we can also squash this bug.

**Testing performed:**

### DIFF
--- a/fs/artifact_fetcher.go
+++ b/fs/artifact_fetcher.go
@@ -22,9 +22,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
 	"strconv"
 
+	sociremote "github.com/awslabs/soci-snapshotter/fs/remote"
 	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/awslabs/soci-snapshotter/soci/store"
 	"github.com/awslabs/soci-snapshotter/util/ioutils"
@@ -35,6 +37,7 @@ import (
 	"golang.org/x/sync/errgroup"
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/errdef"
+	"oras.land/oras-go/v2/registry"
 	"oras.land/oras-go/v2/registry/remote"
 )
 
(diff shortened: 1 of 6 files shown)


sondavidb:
So I did some debugging on failing integration tests, and it looks like the ORAS Resolve call for manifests is actually used when fetching remote objects in FetchSociArtifacts, so we still use ORAS's resolve call in that call path, but when getting layers we no longer do.

Perhaps I should change the commit message then...

fs/artifact_fetcher.go
```diff
@@ -68,6 +71,70 @@ func newArtifactFetcher(refspec reference.Spec, localStore store.BasicStore, rem
...
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+	refDigest, err := ref.Digest()
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+
+	tr := &clientWrapper{r.Client}
+	buildScheme := "https"
```
austinvazquez, line 103: nit: could be const

fs/artifact_fetcher.go
```diff
@@ -68,6 +71,70 @@ func newArtifactFetcher(refspec reference.Spec, localStore store.BasicStore, rem
...
+
+	// Construct the descriptor
+	mediaType, _, _ := mime.ParseMediaType(resp.Header.Get("Content-Type"))
+	if mediaType == "" {
+		mediaType = "application/octet-stream"
+	}
+
+	size, err := sociremote.ParseSize(resp)
+	if err != nil {
+		return ocispec.Descriptor{}, nil
```
austinvazquez, line 118: Did you mean to return the error here?

fs/artifact_fetcher.go
henry118: we shouldn't hardcode `https` here. Use [this logic](https://github.com/awslabs/soci-snapshotter/blob/main/service/resolver/registry.go#L202-L207)

fs/remote/resolver.go
```diff
@@ -450,46 +450,47 @@ func redirect(ctx context.Context, blobURL string, tr http.RoundTripper) (string
...
+		}
+		if method == http.MethodGet {
+			req.Header.Set("Range", "bytes=0-1")
+		}
+
+		resp, err := rt.RoundTrip(req)
+		if err != nil {
+			return nil, err
+		}
+		defer resp.Body.Close()
```
henry118, line 473: defer routines are for the scope of the entire function. But it's better to close the http body for each iteration. we can either don't use defer or wrap the logic in a closure.

fs/fs.go
```diff
@@ -417,31 +417,18 @@ func (fs *filesystem) MountLocal(ctx context.Context, mountpoint string, labels
 	if err != nil {
 		return fmt.Errorf("cannot parse image ref (%s): %w", imageRef, err)
 	}
-	remoteStore, err := newRemoteStore(refspec, client)
+	repo, err := newRemoteStore(refspec, client)
 	if err != nil {
 		return fmt.Errorf("cannot create remote store: %w", err)
 	}
+	remoteStore := newRemoteBlobStore(repo)
```
henry118, line 424: This looks redundant and we are not using `repo` except feeding it to `newRemoteBlobStore`. Probably we can combine the two into the new function.

henry118 (approved):
LGTM

austinvazquez (approved):

Kern-- (approved):

PR #1494, opened 2025-03-20
Add metric for failing to prepare local snapshot

sondavidb:
**Issue #, if available:**

**Description of changes:**
We previously had no way of knowing if MountLocal fails, so adding this metric could help with telemetry for users, as well as help our integration testing.

**Testing performed:**

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -435,6 +435,7 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 
 	// Local snapshot setup failed. Generally means something critical has gone wrong.
 	log.G(lCtx).WithError(err).Warnf("failed to prepare local snapshot; %v", ErrDeferToContainerRuntime)
+	commonmetrics.IncOperationCount(commonmetrics.LocalMountFailureCount, digest.Digest(""))
 	return mounts, nil
 }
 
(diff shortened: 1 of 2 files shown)


snapshot/snapshot.go
```diff
@@ -435,6 +435,7 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 
 	// Local snapshot setup failed. Generally means something critical has gone wrong.
 	log.G(lCtx).WithError(err).Warnf("failed to prepare local snapshot; %v", ErrDeferToContainerRuntime)
+	commonmetrics.IncOperationCount(commonmetrics.LocalMountFailureCount, digest.Digest(""))
 	return mounts, nil
```
austinvazquez, line 439: Do you know why when a critical error occurs we return nil error in this call path?

sondavidb: This is a thing most remote snapshotters do. Whenever we successfully do a prepare call you can see we return ErrAlreadyExists, so that containerd doesn't try to pull it on its own. It will only try to pull an image if there is no error preparing the snapshot, so by returning nil we are signaling to containerd that it needs to pull this layer.

austinvazquez: Gotcha thanks for clarifying.

austinvazquez (approved):

sondavidb:
I wanted to add integration testing but I couldn't think of a good way to force a MountLocal failure. Probably just something we will have to trust works 😬

PR #1501, opened 2025-03-26
Refactor SOCI Index Build API

Kern--:
**Issue #, if available:**

**Description of changes:**

The SOCI index build API had two problems:
1. It required an ArtifactsDB as input and ignored the `WithArtifactsDb` option
2. The build platform was taken as input to the builder rather than a specific index build operation. This meant that we had to create a builder in order to create indexes for multiple platforms rather than calling `Build` for each platform.

This change fixes these by:
1. Removing ArtifactsDB from `NewIndexBuilder`. Callers can set the artifacts db using `soci.WithArtifactsDb` instead.
2. Renaming `BuildOption` to `BuilderOption`. `NewIndexBuilder` takes `BuilderOption` which is the same set of functions as before
3. Creating a new `BuildOption` type that is passed to `Build`. This is for options that are specific to an index build operation rather than all builds. Like:
4. Moving platform from `BuilderOption` -> `BuildOption`. Now you can use a single builder for multiple platforms by using `soci.WithPlatform` on each `Build` call.

Note: This is a backwards incompatible change and clients will need changes in order to update.

**Testing performed:**
`make && make test && make integration`

### DIFF
--- a/soci/soci_index.go
+++ b/soci/soci_index.go
@@ -209,16 +209,15 @@ func GetIndexDescriptorCollection(ctx context.Context, cs content.Store, artifac
 	return descriptors, indexDesc, nil
 }
 
-type buildConfig struct {
+type builderConfig struct {
 	spanSize            int64
 	minLayerSize        int64
 	buildToolIdentifier string
 	artifactsDb         *ArtifactsDb
-	platform            ocispec.Platform
 	optimizations       []Optimization
 }
 
-func (b *buildConfig) hasOptimization(o Optimization) bool {
+func (b *builderConfig) hasOptimization(o Optimization) bool {
 	for _, optimization := range b.optimizations {
 		if o == optimization {
 			return true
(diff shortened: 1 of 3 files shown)


soci/soci_index.go
```diff
@@ -250,53 +249,62 @@ func ParseOptimization(s string) (Optimization, error) {
...
 		c.buildToolIdentifier = tool
 		return nil
 	}
 }
 
-// WithPlatform specifies platform used to build soci indices.
-func WithPlatform(platform ocispec.Platform) BuildOption {
-	return func(c *buildConfig) error {
-		c.platform = platform
+// WithArtifactsDb speicifies the artifacts database
```
austinvazquez, line 288: typo:
```
// WithArtifactsDb specifies the artifacts database
```

soci/soci_index.go
```diff
@@ -250,53 +249,62 @@ func ParseOptimization(s string) (Optimization, error) {
 	return "", fmt.Errorf("optimization %s is not a valid optimization %v", s, Optimizations)
 }
 
-// BuildOption specifies a config change to build soci indices.
-type BuildOption func(c *buildConfig) error
+// BuilderOption is a functional argument that affects a SOCI index builder
+// and all indexes built with that builder.
+type BuilderOption func(c *builderConfig) error
```
austinvazquez, line 254: I know it is pre-existing but do any of the build options return an error?

Kern--: No. The thought was that it's easier to ignore it than to add it later, but I guess since `builderConfig` is not exported it doesn't really matter.

swagatbora90 (approved):

PR #1502, opened 2025-04-01
Attach SOCI index GC to Image

Kern--:
**Issue #, if available:**
N/A

**Description of changes:**
Before this change, SOCI indexes were given a root garbage collection label when using the containerd content store. This meant that in order to remove SOCI indexes, customers would have to know to delete the content in containerd (e.g. with ctr).

This change adds an option to the `Build` API to skip the GC root label and instead allow the caller to manage GC labels. As an example, it updates the CLI's `create` command to set the SOCI index as a content ref for the image. When using the contaienrd content store, the SOCI index will be removed if/when the image gets removed.

**Testing performed:**

I also manually tested creating a SOCI index for an image using the containerd content store and then deleting the image.
Before the change the index still exists.
After the change, the index gets deleted.

### DIFF
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -33,6 +33,7 @@ const (
 	spanSizeFlag        = "span-size"
 	minLayerSizeFlag    = "min-layer-size"
 	optimizationFlag    = "optimizations"
+	sociIndexGCLabel    = "containerd.io/gc.ref.content.soci-index"
 )
 
 // CreateCommand creates SOCI index for an image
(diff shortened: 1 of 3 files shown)


sondavidb (approved):

austinvazquez:
While you have the hood open on this, perhaps it would be good to either expand TestSociCreateGarbageCollection or add a new test case to validate the new default behavior to garbage collect the index after the referenced image is removed?

austinvazquez (approved):
LGTM thanks!

sondavidb (approved):

Issue #715, opened 2023-07-07
[FEATURE] Support multiple namespaces and content stores in ArtifactsDb


sparr:
### Description

With impending support for a second content store in #694, that content store supporting namespaces, and existing support for namespaces in `ctr` and containerd, we need support for namespaces in `soci` and `soci-snapshotter-grpc`. Currently both tools access the same SOCI content regardless of what namespace the runtime is using. When #694 lands, there will be some support for specifying what containerd namespace content should be stored in, and what content store should be accessed. That still leaves all the rest of our tooling ignorant of namespaces, and the artifact db ignorant of content store segregation.

### Describe the solution you'd like

Continue using the top level of the ArtifactsDb to store metadata about content that exists in the soci content store and the default name space. Create a new bucket for each additional content store that will store metadata for the default namespace in that content store. Then in each of those buckets (and at the top level for the soci content store "bucket") create a new bucket for each additional namespace that will store metadata for that namespace in that content store.

![SOCI ArtifactsDB support for namespaces and content stores (1) (1)](https://github.com/awslabs/soci-snapshotter/assets/121299/e23436c9-4640-427d-921c-3d1325f897bb)

If content removal is implemented via #694 and/or something like #645 before this PR lands, update the removal logic to only remove content from a particular content store if it is no longer referenced by metadata in any namespace.

### Describe any alternative solutions/features you've considered

Change nothing. The current implementation works for us now and will likely continue to do so for almost all users indefinitely.

Deprecate ArtifactsDB, use only content store. The metadata stored in the ArtifactsDB could be stored in content store labels instead, if/when SOCI migrates fully to one or more content stores that support labels. The other necessary functionality would be enumeration of artifacts/content, which could be implemented separately and alongside the content store(s). This would eliminate the need to keep two sources of information (artifacts db and content store) in sync. This would be my preferred solution if not for the uncertain future and timeline of the content store migration.

Artifact keys for namespace(s) and content store(s). Artifacts could all live in a single bucket addressed by digest as they do now, with presence in specific namespaces and/or content stores stored in new key/value pairs.

Nested buckets without exceptions. Using the root level to represent the combination of the soci content store and default namespace will maintain backward compatibility with existing databases. This and storing the default namespace within the top level bucket for each content store will also reduce unnecessary layers of storage for the most common use cases. Most uses will need to access 0 or 1 buckets, rather than every case requiring 2 nested buckets without this aspect of the design.
[...]

PR #1504, opened 2025-04-04
Refactor SOCI Init logic

Kern--:
**Issue #, if available:**

**Description of changes:**

This change moves most of the SOCI init logic out of `SociContext.Init` into `fs.fetchSociIndex`. This is to reduce the number of args we need to sent to `SociContext.Init`.

There are 3 logic changes here:

1. Remove the cached error lock on `SociContext`. `sync.Once` guarantees that no call to `Do` will return until exactly 1 executes the function. Therefore, there was never any concurrent access as the write to the cached error will always happen before any read.
2. The SOCI index digest is parsed rather than casted to a digest.Digest.
3. ~~FetchSociArtifacts now uses the Mount context instead of `fs.ctx`. `fs.ctx` is tied to the lifetime of the fs object which is the fs object which is the lifetime of the snapshotter. The mount ctx is tied to the lifetime of the `snapshot.Prepare` call.~~ already merged in #1509 

**Testing performed:**

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -298,7 +298,6 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 
 type sociContext struct {
 	cachedErr            error
-	cachedErrMu          sync.RWMutex
 	bgFetchPauseOnce     sync.Once
 	fetchOnce            sync.Once
 	sociIndex            *soci.Index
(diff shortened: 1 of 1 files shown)


fs/fs.go
```diff
@@ -516,6 +515,12 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	start := time.Now()
 	ctx = log.WithLogger(ctx, log.G(ctx).WithField("mountpoint", mountpoint))
 
+	ns, ok := namespaces.Namespace(fs.ctx)
```
sondavidb, line 518: IIUC this code overwrites this snapshot.Prepare call's namespace with the filesystem's namespace (which I'm not sure it has?). Is there a particular reason for this? I would think that we want to care about the namespace that the user wishes to operate in.

Kern--: It does, but this is not a change in this PR. We already used `fs.ctx` when fetching into the content store which uses a configured namespace or default as set here: https://github.com/awslabs/soci-snapshotter/blob/fadf416a4d5120e340686e298b14688270a8551c/fs/fs.go#L220-L224

The containerd content store already overrides the namespace in all of it's functions (e.g. https://github.com/awslabs/soci-snapshotter/blob/fadf416a4d5120e340686e298b14688270a8551c/soci/store/store.go#L218).

I tried to get rid of the `fs.ctx` namespace on this path, but I was getting issues somewhere due to missing a namespace on a context. Let me double check what exactly those errors were. 

FWIW, I think we need to fix the way we handle namespaces in general. See https://github.com/awslabs/soci-snapshotter/issues/715

Kern--: Looking a bit deeper, this is https://github.com/containerd/containerd/issues/6061 one level along. The namespace comes in as a grpc header, not in the context itself, but grpc headers are not copied across future grpc calls. 

The solution is we need to copy the namespace from the ctx back to the ctx (to move it from the grpc headers to directly on the context). I think we should probably use the containerd namespace interceptor.

To make this work, we need to make the containerd content store use the namespace on the context instead of it's stored namespace.

Kern--: fixed by #1509

austinvazquez (approved):

austinvazquez:
LGTM, TIL you could wrap multiple errors in the same error format. :)

Shubhranshu153 (approved):

Shubhranshu153:
LGTM

sondavidb (approved):

PR #1509, opened 2025-04-07
Always use namespace from request context

sondavidb:
**Issue #, if available:**

**Description of changes:**
Before this change, any calls (mostly to the containerd content store) would use the namespace specified in the config. This means that making changes to the content store within one namespace would still be accessible when using another namespace. This was done because the context would have a namespace attached in the gRPC header, but not in the context struct, so if we didn't explicitly attach a namespace to the context, any calls to the containerd content store would fail.

This change adds middleware to the gRPC to explicitly attach the context from the header to the context, so that it can be passed to store calls.

This change also removes the config option to specify a specific namespace, which is okay since the CLI was never tied to this and still requires a namespace to be used, and the correct namespace to use is the one attached to the containerd context anyway.

**Testing performed:**

### DIFF
--- a/cmd/soci-snapshotter-grpc/main.go
+++ b/cmd/soci-snapshotter-grpc/main.go
@@ -137,8 +137,13 @@ func main() {
 		log.G(ctx).Warn("skipped snapshotter is supported check")
 	}
 
+	serverOpts := []grpc.ServerOption{
+		grpc.ChainUnaryInterceptor(unaryNamespaceInterceptor),
+		grpc.ChainStreamInterceptor(streamNamespaceInterceptor),
+	}
+
 	// Create a gRPC server
-	rpc := grpc.NewServer()
+	rpc := grpc.NewServer(serverOpts...)
 
 	// Configure keychain
 	credsFuncs := []resolver.Credential{dockerconfig.NewDockerConfigKeychain(ctx)}
--- a/cmd/soci/commands/index/rm.go
+++ b/cmd/soci/commands/index/rm.go
@@ -46,8 +46,7 @@ var rmCommand = cli.Command{
 			return fmt.Errorf("please provide either index digests or image ref, but not both")
 		}
 
-		ctx := context.Background()
-		ctx, contentStore, err := store.NewContentStore(ctx, internal.ContentStoreOptions(cliContext)...)
+		contentStore, err := store.NewContentStore(internal.ContentStoreOptions(cliContext)...)
 		if err != nil {
 			return fmt.Errorf("cannot create local content store: %w", err)
 		}
@@ -57,6 +56,9 @@ var rmCommand = cli.Command{
 			return err
 		}
 		if ref == "" {
+			ctx, cancel := internal.AppContext(cliContext)
+			defer cancel()
+
 			byteArgs := make([][]byte, len(args))
 			for i, arg := range args {
 				byteArgs[i] = []byte(arg)
--- a/fs/source/source.go
+++ b/fs/source/source.go
@@ -85,6 +85,9 @@ const (
 	// TargetSociIndexDigestLabel is a label which contains the digest of the soci index.
 	TargetSociIndexDigestLabel = "containerd.io/snapshot/remote/soci.index.digest"
 
+	// TargetNamespace gives us the namespace that the snapshot was created in
+	TargetNamespace = "containerd.io/snapshot/remote/namespace"
+
 	// HasSociIndexDigest is a label that tells if the layer was pulled with a SOCI index.
 	HasSociIndexDigest = "containerd.io/snapshot/remote/has.soci.index.digest"
 )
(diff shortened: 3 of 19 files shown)


cmd/soci-snapshotter-grpc/main.go
```diff
@@ -137,8 +138,10 @@ func main() {
 		log.G(ctx).Warn("skipped snapshotter is supported check")
 	}
 
+	serverOpts := []grpc.ServerOption{grpc.ChainUnaryInterceptor(unaryNamespaceInterceptor)}
```
Kern--, line 141: Can we add a streaming interceptor as well? I don't think we're using it here, but I'm not sure if the containerd stream API would connect to SOCI via a stream as well. If we do add any streaming APIs, it would be nice to not have to go through this debugging again to rediscover the issue.

cmd/soci-snapshotter-grpc/main.go
```diff
@@ -203,6 +206,15 @@ func main() {
 	log.G(ctx).Info("Exiting")
 }
 
+// Copied from https://github.com/containerd/containerd/blob/041743e8afd2bb2189aa904f04bf87b9073892d3/cmd/containerd/server/namespace.go#L34
```
Kern--, line 209: If this was copied, then we need to include containerd's copyright somewhere.

cmd/soci/commands/index/rm.go
```diff
@@ -47,7 +47,7 @@ var rmCommand = cli.Command{
 		}
 
 		ctx := context.Background()
-		ctx, contentStore, err := store.NewContentStore(ctx, internal.ContentStoreOptions(cliContext)...)
+		contentStore, err := store.NewContentStore(internal.ContentStoreOptions(cliContext)...)
```
Kern--, line 50: Before this change, `store.NewContentStore` would set the namespace on the context. If we're not updating the context anymore, we need to set that ourselves or use the context from `internal.NewClient`

This probably applies to some of the other CLI commands as well, so can you double check them all?

soci/store/store.go
```diff
@@ -56,6 +56,11 @@ type Store interface {
 
 type ContentStoreType = config.ContentStoreType
 
+type ContentStoreConfig struct {
+	config.ContentStoreConfig
+	Namespace string
```
Kern--, line 61: Do we still need this here?

sondavidb: Nope, this isn't wired to anything anymore. Got confused because we still need to attach to the namespace in the CLI but looks like we already have it covered [here](https://github.com/awslabs/soci-snapshotter/blob/fadf416a4d5120e340686e298b14688270a8551c/cmd/soci/commands/internal/client.go#L60)

cmd/soci/commands/index/rm.go
sondavidb: Did a walk through and made some changes for this change. In the process I found that we don't actually respect timeouts in a few commands, so I made sure every command honors the `--timeout` flag.

cmd/soci-snapshotter-grpc/main.go
sondavidb: Sure, ended up just moving all this to a new file so it doesn't clutter `main.go` too much. Can reverse this if needed.

cmd/soci-snapshotter-grpc/main.go
sondavidb: Done

Kern-- (dismissed):
LGTM on green

sondavidb:
[TestRestartAfterSigint](https://github.com/awslabs/soci-snapshotter/blob/fadf416a4d5120e340686e298b14688270a8551c/integration/run_test.go#L411-L452) was failing because the context to reload remote snapshots does not have the namespace data. The workaround I did is adding a label to store the namespace before storing the snapshot in our `metadata.db`.

Worth mentioning that our `metadata.db` is not respecting namespaces as-is, which is why we have this issue. If we can fix namespacing in the `metadata.db` we can probably drop this label.

soci/store/store.go
austinvazquez: Thanks I was going to ask how the namespace is set for SOCI CLI create index. :)

austinvazquez (dismissed):

fs/source/source.go
```diff
@@ -85,6 +85,9 @@ const (
 	// TargetSociIndexDigestLabel is a label which contains the digest of the soci index.
 	TargetSociIndexDigestLabel = "containerd.io/snapshot/remote/soci.index.digest"
 
+	// TargetNamespace gives us the namespace that the snapshot was created in
+	TargetNamespace = "contaienrd.io/snapshot/remote/namespace"
```
Kern--, line 89: nit: typo
```suggestion
	TargetNamespace = "containerd.io/snapshot/remote/namespace"
```

Kern-- (approved):

austinvazquez (approved):

fs/source/source.go
sondavidb: oops

PR #1512, opened 2025-04-15
Use functional args for config in integ tests

Kern--:
**Issue #, if available:**
N/A

**Description of changes:**
Before this change, integ tests modeled the SOCI snapshotter config in the string toml form. This change models the config as a config struct and serializes it as needed when running the tests. This should make it easier to override default config without having to expose template variables to every caller.

**Testing performed:**
`make integration`

### DIFF
--- a/config/config.toml
+++ b/config/config.toml
@@ -15,9 +15,7 @@ http_cache_type=""
 filesystem_cache_type=""
 resolve_result_entry=0
 debug=false
-# disable_verification=false
-# Causes TestRunWithDefaultConfig to break, but
-# fine to use in /etc/soci-snapshotter-grpc-config.toml
+disable_verification=false
 max_concurrency=0  # Actually zero
 no_prometheus=false
 mount_timeout_sec=0
--- a/integration/metrics_test.go
+++ b/integration/metrics_test.go
@@ -44,29 +44,20 @@ const (
 	metricsPath        = "/metrics"
 )
 
-const tcpMetricsConfig = `
-metrics_address="` + tcpMetricsAddress + `"
-`
-
-const unixMetricsConfig = `
-metrics_address="` + unixMetricsAddress + `"
-metrics_network="unix"
-`
-
 func TestMetrics(t *testing.T) {
 	tests := []struct {
 		name    string
-		config  string
+		config  snapshotterConfigOpt
 		command []string
 	}{
 		{
 			name:    "tcp",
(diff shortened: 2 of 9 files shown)


integration/metrics_test.go
```diff
@@ -272,9 +262,11 @@ log_fuse_operations = true
 }
 
 func TestFuseOperationCountMetrics(t *testing.T) {
-	const snapshotterConfig = `
-fuse_metrics_emit_wait_duration_sec = 10
-	`
+	var withFuseWaitDuration = func(i int64) snapshotterConfigOpt {
```
sondavidb, line 265: nit: Maybe we could drag this out into `util_test.go` as well?

integration/run_test.go
```diff
@@ -46,22 +45,22 @@ import (
...
 
+	"github.com/awslabs/soci-snapshotter/config"
 	"github.com/awslabs/soci-snapshotter/soci/store"
 	shell "github.com/awslabs/soci-snapshotter/util/dockershell"
 	"github.com/awslabs/soci-snapshotter/util/testutil"
 )
 
 func TestRunWithDefaultConfig(t *testing.T) {
-	b, err := os.ReadFile("../config/config.toml") // example toml file
+	cfg, err := config.NewConfigFromToml("../config/config.toml") // example toml file
```
sondavidb, line 55: Any reason we don't just read the file and use it as the snapshotter TOML? If we read the file and output it to a string we can just not use `getSnapshotterConfigToml`.

nit: While you're here I wonder if we could at least drag the path out as some sort of const

sondavidb (commented):
Overall nothing too major, just want a couple of things looked at before approving

integration/run_test.go
Kern--: Yeah, we can do that. That should fix the weird issue where we had to comment out `disable_verification` in the default config too.

config/config.toml
```diff
@@ -15,7 +15,7 @@ http_cache_type=""
 filesystem_cache_type=""
 resolve_result_entry=0
 debug=false
-# disable_verification=false
+disable_verification=false
 # Causes TestRunWithDefaultConfig to break, but
```
Kern--, line 19: Missed these comments. New version coming soon.

austinvazquez (approved):

sondavidb (approved):

swagatbora90 (approved):

PR #1518, opened 2025-04-17
Consume index annotations

Kern--:
**Issue #, if available:**

**Description of changes:**
This change adds a new mechanism to discover SOCI indexes without the
OCI referrers API via annotations on the image manifest. The new
mechanism is called SOCI v2 while the original is called SOCI v1.
    
By default, SOCI v1 is disabled and SOCI v2 is enabled.

NOTE:
- This needs to be rebased once #1512 is merged
- Integration tests need the corresponding change to add annotations to the image manifests

**Testing performed:**

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -75,8 +75,20 @@ type Config struct {
 }
 type configParser func(*Config)
 
+var parsers = []configParser{parseRootConfig, parseServiceConfig, parseFSConfig}
+
+// NewConfig returns an initialized Config with default values set.
+func NewConfig() *Config {
+	cfg := Config{}
+	ps := append(parsers, defaultPullModes)
+	for _, p := range ps {
+		p(&cfg)
+	}
+	return &cfg
+}
+
 func NewConfigFromToml(cfgPath string) (*Config, error) {
-	cfg := &Config{}
+	cfg := NewConfig()
 	// Get configuration from specified file
 	tree, err := toml.LoadFile(cfgPath)
 	if err != nil && !(os.IsNotExist(err) && cfgPath == defaultConfigPath) {
@@ -85,13 +97,14 @@ func NewConfigFromToml(cfgPath string) (*Config, error) {
 	if err := tree.Unmarshal(cfg); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal config file %q", cfgPath)
 	}
-	parsers := []configParser{parseRootConfig, parseServiceConfig, parseFSConfig}
+	parseConfig(cfg)
+	return cfg, nil
+}
 
+func parseConfig(cfg *Config) {
 	for _, p := range parsers {
 		p(cfg)
 	}
-
-	return cfg, nil
 }
 
 func parseRootConfig(cfg *Config) {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1,0 +1,156 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package config
+
+import (
+	"testing"
+)
+
+func TestConfigDefaults(t *testing.T) {
+	cfg := NewConfig()
+
+	tests := []struct {
+		name     string
+		expected any
+		actual   any
+	}{
+		{
+			name:     "soci v1 enabled",
+			expected: DefaultSOCIV1Enable,
+			actual:   cfg.PullModes.SOCIv1.Enable,
+		},
+		{
+			name:     "soci v2 enabled",
+			expected: DefaultSOCIV2Enable,
+			actual:   cfg.PullModes.SOCIv2.Enable,
+		},
+		{
+			name:     "metrics network",
+			expected: defaultMetricsNetwork,
+			actual:   cfg.MetricsNetwork,
+		},
+		{
+			name:     "cri image service address",
+			expected: DefaultImageServiceAddress,
+			actual:   cfg.CRIKeychainConfig.ImageServicePath,
+		},
+		{
+			name:     "mount timeout",
+			expected: int64(defaultMountTimeoutSec),
+			actual:   cfg.MountTimeoutSec,
+		},
+		{
+			name:     "fuse metric emit wait duration",
+			expected: int64(defaultFuseMetricsEmitWaitDurationSec),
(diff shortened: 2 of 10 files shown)


config/config.go
```diff
@@ -75,8 +75,20 @@ type Config struct {
 }
 type configParser func(*Config)
 
+var parsers = []configParser{parseRootConfig, parseServiceConfig, parseFSConfig}
+
+// NewConfig returns an initialized Config with default values set.
+func NewConfig() *Config {
+	cfg := Config{}
+	ps := append(parsers, defaultPullModes)
```
sondavidb, line 83: Won't this override any of the config put in by the users? Do we maybe want to just give it its own `parsePullModesConfig`?

austinvazquez: It looks like the parsers run twice. Once to apply defaults, then again in `NewConfigFromToml` to apply the user provided configuration from config file.

config/config_test.go
austinvazquez, line 1: Nice thanks for adding this test coverage.

austinvazquez (dismissed):
LGTM, the `fetchSociIndex` implementation reads really well.

config/config.go
Kern--: Yeah, this is weird because the old parsing was:
1. Load from toml
2. Set defaults / fix incompatible changes.

In general, that meant we used zero values as "unset" because by the time we set defaults, we've lost information about whether the value was set to the default explicitly in the toml or not. So in general, we replaces 0 with default value.

This makes it really awkward for default true bools because we don't know whether false is an explicit disable or a missing key (which should default to true). I tried using both custom types / parsers and representing them as strings with special accessors but both make the code harder to read and the semantics of the config harder to understand (e.g. you have to use quotes for bool fields, e.g. `some_config = "true"`).

The solution I landed on is:
1. Initialze default values
2. Load from toml
3. Set defaults / fix incompatible changes.

The trick is that loading from toml doesn't modify any values that aren't present in the config file, so a bool that defaults to true would only become false if the config file set the field to false. I kept the second parse after loading to keep the "replace 0 with default values" semantics from before.

fs/fs.go
```diff
@@ -410,38 +421,107 @@ func (fs *filesystem) fetchSociIndex(ctx context.Context, imageRef, indexDigest,
...
 	}
 
-	var indexDesc ocispec.Descriptor
-	if indexDigest == "" {
-		log.G(ctx).Info("index digest not provided, making a Referrers API call to fetch list of indices")
-		imgDigest, err := digest.Parse(imageManifestDigest)
-		if err != nil {
-			return nil, fmt.Errorf("unable to parse image digest: %w", err)
-		}
+	indexDesc, err := fs.findSociIndexDesc(ctx, imageManifestDigest, indexDigest, remoteStore)
```
swagatbora90, line 424: Naming suggestion: Since with V2, we essentially need to deal with an image index and soci index as well as the image manifests, i think we should follow a naming convention to be clear what we are dealing with. For example, it is not clear if `indexDigest` refers to the soci index Digest or the image index digest. 

First time looking at the code, so may be this is just my learning curve. But using variable names that disamiguate the reference might help with readability.

fs/fs.go
```diff
@@ -410,38 +421,107 @@ func (fs *filesystem) fetchSociIndex(ctx context.Context, imageRef, indexDigest,
...
+	if sociIndexDigest != "" {
+		log.G(ctx).Debug("using provided soci index digest")
+		return parseIndexDigest(sociIndexDigest)
+	}
+	log.G(ctx).Debug("index digest not provided")
+
+	// 2. Try to find an index digest in the manifest labels if SOCI v2 is enabled.
+	if fs.pullmodes.SOCIv2.Enable {
+		log.G(ctx).Debug("checking for soci v2 index annotation")
+		desc, err := findSociIndexDescAnnotation(ctx, imgDigest, remoteStore)
```
swagatbora90, line 454: what if you encounter an error here? should we atleast log that?

Kern--: If the  error is not found, we do log it on line 458. Otherwise we return the error which will eventually get logged.

sondavidb (dismissed):

Shubhranshu153 (dismissed):

fs/fs.go
Kern--: https://github.com/awslabs/soci-snapshotter/issues/1540

config/config_test.go
```diff
@@ -0,0 +1,156 @@
...
+	cfg := NewConfig()
+
+	tests := []struct {
+		name     string
+		expected any
+		actual   any
+	}{
+		{
+			name:     "soci v1 enabled",
+			expected: false,
```
swagatbora90, line 33: Nit: DefaultSOCIV1Enable?

config/config_test.go
```diff
@@ -0,0 +1,156 @@
...
+			actual:   cfg.PullModes.SOCIv1.Enable,
+		},
+		{
+			name:     "soci v2 enabled",
+			expected: true,
```
swagatbora90, line 38: Nit: Nit: DefaultSOCIV2Enable?

fs/fs.go
```diff
@@ -410,38 +421,107 @@ func (fs *filesystem) fetchSociIndex(ctx context.Context, imageRef, indexDigest,
...
+		desc, err := findSociIndexDescReferrer(ctx, imgDigest, remoteStore)
+		if err == nil || !errors.Is(err, ErrNoReferrers) {
+			return desc, err
+		}
+		log.G(ctx).Debug("soci v1 referrers not found")
+	} else {
+		log.G(ctx).Debug("soci v1 is disabled")
+	}
+
+	return ocispec.Descriptor{}, errors.New("no index found or all lazy pull modes are disabled")
```
swagatbora90, line 475: Can we return early if both pull modes are disabled?

fs/fs.go
```diff
@@ -349,6 +359,7 @@ type filesystem struct {
 	mountTimeout                time.Duration
 	fuseMetricsEmitWaitDuration time.Duration
 	pr                          *preresolver
+	pullmodes                   config.PullModes
```
swagatbora90, line 362: Nit: pullModes

fs/fs.go
```diff
@@ -410,38 +421,107 @@ func (fs *filesystem) fetchSociIndex(ctx context.Context, imageRef, indexDigest,
...
+		dg, err := digest.Parse(indexDigestStr)
 		if err != nil {
-			return nil, fmt.Errorf("%w: unable to parse SOCI index digest: %w", snapshot.ErrNoIndex, err)
+			return ocispec.Descriptor{}, fmt.Errorf("unable to parse SOCI index digest: %w", err)
 		}
-		indexDesc = ocispec.Descriptor{
+		return ocispec.Descriptor{
 			Digest: dg,
-		}
+		}, nil
```
swagatbora90, line 512: You can call parseIndexDigest() here

fs/fs.go
Kern--: We already have to check each pull mode. What benefit do we get from an early return?

swagatbora90: The current error message is not clear "no index found or all lazy pull modes are disabled"

swagatbora90 (dismissed):

swagatbora90 (approved):

sondavidb (approved):

PR #1522, opened 2025-04-18
Add SOCI conversion

Kern--:
**Issue #, if available:**

**Description of changes:**
This change adds a command to convert an image to a SOCI-enabled image which contains strong references to both the image and SOCI index in a logically single, deployable artifact.

**Testing performed:**

### DIFF
--- a/cmd/soci/commands/convert.go
+++ b/cmd/soci/commands/convert.go
@@ -20,32 +20,49 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/internal"
 	"github.com/awslabs/soci-snapshotter/config"
 	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/awslabs/soci-snapshotter/soci/store"
+	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/reference"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/urfave/cli"
 )
 
-const (
-	// buildToolIdentifier is placed in annotations of the SOCI index
-	// to help identify how a SOCI index was created
-	buildToolIdentifier = "AWS SOCI CLI v0.2"
-	spanSizeFlag        = "span-size"
-	minLayerSizeFlag    = "min-layer-size"
-	optimizationFlag    = "optimizations"
-	sociIndexGCLabel    = "containerd.io/gc.ref.content.soci-index"
-)
+var ErrInvalidDestRef = errors.New(`the destination image must be a tagged ref of the form "registry/repository:tag"`)
 
-// CreateCommand creates SOCI index for an image
-// Output of this command is SOCI layers and SOCI index stored in a local directory
-// SOCI layer is named as <image-layer-digest>.soci.layer
-// SOCI index is named as <image-manifest-digest>.soci.index
-var CreateCommand = cli.Command{
-	Name:      "create",
-	Usage:     "create SOCI index",
-	ArgsUsage: "[flags] <image_ref>",
+func verifyRef(r string) error {
+	if r == "" {
+		return errors.New("reference cannot be empty")
+	}
+
+	ref, err := reference.Parse(r)
+	if err != nil {
+		return fmt.Errorf("could not parse reference: %w", err)
+	}
+
+	object := ref.Object
+	if object == "" {
+		return errors.New("reference must be tagged")
+	}
+	if strings.Contains(object, "@") {
+		return errors.New("reference must not contain a digest")
+	}
+
+	return nil
+}
+
+// ConvertCommand converts an image into a SOCI enabled image.
+// The new image is added to the containerd content store and can
+// be pushed and deployed like a normal image.
+var ConvertCommand = cli.Command{
+	Name:      "convert",
+	Usage:     "convert an OCI image to a SOCI enabled image",
+	ArgsUsage: "[flags] <image_ref> <dest_ref>",
 	Flags: append(
 		internal.PlatformFlags,
 		cli.Int64Flag{
@@ -67,6 +84,15 @@ var CreateCommand = cli.Command{
 		srcRef := cliContext.Args().Get(0)
 		if srcRef == "" {
 			return errors.New("source image needs to be specified")
+		}
+
+		dstRef := cliContext.Args().Get(1)
+		if dstRef == "" {
+			return errors.New("destination image needs to be specified")
+		}
+		err := verifyRef(dstRef)
+		if err != nil {
+			return fmt.Errorf("%w: %w", ErrInvalidDestRef, err)
 		}
 
 		var optimizations []soci.Optimization
@@ -108,11 +134,6 @@ var CreateCommand = cli.Command{
 			return err
 		}
 
-		ps, err := internal.GetPlatforms(ctx, cliContext, srcImg, cs)
-		if err != nil {
-			return err
-		}
-
 		artifactsDb, err := soci.NewDB(soci.ArtifactsDbPath())
 		if err != nil {
 			return err
@@ -132,25 +153,49 @@ var CreateCommand = cli.Command{
 			return err
 		}
 
-		for _, plat := range ps {
-			batchCtx, done, err := blobStore.BatchOpen(ctx)
-			if err != nil {
-				return err
-			}
-			defer done(ctx)
+		batchCtx, done, err := blobStore.BatchOpen(ctx)
+		if err != nil {
+			return err
+		}
+		defer done(ctx)
 
-			indexWithMetadata, err := builder.Build(batchCtx, srcImg, soci.WithPlatform(plat), soci.WithNoGarbageCollectionLabel())
-			if err != nil {
-				return err
-			}
-
-			if srcImg.Labels == nil {
-				srcImg.Labels = make(map[string]string)
-			}
-			srcImg.Labels[sociIndexGCLabel] = indexWithMetadata.Desc.Digest.String()
-			is.Update(ctx, srcImg, "labels")
+		var platforms []v1.Platform
+		explicitPlatforms := cliContext.StringSlice(internal.PlatformFlagKey)
+		if len(explicitPlatforms) > 0 {
+			platforms, err = internal.GetPlatforms(ctx, cliContext, srcImg, cs)
+		} else {
+			platforms, err = images.Platforms(ctx, cs, srcImg.Target)
+		}
+		if err != nil {
+			return err
 		}
 
-		return nil
+		desc, err := builder.Convert(batchCtx, srcImg,
+			soci.ConvertWithPlatforms(platforms...),
+			// Don't set a GC label on the converted OCI Index. We will create an image
+			// in containerd that will act as the GC root. This way, the OCI index, SOCI indexes, and
+			// images will be removed when the image is deleted in containerd
+			soci.ConvertWithNoGarbageCollectionLabels(),
+		)
+		if err != nil {
+			return err
+		}
+
+		im := images.Image{
+			Name:   dstRef,
+			Target: *desc,
+		}
+		img, err := is.Get(ctx, dstRef)
+		if err != nil {
+			if !errors.Is(err, errdefs.ErrNotFound) {
+				return err
+			}
+			_, err = is.Create(ctx, im)
+			return err
+		}
+		img.Target = *desc
+		_, err = is.Update(ctx, img)
+
+		return err
 	},
 }
--- a/integration/convert_test.go
+++ b/integration/convert_test.go
@@ -1,0 +1,211 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package integration
+
+import (
+	"encoding/json"
+	"slices"
+	"testing"
+
+	"github.com/awslabs/soci-snapshotter/soci"
+	shell "github.com/awslabs/soci-snapshotter/util/dockershell"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+var (
+	convertImages = []string{nginxImage, rabbitmqImage, drupalImage, ubuntuImage}
+)
+
+func validateConversion(t *testing.T, sh *shell.Shell, originalDigest, convertedDigest string) {
+	t.Helper()
+
+	if originalDigest == convertedDigest {
+		t.Fatalf("conversion did not change the digest: %s", originalDigest)
+	}
+
+	var index ocispec.Index
+	content := sh.O("ctr", "content", "get", convertedDigest)
+	err := json.Unmarshal(content, &index)
+	if err != nil {
+		t.Fatalf("failed to decode index: %v", err)
+	}
+	var manifests []ocispec.Descriptor
+	var sociIndexes []ocispec.Descriptor
+	for _, manifest := range index.Manifests {
+		if manifest.ArtifactType == soci.SociIndexArtifactTypeV2 {
+			sociIndexes = append(sociIndexes, manifest)
+		} else if manifest.ArtifactType == "" {
+			manifests = append(manifests, manifest)
+		}
+		// ignore unknown artifacts
+	}
+
+	// We can't verify that manifests and soci indexes are 1-1 because there might be other,
+	// non-soci artifacts in the image (e.g. Docker attestation manifests)
+	if len(manifests) < len(sociIndexes) {
+		t.Errorf("the converted image contains more SOCI indexes than manifests. %d manifests, %d soci indexes", len(manifests), len(sociIndexes))
+	}
+	for _, sociIndexDesc := range sociIndexes {
+		if sociIndexDesc.Annotations == nil {
+			t.Errorf("SOCI index %v has no annotations", sociIndexDesc)
+			continue
+		}
+		if sociIndexDesc.Annotations[soci.IndexAnnotationImageManifestDigest] == "" {
+			t.Errorf("SOCI index %v does not contain image digest", sociIndexDesc)
+		}
+		if sociIndexDesc.Platform == nil {
+			t.Errorf("SOCI index %v does not contain platform", sociIndexDesc)
+			continue
+		}
+
+		manifestIdx := slices.IndexFunc(manifests, func(desc ocispec.Descriptor) bool {
+			return desc.Digest.String() == sociIndexDesc.Annotations[soci.IndexAnnotationImageManifestDigest]
+		})
+		if manifestIdx == -1 {
+			t.Errorf("SOCI index %v does not point to a manifest", sociIndexDesc)
+			continue
+		}
+		manifestDesc := manifests[manifestIdx]
+		if manifestDesc.Annotations == nil {
+			t.Errorf("manifest desc %v has no annotations", manifestDesc)
+			continue
+		}
+		if dg, ok := manifestDesc.Annotations[soci.ImageAnnotationSociIndexDigest]; !ok || dg != sociIndexDesc.Digest.String() {
+			t.Errorf("manifest desc %v does not contain expected soci index digest %v", manifestDesc, sociIndexDesc.Digest)
+		}
+		if manifestDesc.Platform == nil {
+			t.Errorf("manifest desc %v does not contain platform", manifestDesc)
+			continue
+		}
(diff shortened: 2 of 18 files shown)


soci/soci_convert.go
```diff
@@ -0,0 +1,310 @@
...
+
+// annotateImages adds the SOCI index digest to the corresponding image manifest descriptor, modifying `ociIndex` in the process.
+// Adding annotations modifies the image manifest, so annotateImages also pushes the modified image manifests to the blobStore,
+// computes the new image digests, and modified the image descriptors in `ociIndex`
+func (b *IndexBuilder) annotateImages(ctx context.Context, ociIndex *ocispec.Index, sociIndexes []*IndexWithMetadata) error {
+	for i := 0; i < len(ociIndex.Manifests); i++ {
+		manifestDesc := &ociIndex.Manifests[i]
+		// images.Manifest validates the mediatype, no need to do it ourselves like
+		// we did when loading the OCI index
+		manifest, err := images.Manifest(ctx, b.contentStore, *manifestDesc, nil)
```
swagatbora90, line 224: qq: does this return the original manifest? In `newOciIndex`, we copied the contents of the original oci index, but the manifests should still refer to the same digest, right? So if we modify the manifest here, we no longer have access to the original manifest.

cmd/soci/commands/convert.go
```diff
@@ -0,0 +1,166 @@
...
+				return err
+			}
+			optimizations = append(optimizations, optimization)
+		}
+
+		client, ctx, cancel, err := internal.NewClient(cliContext)
+		if err != nil {
+			return err
+		}
+		defer cancel()
```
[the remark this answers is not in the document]
Kern--: no, there's nothing to cancel if there's an error.

cmd/soci/commands/convert.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/urfave/cli"
+)
+
+// ConvertCommand converts an image into a SOCI enabled image.
+// The new image is added to the containerd content store and can
+// be pushed and deployed like a normal image.
+var ConvertCommand = cli.Command{
+	Name:      "convert",
+	Usage:     "convert an OCI image to a SOCI enabled image",
+	ArgsUsage: "[flags] <image_ref> <dest_ref>",
```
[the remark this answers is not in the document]
Kern--: Good point. I'll update this to do more validation that the destination ref has a tag and no digest.

cmd/soci/commands/convert.go
```diff
@@ -0,0 +1,166 @@
...
+
+		artifactsDb, err := soci.NewDB(soci.ArtifactsDbPath())
+		if err != nil {
+			return err
+		}
+
+		builderOpts := []soci.BuilderOption{
+			soci.WithMinLayerSize(minLayerSize),
+			soci.WithSpanSize(spanSize),
+			soci.WithBuildToolIdentifier(buildToolIdentifier),
```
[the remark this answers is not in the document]
Kern--: in `cmd/soci/commands/create.go`. We should probably update it.

soci/soci_convert_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			builder := IndexBuilder{}
+			actual := test.ociIndex
+			var indexes []*IndexWithMetadata
+			for _, index := range test.sociIndexes {
+				indexes = append(indexes, &index)
+			}
+			err := builder.addSociIndexes(context.Background(), &actual, indexes)
```
swagatbora90, line 145: Are you planning to add more test? Specially `annotateImages` method, which seems to be handling a lot of image conversion logic.

soci/soci_convert.go
```diff
@@ -0,0 +1,310 @@
...
+		return nil, errors.New("image does not support any platforms")
+	}
+	if images.IsManifestType(img.Target.MediaType) {
+		// If the image's target descriptor is a single manifest, then it will not
+		// contain a platform because that information is stored in the image config instead.
+		// If we directly use this descriptor in the converted image,
+		// runtimes will not be able to pull the correct manifest.
+		// We know the actual platform by inspecting the image config in `images.Platforms`,
+		// so we add that to the target descriptor.
+		img.Target.Platform = &allPlatforms[0]
```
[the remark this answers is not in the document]
Kern--: Technically yes, I'll add another condition to only set the platform if it's nil.

soci/soci_convert.go
Kern--: Yes, this is intentional. We load the original image manifest, add the annotation, serialize the new manifest and overwrite the digest in the OCI index.

soci/soci_convert.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+	for _, opt := range opts {
+		err := opt(&convertCfg)
+		if err != nil {
+			return nil, err
+		}
+	}
+	convertCfg.platforms = ociutil.DedupePlatforms(convertCfg.platforms)
+
+	// Initialize the OCI Index
```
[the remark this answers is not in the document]
swagatbora90: SGTM!

soci/soci_convert_test.go
Kern--: Not in this PR. All of the other methods have side effects like writing new objects into the blobStore or reading from the content store. I'd rather refactor this to isolate those dependencies rather than mocking them out, but I think that would be a lot for this PR. 

For now, we're validating the rest of the logic via integration tests.

soci/soci_index.go
```diff
@@ -56,10 +64,19 @@ const (
 	// IndexAnnotationDisableXAttrs is the index annotation if the layer has
 	// extended attributes
 	IndexAnnotationDisableXAttrs = "com.amazon.soci.disable-xattrs"
+	// IndexAnnotationImageDigest is the SOCI index descriptor to indicate the digest
+	// of its associated image. This is useful for v2 SOCI indexes which do not contain
+	// a subject field. This annotation goes on a SOCI index descriptor in an OCI index,
+	// not in the SOCI index itself.
+	IndexAnnotationImageDigest = "com.amazon.soci.image-digest"
```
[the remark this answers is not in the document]
Kern--: They do not. 

`IndexAnnotationImageDigest` goes on the SOCI index descriptor and contains the image digest
`ImageAnnotationSociIndexDigest` goes on the image descriptor and contains the image digest.

```
"com.amazon.soci.image-digest"
"com.amazon.soci.index-digest"
```

I agree that it's not visually obvious and could cause confusion. Any suggestions?

cmd/soci/commands/convert.go
```diff
@@ -0,0 +1,166 @@
...
+		srcImg, err := is.Get(ctx, srcRef)
+		if err != nil {
+			return err
+		}
+		spanSize := cliContext.Int64(spanSizeFlag)
+		minLayerSize := cliContext.Int64(minLayerSizeFlag)
+		// Creating the snapshotter's root path first if it does not exist, since this ensures, that
+		// it has the limited permission set as drwx--x--x.
+		// The subsequent oci.New creates a root path dir with too broad permission set.
+		if _, err := os.Stat(config.SociSnapshotterRootPath); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
Kern--: Yes. I created https://github.com/awslabs/soci-snapshotter/issues/1530 to fix this across all the commands.

soci/soci_index.go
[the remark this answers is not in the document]
Kern--: I went with  
```
IndexAnnotationImageManifestDigest = "com.amazon.soci.image-manifest-digest"
```
to make it easier to distinguish.

cmd/soci/commands/convert.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+		if err != nil {
+			return err
+		}
+
+		desc, err := builder.Convert(batchCtx, srcImg,
+			soci.ConvertWithPlatforms(platforms...),
+			// Don't set a GC label on the converted OCI Index. We will create an image
+			// in containerd that will act as the GC root. This way, the OCI index, SOCI indexes, and
+			// iamges will be removed when the image is deleted in containerd
```
coderbirju, line 177: nit: typo

soci/soci_convert.go
```diff
@@ -0,0 +1,313 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return &ociIndexDesc, nil
+}
+
+// buildPlatforms builds a SOCI index for each specified platform
+func (b *IndexBuilder) buildPlatforms(ctx context.Context, img images.Image, platforms []ocispec.Platform) ([]*IndexWithMetadata, error) {
```
Shubhranshu153, line 155: nit: buildSociIndex

soci/soci_convert.go
```diff
@@ -0,0 +1,313 @@
...
+			WithPlatform(platform),
+			// Don't set a GC label on the SOCI indexes because their GC will be attached to
+			// the converted image's OCI index. If we label them, they will not get
+			// gcd with the rest of the converted image
+			WithNoGarbageCollectionLabel(),
+			withIndexVersion(V2),
+		)
+		if err != nil {
+			// Skipping a platform is fine
+			if errors.Is(err, ErrEmptyIndex) {
```
Shubhranshu153, line 168: why skipping is fine here? Like wont it then get pushed and get known while pulling SOCI Image?

soci/soci_convert.go
```diff
@@ -0,0 +1,313 @@
...
+				continue
```
Shubhranshu153, line 169: nit: A debug log here to know we are skipping

soci/soci_convert.go
```diff
@@ -0,0 +1,313 @@
...
+	convertCfg.platforms = ociutil.DedupePlatforms(convertCfg.platforms)
+
+	// Initialize the OCI Index
+	ociIndex, err := b.newOciIndex(ctx, img)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create the SOCI Indexes
+	indexes, err := b.buildPlatforms(ctx, img, convertCfg.platforms)
```
Shubhranshu153, line 113: is build an idempotent process as in if a soci index exist for a platform then it wont re-do and do it for the other platforms. Can there be a use case of potentially removing an platform from SOCI image builds hypothetically.

soci/soci_convert.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	// Apply GC Labels
+	for i, desc := range ociIndex.Manifests {
+		err := store.LabelGCRefContent(ctx, b.blobStore, ociIndexDesc, fmt.Sprintf("m.%d", i), desc.Digest.String())
+		if err != nil {
+			return nil, err
+		}
+	}
+	if convertCfg.gcRoot {
```
Shubhranshu153, line 144: when is this true?

soci/soci_convert.go
```diff
@@ -0,0 +1,313 @@
...
+// 3. Appends the SOCI indexes to the list of manifests in the OCI index
+//
+// Notes:
+// Adding an annotation to an image changes the image digest. This is equivalent to creating a new image.
+// This function will serialize and push the new image manifest to the content store and replaces the original
+// image in the OCI index. The layers will be shared, not duplicated.
+//
+// If the image is a single platform image, this function will create an OCI index so that it can bundle the
+// image and SOCI index into a single artifact.
+func (b *IndexBuilder) Convert(ctx context.Context, img images.Image, opts ...ConvertOption) (*ocispec.Descriptor, error) {
```
Shubhranshu153, line 77: Can source and destination be the same?

cmd/soci/commands/convert.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+		client, ctx, cancel, err := internal.NewClient(cliContext)
+		if err != nil {
+			return err
+		}
+		defer cancel()
+
+		cs := client.ContentStore()
+		is := client.ImageService()
```
Shubhranshu153, line 114: nit: is bit ambiguous.

soci/soci_convert.go
Kern--: `buildSociIndex` is already the name of a function which builds a single index for a specific platforms. I can rename this, but not to that. 

Maybe `buildSociIndexesForPlatforms`?

soci/soci_convert.go
Kern--: Convert always builds SOCI indexes for all platforms. It is idempotent in that if you give the same min layer size, span size, and source image it will produce the same output, but it will redo all the work.

soci/soci_convert.go
Kern--: This is a public API in a library. The CLI always creates an image metdata object in containerd that acts as the GC root, but the library doesn't do that by default and so we can't assume that it's safe to leave it un-rooted.

soci/soci_convert.go
Kern--: Yes, the converted image will get the tag (similar to if you retag latest)

cmd/soci/commands/convert.go
Kern--: This is the convention used in all CLI commands.

integration/convert_test.go
```diff
@@ -0,0 +1,211 @@
...
+	"slices"
+	"testing"
+
+	"github.com/awslabs/soci-snapshotter/soci"
+	shell "github.com/awslabs/soci-snapshotter/util/dockershell"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+var (
+	convertImages = []string{nginxImage, rabbitmqImage, drupalImage, ubuntuImage}
```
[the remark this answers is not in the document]
swagatbora90: Need not block this PR, but I think having test that specifically target a particular type of image might be more helpful

soci/soci_index.go
```diff
@@ -145,17 +211,25 @@ func UnmarshalIndex(b []byte, index *Index) error {
...
+func fromManifest(manifest ocispec.Manifest, index *Index) error {
 	index.MediaType = manifest.MediaType
-	index.ArtifactType = SociIndexArtifactType
+	switch manifest.Config.MediaType {
+	case V1.artifactType:
+		index.ArtifactType = V1.artifactType
+	case V2.artifactType:
+		index.ArtifactType = V2.artifactType
+	default:
+		return fmt.Errorf("unknown index versio: %s", manifest.Config.MediaType)
```
swagatbora90, line 226: typo: version

soci/soci_index.go
```diff
@@ -368,8 +453,9 @@ func NewIndexBuilder(contentStore content.Store, blobStore store.Store, opts ...
 // Returns the SOCI index and its metadata.
 func (b *IndexBuilder) Build(ctx context.Context, img images.Image, opts ...BuildOption) (*IndexWithMetadata, error) {
 	buildCfg := buildConfig{
-		platform: platforms.DefaultSpec(),
-		gcRoot:   true,
+		platform:     platforms.DefaultSpec(),
+		gcRoot:       true,
+		indexVersion: V1,
```
swagatbora90, line 458: Should this be V2 as we are moving to V2 as the default?

util/ociutil/ociutil.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/platforms"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// unknownDocument represents a manifest, manifest list, or index that has not
```
swagatbora90, line 44: typo: UnknownDocument

util/ociutil/ociutil.go
```diff
@@ -0,0 +1,91 @@
...
+// Copied from https://github.com/containerd/containerd/blob/d5534c6014534e02a0893dd6894d88130f45ae02/core/images/image.go#L381
+type UnknownDocument struct {
+	MediaType string          `json:"mediaType,omitempty"`
+	Config    json.RawMessage `json:"config,omitempty"`
+	Layers    json.RawMessage `json:"layers,omitempty"`
+	Manifests json.RawMessage `json:"manifests,omitempty"`
+	FSLayers  json.RawMessage `json:"fsLayers,omitempty"` // schema 1
+}
+
+// validateMediaType returns an error if the byte slice is invalid JSON,
```
swagatbora90, line 55: typo: ValidateMediaType

soci/soci_convert.go
```diff
@@ -0,0 +1,317 @@
...
+		Manifests: []ocispec.Descriptor{
+			img.Target,
+		},
+	}, nil
+}
+
+// annotateImages adds the SOCI index digest to the corresponding image manifest descriptor, modifying `ociIndex` in the process.
+// Adding annotations modifies the image manifest, so annotateImages also pushes the modified image manifests to the blobStore,
+// computes the new image digests, and modified the image descriptors in `ociIndex`
+func (b *IndexBuilder) annotateImages(ctx context.Context, ociIndex *ocispec.Index, sociIndexes []*IndexWithMetadata) error {
```
sondavidb, line 226: Would it be better to just remove the modified index, instead of modifying it in-place? Might make it a little clearer what exactly is being modified

soci/soci_index.go
```diff
@@ -33,6 +33,8 @@ import (
 	"github.com/awslabs/soci-snapshotter/ztoc"
 	"github.com/awslabs/soci-snapshotter/ztoc/compression"
 	"github.com/containerd/containerd/content"
+
+	//"github.com/containerd/containerd/errdefs"
```
sondavidb, line 37: nit: Forgot to remove?

integration/convert_test.go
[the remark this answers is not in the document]
Kern--: I think we could be more intentional about how we select these images, but I think the value is that we're more confident it works generally when we have more examples.

soci/soci_convert.go
```diff
@@ -0,0 +1,317 @@
...
+			indexWithMetadata.Desc.Annotations[IndexAnnotationImageManifestDigest] = manifestDesc.Digest.String()
+		}
+	}
+	return nil
+}
+
+// addSociIndexes modifies the list of manifests in the OCI index to include the SOCI indexes.
+// If the OCI index already contains a SOCI index for the same platform, the old SOCI index is replaced,
+// otherwise, the SOCI index is appended to list of manifests.
+func (b *IndexBuilder) addSociIndexes(_ context.Context, ociIndex *ocispec.Index, sociIndexes []*IndexWithMetadata) error {
```
sondavidb, line 275: Since this is a private function any reason it has to take an unused context var?

soci/soci_index.go
Kern--: No, this is the code path to build a V1 SOCI index.

soci/soci_index.go
```diff
@@ -478,13 +564,19 @@ func (b *IndexBuilder) build(ctx context.Context, img images.Image, buildCfg bui
...
 		Size:      imgManifestDesc.Size,
 	}
+	if buildCfg.indexVersion.version == V2.version {
+		refers = nil
+	}
+
+	index := NewIndex(buildCfg.indexVersion, ztocsDesc, refers, annotations)
 
-	index := NewIndex(ztocsDesc, refers, annotations)
+	//NewIndex(ztocsDesc, refers, annotations)
```
sondavidb, line 573: nit: Forgot to remove?

util/ociutil/ociutil.go
sondavidb, line 1: We have an [idtools](https://github.com/awslabs/soci-snapshotter/tree/main/idtools) package that kind of falls under the same realm of "package we copied from containerd but has internal functions we could use". We should probably move these to a common package in util to help separate their things from ours. (Outside of the scope of this PR but just worth noting IMO.)

sondavidb (dismissed):
Generally LGTM, all my comments are OK to address in a followup PR.

soci/soci_convert.go
Kern--: I was thinking it's easier to take an unused context than to add one later, but you're right that it doesn't matter since this is unexported. 

I think it might make things easier to read in `Convert` if all the steps have the same form, but that's definitely a preference thing.

soci/soci_convert.go
Kern--: Do you mean copy?

I initially wanted all of these functions to return a new index, but indexes have a lot of pointers and arrays and maps, so you would need to be sure to do deep copies to avoid accidentally keeping references. I figured just doing the modifications in place means we don't have to worry about accidentally keeping references because you can't assume the original is unmodified.

integration/convert_test.go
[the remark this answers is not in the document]
Kern--: https://github.com/awslabs/soci-snapshotter/issues/1534

swagatbora90 (approved):
LGTM!

sondavidb (approved):

soci/soci_convert.go
sondavidb: Yeah I kind of see what you mean, I was thinking that since this only modifies the manifests you can have the func take a list of manifests, clone just the manifests list, modify the new list, then return just the manifests, but I can see that being a little clunky to work with unless done carefully.

My main "problem" here is that, aside from the comment, it's not immediately obvious that this modifies the index, but it does make sense in this context. I'm not super opinionated on this, though.

util/ociutil/ociutil.go
Kern--: https://github.com/awslabs/soci-snapshotter/issues/1535

soci/soci_convert.go
Kern--: https://github.com/awslabs/soci-snapshotter/issues/1536

Shubhranshu153 (approved):

PR #1531, opened 2025-04-24
Pass cache options to spancache.Get

Kern--:
**Issue #, if available:**

**Description of changes:**
Before this change, the span manager did not pass cache opts to the span cache when getting keys, but it did pass options when putting keys. This change makes them consistent.

This is problematic with the Direct option because writes are directly to files, but reads use a file descriptor cache. This leads to the following issue:

1. The background fetcher fetches a span, but does not decompress it
2. The background fetcher writes the compressed span to the span cache
3. A later read loads the compressed data from the span cache. This caches the file descriptor
4. The span is decompressed
5. The span is rewritten to the same cache key by opening a temp file, writing the decompressed span data, then renaming it to the original name. NOTE: This is a different file descriptor from step 3!
6. The span is marked decompressed
7. Another read hits in the cache and get's back the fd to the (now deleted) compressed file from step 3.
8. SOCI returns compressed data or get's an unxpected EOF because there isn't as much compressed data as uncompressed data.

The solution here is to pass cache options. Each cache Get opens a new file descriptor, so the caching issue is bypassed.

This was exposed by #1512 because we stopped using the go-toml default values which included `DirectoryCacheConfig.Direct` which was true. We should move the go-toml default values into config parsers because go-toml defaults are deprecated in v2.0. It's a default=true config value which will be easier to add after #1518 

**Testing performed:**

### DIFF
--- a/fs/span-manager/span_manager.go
+++ b/fs/span-manager/span_manager.go
@@ -270,7 +270,7 @@ func (m *SpanManager) getSpanContent(spanID compression.SpanID, offsetStart, off
 		}
 
 		// cache uncompressed span
-		if err := m.addSpanToCache(s.id, uncompSpanBuf, m.cacheOpt...); err != nil {
+		if err := m.addSpanToCache(s.id, uncompSpanBuf); err != nil {
 			return nil, err
 		}
 		if err := s.setState(uncompressed); err != nil {
@@ -328,7 +328,7 @@ func (m *SpanManager) fetchAndCacheSpan(spanID compression.SpanID, uncompress bo
 	}
 
 	// cache span data
-	if err := m.addSpanToCache(spanID, buf, m.cacheOpt...); err != nil {
+	if err := m.addSpanToCache(spanID, buf); err != nil {
 		return nil, err
 	}
 	if err := s.setState(state); err != nil {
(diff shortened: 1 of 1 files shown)


fs/span-manager/span_manager.go
```diff
@@ -409,8 +409,8 @@ func (m *SpanManager) addSpanToCache(spanID compression.SpanID, contents []byte,
 // getSpanFromCache returns the cached span content as an `io.Reader`.
 // `offset` is the offset of the requested contents within the span.
 // `size` is the size of the requested contents.
-func (m *SpanManager) getSpanFromCache(spanID compression.SpanID, offset, size compression.Offset) (io.ReadCloser, error) {
-	rc, err := m.cache.Get(fmt.Sprintf("%d", spanID))
+func (m *SpanManager) getSpanFromCache(spanID compression.SpanID, offset, size compression.Offset, opts ...cache.Option) (io.ReadCloser, error) {
+	rc, err := m.cache.Get(fmt.Sprintf("%d", spanID), opts...)
```
sondavidb, line 413: nit: Since this is only called within the same object can we avoid mutating the func and instead pass `m.opts` here?

Kern--: Sure. I'll fix `addSpanToCache` too.

swagatbora90 (approved):

austinvazquez (approved):

sondavidb (approved):

PR #1546, opened 2025-05-06
Add test for dangling SOCI index v2 reference

Kern--:
**Issue #, if available:**

**Description of changes:**
If an image manifest contains a reference to a SOCI index, but the SOCI index doesn't exist in the registry, the snapshotter should try to use that SOCI index, but then fallback to non-lazy loading when it can't find the SOCI index. This adds a test to verify that.

**Testing performed:**

### DIFF
--- a/integration/pull_modes_test.go
+++ b/integration/pull_modes_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/awslabs/soci-snapshotter/config"
 	"github.com/awslabs/soci-snapshotter/util/dockershell"
 	"github.com/awslabs/soci-snapshotter/util/testutil"
+	"github.com/containerd/containerd/reference"
 	"github.com/containerd/platforms"
 )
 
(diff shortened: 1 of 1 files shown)


integration/pull_modes_test.go
```diff
@@ -252,3 +253,93 @@ func testV1IsNotUsedWhenDisabled(t *testing.T, imgName string) {
 		t.Fatalf("expected no digest, got %s", indexDigestUsed)
 	}
 }
+
+func TestDanglingV2Annotation(t *testing.T) {
```
austinvazquez, line 257: Not a requirement but I've started documenting for some of the lengthier tests the behavior I am working to preserve. The PR description is a pretty good writeup IMO. 

In a perfect world, it should be straightforward to read it from the test itself, but our test framework makes that a little difficult because we are doing some pretty complex stuff.

austinvazquez (approved):

PR #1547, opened 2025-05-06
Test invalid conversions

Kern--:
**Issue #, if available:**

**Description of changes:**
If a SOCI index or image manifest is removed from a converted image, the image should be invalid. This change adds a test to confirm this.

This effectively tests that we are actually putting the SOCI index into the converted Image Index.

**Testing performed:**

### DIFF
--- a/integration/convert_test.go
+++ b/integration/convert_test.go
@@ -18,11 +18,14 @@ package integration
 
 import (
 	"encoding/json"
+	"fmt"
 	"slices"
+	"strings"
 	"testing"
 
 	"github.com/awslabs/soci-snapshotter/soci"
 	shell "github.com/awslabs/soci-snapshotter/util/dockershell"
+	"github.com/containerd/platforms"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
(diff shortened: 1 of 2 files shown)


integration/convert_test.go
```diff
@@ -229,3 +232,72 @@ func TestConvertAndPush(t *testing.T) {
...
+					t.Fatalf("no soci index found")
+				}
+				sh.X("ctr", "content", "delete", index.Manifests[idx].Digest.String())
+			},
+		},
+	}
+
+	for _, imageName := range convertImages {
+		t.Run(imageName, func(t *testing.T) {
+			rebootContainerd(t, sh, "", "")
```
austinvazquez, line 278: Just double checking, this reboot is clearing the cache and needs to occur for each image so no one-time setup is possible?

Kern--: It also starts containerd for the first time. I think we could technically move this out of the loop, but clearing and restarting containerd/soci is not the expensive part of this test.

austinvazquez: Okay yeah thanks was just curious if it was intentional or required. LGTM.

austinvazquez (approved):

integration/util_test.go
```diff
@@ -740,16 +740,28 @@ func generateBasicHtpasswd(user, pass string) ([]byte, error) {
...
 	buffer := new(bytes.Buffer)
 	sh.Pipe(buffer, []string{"ctr", "image", "list", "name==" + ref}, []string{"awk", `NR==2{printf "%s", $3}`})
 	return buffer.String()
 }
 
 func getManifestDigest(sh *shell.Shell, ref string, platform spec.Platform) (string, error) {
-	content := sh.O("ctr", "content", "get", getImageDigest(sh, ref))
-	var index spec.Index
-	err := json.Unmarshal(content, &index)
+	index, err := getImageIndex(sh, ref)
```
swagatbora90, line 764: If the ref points to a manifest here, getImageIndex() would return an error always. Is the logic correct?

Kern--: `getImageDigest` is what you would call on a single manifest.

integration/convert_test.go
```diff
@@ -229,3 +232,72 @@ func TestConvertAndPush(t *testing.T) {
...
+
+			for _, test := range tests {
+				t.Run(test.name, func(t *testing.T) {
+					imageName = fmt.Sprintf("%s/%s", test.repo, imageName)
+					convertedImg := registryConfig.mirror(imageName)
+					sh.X("soci", "convert", "--min-layer-size", "0", img.ref, convertedImg.ref)
+
+					test.modifier(t, sh, convertedImg)
+
+					sh.X("nerdctl", "login", "--username", registryConfig.user, "--password", registryConfig.pass, convertedImg.ref)
```
Shubhranshu153, line 291: nit: may need to make sure the password is not printed in logs when we debug

Shubhranshu153 (approved):

Issue #1536, opened 2025-04-30
[FEATURE] Refactor Convert API


Kern--:
### Description

The Convert API is made up of a series of non-exported functions which apply one step of the conversion process. They are written imperatively with modified inputs, side effects, and subtle data/order dependencies which make them hard to understand. We should refactor these APIs to isolate the side effects and make the core logic easier to test

Kern--:
Another note of potential risk: https://github.com/awslabs/soci-snapshotter/pull/1557#discussion_r2101067395

PR #1557, opened 2025-05-09
Use default platform in convert

Kern--:
**Issue #, if available:**

**Description of changes:**
Before this change, the SOCI cli defaulted to converting all platforms. This is inconsistent with nerdctl's pull, push, and convert functions. It also leads to weird issues like this not working:

```
nerdctl pull docker.io/library/busybox:latest
soci convert docker.io/library/busybox:latest other:latest-soci
```

the solution is either to pull with `--all-platforms` or convert with `--platform`. The latter had an additional bug that convert touched all manifests, even if you didn't convert them.

This change makes SOCI default to single platform conversions and fixes the manifests bug. SOCI convert now behaves like nerdctl pull/convert/push.

**Testing performed:**


integration/convert_test.go
```diff
@@ -216,16 +235,108 @@ func TestConvertAndPush(t *testing.T) {
...
+			t.Fatalf("%s: expected error %q, got nil", name, expectedError)
+		}
+		if !strings.Contains(o, expectedError) {
+			t.Fatalf("%s: expected error %q, got %q", name, expectedError, o)
+		}
+		return true
+	} else if err != nil {
+		t.Fatalf("%s: unexpected error: %v", name, err)
+	}
+	return false
```
Shubhranshu153, line 341: kind of confusing as the name is stating it to validate so successful validation should return true or else false.

Shubhranshu153 (dismissed):

soci/soci_convert.go
```diff
@@ -82,17 +85,26 @@ func (b *IndexBuilder) Convert(ctx context.Context, img images.Image, opts ...Co
...
+	defaultPlatform := platforms.DefaultSpec()
+	if images.IsManifestType(img.Target.MediaType) {
 		// If the image's target descriptor is a single manifest, then it will not
 		// contain a platform because that information is stored in the image config instead.
 		// If we directly use this descriptor in the converted image,
 		// runtimes will not be able to pull the correct manifest.
 		// We know the actual platform by inspecting the image config in `images.Platforms`,
 		// so we add that to the target descriptor.
-		img.Target.Platform = &allPlatforms[0]
+		imagePlatform := allPlatforms[0]
```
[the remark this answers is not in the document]
Kern--: Noted in #1536

integration/convert_test.go
```diff
@@ -216,16 +235,108 @@ func TestConvertAndPush(t *testing.T) {
...
-			sh.X("nerdctl", "push", "--all-platforms", convertedImg.ref)
+			o, err = sh.CombinedOLog(append(push, convertedImg.ref)...)
+			validateErrorOutput(t, "push", string(o), err, test.expectedPushError)
 		})
 	}
 }
+
+func validateErrorOutput(t *testing.T, name string, o string, err error, expectedError string) bool {
+	if expectedError != "" {
+		if err == nil {
```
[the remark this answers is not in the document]
Kern--: Will send a fast follow.

integration/convert_test.go
Kern--: Will send a fast follow.

integration/convert_test.go
```diff
@@ -219,20 +237,109 @@ func TestConvertAndPush(t *testing.T) {
...
+
+	tests := []struct {
+		name                 string
+		pullArgs             []string
+		convert              []string
+		pushArgs             []string
+		expectedConvertError string
+		expectedPushError    string
+	}{
+		{
```
sondavidb, line 251: This is effectively testing all the permutations of converting 3 images yeah? Wonder if we can make it a little more modular with a loop.

```
	var testMatrix [3]string
	mode := []string{"all", "one"}

	for _, i := range mode {
		for _, j := range mode {
			for _, k := range mode {
				temp[0] = i
				temp[1] = j
				temp[2] = k
				t.Run // ...
			}
		}
	}
```

Something along these lines, and then maybe push this out to a helper function. It's also pretty ugly the thought of changing these test cases at all feels a little daunting in its current state.

Shubhranshu153 (approved):

PR #1563, opened 2025-05-13
Add SOCI CLI root flag

austinvazquez:
**Issue #, if available:**
When running unit tests on my local system, I regularly hit test failures for `TestArtifactDB_DoesNotExist`. I found the test setup could be more robust by ensuring db initialization fails to assert the desired behavior.

**Description of changes:**
This change 1) fixes the flaky unit test by ensuring the db initialization will fail during setup and 2) adds the `--root` global flag for the soci command along with refactoring common code between `soci` and `soci-snapshotter-grpc` commands.

**Testing performed:**
Added unit tests.

### DIFF
--- a/cmd/soci/commands/convert.go
+++ b/cmd/soci/commands/convert.go
@@ -19,11 +19,9 @@ package commands
 import (
 	"errors"
 	"fmt"
-	"os"
 	"strings"
 
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/internal"
-	"github.com/awslabs/soci-snapshotter/config"
 	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/awslabs/soci-snapshotter/soci/store"
 	"github.com/containerd/containerd/errdefs"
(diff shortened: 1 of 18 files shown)


cmd/soci/commands/internal/snapshotter.go
```diff
@@ -43,3 +47,25 @@ var SnapshotterFlags = []cli.Flag{
...
+	}
+	return config.DefaultSociSnapshotterRootPath
+}
+
+func EnsureSnapshotterRootPath(c *cli.Context) error {
+	root := SnapshotterRootPathFromContext(c)
+	// Creating the snapshotter's root path first if it does not exist, since this ensures, that
+	// it has the limited permission set as drwx--x--x.
+	// The subsequent oci.New creates a root path dir with too broad permission set.
+	if _, err := os.Stat(root); os.IsNotExist(err) {
```
Kern--, line 63: Can we move this part into the soci package somewhere and call it from the snapshotter as well?

The snapshotter implicitly creates this directory when it creates the socket, but it uses 0700 which I think is more appropriate.

cmd/soci/commands/convert.go
```diff
@@ -118,14 +116,8 @@ var ConvertCommand = cli.Command{
...
-		// Creating the snapshotter's root path first if it does not exist, since this ensures, that
-		// it has the limited permission set as drwx--x--x.
-		// The subsequent oci.New creates a root path dir with too broad permission set.
-		if _, err := os.Stat(config.SociSnapshotterRootPath); os.IsNotExist(err) {
-			if err = os.Mkdir(config.SociSnapshotterRootPath, 0711); err != nil {
-				return err
-			}
-		} else if err != nil {
+
+		if err := internal.EnsureSnapshotterRootPath(cliContext); err != nil {
```
Kern--, line 120: Can we move this into `app.Before` in main.go?

soci/artifacts_test.go
```diff
@@ -98,14 +98,41 @@ func TestGetIndexArtifactEntries(t *testing.T) {
...
+		expected string
+	}{
+		{
+			name:     "default",
+			root:     "",
+			expected: "/var/lib/soci-snapshotter-grpc/artifacts.db",
+		},
+		{
+			name:     "custom",
+			root:     "/tmp",
```
sondavidb, line 114: Isn't /tmp in memory? Any reason we can't just use a tempdir or something? Just want to make sure it gets cleaned up is all.

sondavidb (approved):
LGTM just one small thing I'd want considered for the unit test

soci/artifacts_test.go
austinvazquez: Oh, `ArtifactsDbPath` doesn't actually create any files on disk. It is just some logic to append `artifacts.db` to the provided path or default to `/var/lib/soci-snapshotter-grpc/artifacts.db` if the provided path is empty.

sondavidb: Makes sense, might be worth clearing it up with a different folder name but not worth blocking, LGTM

PR #1564, opened 2025-05-13
Require containerd for soci-snapshotter.service

austinvazquez:
**Issue #, if available:**
For k8s use cases, when SOCI snapshotter is acting as the image endpoint with socket activation, the snapshotter can be started without containerd being up.

**Description of changes:**
This change adds containerd as a startup requirement to ensure when SOCI is started via socket activation that containerd is started first.

If during runtime containerd crashes, SOCI will continue to run albeit some operations like image pulls will fail. This is because when lazy loading, SOCI must continue to handle IO requests for already running containers.

**Testing performed:**

### DIFF
--- a/soci-snapshotter.service
+++ b/soci-snapshotter.service
@@ -29,8 +29,8 @@
 [Unit]
 Description=soci snapshotter containerd plugin
 Documentation=https://github.com/awslabs/soci-snapshotter
-After=network.target
-Before=containerd.service
+After=network.target containerd.service
+Requires=containerd.service
 
 [Service]
 Type=notify


soci-snapshotter.service
```diff
@@ -29,8 +29,8 @@
 [Unit]
 Description=soci snapshotter containerd plugin
 Documentation=https://github.com/awslabs/soci-snapshotter
-After=network.target
-Before=containerd.service
+After=network.target containerd.service
+Requires=containerd.service
```
ginglis13, line 33: In my experience, I only encountered service ordering issues when starting soci-snapshotter with a `RequiredBy=containerd` in the `[Install]` section AND soci configured to use the `containerd` content store:

```
{"error":"cannot create content store: could not connect to containerd socket for 
content store access: failed to dial \"/run/containerd/containerd.sock\": context 
deadline exceeded: connection error: desc = \"transport: error while dialing: 
dial unix:///run/containerd/containerd.sock: timeout\"","level":"fatal",
"msg":"failed to configure filesystem","time":"2025-05-09T22:05:25.537974436Z"}
```

The issue isn't present if using soci's content store. The issue is also fixed by the changes in this PR. It may be worth documenting this discrepancy in more detail somewhere in the project repo.

Kern--: If we add `Requires=containerd.service`, then if containerd goes down, SOCI will go down too. Is that right? Do we want that?

If we're using the containerd content store and containerd goes down, will we be able to re-establish a connection once containerd comes back up?

I wonder if we just shouldn't have any dependency on containerd and recommend socket activation for SOCI by default instead.

PR #1568, opened 2025-05-15
Add test for span arg validation

Kern--:
**Issue #, if available:**

**Description of changes:**
This adds a test that a user cannot specify a min-layer-size or span-size that is negative or larger than what can fit into a signed 64-bit integer.

**Testing performed:**

### DIFF
--- a/integration/create_test.go
+++ b/integration/create_test.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"math"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


austinvazquez:
Not 100% against, but just thinking could these negative tests be unit tests instead?

Kern--:
I think we could, but I wanted to test the UX with the CLI because it can distinguish between negative values and overflows. It would probably be worth

austinvazquez (approved):
Okay yeah not a blocker for me.

Shubhranshu153 (approved):

PR #1571, opened 2025-05-23
Resolve descriptor before unpacking

sondavidb:
**Issue #, if available:**
Partial revert of #1493

**Description of changes:**
In #1493, we introduced a fix where ORAS Resolve calls would fail when pulling images from Public ECR using crictl by implementing our own logic to populate the descriptor size. In doing this we removed "redundant" logic in fs/fs.go, but this removal actually broke the same use case using the ORAS store. (containerd content store doesn't check the expected descriptor size if it is zero, so it passed in this use case.) This is because we would resolve the descriptor in Fetch, but then when storing we would use the unresolved descriptor, which ORAS does not have a special ignore case for if the descriptor size is zero, so it would fail to ingest and error out.

This change re-adds this logic back.

**Testing performed:**
Manual testing via crictl and made sure it worked with my change and failed without.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -393,8 +393,24 @@ func (fs *filesystem) MountLocal(ctx context.Context, mountpoint string, labels 
 		return fmt.Errorf("cannot create fetcher: %w", err)
 	}
 	unpacker := NewLayerUnpacker(fetcher, archive)
+	desc := s.Target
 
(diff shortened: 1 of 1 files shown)


austinvazquez (approved):
LGTM, thanks for finding the issue. I would like to add crictl integration testing which would have helped prevented this. Although that can be an entire effort in its own. For now, let's put the fix back.

PR #1578, opened 2025-05-28
Fix TestOverlayView in userns

### LINKED ISSUE #1573: [Bug] TestOverlayView  may fail with unexpected mount options
### Description

When running tests, it appears that the test `TestOverlayView` may result in the following error:

```
=== RUN   TestOverlayView
    snapshot_test.go:696: expected 1 mount option but got 2
--- FAIL: TestOverlayView (0.07s)
FAIL
```

Unclear. It must be something specific to the test environment.

Please see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1103361 for the full conversation and context.

The test to pass

### Host information

1. OS:  Debian/trixie
2. Snapshotter Version:  0.4.1-4
3. Containerd Version: [1.7.24](https://packages.debian.org/source/testing/containerd)

Applying the following patch:

```
-       if len(m.Options) != 1 {
-               t.Errorf("expected 1 mount option but got %d", len(m.Options))
-       }
+       if len(m.Options) != 1 {
+               t.Fatalf("expected 1 mount option but got %d: %+v", len(m.Options), m.Options)
+       }
```

shows this:

```
=== RUN   TestOverlayView
    snapshot_test.go:696: expected 1 mount option but got 2: [lowerdir=/tmp/TestOverlayView2389272231/001/snapshots/2/fs:/tmp/TestOverlayView2389272231/001/snapshots/1/fs userxattr]
```

I believe the first argument is expected, and the `userxattr` not?

Kern--:
**Issue #, if available:**
Fixes #1573

**Description of changes:**
TestOverlayView assumes that it is not run in a user namespace or that /tmp is a tmpfs. If it is run in a user namespace with a non-tmpfs /tmp, then the mount contains an additional `userxattr` mount option. This change checks if we expect that option and asserts on it as necessary.

**Testing performed:**
```
mkdir tmp
sudo mount --bind tmp /tmp
GO_TEST_FLAGS='-run TestOverlayView' unshare --user --map-current-user make test
```

### DIFF
--- a/snapshot/snapshot_test.go
+++ b/snapshot/snapshot_test.go
@@ -45,6 +45,7 @@ import (
 	"github.com/containerd/containerd/mount"
 	"github.com/containerd/containerd/pkg/testutil"
 	"github.com/containerd/containerd/snapshots"
+	"github.com/containerd/containerd/snapshots/overlay/overlayutils"
 	"github.com/containerd/containerd/snapshots/storage"
 	"github.com/containerd/containerd/snapshots/testsuite"
 	"github.com/containerd/errdefs"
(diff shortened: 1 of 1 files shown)


snapshot/snapshot_test.go
```diff
@@ -709,12 +710,20 @@ func TestOverlayView(t *testing.T) {
...
 	}
-	if len(m.Options) != 1 {
-		t.Errorf("expected 1 mount option but got %d", len(m.Options))
+	requiresUserXattrs, _ := overlayutils.NeedsUserXAttr("/tmp/view2")
+	expectedMountOptionsCount := 1
+	if requiresUserXattrs {
+		expectedMountOptionsCount = 2
+	}
+	if len(m.Options) != expectedMountOptionsCount {
+		t.Errorf("expected %d mount option(s) but got %d: %v", expectedMountOptionsCount, len(m.Options), m.Options)
```
austinvazquez, line 719: +1, thanks for adding the slice output as part of the assertion. :)

PR #1584, opened 2025-06-01
Add imports config

austinvazquez:
**Issue #, if available:**

**Description of changes:**
This change adds the ability to split configuration into multiple files and import them from the root configuration file through `imports` field.

**Testing performed:**


austinvazquez:
Chatted with @Kern-- offline regarding this and the integration test failures. One takeaway from the conversation is it would be good to split the go-toml/v2 upgrade to a seperate PR. Also it is the likely root cause for the integration test failures.

austinvazquez:
Moving this to draft until #1585 is merged.

austinvazquez:
Update on this, I don't think mergo will achieve our goal here. To briefly summarize the issue I am seeing with this implementation, SOCI is having some configuration whose default values do not match the non-Go defaults. This creates an issue where we cannot simply decode a configuration file into a Go Config struct and merge on top of already read configuration.

So instead an alternative approach can be to read and unmarshal each configuration file into a Go map[string]any and walk the tree explicitly setting any nodes in the tree on top of a default Config struct.

config/internal/merge/merge.go
```diff
@@ -0,0 +1,133 @@
...
+		for _, annotation := range annotations {
+			annotation = strings.TrimSpace(annotation)
+			if value, ok = (*src)[annotation]; ok {
+				tomlAnnotation = annotation
+				break
+			}
+		}
+
+		if ok {
+			switch fieldKind {
```
austinvazquez, line 100: Two integration test failures due to not able to merge resolver host configuration. The resolver host configuration is a `map[string]struct{[]struct{}}` so am thinking this switch needs to be more robust to handle nested types like this.

config/fs.go
austinvazquez, line 1: To merge configuration, all fields will be required to have a TOML annotation else it will be ignored. Newer configurations are following snake case; however, for backwards compatibility I also added camel case annotations here so existing configurations continue to work. Thoughts on this approach? Alternatively considered just explicitly having the camel case.

austinvazquez:
@Kern-- , @sondavidb curious if you folks could give this a review.

config/config.go
```diff
@@ -91,20 +97,79 @@ func NewConfig() *Config {
...
+	var (
+		loaded  = map[string]bool{}
+		pending = []string{path}
+	)
+
+	for len(pending) > 0 {
+		path, pending = pending[0], pending[1:]
+
+		// Check if a file at the given path has already been loaded to prevent circular imports.
+		if _, ok := loaded[path]; ok {
```
Kern--, line 113: Do we need to need to canonicalize paths before this?

config/config.go
```diff
@@ -91,20 +97,79 @@ func NewConfig() *Config {
...
-	// Get configuration from specified file
-	if err = toml.NewDecoder(f).Decode(cfg); err != nil {
-		return nil, fmt.Errorf("failed to load config file %q: %w", cfgPath, err)
+// NewConfigFromToml loads the soci-snapshotter service configuration from the provided filepath.
+func NewConfigFromToml(path string) (*Config, error) {
+	mergedCfg := NewConfig()
+
+	var (
+		loaded  = map[string]bool{}
+		pending = []string{path}
```
Kern--, line 106: Do we want to parse the base config via this path? 

I expect that in most cases there will be no imports and so we don't need this merging path. On the other hand, I can see the value in one way of parsing all config rather than a common path and an exception path. What do you think?

config/config.go
```diff
@@ -91,20 +97,79 @@ func NewConfig() *Config {
...
+		cfg := map[string]any{}
+		if err = toml.Unmarshal(data, &cfg); err != nil {
+			return nil, fmt.Errorf("failed to parse config file %q: %w", path, err)
+		}
+
+		if err := merge.Merge(mergedCfg, &cfg); err != nil {
+			return nil, fmt.Errorf("failed to merge config file %q: %w", path, err)
+		}
+
+		if importsVal, ok := cfg["imports"]; ok {
```
Kern--, line 132: Do we want to support recursive imports?

config/config.go
```diff
@@ -91,20 +97,79 @@ func NewConfig() *Config {
...
+					importPaths = append(importPaths, s)
+				}
+			default:
+				return nil, fmt.Errorf("imports must be an array, got %T", importsVal)
+			}
+
+			// Process all import paths
+			for _, s := range importPaths {
+				if strings.Contains(s, "*") {
+					paths, err := filepath.Glob(filepath.Clean(filepath.Join(filepath.Dir(path), s)))
```
Kern--, line 154: What assumptions are we making here about `s`? I don't think this works if the import is absolute. Are there other edge cases?

config/config.go
```diff
@@ -91,20 +97,79 @@ func NewConfig() *Config {
...
+					if err != nil {
+						return nil, fmt.Errorf("failed to resolve import path pattern %q: %w", s, err)
+					}
+					pending = append(pending, paths...)
+				} else {
+					importPath := filepath.Clean(s)
+					if !filepath.IsAbs(importPath) {
+						importPath = filepath.Join(filepath.Dir(path), importPath)
```
Kern--, line 162: Should we allow paths with `..`?

I lean towards no and if you want to access an import that branches somewhere above you in the filesystem, then you should use an absolute path.

config/fs.go
Kern--: TIL you can have multiple values in the toml annotation. 

I think this approach is good. I think the fact that they've been camel case was an accident. I think we should make snake_case the official version (in `config/config.toml` and in the eventual `config dump` command), but support camel case for backwards compatibility like you said.

config/internal/merge/merge.go
```diff
@@ -0,0 +1,230 @@
...
+)
+
+const (
+	structTOMLAnnotation = "toml"
+
+	// Arbitrarily selected based on existing use-case depth. Limiting recursion depth protects against
+	// stack overflows when handling malicious or malformed inputs. Assuming a 400-600 byte implementation
+	// of the merge function and a 1MB stack size, then in the worst case the depth limit would allow
+	// 2-3% consumption of the stack (excluding factors like garbage collection and runtime).
+	maxRecursionDepth = 50
```
Kern--, line 33: I think it would be good to make this even tighter. We could have a test that dumps the default config and then runs it back through the Merge. If it fails with `ErrExceedMaxDepth`, we can increase the recursion depth.

config/internal/merge/merge.go
```diff
@@ -0,0 +1,230 @@
...
+	for i := range dstType.NumField() {
+		fieldVal := dstVal.Field(i)
+		fieldType := dstType.Field(i)
+
+		tomlAnnotation := fieldType.Tag.Get(structTOMLAnnotation)
+		fieldKind := fieldType.Type.Kind()
+
+		if tomlAnnotation == "" {
+			// A struct field with no tag could be an embedded struct that contains more fields at the same map level.
+			if fieldKind == reflect.Struct {
```
Kern--, line 77: What does go-toml do if the field is not embedded? Should we check that the field name is the same as the field type name before recursing?

config/internal/merge/merge.go
```diff
@@ -0,0 +1,230 @@
...
+	maxRecursionDepth = 50
+)
+
+var (
+	ErrDstNotStruct     = errors.New("dst must be a struct or a pointer to a struct")
+	ErrCannotCast       = errors.New("cannot cast value")
+	ErrExceededMaxDepth = fmt.Errorf("exceeded maximum recursion depth of %d", maxRecursionDepth)
+)
+
+func Merge(dst any, src *map[string]any) error {
```
Kern--, line 42: maps are already reference types, does this need to be a double pointer?

config/internal/merge/merge_test.go
```diff
@@ -0,0 +1,516 @@
...
+				t.Fatal("expected error")
+			} else if !reflect.DeepEqual(test.expected, test.dst) {
+				t.Fatalf("invalid config.\nexpected: %v\n  actual: %v", test.expected, test.dst)
+			}
+		})
+	}
+
+	t.Run("negative(dst=value)", func(t *testing.T) {
+		if err := Merge(10, &map[string]any{}); err == nil {
+			t.Fatal("expected error")
```
Kern--, line 471: For negative test cases, we should also assert on the type of error we get back. It's easy to end up with bugs because you get an error, but not the one you expect.

config/internal/merge/merge_test.go
```diff
@@ -0,0 +1,516 @@
...
+				},
+			},
+			expected: &Example{
+				Struct: Struct{
+					IntField: 20,
+				},
+			},
+		},
+		{
+			name: "merge(struct)",
```
Kern--, line 320: Is this the behavior we want? We talked at some point about trying to behave like containerd where when you set a leaf node, it should overwrite the whole struct, not just the specific field. I'm not totally sold that's the right thing to do, but I'd like to understand what the tradeoffs are between the two and why we're going with this specific approach.

config/internal/merge/merge_test.go
```diff
@@ -0,0 +1,516 @@
...
+					},
+				},
+			},
+		},
+		{
+			name:      "negative(src=nil)",
+			dst:       &Example{},
+			src:       nil,
+			expected:  &Example{},
+			expectErr: true,
```
Kern--, line 439: Why not treat a nil map like an empty map instead of erroring?

Kern-- (commented):
Still wrapping my head around the implementation of `mergeWithDepth`. I'll probably have some more comments, but I wanted to get a first draft of review to you.

PR #1585, opened 2025-06-02
Upgrade to github.com/pelletier/go-toml/v2

austinvazquez:
**Issue #, if available:**
Split from #1584 

**Description of changes:**
This change upgrades to github.com/pelletier/go-toml/v2

**Testing performed:**
Added unit tests for two default configurations.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -42,14 +42,15 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/pelletier/go-toml"
+	"github.com/pelletier/go-toml/v2"
 )
 
 const (
 	// DefaultSociSnapshotterRootPath is the default filesystem path for the snapshotter root directory.
 	DefaultSociSnapshotterRootPath = "/var/lib/soci-snapshotter-grpc/"
 
-	defaultConfigPath = "/etc/soci-snapshotter-grpc/config.toml"
+	// DefaultConfigPath is the default filesystem path for the snapshotter configuration file.
+	DefaultConfigPath = "/etc/soci-snapshotter-grpc/config.toml"
 )
 
 type Config struct {
@@ -68,7 +69,7 @@ type Config struct {
 	DebugAddress string `toml:"debug_address"`
 
 	// MetadataStore is the type of the metadata store to use.
-	MetadataStore string `toml:"metadata_store" default:"db"`
+	MetadataStore string `toml:"metadata_store"`
 
 	// SkipCheckSnapshotterSupported is a flag to skip check for overlayfs support needed to confirm if SOCI can work
 	SkipCheckSnapshotterSupported bool `toml:"skip_check_snapshotter_supported"`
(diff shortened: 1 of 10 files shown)


config/config.go
```diff
@@ -79,23 +80,28 @@ var parsers = []configParser{parseRootConfig, parseServiceConfig, parseFSConfig}
...
 	}
-	return &cfg
+
+	return cfg
 }
 
 func NewConfigFromToml(cfgPath string) (*Config, error) {
+	f, err := os.Open(cfgPath)
+	if err != nil {
+		return nil, err
```
pendo324: Should we keep the `failed to load config file %q` error wrapper?

pendo324 (dismissed):

config/config_test.go
```diff
@@ -78,6 +83,11 @@ func TestConfigDefaults(t *testing.T) {
 			expected: int64(defaultFuseTimeoutSec),
 			actual:   cfg.FuseConfig.NegativeTimeout,
 		},
+		{
+			name:     "fuse directory cache direct",
```
Kern--, line 87: Something worth thinking about for the future is that this is a good regression test for existing non-zero defaults, but has no mechanism to ensure new ones get added to the test suite.

Kern-- (approved):
Thanks for this!

config/config.go
austinvazquez: Yeah, let me go ahead and add that to differentiate the open and load errors. It might be important once we support multiple files.

pendo324 (approved):

config/config.go
```diff
@@ -79,23 +80,28 @@ var parsers = []configParser{parseRootConfig, parseServiceConfig, parseFSConfig}
...
 	}
-	return &cfg
+
+	return cfg
 }
 
 func NewConfigFromToml(cfgPath string) (*Config, error) {
+	f, err := os.Open(cfgPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open config file '%q': %w", cfgPath, err)
```
Kern--: nit: `%q` already wraps the string in double quotes.

austinvazquez: Oof, let me get it.

pendo324 (approved):

config/config_test.go
austinvazquez: 👍🏼 , I think this will be to rework the unit test to compare the return value from `NewConfig` and compare it against a known output.

austinvazquez:
Bringing this in to unblock the other PR. We can work to address the remaining work as follow-up.

config/config.go
```diff
@@ -79,23 +80,28 @@ var parsers = []configParser{parseRootConfig, parseServiceConfig, parseFSConfig}
...
+	var initParsers = []configParser{defaultPullModes, defaultDirectoryCacheConfig}
+	for _, p := range append(initParsers, parsers...) {
+		p(cfg)
 	}
-	return &cfg
+
+	return cfg
 }
 
 func NewConfigFromToml(cfgPath string) (*Config, error) {
```
AkihiroSuda, line 94: Seems to have caused a regression:
- https://github.com/awslabs/soci-snapshotter/issues/1621

PR #1588, opened 2025-06-04
Fix SOCI index manifest V2 db references

Kern--:
**Issue #, if available:**
N/A

**Description of changes:**
When we convert images to SOCI index manifest v2, we first create SOCI indexes, then add the annotations to the image manifests. In the process of creating the SOCI indexes, we create an artifacts DB reference to the SOCI index using the original manifest and image ids. This change goes back after everything is done and updates the artifacts DB so the index is associated with the new manifest and image.

**Testing performed:**


soci/soci_convert.go
```diff
@@ -146,6 +146,12 @@ func (b *IndexBuilder) Convert(ctx context.Context, img images.Image, opts ...Co
 		return nil, err
 	}
 
+	// Fix artifacts DB references
```
swagatbora90: nit: isn't it better to call this as `update artifacts DB references`. Using fix gives the impression that something is broken.

soci/artifacts.go
```diff
@@ -418,6 +418,16 @@ func (db *ArtifactsDb) WriteArtifactEntry(entry *ArtifactEntry) error {
...
 }
 
+func (db *ArtifactsDb) fixSociV2ArtifactReference(indexDigest string, manifestDigest string, imageDigest string) error {
+	ae, err := db.GetArtifactEntry(indexDigest)
+	if err != nil {
+		return err
+	}
+	ae.ImageDigest = imageDigest
+	ae.OriginalDigest = manifestDigest
+	return db.WriteArtifactEntry(ae)
```
swagatbora90: Do we care about handling concurrent read/writes here?

soci/soci_convert.go
swagatbora90: same goes for the other references used

Kern--: Fair enough. I'll update that.

soci/artifacts.go
Kern--: I don't think so. The DB is locked from other processes and the individual read/write operations are transactional. So there's technically a TOCTOU between the read and write transactions, but I don't think it's a big risk.

Kern--: After offline discussion: it's easy enough to put these into a transaction so we just don't have any risk.

Shubhranshu153 (approved):

PR #1592, opened 2025-06-05
Prevent pushing SOCI index manifest v2

Kern--:
**Issue #, if available:**
N/A

**Description of changes:**
SOCI index manifest v2 does not have a subject field and instead relies on a connection to an image via an image index. Before this change, you could use `soci push` to push a SOCI index manifest v2 without the rest of the image which is unusable without the image index. This change detects SOCI index manifest v2 and skips the push.

**Testing performed:**
```
$ sudo soci convert --all-platforms --min-layer-size 0 public.ecr.aws/docker/library/busybox:latest public.ecr.aws/docker/library/busybox:latest-soci
$ sudo soci create --min-layer-size 0 public.ecr.aws/docker/library/busybox:latest
$ sudo soci push public.ecr.aws/docker/library/busybox:latest-soci
[WARN] Skippping SOCI index manifest v2. You should push the whole image with higher level tools like finch or nerdctl.
soci: could not find any soci index manifest v1 to push for platform linux/amd64
$ sudo soci push public.ecr.aws/docker/library/busybox:latest
checking if a soci index already exists in remote repository...
pushing soci index with digest: sha256:5ec2624823843d6aeaa3a79cf0ccf3f878b400eca7306f1b1e37bd36aa5f30fa
soci: error pushing graph to remote: failed to perform "Exists" on destination: HEAD "https://public.ecr.aws/v2/docker/library/busybox/blobs/sha256:e4f42ef6191be62c0890258181482b127438843af5651085248fa36848c5f8f0": response status code 401: Unauthorized

$ sudo soci create --min-layer-size 0 public.ecr.aws/docker/library/busybox:latest-soci 
$ sudo soci push public.ecr.aws/docker/library/busybox:latest-soci
[WARN] Skippping SOCI index manifest v2. You should push the whole image with higher level tools like finch or nerdctl.
checking if a soci index already exists in remote repository...
pushing soci index with digest: sha256:99ffe70c7fb431cbf1a281182b8cdb8ea3baa3b67c2cc71b099fd5aa63c66948
soci: error pushing graph to remote: failed to perform "Exists" on destination: HEAD "https://public.ecr.aws/v2/docker/library/busybox/blobs/sha256:e4f42ef6191be62c0890258181482b127438843af5651085248fa36848c5f8f0": response status code 401: Unauthorized
```

### DIFF
--- a/cmd/soci/commands/push.go
+++ b/cmd/soci/commands/push.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"slices"
 	"sort"
 	"strings"
 
@@ -48,6 +49,11 @@ const (
 	// Reference: https://github.com/oras-project/oras-go/blob/850a24737c15927603d1cecddc74c87f5f3377f4/copy.go#L38
 	defaultMaxConcurrentUploads = 3
 )
+
+type pushDescs struct {
+	sociIndexDesc  soci.IndexDescriptorInfo
+	imageIndexDesc *ocispec.Descriptor
+}
 
 // PushCommand is a command to push an image artifacts from local content store to the remote repository
 var PushCommand = cli.Command{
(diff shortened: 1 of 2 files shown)


cmd/soci/commands/push.go
```diff
@@ -194,21 +200,41 @@ if they are available in the snapshotter's local content store.
...
+		var toPush []pushDescs
+		var foundV2Indexes bool
 		for _, platform := range ps {
 			indexDescriptors, imgManifestDesc, err := soci.GetIndexDescriptorCollection(ctx, cs, artifactsDb, img, []ocispec.Platform{platform})
 			if err != nil {
 				return err
 			}
+			v1IndexDescriptors := slices.DeleteFunc(indexDescriptors, func(desc soci.IndexDescriptorInfo) bool {
+				if desc.ArtifactType == soci.SociIndexArtifactTypeV2 {
+					foundV2Indexes = true
```
Shubhranshu153, line 212: the name says v1IndexDescriptors but the variable we check for SOCIIndexV2?

cmd/soci/commands/push.go
```diff
@@ -194,21 +200,41 @@ if they are available in the snapshotter's local content store.
...
+					return true
+				}
+				return false
+			})
 
-			if len(indexDescriptors) == 0 {
-				return fmt.Errorf("could not find any soci indices to push")
+			if foundV2Indexes {
```
Shubhranshu153, line 218: are we imagining scenarios where both v2index and v1index can be found in the same image, if  not skip should immediately return?

cmd/soci/commands/push.go
Kern--: Yes, we need the list of v1 descriptors, but we want to know if we found any v2 so we can give a warning. The logic is if we delete all the v2 indexes we find, we will be left with the v1 indexes.

cmd/soci/commands/push.go
Kern--: You can have both. I think trying to make these mutually exclusive will be hard to eliminate all the edge cases (e.g. if the CLI doesn't know about the v2 indexes, if you want to use v1 but the source image already uses v2, etc)

PR #1607, opened 2025-06-25
Add parallel pull + unpack mode to SOCI

sondavidb:
**Issue #, if available:**

**Description of changes:**
TL;DR this adds a new non-lazy-loaded pull mode that allows for layers to be pulled and unpacked in chunks instead of one chunk per layer.

SOCI has always had a code path for layers that do not have a zTOC, effectively mimicking overlay in its code path. This means that the layer is first fetched, stored if not found in the content store, and then unpacked, all sequentially, and given remote snapshotters' limitations, only one layer at a time can go through these operations. This change adds a new pull mode, `parallel_pull_unpack` which leverages this codepath and modifies it to unpack layers in parallel, and pull in parallel within layers.

An example config is as follows (note this is not necessarily the settings you would want to use for your system):
```
enable=false # Required to enable this
max_concurrent_downloads=-1 # unlimited across images
max_concurrent_downloads_per_image=20 # 20 chunks fetched per image
concurrent_download_chunk_size="8MiB" # Max chunk size requested from upstream repo
max_concurrent_unpacks=-1 # unlimtetd across images
max_concurrent_unpacks_per_image=5 # Unpack 5 layers at a time
discard_unpacked_layers=false # Skip content store ingestion when true, which can improve performance on IOPS-constrained systems
```

The problem with many similar approaches are often the in-memory buffer which makes it very difficult to get meaningful performance gains. This approach instead uses a temporary file as a buffer, offloading the memory to the file and writing to the file in chunks at specified offsets to avoid having to buffer much in memory.

Decompression of each layer is still done sequentially, but the pulls are all done in parallel, thus the decompression process can start sooner for each layer, resulting in large performance gains.

This approach of course has its downsides, particularly for disk-constrained environments, as all writes are buffered to a file. There's room for improvement so this runs better on less-powerful systems, but more powerful systems should expect large performance gains over traditional overlay. (Notably, containerd 2.1 does implement parallel fetching within a layer, but unpacking is still done sequentially, so this approach theoretically should be faster as unpacking also becomes parallelized in this approach.)

**Testing performed:**

### DIFF
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -92,7 +92,7 @@ const (
 	// defaultResponseHeaderTimeoutMsec is the default number of milliseconds before timeout while waiting for response header from a remote endpoint. See `TimeoutConfig.ResponseHeaderTimeout`.
 	defaultResponseHeaderTimeoutMsec = 3_000
 	// defaultRequestTimeoutMsec is the default number of milliseconds that the entire request can take before timeout. See `TimeoutConfig.RequestTimeout`.
-	defaultRequestTimeoutMsec = 30_000
+	defaultRequestTimeoutMsec = 300_000
 
 	// defaults based on a target total retry time of at least 5s. 30*((2^8)-1)>5000
 
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -49,6 +49,11 @@ const (
 	DefaultImageServiceAddress = "/run/containerd/containerd.sock"
 )
 
+// ParallelPullUnpack defaults
+const (
+	Unbounded = -1
+)
+
 // FSConfig defaults
 const (
 	defaultFuseTimeoutSec = 1
@@ -106,4 +111,26 @@ const (
 
 	// DefaultSOCIV2Enable is the default value for whether SOCI v2 is enabled
 	DefaultSOCIV2Enable = true
+
+	// DefaultParallelPullEnable is the default value for whether parallel pull and unpack is enabled
+	DefaultParallelPullUnpackEnable = false
+
+	// Defaults for ParallelPullUnpack.
+	// The default values should mirror default containerd values.
+
+	// defaultMaxConcurrentDownloads sets the default value for how many reads can be done over the network at a time.
+	defaultMaxConcurrentDownloads = Unbounded
+	// defaultMaxConcurrentDownloadsPerImage sets the default value for how many reads a single image can do over the network at a time.
+	// This must be less than or equal to MaxConcurrentDownloads.
+	defaultMaxConcurrentDownloadsPerImage = 3
+
+	// defaultConcurrentDownloadChunkSize sets the default value for the max size, in bytes, of downloaded content per network read.
+	// When unbounded, full layer is downloaded in one request
+	defaultConcurrentDownloadChunkSize = Unbounded
+
+	// defaultMaxConcurrentUnpacks sets the default value for the max number of layers that will be unpacked at once.
+	defaultMaxConcurrentUnpacks = Unbounded
+	// defaultMaxConcurrentUnpacks sets the default value for the max number of layers per image that will be unpacked at once.
+	// This must be less than or equal to MaxConcurrentUnpacks.
+	defaultMaxConcurrentUnpacksPerImage = 1
 )
--- a/config/fs.go
+++ b/config/fs.go
@@ -90,8 +90,9 @@ type DirectoryCacheConfig struct {
 	Direct           bool `toml:"direct"`
 }
 
-func defaultDirectoryCacheConfig(cfg *Config) {
+func defaultDirectoryCacheConfig(cfg *Config) error {
 	cfg.FSConfig.DirectoryCacheConfig.Direct = true
+	return nil
 }
 
 type FuseConfig struct {
@@ -179,7 +180,7 @@ type ContentStoreConfig struct {
 	ContainerdAddress string `toml:"containerd_address"`
 }
 
-func parseFSConfig(cfg *Config) {
+func parseFSConfig(cfg *Config) error {
 	// Parse top level fs config
 	if cfg.MountTimeoutSec == 0 {
 		cfg.MountTimeoutSec = defaultMountTimeoutSec
@@ -197,11 +198,14 @@ func parseFSConfig(cfg *Config) {
 	// Parse nested fs configs
 	parsers := []configParser{parseFuseConfig, parseBackgroundFetchConfig, parseRetryableHTTPClientConfig, parseBlobConfig, parseContentStoreConfig}
 	for _, p := range parsers {
-		p(cfg)
+		if err := p(cfg); err != nil {
+			return err
+		}
 	}
+	return nil
 }
 
-func parseFuseConfig(cfg *Config) {
+func parseFuseConfig(cfg *Config) error {
 	if cfg.FuseConfig.AttrTimeout == 0 {
 		cfg.FuseConfig.AttrTimeout = defaultFuseTimeoutSec
 	}
@@ -213,9 +217,10 @@ func parseFuseConfig(cfg *Config) {
 	if cfg.FuseConfig.NegativeTimeout == 0 {
 		cfg.FuseConfig.NegativeTimeout = defaultFuseTimeoutSec
 	}
+	return nil
 }
 
-func parseBackgroundFetchConfig(cfg *Config) {
+func parseBackgroundFetchConfig(cfg *Config) error {
 	if cfg.BackgroundFetchConfig.FetchPeriodMsec == 0 {
 		cfg.BackgroundFetchConfig.FetchPeriodMsec = defaultBgFetchPeriodMsec
 	}
@@ -230,9 +235,10 @@ func parseBackgroundFetchConfig(cfg *Config) {
 	if cfg.BackgroundFetchConfig.EmitMetricPeriodSec == 0 {
 		cfg.BackgroundFetchConfig.EmitMetricPeriodSec = defaultBgMetricEmitPeriodSec
 	}
+	return nil
 }
 
-func parseRetryableHTTPClientConfig(cfg *Config) {
+func parseRetryableHTTPClientConfig(cfg *Config) error {
 	if cfg.RetryableHTTPClientConfig.TimeoutConfig.DialTimeoutMsec == 0 {
 		cfg.RetryableHTTPClientConfig.TimeoutConfig.DialTimeoutMsec = defaultDialTimeoutMsec
 	}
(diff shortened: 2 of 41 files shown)


config/pull_modes.go
```diff
@@ -37,8 +45,38 @@ type V2 struct {
 	Enable bool `toml:"enable"`
 }
 
-func defaultPullModes(cfg *Config) {
+// ParallelPull modifies behavior for eager image pulls.
+// Set any of the TOML vals to negative to unbound any of these operations.
+type ParallelPullUnpack struct {
```
henry118: hmm.. I didn't notice we had to add so many extra stuff related to parallel mode here. I think it's cleaner to separate the mode flag and the rest of the settings, and move the parsing logic to another file.

fs/parallel_artifact_fetcher.go
```diff
@@ -0,0 +1,307 @@
...
+					defer rc.Close()
+
+					switch {
+					case !isSeekable.Load():
+						errCh <- nil
+					case !ok:
+						isSeekable.Store(false)
+						fallthrough
+					default:
+						errCh <- writeToEntireFile(file, rc)
```
henry118: This seems like a race condition. you may have concurrent go routines reaching line 229 and both end up writing the same file. we should use `compareAndSwap` here to ensure only one go routine successfully sets the flag.

fs/parallel_artifact_fetcher.go
```diff
@@ -0,0 +1,307 @@
...
+		return nil, fmt.Errorf("error truncating temp ingest file at %s: %w", ingestPath, err)
+	}
+
+	eg, egCtx := errgroup.WithContext(ctx)
+	var isSeekable atomic.Bool
+	isSeekable.Store(true)
+	numLoops := f.calcNumLoops(desc.Size)
+
+	for i := range numLoops {
+		if !isSeekable.Load() {
```
henry118, line 175: doubt how effective is this check.. in reality there could be multiple go routines bypass this check and end up performing the download, even not seekable. not a bug but could still impact performance. a better approach is probably set this flag before this for loop. could be follow up task later.

fs/parallel_unpacker.go
```diff
@@ -0,0 +1,143 @@
...
+			defer ingestReader.Close()
+		}
+
+		errGroup.Go(func() error {
+			ticker := time.NewTicker(defaultRetryPeriod)
+			defer ticker.Stop()
+
+			err := lu.fetcher.Store(ctx, desc, ingestReader)
+			// Loop every half second until we can get the lock
+			for errdefs.IsUnavailable(err) {
```
henry118: We need a timeout here.

fs/parallel_unpacker.go
```diff
@@ -0,0 +1,143 @@
...
+			if err := errGroup.Wait(); err != nil {
+				return err
+			}
+
+			rc, _, err = lu.fetcher.Fetch(ctx, desc)
+			if err != nil {
+				return fmt.Errorf("cannot fetch layer: %w", err)
+			}
+			defer rc.Close()
+		}
```
henry118: Not sure I understand this piece. Isn't it effectively making the content store writing and layer unpacking sequential?

fs/artifact_fetcher.go
```diff
@@ -202,7 +250,7 @@ func (f *artifactFetcher) resolve(ctx context.Context, desc ocispec.Descriptor)
 // Store takes in an descriptor and io.Reader and stores it in the local store.
 func (f *artifactFetcher) Store(ctx context.Context, desc ocispec.Descriptor, reader io.Reader) error {
 	err := f.localStore.Push(ctx, desc, reader)
-	if err != nil {
+	if err != nil && !errdefs.IsAlreadyExists(err) && !errors.Is(errdef.ErrAlreadyExists, err) {
```
henry118: The last `errors.Is` has wrong arg order

fs/parallel_unpacker.go
henry118: In fact we can remove the entire for loop here since [OpenWriter](https://github.com/containerd/containerd/blob/99df502341b96dac70160bf8d8070c1cf09e6373/core/content/helpers.go#L131) already handles the "Unavailable" error.

fs/parallel_unpacker.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	defer release()
+
+	var errGroup errgroup.Group
+	if !local && !lu.discardUnpackedLayers {
+		var fetchLocalContentStoreReader bool
+
+		ingestReader, err := lu.controller.GetUnpackIngestReader()
+		if err != nil && errors.Is(err, ErrLayerIngestDoesNotExist) {
+			// Compressed layer is not on disk; so use the remote reader to download and store it in the content store.
```
henry118: Do you really see this happening before? I think it's ok to fail here instead of performing another layer fetch.

fs/parallel_unpacker.go
sondavidb: Artifact from earlier versions of the code. We can remove a lot of this logic and just store them in parallel

sondavidb:
Addressed comments. Next push will be rebasing to ensure that our merge conflicts are solved.

config/defaults.go
```diff
@@ -99,11 +104,33 @@ const (
 	defaultMaxWaitMsec = 300_000
 
 	// DefaultContentStore chooses the soci or containerd content store as the default
-	DefaultContentStoreType = "containerd"
+	DefaultContentStoreType = "soci"
```
henry118: revert this back to "containerd"

config/fs.go
```diff
@@ -277,15 +288,17 @@ func parseBlobConfig(cfg *Config) {
...
+	return nil
 }
 
-func parseContentStoreConfig(cfg *Config) {
+func parseContentStoreConfig(cfg *Config) error {
 	if cfg.ContentStoreConfig.Type == "" {
-		// The snapshotter's default content store is soci until we're confident in the containerd integration
-		cfg.ContentStoreConfig.Type = SociContentStoreType
-	} else if cfg.ContentStoreConfig.Type == ContainerdContentStoreType && cfg.ContentStoreConfig.ContainerdAddress == "" {
+		cfg.ContentStoreConfig.Type = DefaultContentStoreType
```
henry118: The default store should be "SOCI" here.

integration/pull_test.go
```diff
@@ -872,3 +872,134 @@ func TestFullLayerRead(t *testing.T) {
...
+		opts []snapshotterConfigOpt
+	}{
+		{
+			name: "no parallel fetching",
+		},
+		{
+			name: "set chunk size",
+			opts: []snapshotterConfigOpt{
+				func(cfg *config.Config) {
+					cfg.PullModes.Parallel.ConcurrentDownloadChunkSize = 1_000_000
```
henry118, line 890: need to explicitly specify "containerd" content store for these tests.

config/fs.go
sondavidb: Changed and modified this to have a better clarifying comment.

soci/store/store.go
```diff
@@ -181,32 +200,32 @@ func (s *SociStore) BatchOpen(ctx context.Context) (context.Context, CleanupFunc
...
-
+func NewContainerdStore(storeConfig ContentStoreConfig) (*ContainerdStore, error) {
 	containerdStore := ContainerdStore{
-		client: client,
+		ContentStoreConfig: storeConfig,
 	}
 
-	containerdStore.ContentStoreConfig = storeConfig
+	if storeConfig.client == nil {
+		return nil, fmt.Errorf("containerd client is not provided; use WithClient to set it")
```
henry118: It looks like this func is also used by soci cli, which won't pass in the client in. So we should provide a default implementation here - just for the CLI case.

integration/log_test.go
```diff
@@ -0,0 +1,115 @@
...
+			opts: []snapshotterConfigOpt{withParallelPullMode()},
+		},
+		{
+			name: "unbounded parallel image pull",
+			opts: []snapshotterConfigOpt{withParallelPullMode(), withUnboundedPullUnpack()},
+		},
+		{
+			name: "discard unpacked layers",
+			opts: []snapshotterConfigOpt{withDiscardUnpackedLayers()},
+		},
```
henry118: Need explicitly specify containerd store here.

soci/store/store.go
henry118: Actually we can just add WithClient to [this func](https://github.com/awslabs/soci-snapshotter/blob/main/cmd/soci/commands/internal/store.go#L27-L32)

sondavidb: Done, think I did it right

integration/pull_test.go
sondavidb: Done

soci/store/store.go
sondavidb: Oh, just saw this new comment. I think it might be a little easier to work with if we just have nil instantiate a new client.

integration/log_test.go
sondavidb: Done here, did a check and I think all the other parallel pull test cases use that too.

soci/store/store.go
```diff
@@ -181,32 +200,38 @@ func (s *SociStore) BatchOpen(ctx context.Context) (context.Context, CleanupFunc
...
+func NewContainerdStore(storeConfig ContentStoreConfig) (*ContainerdStore, error) {
 	containerdStore := ContainerdStore{
-		client: client,
+		ContentStoreConfig: storeConfig,
 	}
 
-	containerdStore.ContentStoreConfig = storeConfig
+	if containerdStore.client == nil {
+		containerdStore.client = func() (*containerd.Client, error) {
+			containerdClient, err := containerd.New(containerdStore.ContainerdAddress)
```
henry118: This will create a new client instance for each method executed for the store. We need to a global singleton as well, same as the implementation in FS. Make sure to guard it with locks too.

sondavidb: AFAIK every call to the CLI is stateless (it definitely does not require the daemon to be running). I'm not sure that a global client will work for the CLI use case.

henry118: I am talking about the methods of the store itself. The CLI could invoke multiple methods for a single command, like `Exists` followed by `Fetch()`. It will results in multiple client object being created.

sondavidb: Oh ok I understand now. Sure let me see the best way we can do this.

sondavidb: Backtracked a little bit but finally got to a solution I'm kind of happy with. TL;DR made a struct that wraps a client and a mutex that you access with a call to `Client()` which will either instantiate the client or get what's already populated. I scrapped the initial implementation because the global variable solution made it unclear to me exactly what the lifecycle of how populated each variable was and felt a little weird to pass around with having each function manage its own lifecycle. By letting them create the new type with just a function call with a containerd socket address I think this makes it a little easier to deal with. LMK if you think this is sufficient.

soci/store/store.go
```diff
@@ -73,34 +73,83 @@ const (
...
+}
+
+func ErrCouldNotCreateClient(address string) error {
+	return fmt.Errorf("could not create containerd client at %s", address)
+}
+
+// This struct allows SOCI to create a connection to the containerd on-demand
+// instead of on startup, allowing the daemon to start without containerd
+type ContainerdClient struct {
+	ContainerdAddress string
```
sondavidb: Maybe this should be unexported

henry118: Yes

soci/store/store.go
henry118: It works.

sondavidb:
Nice, CI is green now ([link](https://github.com/awslabs/soci-snapshotter/actions/runs/15917682040/job/44898278030)). Will do one more force push to not export the variable in the new struct but all integration tests pass now

henry118 (approved):

Kern-- (approved):

sondavidb:
Merged. Thanks all

PR #1630, opened 2025-07-01
Only instantiate parallel pull structs if enabled

sondavidb:
**Issue #, if available:**

**Description of changes:**
There is currently a cosmetic bug where we will always print the log line "Parallel image pull enabled" even if it is not. This is because when creating the FS we always instantiate structs related to parallel pull despite us having the knowledge that we will not be using these structs. We should avoid these unnecessary instantiations by checking the pull modes, which will then remove this confusing log message from the non-parallel pull mode.`

**Testing performed:**


fs/adaptive_fetch_image_layers.go
```diff
@@ -157,6 +157,9 @@ func newUnpackJobs(ctx context.Context, parallelConfig *config.ParallelConfig, s
 }
 
 func checkParallelPullUnpack(cfg *config.ParallelConfig) error {
+	if cfg == nil {
```
henry118, line 162: IMO it's clearer to change the func param to `config.Parallel` and check the `enabled` bool flag here.

fs/fs.go
```diff
@@ -400,6 +412,10 @@ type filesystem struct {
 }
 
 func (fs *filesystem) MountParallel(ctx context.Context, mountpoint string, labels map[string]string, mounts []mount.Mount) error {
+	if fs.inProgressImageUnpacks == nil {
```
henry118: Can we check the `enabled` flag of parallel mode here instead?

fs/fs.go
```diff
@@ -684,6 +700,10 @@ func (fs *filesystem) getImageManifest(ctx context.Context, dgst string) (*ocisp
 // CleanImage stops all parallel operations for the specific image.
 // Generally this will be called when removing a snapshot for an image.
 func (fs *filesystem) CleanImage(ctx context.Context, imgDigest string) error {
+	if fs.inProgressImageUnpacks == nil {
```
henry118: same here?

fs/adaptive_fetch_image_layers.go
sondavidb: Sure, this required mutating a good amount of function headers but I do think it's right. I was initially thinking the nil check would be better for unit testing but a more realistic case is that the default values are there but the pull mode is disabled. Pushed change.

henry118 (approved):

Kern-- (approved):

PR #1631, opened 2025-07-01
Use own fetcher for ranged GETs

sondavidb:
**Issue #, if available:**

**Description of changes:**
To avoid creating our own fetcher, we have been using the oras-go Repository implementation, which handles a lot of the logic of fetching. The parallel pull mode that was introduced additionally leveraged its Seek call to seek the file at a specific offset, however its implementation will close the http.Response body, which is both an extra round trip and also closes the connection for HTTP 1.1 connections, forcing a lot of extra overhead.

By using our own ranged fetch (which we already have, just abstracted away into the lazy-loaded path, which we should really unify...) we can reduce the overhead for ahead-of-time pulls.

**Testing performed:**

### DIFF
--- a/fs/artifact_fetcher.go
+++ b/fs/artifact_fetcher.go
@@ -126,40 +126,98 @@ func (r *orasBlobStore) Resolve(ctx context.Context, reference string) (ocispec.
 func (r *orasBlobStore) Fetch(ctx context.Context, target ocispec.Descriptor) (io.ReadCloser, error) {
 	rc, err := r.Repository.Fetch(ctx, target)
 	if err != nil {
-		switch retErr := err.(type) {
-		// Redact URLs from ORAS errors, as they might have sensitive info cached
-		case *errcode.ErrorResponse:
-			socihttp.RedactHTTPQueryValuesFromURL(retErr.URL)
-			return nil, retErr
-		// Eat URL errors as a malformed URL might still have credentials.
-		case *url.Error:
-			return nil, errors.New("URL error during fetch")
-		// Otherwise it should be safe to print
-		default:
-			return nil, err
-		}
+		return nil, cleanFetchErrors(err)
 	}
 	return rc, nil
+}
+
+// GetContentWithRange gets the requested content in the byte range [lower, upper]
+func GetContentWithRange(ctx context.Context, realURL string, rt http.RoundTripper, lower, upper int64) (*http.Response, error) {
+	if lower < 0 || upper < lower {
+		return nil, fmt.Errorf("illogical content range [%d, %d]", lower, upper)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, realURL, nil)
+	if err != nil {
+		return nil, err
+	}
+	r := fmt.Sprintf("bytes=%d-%d", lower, upper)
+	req.Header.Set("Range", r)
+
+	resp, err := rt.RoundTrip(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusPartialContent {
+		return resp, nil
+	}
+	return nil, fmt.Errorf("error getting range: %v", err)
+}
+
+// FetchRange returns the response body of the range requested.
+// This assumes the upstream repo supports ranged GET calls.
+// If it does not, return an error.
+// TODO: Unify this with the artifact fetching done in fs/remote/resolver.go
+func (r *orasBlobStore) FetchRange(ctx context.Context, reference string, lower, upper int64) (io.ReadCloser, error) {
+	ref, err := registry.ParseReference(reference)
+	if err != nil {
+		return nil, err
+	}
+
+	tr := &clientWrapper{r.Client}
+	realURL := sociremote.CraftBlobURL(reference, ref)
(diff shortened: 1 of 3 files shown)


sondavidb:
It would be good to test this change but as it stands we can only test if this doesn't break the current parallel pull mode. ~~Possibly a good case to expand our testing suite capabilities.~~ Actually, I think we do have this capability but we need to find a Docker image that lacks this capability.

fs/parallel_artifact_fetcher.go
```diff
@@ -185,58 +184,61 @@ func (f *parallelArtifactFetcher) fetchFromRemoteAndWriteToTempDir(ctx context.C
...
+				// If this layer does not support ranged GET, it is very likely
+				// all other layers of this image do not either.
+				supportRangeGet, err = rs.doInitialFetch(ctx, f.constructRef(desc))
 				if err != nil {
 					return nil, fmt.Errorf("error doing initial authorization for layer: %w", err)
 				}
+			} else {
+				supportRangeGet = false
 			}
 		}
```
henry118: This can be moved before the for loop. this way we can split the range get and regular download implementations into separate routines to simplify this long func. wdyt?

fs/artifact_fetcher.go
```diff
@@ -126,40 +126,98 @@ func (r *orasBlobStore) Resolve(ctx context.Context, reference string) (ocispec.
...
+
+	tr := &clientWrapper{r.Client}
+	realURL := sociremote.CraftBlobURL(reference, ref)
+	resp, err := GetContentWithRange(ctx, realURL, tr, lower, upper)
+	if err != nil {
+		return nil, cleanFetchErrors(err)
+	}
+
+	// Check if upstream allows for ranged GET requests
+	if rangeUnit := resp.Header.Get("Accept-Ranges"); rangeUnit != "bytes" {
```
henry118, line 174: Wonder if the server always sends this header even for range-get requests?

sondavidb: I copied this from the oras-go implementation. I should probably double-check this implementation...

fs/parallel_artifact_fetcher.go
sondavidb: I think I like the idea — let me try to refactor this a little

sondavidb: Did a refactor that I was kind of happy with, LMK if this is clearer.

fs/parallel_artifact_fetcher.go
```diff
@@ -166,79 +157,87 @@ func (f *parallelArtifactFetcher) fetchFromRemoteAndWriteToTempDir(ctx context.C
...
+func (f *parallelArtifactFetcher) multiRequestFetchWrite(ctx context.Context, desc ocispec.Descriptor, file *os.File) error {
+	eg, egCtx := errgroup.WithContext(ctx)
+	reference := f.constructRef(desc)
+
+	blobStore, ok := f.remoteStore.(*orasBlobStore)
+	if !ok {
+		return errors.New("did not pass orasBlobStore type to parallel fetcher")
+	}
+
+	for i := range f.calcNumLoops(desc.Size) {
```
henry118: this is already calculated before. can be passed in as a parameter

sondavidb: Sure makes sense, I didn't at first because it's part of the artifact fetcher but I guess it doesn't make sense to manually calculate this twice.

sondavidb:
Rebased from main to fix merge conflict and addressed comment

sondavidb:
Fixed merge conflicts

fs/artifact_fetcher.go
```diff
@@ -124,40 +124,98 @@ func (r *orasBlobStore) Resolve(ctx context.Context, reference string) (ocispec.
...
+	if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusPartialContent {
+		return resp, nil
+	}
+	return nil, fmt.Errorf("error getting range: %v", err)
+}
+
+// FetchRange returns the response body of the range requested.
+// This assumes the upstream repo supports ranged GET calls.
+// If it does not, return an error.
+// TODO: Unify this with the artifact fetching done in fs/remote/resolver.go
```
Kern--, line 159: Do we have an issue or plan for this?

Kern-- (approved):

fs/artifact_fetcher.go
sondavidb: I want to do this whenever we do our overall FS refactoring. My initial scoping of this problem was that this would be a pretty significant change but the more I think about it the more I feel like it can't be terribly difficult.

sondavidb: #1683

henry118 (approved):

fs/artifact_fetcher.go
sondavidb: ([ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Ranges))

Server sends the `Accept-Ranges` header to advertise ranged GET capabilities. Acceptable header values here are `bytes` or `none`. If it does not have the capabilities, this header will either not be populated or will be `none`. So, this header will never be sent with the value `bytes` unless the server is advertising that it is capable of supporting ranged GETs.

Given this, the implementation here seems to do what is intended.

PR #1636, opened 2025-07-03
Add blank config integ test

### LINKED ISSUE #1623: [FEATURE] Add testing for config present/not present
### Description

#1621 we had a bug where using no config previously worked but no longer worked if the config path was set to default

Integration testing for this scenario

_No response_

_No response_

sondavidb:
**Issue #, if available:**
Fixes #1623 

**Description of changes:**
Adding tests for behavior of snapshotter startup with `--config` flag present. In the process I also moved another startup test to this file (as a separate commit)

**Testing performed:**
`GO_TEST_FLAGS='-run TestStartWithConfigPath'  make integration`

### DIFF
--- a/integration/startup_test.go
+++ b/integration/startup_test.go
@@ -29,6 +29,9 @@ import (
 	"github.com/rs/xid"
 )
 
+// example toml file
+const defaultConfigFileLocation = "../config/config.toml"
+
 // Use a custom metrics config to test that the snapshotter
 // correctly starts up when the metrics address is next to the socket address.
 // This tests a regression with the first implementation of systemd socket activation
@@ -182,4 +185,18 @@ func TestSnapshotterStartupWithBadConfig(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestStartWithDefaultConfig(t *testing.T) {
+	defaultConfigToml, err := os.ReadFile(defaultConfigFileLocation)
+	if err != nil {
+		t.Fatalf("error fetching example toml: %w", err)
+	}
+
+	sh, c := newSnapshotterBaseShell(t)
+	defer c()
+
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false), string(defaultConfigToml))
+	// This will error internally if it fails to boot. If it boots successfully,
+	// the config was successfully parsed and snapshotter is running
 }
--- a/integration/startup_test.go
+++ b/integration/startup_test.go
@@ -17,7 +17,9 @@
 package integration
 
 import (
+	"encoding/json"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


integration/startup_test.go
```diff
@@ -137,3 +142,101 @@ func TestSnapshotterSystemdStartup(t *testing.T) {
...
+		{
+			name:        "should start if config file is present in non-default specified location",
+			defaultPath: false,
+			fileExists:  true,
+		},
+		{
+			name:        "should not start if config file is not present in non-default specified location",
+			defaultPath: false,
+			fileExists:  false,
+			expectErr:   true,
```
Kern--: I think we should try to use error matchers (or strings that should appear in the error message) rather than bools. It's easy for something completely unrelated to fail and not notice because we look for any error.

sondavidb: Yeah good point. Let me address this.

integration/startup_test.go
```diff
@@ -137,3 +143,141 @@ func TestSnapshotterSystemdStartup(t *testing.T) {
 		})
 	}
 }
+
+func TestStartWithDefaultConfig(t *testing.T) {
+	defaultConfigToml, err := os.ReadFile(defaultConfigFileLocation)
+	if err != nil {
+		t.Fatalf("error fetching example toml")
```
Kern--: nit: we should log the error to help debug if this happens.

integration/startup_test.go
```diff
@@ -137,3 +143,141 @@ func TestSnapshotterSystemdStartup(t *testing.T) {
...
+				sh.X("touch", configPath)
+			}
+
+			outR, errR, err := sh.R(snRunCmd...)
+			if err != nil {
+				t.Fatalf("failed to create pipe: %v", err)
+			}
+			reporter := testutil.NewTestingReporter(t)
+			m := testutil.NewLogMonitor(reporter, outR, errR)
+			errMatchCh := make(chan bool, 1)
```
Kern--: nit: not blocking, but a lot of the other tests where we do this just assign to a captured variable like:

```
var foundExpectedError bool
m.Add("find expected error", func( rawL string) {
    if strings.Contains(rawL, expectedError) {
        foundExpectedError = true
   }
})
```

sondavidb:
Rebased to fix merge conflicts. Actual changes next commit.

integration/startup_test.go
sondavidb: Sure, I did also just realize the channel approach means only the first fatal message will be logged, which if there's more than one we want to evaluate all of them. (Though, I don't know why this would ever happen?)

Updated

sondavidb:
Fixed the god awful logic I had before, not sure what I was thinking when I initially wrote this but hopefully this is a lot more readable now 😌

coderbirju (approved):

PR #1638, opened 2025-07-08
Remove image after failing to rebase

sondavidb:
**Issue #, if available:**

**Description of changes:**
If an image is unable to be claimed, the associated image pull with this request will fail, so it would make sense to also remove the rest of the jobs associated with this image.

**Testing performed:**


henry118 (approved):

henry118 (dismissed):

sondavidb:
So some additional context, @henry118 did some additional digging on this bug and found that the real culprit is that this could happen if the manifest was already loaded but the layers are unable to be accessed (e.g. removing all layers after the manifest is fetched then re-fetching without authentication). The root cause here is that the imageJob is created and never removed, which then does not allow subsequent pulls to succeed as the image job is never cleared from memory. So the recent force-push tries to be more vigilant about removing the image job on any failure after it is created.

fs/fs.go
```diff
@@ -477,6 +477,11 @@ func (fs *filesystem) preloadAllLayers(ctx context.Context, desc ocispec.Descrip
 	premountCtx, cancel := context.WithCancelCause(context.Background())
 	premountCtx = namespaces.WithNamespace(premountCtx, ns)
 	imageJob := fs.inProgressImageUnpacks.GetOrAddImageJob(imageDigest, cancel)
+	defer func() {
```
Kern--: nit: this seems error prone, especially without an integ test. Below this line, it's important that err is always assigned and never redeclared. Maybe consider refactoring this to be a normal function call + error check?

fs/fs.go
```diff
@@ -496,7 +501,8 @@ func (fs *filesystem) preloadAllLayers(ctx context.Context, desc ocispec.Descrip
 				}
 			}
 			if startPremounting {
-				layerJob, err := fs.inProgressImageUnpacks.AddLayerJob(imageJob, l.Digest.String())
+				var layerJob *layerUnpackJob
+				layerJob, err = fs.inProgressImageUnpacks.AddLayerJob(imageJob, l.Digest.String())
 				if err != nil {
 					return fmt.Errorf("error adding layer job: %w", err)
```
Kern--: Since we don't assign this new error to the err object, we don't get this additional context in `RemoveImageWithError`. Is that a problem?

fs/fs.go
sondavidb: This is kind of a pattern we've already done but I agree that it's not a good one to follow. Maybe a good argument to refactor some of our other funcs...

Let me do this change.

fs/fs.go
sondavidb: I think we do? At least we never reassign the error afterwards. (But refactor removes this issue anyway.)

henry118 (approved):

Kern-- (approved):

sondavidb:
Would be nice if we could get a good way to test this behavior...

The best way I can think of recreating this bug might be as follows:
1. Pull image from registry (auth and everything)
2. Remove all layers and snapshots, but keep the manifest
4. Logout of nerdctl
5. Pull without credentials

This way we have the manifest which will tell us what to load but then we won't have the needed credentials to pull from the registry, which should be able to test this use case.

Anyway, merging for now. Integration test can be a followup.

PR #1645, opened 2025-07-10
Added Otel Tracing to soci-snapshotter

Swapnanil-Gupta:
**Issue #, if available:**

**Description of changes:**
This PR adds the Otel Stats handler to the SOCI gRPC server along with its own Exporter. This allows SOCI to instrument its gRPC calls and export the traces enabling us to monitor SOCI performance and request latencies. This is demonstrated by adding spans to `FetchSociArtifacts` and `Resolve` functions.

**Testing performed:**
By setting the env variable `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318` and starting a Jaeger server that listens to the same port, we can see traces like this when we pull an image with `crictl` -

For `public.ecr.aws/soci-workshop-examples/tensorflow_gpu:latest`

image

For `public.ecr.aws/soci-workshop-examples/tensortflow:latest`

image

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -39,6 +39,7 @@
 package config
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
@@ -84,8 +85,8 @@ func NewConfig() *Config {
 
 	// Set any defaults which do not align with Go zero values.
 	var initParsers = []configParser{defaultPullModes, defaultDirectoryCacheConfig}
-	for _, p := range append(initParsers, parsers...) {
-		p(cfg)
+	if err := parseConfig(cfg, append(initParsers, parsers...)); err != nil {
+		return nil
 	}
 
 	return cfg
@@ -102,18 +103,26 @@ func NewConfigFromToml(cfgPath string) (*Config, error) {
 	defer f.Close()
 
 	cfg := NewConfig()
+	if cfg == nil {
+		return nil, errors.New("error creating default config")
+	}
 	// Get configuration from specified file
 	if err = toml.NewDecoder(f).Decode(cfg); err != nil {
 		return nil, fmt.Errorf("failed to load config file %q: %w", cfgPath, err)
 	}
-	parseConfig(cfg)
+	if err := parseConfig(cfg, parsers); err != nil {
+		return nil, fmt.Errorf("config file at %q: %w", cfgPath, err)
+	}
 	return cfg, nil
 }
 
-func parseConfig(cfg *Config) {
-	for _, p := range parsers {
-		p(cfg)
+func parseConfig(cfg *Config, cfgParsers []configParser) error {
+	for _, p := range cfgParsers {
+		if err := p(cfg); err != nil {
+			return fmt.Errorf("failed to parse config: %v", err)
+		}
 	}
+	return nil
 }
 
 func parseRootConfig(cfg *Config) error {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -265,6 +265,18 @@ concurrent_download_chunk_size = "1MB"
 				}
 			},
 		},
+		{
+			name: "IncorrectChunkConfig",
+			config: []byte(`
+[pull_modes.parallel_pull_unpack]
+concurrent_download_chunk_size = "badchunksize"
+`),
+			assert: func(t *testing.T, actual *Config, err error) {
+				if err == nil {
+					t.Error("Expected error, got none")
+				}
+			},
+		},
 	}
 
 	for _, test := range tests {
--- a/docs/soci-index-manifest-v2.md
+++ b/docs/soci-index-manifest-v2.md
@@ -45,8 +45,9 @@ There are some use-cases where the dynamic nature of SOCI Index Manifest v1 adds
 If you would like to enable SOCI v1, you can add the following to the soci snapshotter config (located at `/var/lib/soci-snapshotter-grpc/config.toml`):
 
 ```
-[pull_modes.soci_v1]
-enable = true
+[pull_modes]
+  [pull_modes.soci_v1]
+    enable = true
 ```
 
 ## Comparing SOCI Index Manifest v1 and SOCI Index Manifest v2
--- a/integration/pull_test.go
+++ b/integration/pull_test.go
@@ -885,11 +885,7 @@ func TestPullWithParallelism(t *testing.T) {
 		},
 		{
 			name: "set chunk size",
-			opts: []snapshotterConfigOpt{
-				func(cfg *config.Config) {
-					cfg.PullModes.Parallel.ConcurrentDownloadChunkSize = 1_000_000
-				},
-			},
+			opts: []snapshotterConfigOpt{withConcurrentDownloadChunkSize(1_000_000)},
 		},
 	}
 
@@ -902,11 +898,7 @@ func TestPullWithParallelism(t *testing.T) {
 		},
 		{
 			name: "parallel unpacking",
-			opts: []snapshotterConfigOpt{
-				func(cfg *config.Config) {
-					cfg.PullModes.Parallel.MaxConcurrentUnpacks = 3
-				},
-			},
+			opts: []snapshotterConfigOpt{withConcurrentUnpacks(3)},
 		},
 	}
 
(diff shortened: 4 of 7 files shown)


tracing/tracing.go
```diff
@@ -0,0 +1,122 @@
...
+	otlpTracesEndpointEnv = "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT"
+	otlpProtocolEnv       = "OTEL_EXPORTER_OTLP_PROTOCOL"
+	otlpTracesProtocolEnv = "OTEL_EXPORTER_OTLP_TRACES_PROTOCOL"
+	otelTracesExporterEnv = "OTEL_TRACES_EXPORTER"
+	otelServiceNameEnv    = "OTEL_SERVICE_NAME"
+	defaultServiceName    = "soci-snapshotter"
+)
+
+func Init(ctx context.Context) (func(context.Context) error, error) {
+	if err := isDisabled(); err != nil {
```
sondavidb: If tracing is disabled intentionally shouldn't we treat this as not an error? From the code bit in main.go it seems like any error will make this fail. IMO we should return a nil error if it is explicitly disabled, so maybe `isDisabled` should be split into `isDisabled` and something like `checkSetup` or whatever to separate a bad setup from explicitly disabled.

Swapnanil-Gupta: Updated the `isDisabled` method to return a `bool` indicating if tracing is disabled or not, instead of always throwing an error and updated the related code in this commit https://github.com/awslabs/soci-snapshotter/pull/1645/commits/334a4de5f12481d38c7b86cacb033d57deb7a73e.

cmd/soci-snapshotter-grpc/main.go
```diff
@@ -195,6 +198,23 @@ func main() {
 		log.G(ctx).WithError(err).Fatalf("failed to configure snapshotter")
 	}
 
+	log.G(ctx).Info("setting up otel tracing")
```
sondavidb: Info-level seems a little excessive here, since we aren't actually doing anything — maybe debug-level would be more appropriate. We already report when it is successfully setup or errors out so I think it's best to leave this as a debug log (or gone entirely, up to you)

cmd/soci-snapshotter-grpc/main.go
```diff
@@ -195,6 +198,23 @@ func main() {
...
+	tracingDisabled, shutDownTracing, err := tracing.Init(ctx)
+	if err != nil {
+		log.G(ctx).WithError(err).Info("failed to initialize otel tracing")
```
sondavidb: Is there any reason we don't just return error instead of logging an error message? The current behavior means we will log an error but proceed since we don't return the err after getting one.

cmd/soci-snapshotter-grpc/main.go
```diff
@@ -195,6 +198,23 @@ func main() {
...
+	} else if tracingDisabled {
+		log.G(ctx).Info("otel tracing is disabled by env")
+	} else {
+		defer func() {
+			if err := shutDownTracing(ctx); err != nil {
+				log.G(ctx).WithError(err).Errorf("failed to shutdown tracing")
```
sondavidb: nit: `Errorf` -> `Error` — we aren't using any formatting

fs/artifact_fetcher.go
```diff
@@ -257,6 +258,9 @@ func (f *artifactFetcher) Store(ctx context.Context, desc ocispec.Descriptor, re
 }
 
 func FetchSociArtifacts(ctx context.Context, refspec reference.Spec, indexDesc ocispec.Descriptor, localStore store.Store, remoteStore resolverStorage) (*soci.Index, error) {
+	ctx, span := otel.Tracer("").Start(ctx, "soci-snapshotter.fs.artifact_fetcher.FetchSociArtifacts")
```
sondavidb: `soci-snapshotter.fs.artifact_fetcher.FetchSociArtifacts` is a pretty long name. Does it have to be formatted this way? If we want to keep this format (either because we have to or because we think it's the clearest way to do this) we should probably create a dynamic way to generate this string easily. (Alternatively, we can just store this string into a var so that it's a little easier to refer to down the line.)

Not a blocking change, just food for thought 🤷‍♀️

tracing/tracing.go
```diff
@@ -0,0 +1,126 @@
...
+	otlpEndpointEnv       = "OTEL_EXPORTER_OTLP_ENDPOINT"
+	otlpTracesEndpointEnv = "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT"
+	otlpProtocolEnv       = "OTEL_EXPORTER_OTLP_PROTOCOL"
+	otlpTracesProtocolEnv = "OTEL_EXPORTER_OTLP_TRACES_PROTOCOL"
+	otelTracesExporterEnv = "OTEL_TRACES_EXPORTER"
+	otelServiceNameEnv    = "OTEL_SERVICE_NAME"
+	defaultServiceName    = "soci-snapshotter"
+)
+
+func Init(ctx context.Context) (bool, func(context.Context) error, error) {
```
Kern--: nit: would it be more clear to require the caller to decide if tracing is disabled and not call this function?

e.g.
```
if !tracing.Disabled() {
    cleanup, err := tracing.Init(ctx)
    // ...
}
```

tracing/tracing.go
```diff
@@ -0,0 +1,126 @@
...
+	if v := os.Getenv(otelTracesExporterEnv); v != "" && v != "otlp" {
+		return nil, fmt.Errorf("unsupported traces exporter %q", v)
+	}
+
+	v := os.Getenv(otlpTracesProtocolEnv)
+	if v == "" {
+		v = os.Getenv(otlpProtocolEnv)
+	}
+
+	const timeout = 5 * time.Second
```
sondavidb: Style comment: do we want to just do `timeout := 5 * time.Second`? I couldn't find anything in their [style guide](https://google.github.io/styleguide/go/decisions.html#constant-names) on constants and this works perfectly fine but it still sticks out to me some reason🤷

fs/artifact_fetcher.go
Swapnanil-Gupta: As per [containerd](https://github.com/containerd/containerd/blob/main/docs/tracing.md):
```
Span Names
    - Dot-separated notation.
    - Span Names may include relative path to the package.
    - Span Names should include a name that represents the specific component or service performing the operation.
    - For example: "pkg.cri.sbserver.CreateContainer"
        - "pkg.cri.sbserver" - relative path to the package
        - "CreateContainer" - describes the operation that is traced
```
I moved the long names to a separate variable for now. Let me know if this okay or if I should use the `reflect` package to try to generate this dynamically.

cmd/soci-snapshotter-grpc/main.go
Swapnanil-Gupta: From my understanding, tracing should not interfere with the main soci functionality. If we return the error from the main function when it fails to setup the tracer, will that not prevent soci from starting its gRPC server?

sondavidb: Synced offline on this, it doesn't seem like a big deal either way. A hard-failure is louder but maybe undesirable for folks who don't really care about this.

I think the best middle-ground solution we can do here is:
- If all the otel-specific variables are empty, continue launching the snapshotter (optionally maybe a message here too?).
- If the disable var is true, print the disabled message and continue launching the snapshotter.
- If any otel-specific variables are set incorrectly, we should hard-fail.

Does that work? If not LMK, this is my preferred solution but I won't block on this change per se.

Swapnanil-Gupta: Addressed with https://github.com/awslabs/soci-snapshotter/pull/1645/commits/8cf422c19a46bcbb1ca02746a102cca687d2ff10

tracing/tracing.go
Swapnanil-Gupta: Addressed with https://github.com/awslabs/soci-snapshotter/pull/1645/commits/8cf422c19a46bcbb1ca02746a102cca687d2ff10

fs/artifact_fetcher.go
sondavidb: Let's do this for now. If we get too many different variable names we can consider generating dynamically.

PR #1646, opened 2025-07-11
fix: check if commit failed due to content already exists

swagatbora90:
**Issue #, if available:**
Building ztocs layers with containerd store might fail due to blobs already existing

```
INFO[0126] soci: soci: errors encountered while building soci layers; cannot push ztoc to local store: commit failed: blob with expected digest sha256:2fa07ddf9c428cdde1f8be966d338f74cd4ef1be4d2b5649804ff53b70b582a6 exists: already exists
FATA[0126] failed to convert image to SOCI format: exit status 1
```
Currently we only check for ORAS error type https://github.com/awslabs/soci-snapshotter/blob/main/soci/soci_index.go#L654-L657

**Description of changes:**

This PR checks for the containerd errdefs.ErrAlreadyExists when pushing blobs to the containerd store. Also, found that we do check if the content already exists before committing, but the current implementation can have a race condition, since a layer ztoc might be pushed after the check. So we need to validate the return error  writer.Commit()

### DIFF
--- a/soci/store/store.go
+++ b/soci/store/store.go
@@ -334,7 +334,11 @@ func (s *ContainerdStore) Push(ctx context.Context, expected ocispec.Descriptor,
 		return fmt.Errorf("unexpected copy size %d, expected %d: %w", totalWritten, expected.Size, errdefs.ErrFailedPrecondition)
 	}
 
-	return writer.Commit(ctx, expected.Size, expected.Digest)
+	err = writer.Commit(ctx, expected.Size, expected.Digest)
+	if err != nil && !errors.Is(err, errdefs.ErrAlreadyExists) {
+		return err
+	}
+	return nil
 }
 
 // LabelGCRoot labels the target resource to prevent garbage collection of itself.


soci/store/store.go
```diff
@@ -248,7 +248,7 @@ func NewContainerdStore(storeConfig ContentStoreConfig) (*ContainerdStore, error
 	return &containerdStore, nil
 }
 
-// Exists returns true iff the described content exists.
+// Exists returns true if the described content exists.
```
sondavidb: AFAIK "iff" here is used as shorthand for "if and only if"

swagatbora90:
> Can we create a follow up issue to audit all of our errdef/errdefs checks? We have `ErrNotFound` and `ErrAlreadyExists` checks all over and I'm not sure we're always using the right ORAS/containerd version everywhere.

Yes, i can create one.  Though with this PR change, we will handle most cases where `b.blobStore.Push()` is called. I saw that most of the `ErrAlreadyExists` checks are for blobStore.Push(), so should be handled now for both local and containerd store.

soci/store/store.go
```diff
@@ -334,7 +334,12 @@ func (s *ContainerdStore) Push(ctx context.Context, expected ocispec.Descriptor,
 		return fmt.Errorf("unexpected copy size %d, expected %d: %w", totalWritten, expected.Size, errdefs.ErrFailedPrecondition)
 	}
 
-	return writer.Commit(ctx, expected.Size, expected.Digest)
+	err = writer.Commit(ctx, expected.Size, expected.Digest)
+	if err != nil && errors.Is(err, errdefs.ErrAlreadyExists) {
```
sondavidb: nit: For consistency, I think we tend to follow the pattern

```go
if err != nil && !errors.Is(err, errdef.ErrAlreadyExists) {
  return err
}
```

swagatbora90: tbh I have seen all kinds of error check patterns across the codebase, but yeah this one is the most used. I will update to follow this.

sondavidb: It's a mess yeah 😓 I think the big issue is that we don't have a set place for where we should handle this yet. IMHO the store should just push and return an error if it's already there and leave the caller to handle it (since some operations will want to care about if something exists and some will not), but for now we have no real consistency so I'm ok with this.

Maybe a helper `IsAlreadyExists` func that checks for both SOCI and containerd errdefs for it? That might be useful so that we aren't always checking one or the other. (Obviously not a blocker for this, and well out of scope, but just a thought)

sondavidb (approved):

henry118 (approved):

PR #1649, opened 2025-07-14
Use common IsErrAlreadyExists for Push errors

### LINKED ISSUE #1647: Audit error handling for ErrNotFound and ErrAlreadyExists across ORAS/containerd store
There is some inconsistencies in error handling in the soci codebase, mainly related to usage of ORAS vs containerd errdef types. We need to audit all error checks for ErrNotFound, ErrAlreadyExists, and similar sentinel errors. PR #1646 fixes one such case where we were checking for the ORAS error type instead of containerd. 

Also, it will be nice to standardize the error handling patterns throughout the codebase.  Some locations use direct equality checks, others use errors.Is(), and some use library-specific helper functions like errdefs.IsAlreadyExists(). We should probably standardize on `errors.Is()` for all sentinel error checks.

sondavidb:
**Issue #, if available:**
Closes #1647 

**Description of changes:**
containerd and ORAS both return a var named ErrAlreadyExists when pushing to their respective local content stores. This leads to some pretty cumbersome error checking cases. This change should make it a little easier to work with.

This also changes the push behavior to return an error if content already exists. This is to match the ORAS behavior for if content already exists and relies on the caller to catch this error.

**Testing performed:**

### DIFF
--- a/fs/artifact_fetcher.go
+++ b/fs/artifact_fetcher.go
@@ -32,7 +32,6 @@ import (
 	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/awslabs/soci-snapshotter/soci/store"
 	"github.com/awslabs/soci-snapshotter/util/ioutils"
-	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/reference"
 	"github.com/containerd/containerd/remotes/docker"
 	"github.com/containerd/log"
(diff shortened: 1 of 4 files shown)


fs/artifact_fetcher.go
```diff
@@ -32,7 +32,6 @@ import (
 	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/awslabs/soci-snapshotter/soci/store"
 	"github.com/awslabs/soci-snapshotter/util/ioutils"
```
Kern--, line 34: Looks like another instance was missed on line 322

soci/soci_index.go
```diff
@@ -652,7 +652,7 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 	}
```
Kern--, line 651: Looks like an instance was missed on line 489.

Kern-- (approved):

coderbirju (approved):

PR #1651, opened 2025-07-18
Hard-fail on config parsing errors

### LINKED ISSUE #1650: [Bug] config parsing errors silently dropped
### Description

I believe I've found a bug in config parsing for soci-snapshotter. Errors in parsing config are dropped and default values are set. Is this behavior expected? With config

```
[content_store]
type="containerd"

enable = true
concurrent_download_chunk_size="xzy" # clearly invalid
```

soci starts just fine with `"concurrent_download_chunk_size":-1`. It looks like the error from various config parsers is being dropped and not returned in parseConfig https://github.com/awslabs/soci-snapshotter/blob/8d8d1200b8eba2daf31a95ce8585a44a553d91cc/config/config.go#L113-L117. Each parser https://github.com/awslabs/soci-snapshotter/blob/8d8d1200b8eba2daf31a95ce8585a44a553d91cc/config/config.go#L79
has a return signature error but it's unhandled.

A naive patch that I observed prints the error for my invalid config:

```
diff --git a/config/config.go b/config/config.go
index 25251c0f..25908d9c 100644
--- a/config/config.go
+++ b/config/config.go
@@ -112,7 +112,10 @@ func NewConfigFromToml(cfgPath string) (*Config, error) {

 func parseConfig(cfg *Config) {
        for _, p := range parsers {
-               p(cfg)
+               err := p(cfg)
+               if err != nil {
+                       fmt.Printf("Failed to parse config: %v\n", err)
+               }
        }
 }
```

See above

The config parsing fails

### Host information

1. OS: Fedora 40
2. Snapshotter Version: built from tip of `main`
3. Containerd Version: n/a

_No response_

sondavidb:
**Issue #, if available:**
Closes #1650 

**Description of changes:**
We had an issue where a bad config value will allow the snapshotter to continue running. This change fixes this logic so that it will hard fail instead of silently fail. I also added integration and unit tests for this

This change also includes a small refactor to move some parallel pull config changes into helper funcs

**Testing performed:**
`GO_TEST_FLAGS='-run TestSnapshotterStartupWithBadConfig -count 1'  make integration`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -39,6 +39,7 @@
 package config
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
@@ -84,8 +85,8 @@ func NewConfig() *Config {
 
 	// Set any defaults which do not align with Go zero values.
 	var initParsers = []configParser{defaultPullModes, defaultDirectoryCacheConfig}
-	for _, p := range append(initParsers, parsers...) {
-		p(cfg)
+	if err := parseConfig(cfg, append(initParsers, parsers...)); err != nil {
+		return nil
 	}
 
 	return cfg
@@ -102,18 +103,26 @@ func NewConfigFromToml(cfgPath string) (*Config, error) {
 	defer f.Close()
 
 	cfg := NewConfig()
+	if cfg == nil {
+		return nil, errors.New("error creating default config")
+	}
 	// Get configuration from specified file
 	if err = toml.NewDecoder(f).Decode(cfg); err != nil {
 		return nil, fmt.Errorf("failed to load config file %q: %w", cfgPath, err)
 	}
-	parseConfig(cfg)
+	if err := parseConfig(cfg, parsers); err != nil {
+		return nil, fmt.Errorf("config file at %q: %w", cfgPath, err)
+	}
 	return cfg, nil
 }
 
-func parseConfig(cfg *Config) {
-	for _, p := range parsers {
-		p(cfg)
+func parseConfig(cfg *Config, cfgParsers []configParser) error {
+	for _, p := range cfgParsers {
+		if err := p(cfg); err != nil {
+			return fmt.Errorf("failed to parse config: %v", err)
+		}
 	}
+	return nil
 }
 
 func parseRootConfig(cfg *Config) error {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -265,6 +265,18 @@ concurrent_download_chunk_size = "1MB"
 				}
 			},
 		},
+		{
+			name: "IncorrectChunkConfig",
+			config: []byte(`
+[pull_modes.parallel_pull_unpack]
+concurrent_download_chunk_size = "badchunksize"
+`),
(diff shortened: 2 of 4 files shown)


henry118 (approved):

ginglis13 (approved):
lgtm

config/config.go
```diff
@@ -84,8 +85,9 @@ func NewConfig() *Config {
 
 	// Set any defaults which do not align with Go zero values.
 	var initParsers = []configParser{defaultPullModes, defaultDirectoryCacheConfig}
-	for _, p := range append(initParsers, parsers...) {
-		p(cfg)
+	err := parseConfig(cfg, append(initParsers, parsers...))
+	if err != nil {
+		return nil
 	}
```
pendo324: nit: we're mixing just assigning to err and the "if scoped" error handling patterning in this code for no reason. Since error handling already proved problematic here, can we just make all of the error handling use the if pattern?

```suggestion
	if err := parseConfig(cfg, append(initParsers, parsers...)); err != nil {
		return nil
	}
```

config/config.go
```diff
@@ -102,18 +104,27 @@ func NewConfigFromToml(cfgPath string) (*Config, error) {
...
+	}
 	// Get configuration from specified file
 	if err = toml.NewDecoder(f).Decode(cfg); err != nil {
 		return nil, fmt.Errorf("failed to load config file %q: %w", cfgPath, err)
 	}
-	parseConfig(cfg)
+	err = parseConfig(cfg, parsers)
+	if err != nil {
+		return nil, fmt.Errorf("config file at %q: %w", cfgPath, err)
+	}
```
pendo324: Same as above
```suggestion
	if err := parseConfig(cfg, parsers); err != nil {
		return nil, fmt.Errorf("config file at %q: %w", cfgPath, err)
	}
```

pendo324 (approved):

ginglis13:
@sondavidb merging this PR will close the wrong issue - you want to reference #1650

sondavidb:
> @sondavidb merging this PR will close the wrong issue - you want to reference https://github.com/awslabs/soci-snapshotter/issues/1650

whoops, thanks

pendo324 (approved):

config/config.go
```diff
@@ -102,18 +103,26 @@ func NewConfigFromToml(cfgPath string) (*Config, error) {
 	defer f.Close()
 
 	cfg := NewConfig()
```
Shubhranshu153, line 105: should not NewConfig return error describing what was ther error?

sondavidb: Probably, but our whole config parsing is weird to begin with...I think we should probably refactor our config parsing logic because our current setup also means we can't use default values (like 0) without some weird logic. (And this is why unbounded pulls in parallel pull mode is -1 instead of 0 🙂)

sondavidb: Anyway, if this is not blocking I'll resolve this to merge — we can change our messy logic in a followup PR

PR #1653, opened 2025-07-24
Do not allow parallel pull mode with lazy-loading

sondavidb:
**NOTE:** This is potentially a breaking change as users may be expecting the old order of precedence when pulling images.

**Issue #, if available:**

**Description of changes:**
Previously, the order of precedence would be parallel pull -> SOCI v2 -> SOCI v1, which was not the intended order of precedence. This change makes it so that the lazy-loading modes take precedence over parallel pull modes.

To further avoid confusion/ambiguous behavior, this change also enforces that parallel pull mode will never be used if any lazy-loaded mode is enabled. We can revisit this in the future, but for now it is probably best for clarity's sake to ensure that only one path or the other can be taken.

**Testing performed:**

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -114,15 +114,12 @@ func NewSociSnapshotterService(ctx context.Context, root string, serviceCfg *con
 
 	var snapshotter snapshots.Snapshotter
 
-	snOpts := []snbase.Opt{snbase.WithAsynchronousRemove}
+	snOpts := []snbase.Opt{snbase.WithAsynchronousRemove, snbase.WithPullModes(&serviceCfg.PullModes)}
 	if serviceCfg.MinLayerSize > -1 {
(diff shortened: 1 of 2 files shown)


sondavidb:
We had some discussions on this, and the TL;DR seems to be that we don't think this is the right thing to do for now. Having parallel pull mode take precedence over lazy-loading (which is the current behavior) means a user explicitly using the parallel pull mode can set up their config with only parallel pull-related configs and it will work. If we make lazy-loading take precedence a user will also have to explicitly disable SOCIv2 on top of this, which is not super intuitive. Meanwhile someone who wants to use SOCIv1 does not have to explicitly disable SOCIv2 to get their workflows running.

An alternative to this would be to allow all pull modes, all at once, but there's a bit of ambiguity/risk there that has yet to be fully evaluated so for now we'll close this and revisit this in the future. I think if we reopen this idea, we should explicitly allow all pull modes to be enabled at once, to reduce friction and minimize unintuitive workflows.

PR #1654, opened 2025-07-24
Introduce async digest verification for layer unpacking

henry118:
**Issue #, if available:**

During testing we found that the digest calculation is expensive. And it's part of the sync operation of a layer's unpack operation.

**Description of changes:**

This patch makes the digest calculation as a parallel operation of the layer's unpack process.

**Testing performed:**

```
$ go test -benchmem -run=^$ -bench ^BenchmarkTeeReader$ github.com/awslabs/soci-snapshotter/fs
goos: linux
goarch: amd64
pkg: github.com/awslabs/soci-snapshotter/fs
cpu: Intel(R) Xeon(R) Platinum 8275CL CPU @ 3.00GHz
BenchmarkTeeReader-4           1        45820754476 ns/op        1173056 B/op       4121 allocs/op
PASS
ok      github.com/awslabs/soci-snapshotter/fs  45.831s

$ go test -benchmem -run=^$ -bench ^BenchmarkAsyncTeeReader$ github.com/awslabs/soci-snapshotter/fs
goos: linux
goarch: amd64
pkg: github.com/awslabs/soci-snapshotter/fs
cpu: Intel(R) Xeon(R) Platinum 8275CL CPU @ 3.00GHz
BenchmarkAsyncTeeReader-4              1        10633266274 ns/op        1300856 B/op       5338 allocs/op
PASS
ok      github.com/awslabs/soci-snapshotter/fs  10.644s
```


fs/unpacker.go
```diff
@@ -46,56 +47,192 @@ type Archive interface {
...
+				return
+			}
+			bp.Put(b)
+		}
+	}()
+
+	return pr
+}
+
+type noopReadCloser struct {
```
sondavidb: Can't we just use [io.NopCloser](https://cs.opensource.google/go/go/+/go1.24.5:src/io/io.go;l=682)? (and if we have to use this maybe a comment would help clarify why)

henry118: Thanks for point it out. updated.

fs/adaptive_fetch_image_layers.go
```diff
@@ -563,6 +564,7 @@ func newImageUnpackJob(imageDigest string, opts ...imageUnpackOption) *imageUnpa
 		concurrentDownloadsLimiter:       NewSemaphoreWithNil(unlimited),
 		concurrentUnpacksLimiter:         NewSemaphoreWithNil(unlimited),
 		layers:                           make(map[string][]*layerUnpackJob),
+		bufferPool:                       newbufferPool(64 * 1024),
```
sondavidb, line 567: This is only 64KB so presumably this doesn't need to be configurable, but floating the idea of making it configurable anyway, just in case others disagree.

fs/unpacker.go
```diff
@@ -46,56 +47,184 @@ type Archive interface {
...
+		return
+	}
+	sig := make(chan struct{}, 1)
+	go func() {
+		close(sig)
+		io.Copy(av.v, reader)
+		close(av.waitCh)
+		reader.Close()
+	}()
+	<-sig
```
sondavidb: Maybe a silly question — why do we need `sig` here? Is there any reason we can't just start the goroutine and quit?

fs/unpacker.go
```diff
@@ -46,56 +47,184 @@ type Archive interface {
...
+	av.started = true
+}
+
+func (av *asyncVerifier) Verified(ctx context.Context) bool {
+	if av == nil || av.v == nil || !av.started {
```
sondavidb: If it hasn't started I don't think we should say it has been verified. Either error out or wait on it IMO.

fs/parallel_artifact_fetcher.go
```diff
@@ -273,9 +278,25 @@ func (f *parallelArtifactFetcher) fetchFromRemoteAndWriteToTempDir(ctx context.C
 	if err != nil {
 		return nil, fmt.Errorf("error reopening file at %s after writing: %w", ingestPath, err)
 	}
+
+	// start async verification of the blob digest
+	if err = f.asyncVerifyBlobDigest(ctx, ingestPath); err != nil {
```
sondavidb, line 283: This stuck out to me a little but I guess thinking about it we should verify the digest of every object we download so it makes sense to verify the digest here. Should we also do this in the non-parallel artifact fetcher? Or does this not work without having the file on disk?

fs/unpacker_test.go
```diff
@@ -210,3 +213,142 @@ func getFakeMounts() []mount.Mount {
 		},
 	}
 }
+
+// Test that AsyncTeeReader copies data to both the returned reader and the writer.
+func TestAsyncTeeReaderBasic(t *testing.T) {
```
sondavidb: nit: should we combine this with the test below? They look to do the same thing.

fs/unpacker.go
```diff
@@ -46,56 +47,184 @@ type Archive interface {
...
+	return &asyncVerifier{
+		v:       v,
+		waitCh:  make(chan struct{}, 1),
+		started: false,
+	}
+}
+
+func (av *asyncVerifier) AsyncVerify(reader io.ReadCloser) {
+	if av == nil || av.v == nil {
+		return
```
Kern--, line 73: This should call reader.Close() before returning, right?

fs/unpacker.go
```diff
@@ -46,56 +47,184 @@ type Archive interface {
...
 
+	var (
+		ch     = make(chan *chunk, 16)
+		pr, pw = io.Pipe()
+	)
+
+	// writer goroutine writing chunks to w
+	go func() {
+		for c := range ch {
+			if len(*c.data) == 0 || c.n == 0 {
```
Kern--: nit: do we need to special case this? Shouldn't write handle an empty slice properly?

fs/unpacker.go
```diff
@@ -46,56 +47,184 @@ type Archive interface {
...
+	go func() {
+		defer close(ch)
+		for {
+			b := bp.Get()
+			n, err := r.Read(*b)
+			if n > 0 {
+				c := &chunk{data: bp.Get(), n: n}
+				copy((*c.data), (*b)[:n])
+				ch <- c
+				pw.Write((*b)[:n])
```
Kern--: We should handle write errors here.

fs/unpacker.go
```diff
@@ -46,56 +47,184 @@ type Archive interface {
...
 		decompressStream: decompressStream,
+		bufferPool:       bufPool,
 	}
 }
 
-func (la *layerArchive) Apply(ctx context.Context, root string, r io.Reader, opts ...archive.ApplyOpt) (int64, error) {
-	// Decompress first, then apply.
-	if la.verifier.compressed != nil {
-		r = io.TeeReader(r, la.verifier.compressed)
+func AsyncTeeReader(r io.Reader, w io.WriteCloser, bp *bufferPool) io.Reader {
```
Kern--, line 139: Did you consider an `AsyncWriter` + normal io.TeeReader for this? I'm not sure if there's a use-case for an async writer outside of a TeeReader, but maybe that would be a smaller abstraction? What do you think?

fs/parallel_artifact_fetcher.go
sondavidb: Ah ok, I forgot to comment on this before but we make this asynchronous by opening another file descriptor, by the time we get to the layerArchive we lose this ref. So it's probably best to just use our own fd here.

fs/adaptive_fetch_image_layers.go
henry118: I was trying to avoid adding additional configs since there are already enough knobs for the fast-pull mode. Also, I think it would be difficult for users to come up with a reasonable value for this.

fs/unpacker.go
henry118: You're right. It's not really needed. Removed in the latest revision.

fs/unpacker.go
henry118: There is already bufio in golang but the challenge with it is that golang grows the buffer cap as needed opposed to limiting the cap to a fixed size. I considered to implement a AsyncWriter with circular buffers, but I don't really see other use cases other than this. Also IMHO the AsyncTeeReader implementation is also simple enough. So I prefer keeping it as current form now.

fs/unpacker.go
henry118: updated

fs/unpacker_test.go
henry118: updated

fs/unpacker.go
henry118: updated

fs/unpacker.go
henry118: updated

fs/unpacker.go
henry118: updated

fs/unpacker.go
```diff
@@ -46,56 +47,177 @@ type Archive interface {
...
+	}()
+	av.started = true
+}
+
+func (av *asyncVerifier) Verified(ctx context.Context) bool {
+	if av == nil || av.v == nil {
+		return true
+	}
+	if !av.started {
+		return false
```
sondavidb, line 88: Is there ever a race condition in the code path where it has not started at this point? I don't think so but just something to think about

sondavidb (approved):
LGTM

fs/unpacker.go
henry118: Right. not in current implementation. since the async operation is always started by the fetch much earlier.

fs/unpacker_test.go
```diff
@@ -210,3 +213,136 @@ func getFakeMounts() []mount.Mount {
 		},
 	}
 }
+
+func TestAsyncTeeReader(t *testing.T) {
```
Kern--, line 217: Is it worth having a test where the input is larger than the buffer size?

fs/unpacker_test.go
```diff
@@ -210,3 +213,136 @@ func getFakeMounts() []mount.Mount {
...
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if tc.experror && err == nil {
+				t.Fatal("expected error, got nil")
+			}
+			if string(out) != tc.expected {
+				t.Errorf("unexpected output: %q", out)
+			}
+			w.Wait()
+			if w.Written() != int64(len(tc.expected)) {
```
Kern--: Is it worth storing and checking the data the writer received to make sure we get the same data as the reader?

fs/unpacker_test.go
henry118: Added test

fs/unpacker_test.go
henry118: Yep updated tests

PR #1660, opened 2025-07-29
Update dependabot.yml and remove custom workflow

sondavidb:
**Issue #, if available:**

**Description of changes:**
Since June 25 2024, dependabot has added multi-directory support for dependencies. As the existing bump-deps.yml was meant as a workaround for this, we can safely remove it and rely solely on dependabot instead.

Doc ref for dependabot changes: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories

**Testing performed:**
Enabled dependabot on personal fork and ensured the created PR updates all dependencies at once: https://github.com/sondavidb/soci-snapshotter/pull/33


sondavidb:
This can be merged as-is but this might also not be a bad time to start grouping our various dependencies into different groups, e.g. if we want to start including k8s deps in here they can be grouped updates to avoid breaking things or whatever. 🤷

.github/workflows/bump-deps.yml
```diff
@@ -1,41 +0,0 @@
-# adapted based on
```
Kern--, line 1: If we're removing this, should we also remove scripts/bump-deps.sh?

Kern-- (commented):
How confident are we that we have proper support of kubernetes? Do we have confirmation from the EKS folks on what versions we should be targetting? If we bump the deps are we going to lose v1alpha1 version support?

.github/workflows/bump-deps.yml
sondavidb: Done

Shubhranshu153 (approved):

henry118 (approved):

PR #1663, opened 2025-07-31
Move to math/rand/v2 and deprecate containerd/seed

### LINKED ISSUE #882: [Bug] Dependency on github.com/containerd/containerd/pkg/seed [Deprecated]
### Description

containerd's `github.com/containerd/containerd/pkg/seed` package has been marked deprecated since `v1.7.0`. See https://pkg.go.dev/github.com/containerd/containerd@v1.7.0/pkg/seed.

The SOCI CLI currently takes dependency on this package. The task here is to research the expected behavior and remove the dependency on the deprecated package by either 1) updating containerd's implementation or 2) implementing a local up-to-date implementation for SOCI CLI.

_No response_

SOCI CLI should not be reliant on deprecated packages.

### Host information

N/A

_No response_

sondavidb:
**Issue #, if available:**
Closes #882 

**Description of changes:**
Go 1.22 introduced math/rand/v2, which introduced breaking changes from the math/rand library. Notably, it removed ambiguity on who seeds the RNG used for global random number calls. By moving away from this library we can remove our usage of containerd/seed, which has been deprecated anyway.

This process included moving away from ThreadsafeRandom in our testutil package. The idea made sense (we want to seed the RNG but make it threadsafe) but in practice it never really worked properly since the order the tests run in aren't really deterministic, and if it's not deterministic then there's no reason to seed it to begin with. By leveraging the new rand.Source and ensuring we create a new Random variable for every instance we need it, we can achieve the actual end goal of deterministic randomness in our testing suite.

With that said, I did make a best effort to ensure that we are always creating new rand.Rand calls within single-threaded operations, but it's possible that I missed a couple cases.

There's still a couple of considerations even after this commit:

1. We still use the global RNG in our jitter function which is used for our HTTP backoff strategy. I can't imagine this is particularly impactful, but it's still worth noting that we are not making that deterministic within our tests.
2. If we consume any libraries that still use math/rand, we might lose reproducible behavior as Go 1.20 has math/rand calls be nondeterministic unless it is seeded, which we always did.

In spite of these caveats, this is hopefully still a net positive :)

**Testing performed:**


util/testutil/util.go
```diff
@@ -158,7 +127,10 @@ func RandomUInt64() (uint64, error) {
 // RandomByteData returns a byte slice with `size` populated with random generated data
 func RandomByteData(size int64) []byte {
 	b := make([]byte, size)
-	r.Read(b)
+	r := NewSetSeedRand()
+	for i := range b {
+		b[i] = byte(r.Int64())
+	}
```
sondavidb: ref: https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/math/rand/rand.go;l=272-302

sondavidb:
Synced offline w/ @Kern-- and we decided that it didn't really make sense to have global helper functions for randomness for the same reason that `math/rand` having a global seeder didn't make sense — if we want deterministic randomness it's not possible when relying on a global value that gets referred to from many operations. By scoping these functions behind a random object the caller can then determine how much randomness they want. So, I opted to make a wrapper for rand.Rand that would also implement the previous global helper functions, as well as Read.

The other important change I made is that instead of using two seeds, I use one predetermined seed and get the other seed from hashing the test name calling it. In doing so we get better randomness across tests, but keep determinism between runs. I considered allowing an optional identifier in place of `t.Name()` but at that point it felt like it was doing too much. I guess we can always add this in the future 🤷

sondavidb:
> Does this PR evaluate where the non-global random gets instantiated

I made a good effort to do so, I combed through every place that called the helper function and replaced it with a new instantiation. I ignored some places this got called where randomness didn't matter, though maybe if we don't need a random we shouldn't be calling this function. Let me reiterate further on this and just stop using random where we don't need it.

sondavidb:
Ok, I think newest changes should remove more or less all the unnecessary random calls, I think there's still [one call](https://github.com/awslabs/soci-snapshotter/pull/1663/files#diff-ecf50918f0b7a9a1b042c4498d9a03a763fa150b2a0dacf34547d49e3f7a9695R56) I don't really like because the intention is to give bad data, but since we have to implement the reader interface here it feels better to get a bunch of garbage data than, say, filling it with a bunch of zero chars.

Kern-- (approved):

Shubhranshu153 (dismissed):

sondavidb:
Rebased due to merge conflicts

PR #1664, opened 2025-08-02
Enable copyloopvar rule in Go linter

sondavidb:
**Issue #, if available:**

**Description of changes:**
We had quite a few instances of this pre-Go1.22 logic that I found while doing my random refactor, figured it might be good to start enforcing this rule in the linter

**Testing performed:**


sondavidb:
~~So this was a bad change because removing every instance of this breaks Goroutines that get spun up within the loop....I'll look into this, might close the PR if it's too clunky to work with.~~ EDIT: NVM, it was because a variable was declared outside of the scope of the loop but the linter still caught that. I think it should be fine then 🙂

Shubhranshu153 (approved):

PR #1672, opened 2025-08-05
Migrate soci to urfave/cli/v2

Kern--:
**Issue #, if available:**

**Description of changes:**
Partial carry of #1587 (thanks @austinvazquez!)

V2 and V3 both have a bunch of changes, so I want to split the original PR in 2 to make it easier to review. 

The migration guide is here: https://cli.urfave.org/migrate-v1-to-v2/

**Testing performed:**

### DIFF
--- a/cmd/soci/commands/internal/store.go
+++ b/cmd/soci/commands/internal/store.go
@@ -18,13 +18,13 @@ package internal
 
 import (
 	"github.com/awslabs/soci-snapshotter/soci/store"
-	"github.com/urfave/cli"
+	"github.com/urfave/cli/v2"
 )
 
 // ContentStoreOptions builds a list of content store options from a CLI context.
 func ContentStoreOptions(context *cli.Context) []store.Option {
 	return []store.Option{
-		store.WithType(store.ContentStoreType(context.GlobalString("content-store"))),
-		store.WithContainerdAddress(context.GlobalString("address")),
+		store.WithType(store.ContentStoreType(context.String("content-store"))),
+		store.WithContainerdAddress(context.String("address")),
 	}
 }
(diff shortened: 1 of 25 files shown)


Kern--:
~~Looks like there's a bug in urfave/cli/v2 where you can't set an integer with a default value to 0 (it just resets to the default): https://github.com/urfave/cli/issues/1050~~

~~Seems like it was fixed in v3, but that means we can't update in two steps :(~~

Actually, I think that's wrong. I think the real issue is that in V2, flags need to come before args.

sondavidb:
> I think the real issue is that in V2, flags need to come before args.

Yeah, this is also changed in v3 where flag ordering doesn't matter there (I think?) so this might not be worth trying to split into two different PRs from that alone

cmd/soci/commands/internal/store.go
```diff
@@ -17,14 +17,15 @@
...
+	"github.com/urfave/cli/v2"
 )
 
 // ContentStoreOptions builds a list of content store options from a CLI context.
 func ContentStoreOptions(context *cli.Context) []store.Option {
 	return []store.Option{
-		store.WithType(store.ContentStoreType(context.GlobalString("content-store"))),
-		store.WithContainerdAddress(context.GlobalString("address")),
+		store.WithType(store.ContentStoreType(context.String("content-store"))),
+		store.WithContainerdAddress(config.TrimSocketAddress(context.String("address"))),
```
sondavidb: We already trim the socket address in the helper func ([ref](https://github.com/awslabs/soci-snapshotter/blob/bdd630e2f63811782606ddfce57cf4cab783663b/soci/store/store.go#L149-L154))

Kern--:
> Yeah, this is also changed in v3 where flag ordering doesn't matter there (I think?) so this might not be worth trying to split into two different PRs from that alone

I see discussion on the order of precedent of alt sources, but I don't see any indication that flags can come after args in v3.

sondavidb:
> I don't see any indication that flags can come after args in v3.

I think here? https://github.com/urfave/cli/pull/1987

sondavidb (approved):

Shubhranshu153 (approved):

PR #1675, opened 2025-08-08
Remove filenames from `metadata` error logs

Swapnanil-Gupta:
**Issue #, if available:**

**Description of changes:**
Remove filenames from being logged when certain errors occur in:
- `metadata/reader`
- `metadata/db`

**Testing performed:**

### DIFF
--- a/metadata/reader.go
+++ b/metadata/reader.go
@@ -192,15 +192,15 @@ func (r *reader) initNodes(toc ztoc.TOC) error {
 				if isLink {
 					id, err = getIDByName(md, cleanLinkName, r.rootID)
 					if err != nil {
-						return fmt.Errorf("%q is a hardlink but cannot get link destination %q: %w", ent.Name, ent.Linkname, err)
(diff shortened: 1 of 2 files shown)


metadata/reader.go
```diff
@@ -234,15 +234,15 @@ func (r *reader) initNodes(toc ztoc.TOC) error {
 					}
 					// Write the new node object to the node bucket.
 					if err := writeNodeEntry(b, attrFromZtocEntry(&ent, &attr)); err != nil {
-						return fmt.Errorf("failed to set attr to %d(%q): %w", id, ent.Name, err)
+						return fmt.Errorf("failed to set attr to %d: %w", id, err)
```
Shubhranshu153, line 237: can err hold file name or other identifiable data?

Swapnanil-Gupta: No they only contain ids (eg: `fsId`) which should be safe to log and could help in debugging.

Kern-- (approved):

PR #1676, opened 2025-08-12
Migrate soci command to urfave/cli/v3

Kern--:
**Issue #, if available:**

**Description of changes:**
closes #1587 

Migrate soci command to urfave/cli/v3

migration instructions: https://cli.urfave.org/migrate-v2-to-v3/

**Testing performed:**

### DIFF
--- a/cmd/soci-snapshotter-grpc/main.go
+++ b/cmd/soci-snapshotter-grpc/main.go
@@ -72,7 +72,7 @@ import (
 	sddaemon "github.com/coreos/go-systemd/v22/daemon"
 	metrics "github.com/docker/go-metrics"
 	"github.com/sirupsen/logrus"
-	"github.com/urfave/cli/v2"
+	"github.com/urfave/cli/v3"
 	bolt "go.etcd.io/bbolt"
 	"golang.org/x/sys/unix"
 	"google.golang.org/grpc"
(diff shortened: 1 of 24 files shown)


Shubhranshu153 (dismissed):

cmd/soci-snapshotter-grpc/main.go
```diff
@@ -87,147 +87,146 @@ const (
...
+				log.G(ctx).Debug("Closing the snapshotter")
+				rs.Close()
+			}
+			log.G(ctx).Info("Exiting")
+			return nil
+		},
 	}
 
-	if err := app.Run(os.Args); err != nil {
+	if err := app.Run(context.TODO(), os.Args); err != nil {
```
sondavidb: We should create a new context here w/ a cancel and defer the cancel call instead (the context in question should also be tied to context.Background instead of context.TODO)

cmd/soci/commands/index/list.go
```diff
@@ -31,7 +32,17 @@ import (
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/platforms"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
-	"github.com/urfave/cli/v2"
+	"github.com/urfave/cli/v3"
+)
+
+const (
```
sondavidb, line 38: I like this pattern, should we apply it more consistently across all of our commands?

cmd/soci/commands/internal/platform.go
```diff
@@ -39,26 +40,25 @@ var PlatformFlags = []cli.Flag{
...
 // This method is not suitable for situations where the default should be all supported platforms (e.g. the `soci index list` command)
-func GetPlatforms(ctx context.Context, cliContext *cli.Context, img images.Image, cs content.Store) ([]ocispec.Platform, error) {
-	if cliContext.Bool(AllPlatformsFlagKey) {
+func GetPlatforms(ctx context.Context, cmd *cli.Command, img images.Image, cs content.Store) ([]ocispec.Platform, error) {
+	allPlatforms := cmd.Bool(AllPlatformsFlagKey)
+	if allPlatforms {
 		return images.Platforms(ctx, cs, img.Target)
 	}
-	ps := cliContext.StringSlice(PlatformFlagKey)
-	if len(ps) == 0 {
```
sondavidb, line 59: Any reason we removed this check?

cmd/soci/main.go
```diff
@@ -44,63 +45,62 @@ import (
...
-		commands.PushCommand,
-		commands.RebuildDBCommand,
-	}
-
-	app.Before = func(context *cli.Context) error {
-		return soci.EnsureSnapshotterRootPath(context.String("root"))
 	}
 
-	if err := app.Run(os.Args); err != nil {
+	ctx := context.TODO()
```
sondavidb: Same here, we should attach a proper context here.

cmd/soci/commands/internal/platform.go
Kern--: This is already handled by the logic below, there's no need to have a special case here.

cmd/soci/commands/index/list.go
Kern--: Yeah, I think that makes sense.

cmd/soci-snapshotter-grpc/main.go
```diff
@@ -87,150 +87,152 @@ const (
...
+			}
+			log.G(ctx).Info("Exiting")
+			return nil
+		},
 	}
 
-	if err := app.Run(os.Args); err != nil {
+	ctx, cancel := context.WithCancel(context.Background())
+	if err := app.Run(ctx, os.Args); err != nil {
+		cancel()
```
sondavidb, line 231: nit: Any reason we do this instead of just immediately deferring cancel? Not that opinionated if it's a style choice.

cmd/soci/main.go
```diff
@@ -44,64 +45,66 @@ import (
...
-		commands.RebuildDBCommand,
-	}
-
-	app.Before = func(context *cli.Context) error {
-		return soci.EnsureSnapshotterRootPath(context.String("root"))
 	}
 
-	if err := app.Run(os.Args); err != nil {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
sondavidb: Here too, but we're also deferring while also explicitly calling it which doesn't feel right — we should just do one or the other.

sondavidb (approved):
Small style thing but LGTM otherwise.

cmd/soci/main.go
Kern--: Whoops this was left over. I will remove this.

cmd/soci-snapshotter-grpc/main.go
Kern--: defer statements won't get called after `os.Exit`. I'm not sure it really matters since we'll probably exit too fast to process a cancellation anyway, but deferring seems more wrong.

sondavidb: Oh huh, that's kind of weird but you're right yeah. Maybe this could be a case to move this code out of main into its own function but obviously not really in scope here.

sondavidb (approved):

Shubhranshu153:
has conflicts.

Shubhranshu153 (approved):

Shubhranshu153 (approved):

PR #1682, opened 2025-08-19
Use cancellable contexts when possible

sondavidb:
**Issue #, if available:**

**Description of changes:**
Functionally this shouldn't change much if anything, it just ensures that we are passing meaningful contexts to more function calls. This mostly affects our tests where we were very liberal with using context.Background when we should ensure that all functions terminate by the end of the test case.

This also includes changes to use log.L instead of log.G with context.Background, which shouldn't make a difference since log.G with an empty context works the same as log.L (I think...)

**Testing performed:**
`make test` and ran a couple of integration tests

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -258,7 +258,7 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 	var bgFetcher *bf.BackgroundFetcher
 
 	if !cfg.BackgroundFetchConfig.Disable {
-		log.G(context.Background()).WithFields(logrus.Fields{
+		log.G(ctx).WithFields(logrus.Fields{
 			"fetchPeriod":      bgFetchPeriod,
 			"silencePeriod":    bgSilencePeriod,
 			"maxQueueSize":     bgMaxQueueSize,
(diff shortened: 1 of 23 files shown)


fs/fs.go
```diff
@@ -273,9 +273,9 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 		if err != nil {
 			return nil, fmt.Errorf("cannot create background fetcher: %w", err)
 		}
-		go bgFetcher.Run(context.Background())
+		go bgFetcher.Run(ctx)
```
sondavidb, line 276: Attached the background fetcher to the context of the filesystem here, this made more sense to me than allowing it to run indefinitely since if the filesystem context gets cancelled, the background fetcher should also cease operations.

fs/fs.go
```diff
@@ -258,7 +258,7 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 	var bgFetcher *bf.BackgroundFetcher
 
 	if !cfg.BackgroundFetchConfig.Disable {
-		log.G(context.Background()).WithFields(logrus.Fields{
+		log.L.WithFields(logrus.Fields{
```
Kern--: I'm not sure why we weren't using `log.G(ctx)` here or below, but I think we probably should use that instead.

sondavidb: Yeah makes sense. We used log.L a little further down as well but it was for the context of trace FUSE metrics so I think not attaching that to the current Mount context is fine, since FUSE traces are probably not useful to attach to a specific mount call. Unless you think this is useful extra info?

fs/layer/layer.go
```diff
@@ -300,7 +300,7 @@ func (r *Resolver) Resolve(ctx context.Context, hosts []docker.RegistryHost, ref
 	}
 
 	// log ztoc info
-	log.G(context.Background()).WithFields(logrus.Fields{
+	log.L.WithFields(logrus.Fields{
```
sondavidb: Switched this to log.G(ctx) since we have one and this is a synchronous log call for downloading a specific zTOC

Kern-- (dismissed):

sondavidb:
Went ahead and also removed all the TODO contexts, and removed use of `errgroup.WithContext` with just a nil variable, which should functionally be the same.

Kern-- (approved):

Swapnanil-Gupta (approved):

PR #1579, opened 2025-05-28
Add -config-dump flag

ginglis13:
**Issue #, if available:**

Closes #1570

**Description of changes:**

Add a new flag to dump the contents of the resolved soci-snapshotter-grpc configuration

**Testing performed:**

Default config:

<details>

```
./out/soci-snapshotter-grpc --config-dump
http_cache_type = ''
filesystem_cache_type = ''
resolve_result_entry = 0
debug = false
disable_verification = false
max_concurrency = 100
no_prometheus = false
mount_timeout_sec = 30
fuse_metrics_emit_wait_duration_sec = 60
metrics_address = ''
metrics_network = 'tcp'
debug_address = ''
metadata_store = 'db'
skip_check_snapshotter_supported = false

[http]
  DialTimeoutMsec = 3000
  ResponseHeaderTimeoutMsec = 3000
  RequestTimeoutMsec = 30000
  MaxRetries = 8
  MinWaitMsec = 30
  MaxWaitMsec = 300000

[blob]
  valid_interval = 60
  fetching_timeout_sec = 300
  max_retries = 8
  min_wait_msec = 30
  max_wait_msec = 300000
  check_always = false
  force_single_range_mode = false
  max_span_verification_retries = 0

[directory_cache]
  max_lru_cache_entry = 0
  max_cache_fds = 0
  sync_add = false
  direct = true

[fuse]
  attr_timeout = 1
  entry_timeout = 1
  negative_timeout = 1
  log_fuse_operations = false

[background_fetch]
  disable = false
  silence_period_msec = 30000
  fetch_period_msec = 500
  max_queue_size = 100
  emit_metric_period_sec = 10

[content_store]
  type = 'soci'
  containerd_address = ''

[pull_modes]
  [pull_modes.soci_v1]
    enable = false

  [pull_modes.soci_v2]
    enable = true

[kubeconfig_keychain]
  enable_keychain = false
  kubeconfig_path = ''

[cri_keychain]
  enable_keychain = false
  image_service_path = '/run/containerd/containerd.sock'

[resolver]

[snapshotter]
  min_layer_size = 0
  allow_invalid_mounts_on_restart = false
```

</details>

Create `/etc/soci-snapshotter-grpc/config.toml`

```
[snapshotter]
min_layer_size = 111
```

Dump config:

```
./out/soci-snapshotter-grpc --config-dump

...

[snapshotter]
  min_layer_size = 111
  allow_invalid_mounts_on_restart = false
...
```

Start snapshotter as normal:

```
[...]

PR #1693, opened 2025-08-28
Add soci-snapshotter-grpc config dump command

Kern--:
**Issue #, if available:**
Closes https://github.com/awslabs/soci-snapshotter/pull/1579

I rewrote that PR now that we support urfave in the snapshotter and go-toml/v2 everywhere.

**Description of changes:**

**Testing performed:**

### DIFF
--- a/cmd/soci-snapshotter-grpc/config.go
+++ b/cmd/soci-snapshotter-grpc/config.go
@@ -1,0 +1,47 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"os"
+
+	"github.com/awslabs/soci-snapshotter/config"
+	"github.com/containerd/log"
+	"github.com/pelletier/go-toml/v2"
+	"github.com/urfave/cli/v3"
+)
+
+var ConfigCommand = &cli.Command{
+	Name:  "config",
+	Usage: "Manage configuration",
+	Commands: []*cli.Command{
+		{
+			Name:  "dump",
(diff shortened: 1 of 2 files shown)


cmd/soci-snapshotter-grpc/config.go
```diff
@@ -0,0 +1,47 @@
...
+	Name:  "config",
+	Usage: "Manage configuration",
+	Commands: []*cli.Command{
+		{
+			Name:  "dump",
+			Usage: "Dump configuration",
+			Action: func(ctx context.Context, cmd *cli.Command) error {
+				cfg, err := config.NewConfigFromToml(cmd.String("config"))
+				if err != nil {
+					log.G(ctx).WithError(err).Fatal(err)
```
sondavidb, line 39: I think we should return here?

cmd/soci-snapshotter-grpc/config.go
sondavidb, line 1: nit: Is this the right place for it? This feels like it should be a CLI command. I understand it conceptually since the CLI doesn't actually use the config here but it feels weird that the daemon also lets you have a config dump command that doesn't actually run the daemon.

sondavidb, line 1: Also should we add a commands subfolder to mimic the CLI? Not sure if that would just confuse people though.

cmd/soci-snapshotter-grpc/config.go
Kern--: `Fatal` calls `os.Exit(1)`. I can make that explicit if you prefer.

cmd/soci-snapshotter-grpc/config.go
Kern--: This behaves the same as `containerd config dump`. The CLI is probably not the right place for it because there is no need for the CLI on server side today.

cmd/soci-snapshotter-grpc/config.go
sondavidb: Oh ok that makes sense then, I think because it's fatal it should imply that it ceases operations so I don't think we need to clarify this.

cmd/soci-snapshotter-grpc/config.go
sondavidb: Oh ok, if containerd has a precedent for this then I'm okay mimicking that.

Kern--: I think maybe we should consider that if the number of commands grows. I'm not sure adding more structure makes things simpler right now. What do you think?

sondavidb: More or less what I'm thinking yeah, because this is literally one command I'm OK leaving it as is

PR #1702, opened 2025-09-04
Add soci-snapshotter-grpc config default command

### LINKED ISSUE #568: [FEATURE] Add `soci-snapshotter-grpc config` command
### Description

Containerd has a [config command](https://github.com/containerd/containerd/blob/main/cmd/containerd/command/config.go) that can dump either the default config or the actual config as parsed from the config files. We should consider a similar command.

I think this is open for suggestion, but I would like the commands to only dump the actual configs with a not about where it's getting it's data. E.g.

```
$ soci-snapshotter-grpc config
# config based on /etc/soci-snapshotter-grpc/config.toml
metrics_address = "localhost:1339"
...
```
or if there is no config file
```
$ soci-snapshotter-grpc config
# config based on the default config
metrics_address = ""
...
```

The important piece is that the output should be valid toml and nothing else so that you can just write it to a file. E.g.
```
$ soci-snapshotter-gprc config > /etc/soci-snapshotter-grpc/config.toml
```

Should write out a config to `/etc/soci-snapshotter-grpc/config.toml` with all of the non-nullable config parameters set. 

_No response_

_No response_

### LINKED ISSUE #1570: [FEATURE] Add `config` subcommand
### Description

I would like to dump soci's config as it would be loaded by the snapshotter. The same behavior as containerd's equivalent `config` subcommand would be preferred.

```
soci-snapshotter-grpc config
NAME:
   soci-snapshotter-grpc config - Information on the soci config

USAGE:
   soci-snapshotter-grpc config command [command options] [arguments...]

COMMANDS:
   default  See the output of the default config
   dump     See the output of the final main config with imported in subconfig files

OPTIONS:
   --help, -h  show help
```

_No response_

_No response_

sondavidb:
**Issue #, if available:**
Closes #1570
Closes #568 

**Description of changes:**
Add `soci-snapshotter-grpc config default` command

**Testing performed:**
Just a manual visual verification that it looked correct. Seems to align with our defaults.
<details>
<summary>Output of "soci-snapshotter-grpc config default"</summary>

```
$ out/soci-snapshotter-grpc config default
http_cache_type = ''
filesystem_cache_type = ''
resolve_result_entry = 0
debug = false
disable_verification = false
max_concurrency = 100
no_prometheus = false
mount_timeout_sec = 30
fuse_metrics_emit_wait_duration_sec = 60
metrics_address = ''
metrics_network = 'tcp'
debug_address = ''
metadata_store = 'db'
skip_check_snapshotter_supported = false

[http]
  DialTimeoutMsec = 3000
  ResponseHeaderTimeoutMsec = 3000
  RequestTimeoutMsec = 300000
  MaxRetries = 8
  MinWaitMsec = 30
  MaxWaitMsec = 300000

[blob]
  valid_interval = 60
  fetching_timeout_sec = 300
  max_retries = 8
  min_wait_msec = 30
  max_wait_msec = 300000
  check_always = false
  force_single_range_mode = false
  max_span_verification_retries = 0

[directory_cache]
  max_lru_cache_entry = 0
  max_cache_fds = 0
  sync_add = false
  direct = true

[fuse]
  attr_timeout = 1
  entry_timeout = 1
  negative_timeout = 1
  log_fuse_operations = false

[background_fetch]
  disable = false
  silence_period_msec = 30000
  fetch_period_msec = 500
  max_queue_size = 100
  emit_metric_period_sec = 10

[content_store]
  type = 'soci'
  containerd_address = '/run/containerd/containerd.sock'

[pull_modes]
  [pull_modes.soci_v1]
    enable = false

  [pull_modes.soci_v2]
    enable = true

    max_concurrent_downloads = -1
    max_concurrent_downloads_per_image = 3
    concurrent_download_chunk_size = ''
    max_concurrent_unpacks = -1
    max_concurrent_unpacks_per_image = 1
    discard_unpacked_layers = false
    enable = false

[kubeconfig_keychain]
  enable_keychain = false
  kubeconfig_path = ''

[cri_keychain]
  enable_keychain = false
  image_service_path = '/run/containerd/containerd.sock'

[resolver]

[snapshotter]
  min_layer_size = 0
  allow_invalid_mounts_on_restart = false
```
</details>

### DIFF
--- a/cmd/soci-snapshotter-grpc/config.go
+++ b/cmd/soci-snapshotter-grpc/config.go
@@ -39,8 +39,7 @@ var ConfigCommand = &cli.Command{
 					log.G(ctx).WithError(err).Fatal(err)
 				}
 
-				toml.NewEncoder(os.Stdout).SetIndentTables(true).Encode(cfg)
-				return nil
+				return toml.NewEncoder(os.Stdout).SetIndentTables(true).Encode(cfg)
 			},
 		},
 	},
--- a/cmd/soci-snapshotter-grpc/config.go
+++ b/cmd/soci-snapshotter-grpc/config.go
@@ -42,5 +42,13 @@ var ConfigCommand = &cli.Command{
 				return toml.NewEncoder(os.Stdout).SetIndentTables(true).Encode(cfg)
 			},
 		},
+		{
+			Name:  "default",
+			Usage: "Print default configuration",
+			Action: func(ctx context.Context, cmd *cli.Command) error {
+				cfg := config.NewConfig()
+				return toml.NewEncoder(os.Stdout).SetIndentTables(true).Encode(cfg)
+			},
+		},
 	},
 }
--- a/config/config.toml
+++ b/config/config.toml
@@ -1,59 +1,59 @@
-# An example config showing all of the toml variables used.
-
-# Copy to /etc/soci-snapshotter-grpc/config.toml
-# to use on your system.
-
-# NOTE: Many variables set to zero are just an indicator
-# to use the built-in default. These values may change over time,
-# which is why the config uses zeroes.
-# Please see docs/config.md for more info on what these variables do.
-
-# TODO: Can we put these inside dedicated TOML vars to refer to them?
-
-# config/fs.go FSConfig
-http_cache_type=""
-filesystem_cache_type=""
-resolve_result_entry=0
-debug=false
-disable_verification=false
-max_concurrency=0  # Actually zero
-no_prometheus=false
-mount_timeout_sec=0
-fuse_metrics_emit_wait_duration_sec=0
-
-## config/config.go Config
-
-metrics_address=""
-metrics_network="" # Uses default metrics network
-# no_prometheus=true # Defined above, can't be redeclared
-debug_address=""
-metadata_store="db"
-skip_check_snapshotter_supported=false
-
-[pull_modes.parallel_pull_unpack]
-enable=false
-max_concurrent_downloads=-1
-max_concurrent_downloads_per_image=3
-concurrent_download_chunk_size=""
-max_concurrent_unpacks=-1
-max_concurrent_unpacks_per_image=1
-discard_unpacked_layers=false
-
-[pull_modes.parallel_pull_unpack.decompress_streams."gzip"]
-path = "/usr/bin/gzip"
-args = ["-d", "-c"]
+http_cache_type = ''
+filesystem_cache_type = ''
+resolve_result_entry = 0
+debug = false
+disable_verification = false
+max_concurrency = 100
+no_prometheus = false
+mount_timeout_sec = 30
+fuse_metrics_emit_wait_duration_sec = 60
+metrics_address = ''
+metrics_network = 'tcp'
+debug_address = ''
+metadata_store = 'db'
+skip_check_snapshotter_supported = false
 
 [http]
-MaxRetries=0
-MinWaitMsec=0
-MaxWaitMsec=0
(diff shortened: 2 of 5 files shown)


cmd/soci-snapshotter-grpc/config.go
```diff
@@ -43,5 +43,14 @@ var ConfigCommand = &cli.Command{
 				return nil
 			},
 		},
+		{
+			Name:  "default",
+			Usage: "Print default configuration",
+			Action: func(ctx context.Context, cmd *cli.Command) error {
+				cfg := config.NewConfig()
+				toml.NewEncoder(os.Stdout).SetIndentTables(true).Encode(cfg)
```
Kern--: We should probably return the error from Encode (my previous PR should have done that too.)

sondavidb:
Realized that we still have the manual config in `config/config.toml` which is no longer necessary after this change. The integration test is still useful since it ensures our default TOML still runs but I've added a Makefile target to regenerate the config file easily, and a check in our prebuild workflow that will ensure this stays updated.

config/config.toml
```diff
@@ -1,151 +1,85 @@
...
-[pull_modes.parallel_pull_unpack]
-enable=false
-max_concurrent_downloads=-1
-max_concurrent_downloads_per_image=3
-concurrent_download_chunk_size=""
-max_concurrent_unpacks=-1
-max_concurrent_unpacks_per_image=1
-discard_unpacked_layers=false
-
-[pull_modes.parallel_pull_unpack.decompress_streams."gzip"]
```
Kern--, line 42: Should this config still be printed?

sondavidb: Probably not. By default these aren't initialized. With that said, we do also store default values for `parallel_pull_unpack` mode even when it's disabled so maybe not a bad idea to include this.

sondavidb: Yeah I decided to just do it since we do have default values for `parallel_pull_unpack`, and even if the binary is missing on their system for some reason this won't instantiate unless they enable the pull mode.

sondavidb:
So there was a legitimate linter issue, but there was a really weird [error](https://github.com/awslabs/soci-snapshotter/actions/runs/17502330483/job/49717974028?pr=1702#step:5:617) that occurred in running x86 + 1.7.27. I got worried because it mentioned concurrent map write + iteration, but it looks like it's related to our LogMonitor funcs in a lazy-loading test, and not related my changes. Really weird regardless. We should probably keep an eye on this...

sondavidb:
It does point out [shell.go line 146](https://github.com/awslabs/soci-snapshotter/blob/bf29e77745b3d06dce9effe3d5e38496ea009ab4/util/testutil/shell.go#L146) as being the culprit for iterating. Looking at the code it seems like our scanLog function will always iterate through the scanner as long as we are good to continue scanning. Our log monitors when created immediately start scanning constantly. If my understanding of the Scanner is correct, it seems like this will be constantly checking the stdout/stderr from the moment the testmonitor is created, and the monitor is always added after creation. So I'm surprised this hasn't happened before. I feel like I'm missing something maybe?

Anyway, none of this is relevant to the PR at hand. I'll look into this later.

sondavidb:
All test failures are network errors. Probably not relevant. Rerunning.

Kern-- (dismissed):

sondavidb:
Found that my change to add default decompression streams broke default behavior, didn't break anything in the tests but in practice it limited the decompressor used by end users if they don't specify one which is not great. Reverted the change.

henry118 (approved):

Kern-- (approved):

Issue #951, opened 2023-11-20
[Bug] Dedupe Span Cache


sondavidb:
### Description

We currently have two separate ways of caching span cache. One is our [span cache](https://github.com/awslabs/soci-snapshotter/blob/6dcfca7bc0c70b00a5f23813d468379d6fbc8de5/fs/layer/layer.go#L272), which is what we're using. The other is the [http cache](https://github.com/awslabs/soci-snapshotter/blob/6dcfca7bc0c70b00a5f23813d468379d6fbc8de5/fs/layer/layer.go#L375), which is not used after creation. We should probably remove the latter.

### Describe the results you expected

a

### Host information

1. OS: 
2. Snapshotter Version: 
3. Containerd Version:

Kern--:
This might either be fixed by or the behavior changed by #534, so I think we should block this for now.

PR #1710, opened 2025-09-12
Fix resource cleanup

Shubhranshu153:
**Issue #, if available:**
1. Spancache and httpcache are not cleanedup
2. Background fetcher doesn't stop when layer is getting cleanedup

**Description of changes:**
1. Httpcache is not used in the codebase, Removed it.
2. Spancache eviction was not called as the done was a idempotent call instead of reducing the count. And we dont call remove of the cache for requesting eviction.
3. When all references of spancache is removed then the spanmanager has a finalizer is called which releases the cache resources. This is non-deterministic and scheduled by go runtime. 
4. Background fetcher doesnt fetch spans on restarts.

TODO:
1. Add test to verify background fetcher is stopped
2. Add cache cleanup on immediately after eviction rather than depending on finalizer, as it doesnt work well on restart once the context is lost.
3. Cleanup of resources on restarts
5. Add test to verify resource cleanup

**Testing performed:**
1. Verification that cache gets cleared
```
Before cleanup
sudo du -chx --max-depth=1 /var/lib/soci-snap
shotter-grpc/soci/
200M    /var/lib/soci-snapshotter-grpc/soci/spancache
200M    /var/lib/soci-snapshotter-grpc/soci/
200M    total
```

```
After removing of the image (before the fix this won't get cleanedup
sudo du -chx --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/
4.0K    /var/lib/soci-snapshotter-grpc/soci/spancache
8.0K    /var/lib/soci-snapshotter-grpc/soci/
8.0K    total

```
3. Verification that bg fetcher doesnt get called on restarts

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -1228,8 +1228,13 @@ func (fs *filesystem) Unmount(ctx context.Context, mountpoint string) error {
 	delete(fs.layer, mountpoint)
 	// If the mountpoint is an id-mapped layer, it is pointing to the
 	// underlying layer, so we cannot call done on it.
+	// We do a evict call to call the registered evict functions.
+	// This cleans up layer resources and removes it from the resolver cache,
+	// preparing the layer for removal and freeing associated memory done on
+	// call of the finalizer in span manager
 	if !isIDMappedDir(mountpoint) {
 		l.Done()
+		fs.resolver.Evict(l.GetCacheRefKey())
 	}
 	fs.layerMu.Unlock()
 	fs.metricsController.Remove(mountpoint)
--- a/fs/fs_test.go
+++ b/fs/fs_test.go
@@ -93,6 +93,7 @@ func (l *breakableLayer) SkipVerify()                          {}
 func (l *breakableLayer) ReadAt([]byte, int64, ...remote.Option) (int, error) {
 	return 0, fmt.Errorf("fail")
 }
+func (l *breakableLayer) GetCacheRefKey() string { return "" }
 func (l *breakableLayer) BackgroundFetch() error { return fmt.Errorf("fail") }
 func (l *breakableLayer) Check() error {
 	if !l.success {
--- a/fs/layer/layer.go
+++ b/fs/layer/layer.go
@@ -104,6 +104,9 @@ type Layer interface {
 	// Done releases the reference to this layer. The resources related to this layer will be
 	// discarded sooner or later. Queries after calling this function won't be serviced.
 	Done()
+
+	// GetCacheRefKey returns the reference key for the cache used by the layer
+	GetCacheRefKey() string
 }
 
 // Info is the current status of a layer.
@@ -226,6 +229,12 @@ func newCache(root string, cacheType string, cfg config.FSConfig) (cache.BlobCac
 	)
 }
 
+func (r *Resolver) Evict(name string) {
+	r.layerCacheMu.Lock()
+	r.layerCache.Remove(name)
+	r.layerCacheMu.Unlock()
+}
+
 // Resolve resolves a layer based on the passed layer blob information.
 func (r *Resolver) Resolve(ctx context.Context, hosts []docker.RegistryHost, refspec reference.Spec, desc, sociDesc ocispec.Descriptor, opCounter *FuseOperationCounter, disableVerification bool, metadataOpts ...metadata.Option) (_ Layer, retErr error) {
 	name := refspec.String() + "/" + desc.Digest.String()
@@ -339,7 +348,7 @@ func (r *Resolver) Resolve(ctx context.Context, hosts []docker.RegistryHost, ref
 	}
 	disableXAttrs := getDisableXAttrAnnotation(sociDesc)
 	// Combine layer information together and cache it.
-	l := newLayer(r, desc, blobR, vr, bgLayerResolver, opCounter, disableXAttrs)
+	l := newLayer(r, desc, name, blobR, vr, bgLayerResolver, opCounter, disableXAttrs)
 	r.layerCacheMu.Lock()
 	cachedL, done2, added := r.layerCache.Add(name, l)
 	r.layerCacheMu.Unlock()
@@ -387,6 +396,7 @@ func (r *Resolver) resolveBlob(ctx context.Context, hosts []docker.RegistryHost,
 func newLayer(
 	resolver *Resolver,
 	desc ocispec.Descriptor,
+	cacheRefKey string,
 	blob *blobRef,
 	r reader.Reader,
 	bgResolver backgroundfetcher.Resolver,
@@ -396,6 +406,7 @@ func newLayer(
 	return &layer{
 		resolver:             resolver,
 		desc:                 desc,
+		cacheRefKey:          cacheRefKey,
 		blob:                 blob,
 		r:                    r,
 		bgResolver:           bgResolver,
@@ -405,9 +416,10 @@ func newLayer(
 }
 
 type layer struct {
-	resolver *Resolver
-	desc     ocispec.Descriptor
-	blob     *blobRef
+	resolver    *Resolver
+	desc        ocispec.Descriptor
+	cacheRefKey string
+	blob        *blobRef
 
 	bgResolver backgroundfetcher.Resolver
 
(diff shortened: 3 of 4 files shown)


fs/fs.go
```diff
@@ -1230,6 +1230,7 @@ func (fs *filesystem) Unmount(ctx context.Context, mountpoint string) error {
 	// underlying layer, so we cannot call done on it.
 	if !isIDMappedDir(mountpoint) {
 		l.Done()
+		fs.resolver.Evict(l.GetCacheRefKey())
```
Shubhranshu153, line 1237: Add a comment why we call evict after Done is called.

snapshot/snapshot.go
```diff
@@ -752,8 +751,10 @@ func (o *snapshotter) unmountSnapshotDirectory(ctx context.Context, dir string)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
 			// nothing to unmount as directory does not exist
+			log.L.Infof("not mounted err ")
```
Shubhranshu153: remove the logs

util/lrucache/lrucache.go
```diff
@@ -89,7 +89,7 @@ func (c *Cache) Add(key string, value interface{}) (cachedValue interface{}, don
 	if o, ok := c.cache.Get(key); ok {
```
Shubhranshu153, line 89: This change has a issue where a double call to done will evict the cache. 
Need to add a guardrail around it.

Kern--: Did the`sync.Once` that was removed not act as that guardrail?

Shubhranshu153: sync.Once can be called only once ever. so if there are 2 reference created the count will become 3 and the decremented once so count will become 2 . Any further calls to done will be inconsequential making the logic incorrect.

Kern--: You get a new closure over a new sync.Once on every call to `Get`. Does that not work?

sondavidb:
Merging this should close #951

fs/fs.go
```diff
@@ -1228,8 +1228,10 @@ func (fs *filesystem) Unmount(ctx context.Context, mountpoint string) error {
 	delete(fs.layer, mountpoint)
 	// If the mountpoint is an id-mapped layer, it is pointing to the
 	// underlying layer, so we cannot call done on it.
+	// We do a evict call to call the registered evict functions.
```
sondavidb, line 1231: nit: This comment only states what the function does — we should explain why we care about evicting.

sondavidb (commented):
Is it possible to add a unit test? Not sure how useful a mock of this would be though.

fs/fs.go
Kern--: Can you also clarify why this is safe?

IIRC, the idea is that once it's removed from the cache, it can be eligible for garbage collection which may eventually call the finalizer, right?

In the future, if we fix everything so that cleanup is deterministic, does it still make sense to call `Evict` manually? Do we need to be more careful that eviction only happens if the refcount is 0?

wswsmao:
It appears the changes haven't taken effect. Testing with commit 00edc4f614002399d745b29d5633aebcf9b949f9

**Steps:**

1. **Pull the image**
```bash
nerdctl pull --snapshotter soci example.com/test:soci
```

2. **Wait for background process**
```bash
Every 1.0s: du -h --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/spancache                   VM-33-248-tlinux: Wed Nov  5 02:42:00 2025

33M     /var/lib/soci-snapshotter-grpc/soci/spancache/1231516239
33M     /var/lib/soci-snapshotter-grpc/soci/spancache
```

3. **Remove the image**
```bash
nerdctl images -q | xargs nerdctl rmi -f
```

**spancache status:**
```bash
Every 1.0s: du -h --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/spancache                   VM-33-248-tlinux: Wed Nov  5 02:42:16 2025

33M     /var/lib/soci-snapshotter-grpc/soci/spancache/1231516239
33M     /var/lib/soci-snapshotter-grpc/soci/spancache
```

wswsmao:
> It appears the changes haven't taken effect. Testing with commit [00edc4f](https://github.com/awslabs/soci-snapshotter/commit/00edc4f614002399d745b29d5633aebcf9b949f9)
> 
> **Steps:**
> 
> 1. **Pull the image**
> 
> ```shell
> nerdctl pull --snapshotter soci example.com/test:soci
> ```
> 
> 2. **Wait for background process**
> 
> ```shell
> Every 1.0s: du -h --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/spancache                   VM-33-248-tlinux: Wed Nov  5 02:42:00 2025
> 
> 33M     /var/lib/soci-snapshotter-grpc/soci/spancache/1231516239
> 33M     /var/lib/soci-snapshotter-grpc/soci/spancache
> ```
> 
> 3. **Remove the image**
> 
> ```shell
> nerdctl images -q | xargs nerdctl rmi -f
> ```
> 
> **spancache status:**
> 
> ```shell
> Every 1.0s: du -h --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/spancache                   VM-33-248-tlinux: Wed Nov  5 02:42:16 2025
> 
> 33M     /var/lib/soci-snapshotter-grpc/soci/spancache/1231516239
> 33M     /var/lib/soci-snapshotter-grpc/soci/spancache
> ```

maybe we can use this https://github.com/awslabs/soci-snapshotter/pull/1745

Shubhranshu153:
@wswsmao

A finalizer is called by golang to clean up the span cache. It is not a deterministic call. So we need to wait for sometime to see the cache getting removed. As a result we dont immediately see a cleanup of the cache. We need to make sure all the raw pointers are released before we evict the span. 

For testing with nerdctl after you do a cleanup wait for 10~20 seconds to check if the spancache is cleaned up.


Also thanks for working on it. Will review your PR.

fs/fs.go
Shubhranshu153: I think evict is still needed to be called here manually as when we create the layer we already increment the layer and dont hold an reference for that.
```
	rc.initialize() // Keep this object having at least 1 ref count (will be decreased in OnEviction)
	rc.inc()        // The client references this object (will be decreased on "done")
```

When we release the cache got  we decrement it once by calling l.done()  which only releases the reference. But when we want to unmount we need to also evict the cache completely. Hence we need to evict it.

Eviction only happens if the refcount is 0 as the reference of the layer would increment the count and decremented only when done is called for the respective reference.

Shubhranshu153:
Pull the image
```
Nov 06 04:43:26 lima-finch soci-snapshotter-grpc[8070]: {"level":"debug","msg":"Spancache finalizer called. Cleaning up spancache","time":"2025-11-06T04:43:26.268460162-08:00"}
```
Remove the image
```
sudo nerdctl rmi $(sudo nerdctl image ls -q)
```


Finalizer call, finalizer calls are made async and handled by golang. (To see this log soci-snapshooter needs to be run in debug mode)
```
Nov 06 04:43:26 lima-finch soci-snapshotter-grpc[8070]: {"level":"debug","msg":"Spancache finalizer called. Cleaning up spancache","time":"2025-11-06T04:43:26.268460162-08:00"}
```



Cache
```
[shubhum@lima-finch soci-snapshotter]$ sudo du -chx --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/
85M     /var/lib/soci-snapshotter-grpc/soci/spancache
85M     /var/lib/soci-snapshotter-grpc/soci/
85M     total
[shubhum@lima-finch soci-snapshotter]$ sudo du -chx --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/
4.0K    /var/lib/soci-snapshotter-grpc/soci/spancache
8.0K    /var/lib/soci-snapshotter-grpc/soci/
8.0K    total
[shubhum@lima-finch soci-snapshotter]$ 
```


The case it doesnt handle at the moment is restarts the snapshotter. From my testings the case you are seeing may be a case when we didnt actually wait for the cache to get filled up initially during the test and hence we dont see the cleanup compared to it. @wswsmao

Shubhranshu153:
This fix has been addressed by a different PR.

PR #1718, opened 2025-10-06
ci: Properly check if Docker Compose is installed

sondavidb:
**Issue #, if available:**

**Description of changes:**
The `compose.Supported()` only checks the Docker version, not the Docker Compose version. So, if a user did not have Docker Compose installed, it erroneously passed when it should not have, causing an unclear error message down the line (`unknown shorthand flag: 'f' in -f`) which was difficult to debug for unfamiliar users. So I changed the Supported check to make sure it caught this issue.

Additionally, `docker compose version` would only return `exit status 1` if Docker Compose was not installed, so I amended the error message to ensure it is a little more clear when it fails. (The previous check for docker version would say `executable file not found in $PATH` if it failed, which is descriptive enough for an error message, so that remained untouched.)

**Testing performed:**
Manually did `chmod -x ~/.docker/cli-plugins/docker-compose` and ensured that integration tests run with executable perms and do not without. Also verified the same with `/usr/bin/docker`.

### DIFF
--- a/util/dockershell/compose/compose.go
+++ b/util/dockershell/compose/compose.go
@@ -52,7 +52,13 @@ const (
 
 // Supported checks if this pkg can run on the current system.
 func Supported() error {
-	return exec.Command("docker", "--version").Run()
+	if err := exec.Command("docker", "--version").Run(); err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


Swapnanil-Gupta:
Thanks!

Shubhranshu153 (approved):

Issue #1669, opened 2025-08-05
Skip ztoc generation if one already exists.


Kern--:
### Discussed in https://github.com/awslabs/soci-snapshotter/discussions/1665

<div type='discussions-op-text'>

<sup>Originally posted by **wsmlby** August  3, 2025</sup>
Hey team,
I have been testing soci and find it amazing how much it can accelerate the launch speed.
I am wondering if there is something I can help to improve the image convert speed.

I noticed that if i convert 2 images with only the last layer different from another, one after another, converting the image still takes the same amount of time. 
The output shows the last layer being skipped due to small size, but the 2nd convert still take non-trivial amount of time(and 100% CPU during that time)
I wonder if there is any obvious reason I am missing that soci convert don't have an option to skip layer convert if it has been converted.

Looking at [soci_index](https://github.com/awslabs/soci-snapshotter/blob/70d0a37970ebd201c0226a8e177d27a92de2c260/soci/soci_index.go#L594),
It looks this can be a very natural place to add a optional check to skip re-generating the ztoc for a layer that already have a ztoc layer generated. 

For images with large base layers but only update the last few layers, I assume this can hugely accelerating the convert process. 

Another related question is about the output when running the convert: 
The layer sha256 doesn't seem to be related to the layer sha256 seen by nerdctl inspect, am I missing something here? 

```

ztoc skipped - layer sha256:7d61afc7a3ac0a2b20643507f9ed50f42939d6838db55de1736527a711e98038 (application/vnd.docker.image.rootfs.diff.tar.gzip) size 63861 is less than min-layer-size 10485760
ztoc skipped - layer sha256:4ace4985aaa6c68a61c158cd4c17ccf842a62940f9857b2760873d2f75b06f60 (application/vnd.docker.image.rootfs.diff.tar.gzip) size 3031 is less than min-layer-size 10485760
ztoc skipped - layer sha256:8bd2762ffdd9e59311bc1d8909932d3363ca4897858f7f71aa837969b22754d8 (application/vnd.docker.image.rootfs.diff.tar.gzip) size 1683 is less than min-layer-size 10485760
ztoc skipped - layer sha256:2a5ee6fadd42bc8a2088ecb52dc294274388aaa285d21507e399dbebfdb54f49 (application/vnd.docker.image.rootfs.diff.tar.gzip) size 1523 is less than min-layer-size 10485760
[...]

PR #1720, opened 2025-10-07
feat: skip existing ztoc generation

Swapnanil-Gupta:
**Issue #, if available:**
Closes https://github.com/awslabs/soci-snapshotter/issues/1669

**Description of changes:**
If a ztoc for a layer already exists in the blob store and a corresponding entry exists in the artifacts store, soci will now skip regenerating the ztoc for subsequent `soci create` or `soci convert` calls. To do this, the `span-size` used to generate the ztoc is now being written to the artifact store and also in the layer annotations, so that `rebuild-db` can regenerate the artifact entries properly and take advantage of this feature.  To fall back to the previous behavior `-skip-existing-ztoc-check` flag can be used during `soci create` or `soci convert`.

**Testing performed:**

### DIFF
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -31,11 +31,12 @@ import (
 const (
 	// buildToolIdentifier is placed in annotations of the SOCI index
 	// to help identify how a SOCI index was created
-	buildToolIdentifier = "AWS SOCI CLI v0.2"
-	spanSizeFlag        = "span-size"
-	minLayerSizeFlag    = "min-layer-size"
-	optimizationFlag    = "optimizations"
-	sociIndexGCLabel    = "containerd.io/gc.ref.content.soci-index"
+	buildToolIdentifier       = "AWS SOCI CLI v0.2"
+	spanSizeFlag              = "span-size"
+	minLayerSizeFlag          = "min-layer-size"
+	optimizationFlag          = "optimizations"
+	skipExistingZtocCheckFlag = "skip-existing-ztoc-check"
+	sociIndexGCLabel          = "containerd.io/gc.ref.content.soci-index"
 )
 
 // CreateCommand creates SOCI index for an image
@@ -61,6 +62,11 @@ var CreateCommand = &cli.Command{
 		&cli.StringSliceFlag{
 			Name:  optimizationFlag,
 			Usage: fmt.Sprintf("(Experimental) Enable optional optimizations. Valid values are %v", soci.Optimizations),
+		},
+		&cli.BoolFlag{
+			Name:  skipExistingZtocCheckFlag,
+			Usage: "Skip checking if zTOCs already exist for layers or not. Defaults to false.",
+			Value: false,
 		},
 	),
 	Action: func(ctx context.Context, cmd *cli.Command) error {
@@ -93,6 +99,7 @@ var CreateCommand = &cli.Command{
 
 		spanSize := cmd.Int64(spanSizeFlag)
 		minLayerSize := cmd.Int64(minLayerSizeFlag)
+		skipExistingZtocCheck := cmd.Bool(skipExistingZtocCheckFlag)
 
 		blobStore, err := store.NewContentStore(internal.ContentStoreOptions(ctx, cmd)...)
 		if err != nil {
@@ -118,6 +125,7 @@ var CreateCommand = &cli.Command{
 			soci.WithBuildToolIdentifier(buildToolIdentifier),
 			soci.WithOptimizations(optimizations),
 			soci.WithArtifactsDb(artifactsDb),
+			soci.WithSkipExistingZtocCheck(skipExistingZtocCheck),
 		}
 
 		builder, err := soci.NewIndexBuilder(cs, blobStore, builderOpts...)
--- a/integration/convert_test.go
+++ b/integration/convert_test.go
@@ -17,21 +17,86 @@
 package integration
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
+	"path/filepath"
 	"slices"
 	"strings"
 	"testing"
 
+	"github.com/awslabs/soci-snapshotter/config"
 	"github.com/awslabs/soci-snapshotter/soci"
 	shell "github.com/awslabs/soci-snapshotter/util/dockershell"
+	"github.com/awslabs/soci-snapshotter/util/testutil"
 	"github.com/containerd/platforms"
+	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 var (
 	convertImages = []string{nginxImage, rabbitmqImage, drupalImage, ubuntuImage}
 )
+
+func TestConvertWithExistingZtocCheck(t *testing.T) {
+	sh, done := newSnapshotterBaseShell(t)
+	defer done()
+	rebootContainerd(t, sh, "", "")
+
+	// if we build an index for the image first and then convert it
+	// without the --skip-existing-ztoc-check flag, all ztoc's should be skipped
+	image := dockerhub(nginxAlpineImage)
+	indexDigest := buildIndex(sh, image)
+	if indexDigest == "" {
+		t.Fatal("failed to get soci index for test image")
+	}
+	contentStoreBlobPath, _ := testutil.GetContentStoreBlobPath(config.DefaultContentStoreType)
+	parsedDigest, err := digest.Parse(indexDigest)
+	if err != nil {
+		t.Fatalf("cannot parse digest: %v", err)
+	}
+	checkpoints := fetchContentFromPath(sh, filepath.Join(contentStoreBlobPath, parsedDigest.Encoded()))
+	var index soci.Index
+	err = soci.DecodeIndex(bytes.NewReader(checkpoints), &index)
+	if err != nil {
+		t.Fatalf("cannot get index data: %v", err)
+	}
+
+	testCases := []struct {
+		name                      string
+		skipExistingZtocCheckFlag bool
+		numZtocSkipped            int
+	}{
+		{
+			name:                      "test soci convert without --skip-existing-ztoc-check flag",
+			skipExistingZtocCheckFlag: false,
+			numZtocSkipped:            len(index.Blobs),
+		},
+		{
+			name:                      "test soci convert with --skip-existing-ztoc-check flag",
+			skipExistingZtocCheckFlag: true,
+			numZtocSkipped:            0,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			args := []string{"soci", "convert", "--min-layer-size=0", "--platform", platforms.Format(image.platform)}
+			if tc.skipExistingZtocCheckFlag {
+				args = append(args, "--skip-existing-ztoc-check")
+			}
+			args = append(args, image.ref, image.ref+"-soci")
+			b, err := sh.CombinedOLog(args...)
+			if err != nil {
+				t.Fatalf("expected no error, got %v", err)
+			}
+
+			numSkipped := strings.Count(string(b), "already exists")
+			if numSkipped != tc.numZtocSkipped {
+				t.Fatalf("expected %v ztoc to be skipped, got %v", tc.numZtocSkipped, numSkipped)
+			}
+		})
+	}
+}
 
 func validateConversion(t *testing.T, sh *shell.Shell, originalDigest, convertedDigest string) {
 	t.Helper()
--- a/integration/create_test.go
+++ b/integration/create_test.go
@@ -34,6 +34,66 @@ import (
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
+func TestCreateWithExistingZtocCheck(t *testing.T) {
+	sh, done := newSnapshotterBaseShell(t)
+	defer done()
+	rebootContainerd(t, sh, "", "")
+
+	// if we build an index for the image first and then create it again
+	// without the --skip-existing-ztoc-check flag, all ztoc's should be skipped
+	image := dockerhub(nginxAlpineImage)
+	indexDigest := buildIndex(sh, image)
+	if indexDigest == "" {
+		t.Fatal("failed to get soci index for test image")
+	}
+	contentStoreBlobPath, _ := testutil.GetContentStoreBlobPath(config.DefaultContentStoreType)
+	parsedDigest, err := digest.Parse(indexDigest)
+	if err != nil {
+		t.Fatalf("cannot parse digest: %v", err)
+	}
+	checkpoints := fetchContentFromPath(sh, filepath.Join(contentStoreBlobPath, parsedDigest.Encoded()))
+	var index soci.Index
+	err = soci.DecodeIndex(bytes.NewReader(checkpoints), &index)
+	if err != nil {
+		t.Fatalf("cannot get index data: %v", err)
+	}
+
+	testCases := []struct {
+		name                      string
+		skipExistingZtocCheckFlag bool
+		numZtocSkipped            int
+	}{
+		{
+			name:                      "test soci create without --skip-existing-ztoc-check flag",
+			skipExistingZtocCheckFlag: false,
+			numZtocSkipped:            len(index.Blobs),
+		},
+		{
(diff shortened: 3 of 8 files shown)


soci/soci_index.go
```diff
@@ -665,21 +691,46 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
...
+		return nil
+	}
+	var existingZtoc *ocispec.Descriptor
+	cfg.artifactsDb.Walk(func(ae *ArtifactEntry) error {
+		if ae.Type == ArtifactEntryTypeLayer && ae.OriginalDigest == layerDesc.Digest.String() && ae.SpanSize == cfg.spanSize {
+			existingZtoc = &ocispec.Descriptor{
+				Digest: digest.Digest(ae.Digest),
+				Size:   ae.Size,
+			}
+			return fmt.Errorf("found existing ztoc for layer %s", layerDesc.Digest)
```
sondavidb, line 736: This error ends up being discarded, either log it, return nil, or have the parent function return an error (IMO return nil but not opposed to any solution)

soci/soci_index_test.go
```diff
@@ -30,6 +30,95 @@ import (
...
+			existingZtoc := getExistingZtocForLayer(tc.layerDesc, tc.builderConfig)
+			if existingZtoc == nil && tc.existingZtoc == nil {
+				return
+			}
+			if existingZtoc == nil || tc.existingZtoc == nil {
+				t.Fatalf(
+					"mismatch between expected and returned ztoc descriptor. expected: %v, returned: %v", tc.existingZtoc, existingZtoc)
+			}
+			if existingZtoc.Digest != tc.existingZtoc.Digest || existingZtoc.Size != tc.existingZtoc.Size {
+				t.Fatalf("returned ztoc descriptor is invalid")
```
sondavidb: Maybe mention it's specifically a mismatch?

soci/soci_index.go
Swapnanil-Gupta: I thought by returning an error once an artifact entry is found will cause the `cfg.artifactsDb.Walk` to exit early as we do not need to walk the db any further once a match is found. But yes I agree this error ends up getting discarded.

sondavidb: Yeah you're right. A comment would help clarify this I think but the behavior is fine by me

integration/index_test.go
```diff
@@ -335,3 +335,51 @@ func TestSociIndexRemove(t *testing.T) {
...
+			t.Fatal("failed to get soci index for nginx alpine 2 test image")
+		}
+
+		sh.X("soci", "index", "rm", imgDigest)
+
+		ztocsBeforeRebuild, err := getZtocDigestsForImage(sh, img2)
+		if err != nil {
+			t.Fatalf("could not get ztoc digests before rebuild-db: %v", err)
+		}
+		sh.X("soci", "--content-store", string(store.ContainerdContentStoreType), "rebuild-db")
```
sondavidb: Do we need to use the containerd content store here? We already explicitly use the containerd content store in calls to SyncWithLocalStore in rebuild-db.

integration/util_soci_test.go
```diff
@@ -78,6 +80,28 @@ func withIndexBuildConfig(newIbc indexBuildConfig) indexBuildOption {
...
+}
+
+// withRebuildDbBeforeCreate syncs the artifact store with the content store before calling "soci create"
+// We do this because the artifact store could be out of sync with the content store when 'soci create' is called.
+// This is problematic in cases where we create soci indexes for some images, delete those indexes and immediately recreate
+// them (like in TestSociIndexRemove) - as there could be ztoc entries in the artifact store which are not present in the
+// content store, causing 'soci create/convert' without --skip-existing-ztoc-check flag to throw an error.
+//
+// We can run this by default and probably remove this option in the future when the race condition with rebuild-db is solved.
+func withRunSociRebuildDbBeforeSociCreate() indexBuildOption {
```
sondavidb: nit: `withRunRebuildDbBeforeCreate`?

cmd/soci/commands/create.go
```diff
@@ -31,11 +31,12 @@ import (
...
-	buildToolIdentifier = "AWS SOCI CLI v0.2"
-	spanSizeFlag        = "span-size"
-	minLayerSizeFlag    = "min-layer-size"
-	optimizationFlag    = "optimizations"
-	sociIndexGCLabel    = "containerd.io/gc.ref.content.soci-index"
+	buildToolIdentifier       = "AWS SOCI CLI v0.2"
+	spanSizeFlag              = "span-size"
+	minLayerSizeFlag          = "min-layer-size"
+	optimizationFlag          = "optimizations"
+	skipExistingZtocCheckFlag = "skip-existing-ztoc-check"
```
sondavidb: nit: name feels a little verbose — maybe drop `check` at the end? Not a big deal either way

integration/index_test.go
```diff
@@ -242,10 +242,10 @@ func TestSociIndexRemove(t *testing.T) {
...
 	defer done()
 	rebootContainerd(t, sh, getContainerdConfigToml(t, false, `
-[plugins."io.containerd.gc.v1.scheduler"]
-	deletion_threshold = 1
-	startup_delay = "10ms"
-`), "")
+	[plugins."io.containerd.gc.v1.scheduler"]
+		deletion_threshold = 1
+		startup_delay = "10ms"
+	`), "")
```
sondavidb: Shouldn't change this — the indentation is intentional since we're using a raw string to write to a file. (Really we should be using our helper function instead of doing this, which we moved to for most of our tests, but out of scope for this PR.)

soci/soci_index.go
```diff
@@ -620,6 +630,25 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 		return nil, errUnsupportedLayerFormat
 	}
 
+	existingZtoc := getExistingZtocForLayer(desc, b.config)
+	if existingZtoc != nil {
+		reader, err := b.blobStore.Fetch(ctx, *existingZtoc)
+		if err != nil {
+			if errors.Is(err, errdefs.ErrNotFound) {
+				return nil, fmt.Errorf("a ztoc entry was found in the artifact store but not in the content store (try tunning \"soci rebuild-db\" first or try again with \"--skip-existing-ztoc-check\" flag): %w", err)
```
Kern--: nit:
```suggestion
				return nil, fmt.Errorf("a ztoc entry was found in the artifact store but not in the content store (try running \"soci rebuild-db\" first or try again with \"--skip-existing-ztoc-check\" flag): %w", err)
```

cmd/soci/commands/create.go
Kern--: personally, I would go with `--force` and document it as building the ztocs and index even if they already exist.

soci/artifacts.go
```diff
@@ -303,6 +307,15 @@ func (db *ArtifactsDb) addNewArtifacts(ctx context.Context, blobStorePath string
 				return err
 			}
 			for _, zt := range sociIndex.Blobs {
+				var spanSize int64
+				spanSizeStr, ok := zt.Annotations[IndexAnnotationSociSpanSize]
+				if ok {
+					parsedSpanSize, err := strconv.ParseInt(spanSizeStr, 10, 64)
```
Kern--: nit, we don't need to introduce a new variable here and an extra assignment.
```suggestion
					spanSize, err = strconv.ParseInt(spanSizeStr, 10, 64)
```

soci/soci_index.go
```diff
@@ -665,21 +694,48 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
...
+		IndexAnnotationSociSpanSize:        fmt.Sprintf("%d", b.config.spanSize),
 	}
-	b.maybeAddDisableXattrAnnotation(&ztocDesc, toc)
-	return &ztocDesc, err
+	b.maybeAddDisableXattrAnnotation(ztocDesc, toc)
+}
+
+// getExistingZtocForLayer returns a ztoc descriptor for the provided layer if an entry corresponding to the
+// layer already exists in the artifact store.
+func getExistingZtocForLayer(layerDesc ocispec.Descriptor, cfg *builderConfig) *ocispec.Descriptor {
```
Kern--: Should this be a function on the index builder instead of taking in the cfg as a parameter?

soci/soci_index.go
```diff
@@ -665,21 +694,48 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
...
+}
 
+func (b *IndexBuilder) addSociLayerAnnotations(layerDesc *ocispec.Descriptor, ztocDesc *ocispec.Descriptor, toc *ztoc.Ztoc) {
 	ztocDesc.MediaType = SociLayerMediaType
 	ztocDesc.Annotations = map[string]string{
-		IndexAnnotationImageLayerMediaType: desc.MediaType,
-		IndexAnnotationImageLayerDigest:    desc.Digest.String(),
+		IndexAnnotationImageLayerMediaType: layerDesc.MediaType,
+		IndexAnnotationImageLayerDigest:    layerDesc.Digest.String(),
+		IndexAnnotationSociSpanSize:        fmt.Sprintf("%d", b.config.spanSize),
```
Kern--: nit: `strconv.FormatInt` is generally preferable to `fmt.Sprintf`

integration/index_test.go
Swapnanil-Gupta: Not needed since it is the default. Removed it.

cmd/soci/commands/create.go
Swapnanil-Gupta: Changed to `skip-existing-ztoc`.

Swapnanil-Gupta: @Kern-- `--force` sounds like it will override other flags like `--min-layer-size` as well so I went for something a little verbose.

Kern--: `docker compose up` has a `--force-recreate` flag to rebuild containers even if the config hasn't changed. Something like that could be an option. I feel like this only needs the extra word because building the container is not the primary action, though.

I still think `soci convert --force` is clear and succinct enough. I would expect this to "force" the min layer size to what I set.

At the least, the current update is inverted. `skip-existing-ztoc` is the default behavior.

sondavidb: I do like `--force` as well. Verbosity aside, since we're explicitly using the flag in `create` it makes the point clear which is that using this flag will make a new ztoc, no matter what. `skip-existing-ztoc-check` is what it actually does, but to the end user they care more about the behavior and not what it does under the hood, so I feel `force` is more appropriate.

Swapnanil-Gupta: Changed to `--force` or `-f`.

soci/artifacts.go
```diff
@@ -468,6 +481,11 @@ func loadArtifact(artifactBkt *bolt.Bucket, digest string) (*ArtifactEntry, erro
 	if err != nil {
 		return nil, err
 	}
+	encodedSpanSize := artifactBkt.Get(bucketKeySpanSize)
+	spanSize, err := dbutil.DecodeInt(encodedSpanSize)
```
Kern--: `dbutil.DecodeInt` returns an error if the input is nil. AFAICT, this will break for artifacts created before updating SOCI.

soci/artifacts.go
```diff
@@ -310,6 +314,14 @@ func (db *ArtifactsDb) addNewArtifacts(ctx context.Context, blobStorePath string
 				return err
 			}
 			for _, zt := range sociIndex.Blobs {
+				var spanSize int64
```
Kern--, line 317: Should we distinguish between 0 and not set? Maybe by defaulting to -1?

I think we'll probably want to update `soci ztoc ls` to include span size at some point (don't need to be this PR) so distinguishing them would be useful.

Kern--:
Just a note about your commits, we only allow rebase merges, so I think the merge commits in your history are going to break. 

In general, perfer rebasing when updating from main (e.g. `git pull --rebase origin main`)

soci/artifacts.go
Swapnanil-Gupta: Since it is not possible to set span size to 0, can 0 not act as the default value when it's not set?

Kern--: You're right. I was thinking of 0 as a reasonable value for min layer size. I think this is fine.

Swapnanil-Gupta:
> Just a note about your commits, we only allow rebase merges, so I think the merge commits in your history are going to break.
> 
> In general, perfer rebasing when updating from main (e.g. `git pull --rebase origin main`)

Thanks. I have gone ahead and squashed the commits.

Kern-- (approved):

integration/convert_test.go
```diff
@@ -17,22 +17,87 @@
...
+	rebootContainerd(t, sh, "", "")
+
+	// if we build an index for the image first and then convert it
+	// without the --force flag, all ztoc's should be skipped
+	image := dockerhub(nginxAlpineImage)
+	indexDigest := buildIndex(sh, image)
+	if indexDigest == "" {
+		t.Fatal("failed to get soci index for test image")
+	}
+	contentStoreBlobPath, _ := testutil.GetContentStoreBlobPath(config.DefaultContentStoreType)
```
sondavidb, line 53: Do we want to test across multiple content stores?

integration/create_test.go
```diff
@@ -34,6 +34,66 @@ import (
...
+	rebootContainerd(t, sh, "", "")
+
+	// if we build an index for the image first and then create it again
+	// without the --force flag, all ztoc's should be skipped
+	image := dockerhub(nginxAlpineImage)
+	indexDigest := buildIndex(sh, image)
+	if indexDigest == "" {
+		t.Fatal("failed to get soci index for test image")
+	}
+	contentStoreBlobPath, _ := testutil.GetContentStoreBlobPath(config.DefaultContentStoreType)
```
sondavidb, line 49: Same here

integration/index_test.go
```diff
@@ -335,3 +335,51 @@ func TestSociIndexRemove(t *testing.T) {
 		}
 	})
 }
+
+func TestSociIndexRemoveAndRebuildWithSharedLayers(t *testing.T) {
+	sh, done := newSnapshotterBaseShell(t)
+	defer done()
+	rebootContainerd(t, sh, "", "")
+
+	t.Run("soci index rm with shared layers should not affect the other image", func(t *testing.T) {
```
sondavidb, line 344: maybe specify with containerd store, sincethis doesn't work with soci store?

Swapnanil-Gupta:
> All the stuff I left are nits really, ok to address this in a followup PR I think (but I do think we should ensure this works across all content stores)

Sure. I can follow this up with another PR to test different content stores.

soci/soci_index.go
```diff
@@ -620,6 +630,25 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 		return nil, errUnsupportedLayerFormat
 	}
 
+	existingZtoc := b.getExistingZtocForLayer(desc)
+	if existingZtoc != nil {
+		reader, err := b.blobStore.Fetch(ctx, *existingZtoc)
+		if err != nil {
+			if errors.Is(err, errdefs.ErrNotFound) {
+				return nil, fmt.Errorf("a ztoc entry was found in the artifact store but not in the content store (try running \"soci rebuild-db\" first or try again with \"--force\" flag): %w", err)
+			}
```
henry118, line 639: I don't fully understand the need of this logic. 

It's a regression of the current image conversion process and breaking the nerdctl canary https://github.com/containerd/nerdctl/issues/4622 and https://github.com/awslabs/soci-snapshotter/issues/1789.

Why can't we continue the conversion in case of "not found" and overwrite the existing ztoc?

@Kern-- @sondavidb @Swapnanil-Gupta

sondavidb: I think this leans on the side of caution where we may not want to override a zTOC that we think might exist but can't find it.

That said, the content store is more reliable than our artifacts DB. Perhaps we can continue with making the zTOC as you suggested?

Swapnanil-Gupta: We saw this issue in our tests. I think we were very intentional about placing that error and in requiring a `soci rebuild-db` because SOCI's artifacts DB might be out of sync with containerd's content store. We modified the failing tests to include that extra step.

I can create a follow-up PR if it wasn't the right choice.

Issue #1727, opened 2025-10-08
[Bug] `soci ztoc ls` lists ztocs from other images if they share layers


Swapnanil-Gupta:
### Description

When two images share layers and soci indexes have been created for both with different span sizes, soci's artifact store will contain multiple ztoc entries for the same layer digest (shared by the images). And because `soci ztoc ls` walks the artifact store looking for entries that match a particular layer digest, ztocs from other images will be listed in the output.

### Steps to reproduce the bug

- Pull 2 images that share layers
- Create soci indexes for both with different span sizes
- list the ztocs for one image and examine the output

As seen below ztocs from images (corresponding to the same layer digest `f56be85fc22e46face30e2c3de3f7fe7c15f8fd7c4e5add29d7f64b87abdaa09`) are listed in the output. 

```
➜  soci-snapshotter git:(feat-skip-existing-ztoc-gen) sudo ./out/soci create docker.io/library/nginx:1.23-alpine3.17 --min-layer-size=0 --span-size=10000
layer sha256:59b9d2200e632e457f800814693b3a01adf09a244c38ebe8d3beef5c476c4c55 -> ztoc sha256:af707474537b7585d94f41768d2c875b9750ed61214c6cd9a1b9744aecb785d4
layer sha256:1f21f983520d9a440d410ea62eb0bda61a2b50dd79878071181b56b82efa9ef3 -> ztoc sha256:daf3383edbc91b08ad03566695e11013ba622032280827afc396b96a0020f52a
layer sha256:3e1e579c95fece6bbe0cb9c8c2949512a3f8caaf9dbe6219dc6495abb9902040 -> ztoc sha256:d00b2d06a45af1ae3fb913e478fed5b60698a9cf4ec4991021eab13a779361f6
layer sha256:547a97583f72a32903ca1357d48fa302e91e8f83ffa18e0c40fd87adb5c06025 -> ztoc sha256:3859fc16dfeaa1b8a5c576dccc1a75beed08384ddb72b15d29d07da5abaafd34
layer sha256:2ce963c369bc5690378d31c51dc575c7035f6adfcc1e286051b5a5d9a7b0cc5c -> ztoc sha256:21bcb6b9a97df169315b76e3552811f754a11483bf70d774799da662e35b42ce
[...]

PR #1728, opened 2025-10-09
fix!: ztocs from other images showing up in `ztoc ls` when these images share layers

Swapnanil-Gupta:
**Issue #, if available:**
Fixes https://github.com/awslabs/soci-snapshotter/issues/1727

**Description of changes:**
- Ztoc listing will now happen via the soci index
- Removed the `verbose` flag as it is not being used (which makes it a breaking fix)
- Added a test to ensure that ztocs from other images do not show up in `ztoc ls --image-ref`

**Testing performed:**

### DIFF
--- a/cmd/soci/commands/ztoc/list.go
+++ b/cmd/soci/commands/ztoc/list.go
@@ -19,13 +19,13 @@ package ztoc
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"text/tabwriter"
 
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/internal"
 	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/containerd/containerd/images"
-	"github.com/containerd/platforms"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/urfave/cli/v3"
 )
@@ -44,11 +44,6 @@ var listCommand = &cli.Command{
 			Usage: "filter ztocs to those that are associated with a specific image",
 		},
 		&cli.BoolFlag{
-			Name:    "verbose",
-			Aliases: []string{"v"},
-			Usage:   "display extra debugging messages",
-		},
-		&cli.BoolFlag{
 			Name:    "quiet",
 			Aliases: []string{"q"},
 			Usage:   "only display the index digests",
@@ -62,7 +57,6 @@ var listCommand = &cli.Command{
 
 		ztocDgst := cmd.String("ztoc-digest")
 		imgRef := cmd.String("image-ref")
-		verbose := cmd.Bool("verbose")
 		quiet := cmd.Bool("quiet")
 
 		var artifacts []*soci.ArtifactEntry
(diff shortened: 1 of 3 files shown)


cmd/soci/commands/ztoc/list.go
```diff
@@ -82,52 +76,58 @@ var listCommand = &cli.Command{
...
+			indexInfos, _, err := soci.GetIndexDescriptorCollection(ctx, cs, db, img, supportedPlatforms)
+			if err != nil {
+				return fmt.Errorf("failed to get soci indexes for the image: %w", err)
+			}
+
+			var ztocDescs []ocispec.Descriptor
+			for _, indexInfo := range indexInfos {
+				readerAt, err := cs.ReaderAt(ctx, indexInfo.Descriptor)
+				if err != nil {
+					return fmt.Errorf("failed to read a soci index from the content store: %w", err)
```
sondavidb: Maybe specify the hash of the index?

cmd/soci/commands/ztoc/list.go
```diff
@@ -82,52 +76,58 @@ var listCommand = &cli.Command{
...
+				}
+				var sociIndex soci.Index
+				err = soci.DecodeIndex(io.NewSectionReader(readerAt, 0, indexInfo.Descriptor.Size), &sociIndex)
+				if err != nil {
+					return fmt.Errorf("failed to decode a soci index: %w", err)
```
sondavidb: Same here

cmd/soci/commands/ztoc/list.go
```diff
@@ -82,52 +76,58 @@ var listCommand = &cli.Command{
...
-								artifacts = append(artifacts, ae)
-							}
-						}
-					}
+			// at this point we already have the ztoc digests for the image
+			// but we have to query to artifacts db to get the associated layer digests
+			for _, ztocDesc := range ztocDescs {
+				entry, err := db.GetArtifactEntry(string(ztocDesc.Digest))
+				if err != nil {
+					return fmt.Errorf("failed to get ztoc from artifacts store: %w", err)
```
sondavidb: Could this happen if we aren't synced w/ the content store? Should we suggest rebuild-db here?

cmd/soci/commands/ztoc/list.go
```diff
@@ -82,52 +76,58 @@ var listCommand = &cli.Command{
...
 				}
-				return nil
-			})
+				if ztocDgst == "" || entry.Digest == ztocDgst {
```
sondavidb: nit: Maybe just me but the ordering of `ztocDgst` being inconsistent messed w/ my understanding of this, I think below ordering is a little clearer

```suggestion
				if ztocDgst == "" || ztocDgst == entry.Digest {
```

integration/ztoc_test.go
```diff
@@ -87,10 +89,22 @@ func TestSociZtocList(t *testing.T) {
 				if blob.MediaType != soci.SociLayerMediaType {
 					continue
 				}
-
-				ztocExistChecker(t, outputLines, img, blob)
+				// if two images share the same ztoc, they will share the same entry in the artifacts db, leading to the ztoc
+				// getting printed only once in the "ztoc ls" output.
```
sondavidb: I think this edge case should be handled in `verifyZtocListing` TBH.

integration/ztoc_test.go
```diff
@@ -153,9 +168,57 @@ func TestSociZtocList(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		t.Run("nginx alpine 1", func(t *testing.T) {
+			output := strings.Trim(string(sh.O("soci", "ztoc", "list", "--image-ref", nginxAlpineImg.ref)), "\n")
+			outputLines := strings.Split(output, "\n")
+			ztocOutput := outputLines[1:]
+
+			var ztcoBlobs []*v1.Descriptor
```
sondavidb:
```suggestion
			var ztocBlobs []*v1.Descriptor
```

integration/ztoc_test.go
```diff
@@ -153,9 +168,57 @@ func TestSociZtocList(t *testing.T) {
...
+		nginxAlpineSociIndex, err := sociIndexFromDigest(sh, nginxAlpineIndexDigest)
+		if err != nil {
+			t.Fatal(err)
+		}
+		nginxAlpine2SociIndex, err := sociIndexFromDigest(sh, nginxAlpine2IndexDigest)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		t.Run("nginx alpine 1", func(t *testing.T) {
```
sondavidb: Any reason these tests need to run separately? Seems to make more sense to run these two in a loop for code reuse and lump them with the same test run.

cmd/soci/commands/ztoc/list.go
Swapnanil-Gupta: There is a possibility. Added the suggestion to the error message.

cmd/soci/commands/ztoc/list.go
```diff
@@ -82,52 +76,58 @@ var listCommand = &cli.Command{
...
+				for _, blob := range sociIndex.Blobs {
+					if blob.MediaType != soci.SociLayerMediaType {
+						continue
+					}
+					ztocDescs = append(ztocDescs, blob)
 				}
 			}
-			if len(layers) == 0 {
-				return fmt.Errorf("no image layers. could not filter ztoc")
+			if len(ztocDescs) == 0 {
```
Kern--: With the old behavior, no indexes/ztocs for an image was not considered an error, but now it is. Is that what we want?

cmd/soci/commands/ztoc/list.go
```diff
@@ -82,52 +76,58 @@ var listCommand = &cli.Command{
...
+			indexInfos, _, err := soci.GetIndexDescriptorCollection(ctx, cs, db, img, supportedPlatforms)
+			if err != nil {
+				return fmt.Errorf("failed to get soci indexes for the image: %w", err)
+			}
+
+			var ztocDescs []ocispec.Descriptor
+			for _, indexInfo := range indexInfos {
+				readerAt, err := cs.ReaderAt(ctx, indexInfo.Descriptor)
+				if err != nil {
+					return fmt.Errorf("failed to read a soci index with digest %s from the content store: %w", indexInfo.Descriptor.Digest, err)
```
Kern--: In the other PR, we added a note to try `rebuild-db` to fix issues like this, should we add that note here too?

Swapnanil-Gupta: ~~I think we were suggesting that when we were sure that an artifact exists in the content store but not in soci's artifact store or vice versa. I don't think at this point we know whether this artifact exists in soci's artifact store or not?~~

Nvm. `soci.GetIndexDescriptorCollection` gets the artifacts from the artifacts store, so it makes sense to suggest `rebuild-db` here.

cmd/soci/commands/ztoc/list.go
Swapnanil-Gupta: Removed the check and error.

sondavidb (approved):
LGTM after squashing commits

PR #1729, opened 2025-10-10
Fix race condition in rebuild-db

sondavidb:
**Issue #, if available:**

**Description of changes:**
The containerd content store (and likely other content stores) implement Walk() by going through the content laid out on disk and running the provided function on each artifact found. However, as the content store has no lock on it while walking, there is a gap between finding content and fetching said content in which the content may disappear.

Fixing this is simple — do a no-op if the content is not found in the content store. This effectively mirrors the behavior of not finding said content in the store to begin with, which more accurately reflects the state of the disk.

This change also simplifies addNewArtifacts by using the native Walk() function, instead of walking through the disk ourselves. (containerd's content store does the same thing under the hood anyway.)

**Testing performed:**

### DIFF
--- a/soci/artifacts.go
+++ b/soci/artifacts.go
@@ -253,7 +253,12 @@ func (db *ArtifactsDb) addNewArtifacts(ctx context.Context, blobStorePath string
 			return nil
 		}
 		f, err := os.Open(path)
-		if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			// This check attempts to mitigate a race between finding content and getting said content.
+			// If content is provided there's no guarantee it will still be there by the time we want to fetch it.
+			// Returning should be safe as it's effectively a no-op.
+			return nil
+		} else if err != nil {
 			return err
 		}
 		defer f.Close()


soci/artifacts.go
```diff
@@ -231,36 +230,30 @@ func (db *ArtifactsDb) RemoveOldArtifacts(ctx context.Context, blobStore store.S
...
 		}
-		info, err := d.Info()
-		if err != nil {
-			return err
+		desc := ocispec.Descriptor{
+			Digest: info.Digest,
+			Size:   info.Size,
 		}
-		// skip: entry is an empty config
-		if info.Size() < 10 {
```
Swapnanil-Gupta, line 252: Are we sure about removing this size check?

sondavidb: It's still there, diff just doens't make it super obvious: https://github.com/awslabs/soci-snapshotter/pull/1729/files#diff-4df543d29c1b805f5ff3616859d647084d559e3268bfb2aea03f21a3719fc0c0R235

sondavidb:
Oops...I forgot that content.Store implements the containerd content store, NOT our blob store. The reason we implemented this via a file walk was because of this, and in particular because ORAS doesn't have a content store Walk() func. Reverted the cs.Walk changes and just kept the ErrNotExist error.

Kern-- (approved):

Swapnanil-Gupta (approved):

PR #1741, opened 2025-10-30
Upgrade to nerdctl v2.1.6

sondavidb:
**Issue #, if available:**

**Description of changes:**
Upgraded nerdctl to v2.1.6 from v1.7.7.

This has mostly not affected our testing suite but using default namespaces now outputs things to stderr, which makes our TestNetworkRetry test fail. I put in a workaround but I'm not very happy with it. Ideally we figure out how to suppress it in nerdctl but I dug quite a bit and couldn't find out how to. Suggestions are appreciated but for now I just dispose the first failure channel message.

This upgrade also adds the `--userns-remap` flag which allows us to remove our Makefile's `nerdctl-with-idmapping` target + its associated patch.

Finally, I noticed some of the idmapping checks being weird. I added some formatting outputs to the `stat` command to make it easier to work with. Made it a separate commit for clarity.

**Testing performed:**
`make integration`


integration/run_test.go
```diff
@@ -351,21 +351,33 @@ func TestNetworkRetry(t *testing.T) {
...
-					t.Fatalf("expected Read request to succeed; got data in stderr: %s", data)
-				} else if len(successChannel) > 0 {
-					t.Fatalf("expected Read request to fail; got data in (stdout, stderr) : (%s, %s)", data, <-successChannel)
-				}
-			case data := <-successChannel:
-				if !tt.config.expectedSuccess {
-					t.Fatalf("expected Read request to fail; got data in stdout: %s", data)
-				} else if len(failureChannel) > 0 {
-					t.Fatalf("expected Read request to succeed; got data in (stdout, stderr) : (%s, %s)", data, <-failureChannel)
+			// TODO: Eliminate this check as it makes an assumption about the first warning.
```
Shubhranshu153: can we simplify it into 2 parts. (readability)
1. Get the first message make sure that its the warning message expected
2. Continue the test without changing the earlier logic

sondavidb: Sure let me see what I can do. I did just realize this logic is faulty regardless (the failure tests should not be passing in this scenario).

sondavidb: Fixed with your suggestions and amended the test. should pass properly now.

Shubhranshu153 (approved):

Kern-- (approved):

PR #1745, opened 2025-11-05
fix fscache not cleanup

wswsmao:
**Issue #, if available:**
1. When users run `nerdctl images rmi` to remove images, fscache is not cleaned up
2. When the snapshotter exits, fscache is not cleared
3. After restarting the snapshotter, a new batch of fscache directories is created with duplicate content
4. The snapshotter does not automatically restore fscache upon restart

**Description of changes:**
This patch implements explicit resource lifecycle management with immediate eviction support to fix resource leaks during unmount operations.

The core change enhances the LRU cache to support both deferred and immediate eviction modes. The `done` callback now accepts a boolean parameter: `false` for lazy cleanup and `true` for immediate removal. When immediate eviction is requested, cache entries are finalized and removed synchronously instead of waiting for LRU eviction.

A new `Close()` method is introduced alongside the existing `Done()` to distinguish between normal resource release and unmount scenarios. During filesystem unmount, layers call `Close()` to trigger immediate resource eviction, preventing the resource leaks observed in the original issue. Error paths are also updated to use immediate eviction for invalid or corrupted resources, ensuring they don't occupy cache space.

Additionally, the patch adds explicit cleanup of span manager resources in the reader's `Close()` method and improves snapshot restoration robustness by pre-creating required directories before mount operations. Overall, this provides deterministic cleanup timing with clear "release when convenient" vs. "release immediately" semantics, which is critical for proper resource management during unmounts and error handling. and error handling.

**Testing performed:**
1. Run an image
```bash
nerdctl run --snapshotter soci --rm -it env12.com/ocs9:soci echo "hello"
```

Observed:
```bash
Every 1.0s: du -h --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/spancache                   VM-33-248-tlinux: Wed Nov  5 09:35:08 2025

25M     /var/lib/soci-snapshotter-grpc/soci/spancache/2068117739
25M     /var/lib/soci-snapshotter-grpc/soci/spancache
```

2. Remove the image
```bash
nerdctl images -q | xargs nerdctl rmi -f
Untagged: env12.com/ocs9:soci@sha256:979fe973858c87e1fb2152a9b17563f6fc59abc5c2f85293d15a7c827bbb6a02
Deleted: sha256:3a6ef9931f24a0ccd89e429366e0341d81b58d97bf7115612494f4249dd8d351
```

Observed:
```bash
Every 1.0s: du -h --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/spancache                   VM-33-248-tlinux: Wed Nov  5 09:35:27 2025

4.0K    /var/lib/soci-snapshotter-grpc/soci/spancache
```

3. Re-run the image
```bash
nerdctl run --snapshotter soci --rm -it env12.com/ocs9:soci echo "hello"
env12.com/ocs9:soci:                                                              resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:979fe973858c87e1fb2152a9b17563f6fc59abc5c2f85293d15a7c827bbb6a02:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:fb1079429352f239c48ea35304477b270a42932d000b0707e4d96990f68afb2f: done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:ba7df7c2d4c09ddea132cac1bcc9d77c461db75fb49d681837695bbec4df6142:   done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 0.6 s                                                                    total:  3.2 Ki (5.3 KiB/s)                                       
hello
```

Observed:
```bash
Every 1.0s: du -h --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/spancache                   VM-33-248-tlinux: Wed Nov  5 09:35:59 2025

71M     /var/lib/soci-snapshotter-grpc/soci/spancache/3428176031
71M     /var/lib/soci-snapshotter-grpc/soci/spancache
```

4. Kill the snapshotter process (Ctrl+C)

Observed:
```bash
Every 1.0s: du -h --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/spancache                   VM-33-248-tlinux: Wed Nov  5 09:36:17 2025

4.0K    /var/lib/soci-snapshotter-grpc/soci/spancache
```

5. Restart the snapshotter process and run the container again

Observed:
```bash
Every 1.0s: du -h --max-depth=1 /var/lib/soci-snapshotter-grpc/soci/spancache                   VM-33-248-tlinux: Wed Nov  5 09:36:31 2025

25M     /var/lib/soci-snapshotter-grpc/soci/spancache/1469904631
25M     /var/lib/soci-snapshotter-grpc/soci/spancache
```

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -1228,8 +1228,13 @@ func (fs *filesystem) Unmount(ctx context.Context, mountpoint string) error {
 	delete(fs.layer, mountpoint)
 	// If the mountpoint is an id-mapped layer, it is pointing to the
 	// underlying layer, so we cannot call done on it.
+	// We do a evict call to call the registered evict functions.
+	// This cleans up layer resources and removes it from the resolver cache,
+	// preparing the layer for removal and freeing associated memory done on
+	// call of the finalizer in span manager
 	if !isIDMappedDir(mountpoint) {
 		l.Done()
+		fs.resolver.Evict(l.GetCacheRefKey())
 	}
 	fs.layerMu.Unlock()
 	fs.metricsController.Remove(mountpoint)
--- a/fs/fs_test.go
+++ b/fs/fs_test.go
@@ -93,6 +93,7 @@ func (l *breakableLayer) SkipVerify()                          {}
 func (l *breakableLayer) ReadAt([]byte, int64, ...remote.Option) (int, error) {
 	return 0, fmt.Errorf("fail")
 }
+func (l *breakableLayer) GetCacheRefKey() string { return "" }
 func (l *breakableLayer) BackgroundFetch() error { return fmt.Errorf("fail") }
 func (l *breakableLayer) Check() error {
 	if !l.success {
--- a/fs/layer/layer.go
+++ b/fs/layer/layer.go
@@ -104,6 +104,9 @@ type Layer interface {
 	// Done releases the reference to this layer. The resources related to this layer will be
 	// discarded sooner or later. Queries after calling this function won't be serviced.
 	Done()
+
+	// GetCacheRefKey returns the reference key for the cache used by the layer
+	GetCacheRefKey() string
 }
 
 // Info is the current status of a layer.
@@ -226,6 +229,12 @@ func newCache(root string, cacheType string, cfg config.FSConfig) (cache.BlobCac
 	)
 }
 
+func (r *Resolver) Evict(name string) {
+	r.layerCacheMu.Lock()
+	r.layerCache.Remove(name)
+	r.layerCacheMu.Unlock()
+}
+
 // Resolve resolves a layer based on the passed layer blob information.
 func (r *Resolver) Resolve(ctx context.Context, hosts []docker.RegistryHost, refspec reference.Spec, desc, sociDesc ocispec.Descriptor, opCounter *FuseOperationCounter, disableVerification bool, metadataOpts ...metadata.Option) (_ Layer, retErr error) {
 	name := refspec.String() + "/" + desc.Digest.String()
@@ -339,7 +348,7 @@ func (r *Resolver) Resolve(ctx context.Context, hosts []docker.RegistryHost, ref
 	}
 	disableXAttrs := getDisableXAttrAnnotation(sociDesc)
 	// Combine layer information together and cache it.
-	l := newLayer(r, desc, blobR, vr, bgLayerResolver, opCounter, disableXAttrs)
+	l := newLayer(r, desc, name, blobR, vr, bgLayerResolver, opCounter, disableXAttrs)
 	r.layerCacheMu.Lock()
 	cachedL, done2, added := r.layerCache.Add(name, l)
 	r.layerCacheMu.Unlock()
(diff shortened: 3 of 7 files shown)


util/lrucache/lrucache.go
```diff
@@ -110,12 +110,18 @@ func (c *Cache) Remove(key string) {
...
 
-func (c *Cache) decreaseOnceFunc(rc *refCounter) func() {
+func (c *Cache) decreaseOnceFunc(rc *refCounter) func(bool) {
 	var once sync.Once
-	return func() {
+	return func(evict bool) {
 		c.mu.Lock()
 		defer c.mu.Unlock()
 		once.Do(func() { rc.dec() })
+		if evict {
```
sondavidb: Seems that the decrement is locked behind a sync.Once but the evict isn't. So this means that `done(true)` can be called at anytime after `done(false)` and on that call it will do the eviction. If this is intentional I wonder if a separate cleanup function is the right call here. Having a `Close()` call that's functionally the same as `Done()` is a little strange, in particular because calling `Done()` signifies that we are done with the layer, so having a separate call that can also say the layer is done being used seems to be a strange pattern.

wswsmao: The main benefit of the current approach is that it provides fine-grained control over resource lifecycle while maintaining backward compatibility. The `done(false)` path handles the normal case where we're just releasing a reference, while `done(true)` addresses a specific problem we encountered: when unmounting a layer, we need to immediately evict it from the cache to prevent resource leaks, especially for resources like file descriptors and C-allocated memory in the compression libraries. The dual behavior in a single callback keeps the refcounting logic centralized and ensures that the eviction path properly decrements references before removing from cache, which is critical for correctness.

That said, I understand your concern about the API semantics. The `sync.Once` only protecting the decrement but not the eviction logic does create a potential confusion point where someone might call `done(false)` first and then `done(true)` later, expecting both behaviors to be idempotent together. While this works correctly from a safety perspective due to the global mutex and the finalize-once pattern, it's not immediately obvious from the API signature.

I'm open to refactoring toward your suggested approach of separating `release()` and `evict()` callbacks. However, this would require touching quite a few callsites across the codebase - not just in `cache.go` and `layer.go`, but also in all the places that currently use these callbacks. Before I go down that path, I wanted to check if you'd be comfortable with this level of change, or if you'd prefer a middle-ground approach where we keep the `done(bool)` signature but move the eviction logic inside the `once.Do` block. That way, the first call wins regardless of the bool value, making the behavior more predictable even if the semantics remain somewhat mixed. 

Which direction you'd prefer.

cache/cache.go
```diff
@@ -214,7 +214,7 @@ func (dc *directoryCache) Get(key string, opts ...Option) (Reader, error) {
 			return &reader{
```
Shubhranshu153, line 214: Thank You for your contribution.

The logic works, just i feel it is complicating the logic a bit. Evicting the top level reference would evict all child references as a result we should not be required to handle it individually. Can you try to merge the PR (https://github.com/awslabs/soci-snapshotter/pull/1710/files)  i had with the restart logic here (i will also will merge it and check). We can merge the logic and put it in this PR.

I believe the changes in restoreRemoteSnapshot  is for cases when the snapshotter restarts?

wswsmao: > I believe the changes in restoreRemoteSnapshot is for cases when the snapshotter restarts?

Yes, As I described at the beginning of the pr, the snapshotter does not automatically restore fscache upon restart

wswsmao: > The logic works, just i feel it is complicating the logic a bit. Evicting the top level reference would evict all child references as a result we should not be required to handle it individually. Can you try to merge the PR (https://github.com/awslabs/soci-snapshotter/pull/1710/files) i had with the restart logic here (i will also will merge it and check). We can merge the logic and put it in this PR.

merge done

Kept the explicit spanManager.Close() call to make spancache cleanup immediate and observable.

Shubhranshu153: Thank you for working on it.
Apologies for the delay i was out for some time.

sondavidb:
Thanks!

PR #1752, opened 2025-11-06
Remove patch version requirement in go.mod

sondavidb:
**Issue #, if available:**

**Description of changes:**
Noticed our benchmarking script failed as we didn't have the right Go version. I think we probably don't need to mandate the absolute latest of Go so I just put the version to 1.24.0 and updated our Benchmarking script.

**Testing performed:**

### DIFF
--- a/.github/workflows/benchmark-visualization.yml
+++ b/.github/workflows/benchmark-visualization.yml
@@ -19,7 +19,7 @@ permissions:
   deployments: write
 
 env:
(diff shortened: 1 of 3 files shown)


Kern-- (approved):
It's usually one of the x packages which forces us to upgrade, but it seems they only depend on the initial minor version, not specific patch versions.

pendo324 (approved):

PR #1778, opened 2025-11-18
Make LogMonitor threadsafe

sondavidb:
**Issue #, if available:**

**Description of changes:**
Our LogMonitor is not threadsafe, which occasionally leads to test failures such at [this](https://github.com/awslabs/soci-snapshotter/actions/runs/19478810665/job/55745386314?pr=1775#step:5:447) one. Small change to help avoid this issue.

**Testing performed:**

### DIFF
--- a/integration/log_test.go
+++ b/integration/log_test.go
@@ -22,8 +22,6 @@ import (
 	"time"
 )
 
-const sensitiveInfoMonitorKey = "sensitive-info-monitor"
-
 // TestSnapshotterDoesNotLogSensitiveInformation verifies that the snapshotter
 // doesn't log sensitive information during various operations.
 func TestSnapshotterDoesNotLogSensitiveInformation(t *testing.T) {
@@ -59,10 +57,9 @@ func TestSnapshotterDoesNotLogSensitiveInformation(t *testing.T) {
 				done()
 			})
 
-			logMonitor := rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, tc.opts...))
-			logMonitor.Add(sensitiveInfoMonitorKey, sensitiveInfoMonitor(t, sensitivePatterns...))
+			logMonitor := rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, tc.opts...), sensitiveInfoMonitor(t, sensitivePatterns...))
 			t.Cleanup(func() {
-				logMonitor.Remove(sensitiveInfoMonitorKey)
+				logMonitor.Cleanup(t)
 			})
 
 			copyImage(sh, dockerhub(image), regConfig.mirror(image))
--- a/integration/util_test.go
+++ b/integration/util_test.go
@@ -852,7 +852,7 @@ func getReferrers(sh *shell.Shell, regConfig registryConfig, imgName, digest str
 	return &index, nil
 }
 
-func rebootContainerd(t *testing.T, sh *shell.Shell, customContainerdConfig, customSnapshotterConfig string) *testutil.LogMonitor {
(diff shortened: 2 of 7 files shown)


sondavidb:
Had issues with a hanging test for TestNetworkRetry, turns out running arbitrary functions that can block while holding a mutex isn't great...

Ended up reworking TestNetworkRetry further so that we only call the channels as many times as we expect to read it.

Kern--:
Could we make this simpler by making `LogMonitor` have an explicit `Start` and making it an error to add monitor funcs after starting?

I think the current implementation still has a race where the monitor can process log lines before monitor functions are added.

sondavidb:
Ugh, so I did a lot more than I wanted to...

Trying to make LogMonitor have a dedicated Start func required a refactor of rebootContainerd. I ended up having to refactor our other LogMonitor helper functions to accommodate for this, as well as our checkSnapshotterStartup function, but then we also had to ensure we could clean up all the functions that were added but there was no reason to have a key so it ended up being an array, but then there was no way to remove specific LogMonitors so I made Remove a Cleanup function instead, and in trying to make it use all the resources I found that it doesn't fully clean up everything until pipes are closed, so I had to workaround that by having a dangling goroutine in the hopes that the reader pipe would eventually close...

TL;DR this ended up being a lot more work than anticipated, BUT maybe this is ultimately a bit of a nicer interface. @Kern-- WDYT?

integration/log_test.go
```diff
@@ -59,10 +57,9 @@ func TestSnapshotterDoesNotLogSensitiveInformation(t *testing.T) {
 				done()
 			})
 
-			logMonitor := rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, tc.opts...))
-			logMonitor.Add(sensitiveInfoMonitorKey, sensitiveInfoMonitor(t, sensitivePatterns...))
+			logMonitor := rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, tc.opts...), sensitiveInfoMonitor(t, sensitivePatterns...))
```
Shubhranshu153, line 60: reboot containerd spitting out a logMonitor is unintuitive can we separate the calls.
my understanding is reboot Containerd is a setup function and log monitor is a init function

integration/util_test.go
```diff
@@ -852,7 +852,7 @@ func getReferrers(sh *shell.Shell, regConfig registryConfig, imgName, digest str
 	return &index, nil
 }
 
-func rebootContainerd(t *testing.T, sh *shell.Shell, customContainerdConfig, customSnapshotterConfig string) *testutil.LogMonitor {
+func rebootContainerd(t *testing.T, sh *shell.Shell, customContainerdConfig, customSnapshotterConfig string, monitorFuncs ...func(string)) *testutil.LogMonitor {
```
Shubhranshu153, line 855: this func is doing too many things which can make it difficult to maintain and does nt really follow single responsibility principle.

integration/log_test.go
sondavidb: `rebootContainerd` has always returned a logMonitor function. I think we could made the case for it before this change but with this change it requires that logMonitors are registered before starting. This is to ensure we do not miss logs. Not to mention we use a log monitor to check for snapshotter startup before returning from the func at all.

One option to change this is to split this function into one that just restarts the appropriate snapshotter, and have a separate function call if we care about logMonitors.

integration/util_test.go
sondavidb: Yeah, though I don't think this PR exacerbates the problem. The point of this PR is just to make the logMonitors deterministic as right now we run into races every so often.

I do think we can do a refactor if we think it's useful to do so. Can do so here or as a followup.

PR #1781, opened 2025-11-21
Make sure draining the resp body for range-gets

henry118:
**Issue #, if available:**

**Description of changes:**

**Testing performed:**

### DIFF
--- a/fs/parallel_artifact_fetcher.go
+++ b/fs/parallel_artifact_fetcher.go
@@ -301,6 +301,7 @@ func writeToFileRange(file *os.File, rsc io.ReadCloser, lower, upper int64) erro
 	// Reading any more or less will result in an incorrect file being created,
 	// so use io.CopyN to guarantee we read exactly lower - upper + 1 bytes.
 	_, err := io.CopyN(w, rsc, readSize)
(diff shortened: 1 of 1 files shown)


fs/parallel_artifact_fetcher.go
```diff
@@ -301,6 +301,7 @@ func writeToFileRange(file *os.File, rsc io.ReadCloser, lower, upper int64) erro
 	// Reading any more or less will result in an incorrect file being created,
 	// so use io.CopyN to guarantee we read exactly lower - upper + 1 bytes.
 	_, err := io.CopyN(w, rsc, readSize)
+	io.Copy(io.Discard, rsc) // Drain remaining data
```
Kern--, line 304: Why would there be more data in the response? Doesn't this get called if we successfully made a ranged get?

Kern-- (approved):

fs/parallel_artifact_fetcher.go
henry118: This is just a safe-guard to ensure the body is drained.

sondavidb (approved):

PR #1792, opened 2025-11-28
Support specifying content directory via root option

wswsmao:
**Issue #, if available:**

**Description of changes:**

Currently, the `root` option only controls the snapshotter root directory, while the content directory remains hardcoded to the default path (`/var/lib/soci-snapshotter-grpc/content`). 

This inconsistency forces users to manage two separate locations, which becomes problematic when `/var` has insufficient space - the hardcoded content path may cause the snapshotter to fail.

This PR unifies both directories under the `root` option, allowing users to control all data locations through a single configuration.

**Testing performed:**

**Before this PR:**
When `root` is set to `/mock/soci-snapshotter`:
```
Every 1.0s: du -h --max-depth=1 /mock/soci-snapshotter                                          VM: Fri Nov 28 07:58:20 2025

16K     /mock/soci-snapshotter/snapshotter
0       /mock/soci-snapshotter/soci
0       /mock/soci-snapshotter/unpack
32K     /mock/soci-snapshotter


Every 1.0s: du -h --max-depth=1 /var/lib/soci-snapshotter-grpc                                  VM: Fri Nov 28 07:58:12 2025

16K     /var/lib/soci-snapshotter-grpc/content
20K     /var/lib/soci-snapshotter-grpc

```

**After this PR:**
When `root` is set to `/mock/soci-snapshotter`:
```
Every 1.0s: du -h --max-depth=1 /mock/soci-snapshotter                                          VM: Fri Nov 28 08:04:37 2025

16K     /mock/soci-snapshotter/snapshotter
8.0K    /mock/soci-snapshotter/content
0       /mock/soci-snapshotter/soci
0       /mock/soci-snapshotter/unpack
40K     /mock/soci-snapshotter
```

### DIFF
--- a/soci/store/store_test.go
+++ b/soci/store/store_test.go
@@ -18,6 +18,7 @@ package store
 
 import (
 	"context"
+	"path/filepath"
 	"testing"
 
 	"github.com/awslabs/soci-snapshotter/config"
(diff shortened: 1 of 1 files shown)


sondavidb:
Changes LGTM, thanks for this! Can we add a unit or integration test for this?

wswsmao:
> Changes LGTM, thanks for this! Can we add a unit or integration test for this?

ok, I have add unit test in commit [460c317](https://github.com/awslabs/soci-snapshotter/pull/1792/commits/460c3173801f674539fb6061a93529269a7c0eaa)

sondavidb:
So the changes on paper LGTM. I think this should have been the behavior from the start. But I wonder if there are any concerns for breaking existing workflows...we should at least call this out in our next SOCI release IMO.

cmd/soci/commands/internal/store.go
```diff
@@ -27,8 +28,16 @@ import (
 func ContentStoreOptions(ctx context.Context, cmd *cli.Command) []store.Option {
 	contentStore := cmd.String("content-store")
 	address := cmd.String("address")
+	root := cmd.String("root")
+	var snapshotterRoot string
+	if root != "" {
```
Kern--: The root flag already sets the default snapshotter directory value. Can't we use the flag directly like in rebuild-db below?

wswsmao: done

Kern-- (approved):

sondavidb:
Thanks!

Issue #1800, opened 2025-12-03
[Bug] `nerdctl push` failing post `soci convert`. convert is possibly deleting layers


chiragjn:
### Description

It seems like this not reproducible with every image but it does happen with a few images consistently. It is also not clear if some bug in nerdctl or containerd is causing this

soci version  0.12.0
containerd version 2.2.0
nerdctl version 2.2.0

### Steps to reproduce the bug

First start with a clean slate

```
sudo nerdctl system prune --all --volumes --force
sudo ctr content rm $(sudo ctr content ls -q)
sudo soci rebuild-db
sudo nerdctl images
sudo ctr images ls
sudo ctr content ls
sudo ctr snapshots ls
sudo soci index ls
```

Pull in a public image

```
sudo nerdctl pull --platform linux/amd64 public.ecr.aws/truefoundrycloud/jupyter:0.4.4-cu124-py3.11.10-sudo
```

Inspect
```
sudo ctr images inspect public.ecr.aws/truefoundrycloud/jupyter:0.4.4-cu124-py3.11.10-sudo
```

```
public.ecr.aws/truefoundrycloud/jupyter:0.4.4-cu124-py3.11.10-sudo
│    Created: 2025-12-03 02:13:03.59592606 +0000 UTC
│    Updated: 2025-12-03 02:13:03.59592606 +0000 UTC
└── application/vnd.docker.distribution.manifest.v2+json @sha256:93e4bb685b627e0e46a58db974ceb51ea25ba1f8cefed2e6d57e4bedb3bd1514 (6208 bytes)
    ├── application/vnd.docker.container.image.v1+json @sha256:293014260834bb3c19680d4162e5d8a166830115fde8e1480736e7845c838b18 (36444 bytes)
    ├── application/vnd.docker.image.rootfs.diff.tar.gzip @sha256:3c645031de2917ade93ec54b118d5d3e45de72ef580b8f419a8cdc41e01d042c (29533419 bytes)
    ├── application/vnd.docker.image.rootfs.diff.tar.gzip @sha256:0d6448aff88945ea46a37cfe4330bdb0ada228268b80da6258a0fec63086f404 (4623694 bytes)
    ├── application/vnd.docker.image.rootfs.diff.tar.gzip @sha256:0a7674e3e8fe69dcd7f1424fa29aa033b32c42269aab46cbe9818f8dd7154754 (57593640 bytes)
[...]

PR #1804, opened 2025-12-04
add garbage collection labels to image mainfest

coderbirju:
**Issue #, if available:**
#1800 

**Description of changes:**
- adds missing labels to image manifest after `soci image convert`
- adds test to verify that all the labels exist on the converted image ~and the digest associated with the config labels points to the image. (Verified by doing a content get on this extracted digest)~

We now follow the following steps to ensure all the digests are referrenced in the labels
- make a map of digest(key) -> label(list)*
- iterate the layer digests and config digest from the manifest
- verify that there is atleast one label associated with the digests

*this ensures theres atleast one reference to each layer even if we end up adding multiple lables to layers in the future

**Testing performed:**
- soci convert now works in-place
- deleting image tags of original image does not corrupt converted images

```
sudo nerdctl pull jrottenberg/ffmpeg:latest
sudo soci convert jrottenberg/ffmpeg:latest jrottenberg/ffmpeg:latest
sudo ctr i inspect --content docker.io/jrottenberg/ffmpeg:latest
```
gives us a valid output 
```
docker.io/jrottenberg/ffmpeg:latest
│    Created: 2025-12-04 02:02:17.117510483 +0000 UTC
│    Updated: 2025-12-04 02:03:07.45875418 +0000 UTC
└── application/vnd.oci.image.index.v1+json @sha256:b2333f366fec96b22948ee73e77fa36dbdbdbed60da2f67ffe9d16fe5f7b8fac (803 bytes)
    ├── application/vnd.oci.image.manifest.v1+json @sha256:0b940c1d2b3a1069c42f179e09117dc2cd28fcbebe723a8c21d50d8399e584b0 (1187 bytes)
    │   │    Platform: linux/amd64
    │   ├── application/vnd.oci.image.config.v1+json @sha256:a8bf3d68fbdb7e4a79769f599d4172c46344a983e621a4eb340eaa32edd34ac7 (2977 bytes)
    │   ├── application/vnd.docker.image.rootfs.diff.tar.gzip @sha256:20043066d3d5c78b45520c5707319835ac7d1f3d7f0dded0138ea0897d6a3188 (29724688 bytes)
    │   ├── application/vnd.docker.image.rootfs.diff.tar.gzip @sha256:716e71fe910f5c1a05201f410f410e7e2c0276701158dde03d00568c05e0922a (1886898 bytes)
    │   ├── application/vnd.docker.image.rootfs.diff.tar.gzip @sha256:3017d0488c3c00fee2ef3c8c23b653fcd36787eb1ed0866a8b0c8127852587e3 (16156 bytes)
    │   ├── application/vnd.docker.image.rootfs.diff.tar.gzip @sha256:7cd4e390adb98b5b44c4c0ad7a43187784d7c55cd84972c5d2d0b0011f95f41f (21898 bytes)
    │   └── application/vnd.docker.image.rootfs.diff.tar.gzip @sha256:503ba9a45cd098fcbafd665e3fa1a61099502a896cae0cfa397265b0c396f517 (51234035 bytes)
    └── application/vnd.oci.image.manifest.v1+json @sha256:341bca01870c073893654c02bd7517c7c697e07898862ed16d2fbb6ad7ff920d (1114 bytes)
        │    Platform: linux/amd64
        ├── application/vnd.amazon.soci.index.v2+json @sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a (2 bytes)
        ├── application/octet-stream @sha256:463bedc6834a57f54f0f767a02177609f5721c2aca204fe0c615854e2a8f0fb1 (1216912 bytes)
        └── application/octet-stream @sha256:231dfcc2b681dd2df5d89b645fc09294edb524359fc632d874aa225ffe256682 (1124248 bytes)
```

previously it would error out  : `ctr: content digest sha256:a8bf3d68fbdb7e4a79769f599d4172c46344a983e621a4eb340eaa32edd34ac7: not found`

### DIFF
--- a/integration/convert_test.go
+++ b/integration/convert_test.go
@@ -101,6 +101,57 @@ func TestConvertWithForceRecreateZtocs(t *testing.T) {
 func validateConversion(t *testing.T, sh *shell.Shell, originalDigest, convertedDigest string) {
 	t.Helper()
 
+	// Helper function to get digest label mapping
+	getDigestLabelMapping := func(manifestDigest string) (map[string][]string, error) {
+		digestMap := make(map[string][]string)
+
+		manifestLabelOutput := sh.O("ctr", "content", "label", manifestDigest)
+		manifestLabels := strings.Split(strings.TrimSpace(string(manifestLabelOutput)), ",")
+
+		if len(manifestLabels) <= 0 {
+			return nil, fmt.Errorf("manifest does not contain any labels")
+		}
+
+		for _, label := range manifestLabels {
+			keyVal := strings.Split(label, "=")
+			if len(keyVal) != 2 {
+				continue
+			}
+			digestMap[keyVal[1]] = append(digestMap[keyVal[1]], keyVal[0])
+		}
+		return digestMap, nil
+	}
+
+	// Helper function to verify image layers contain required labels
+	verifyImageLayersContainsLabel := func(digestMap map[string][]string, manifest ocispec.Manifest) error {
+		// Helper function to check if a digest has the required label
+		checkDigestLabels := func(digest string) error {
+			if labels, ok := digestMap[digest]; ok {
+				for _, label := range labels {
+					if strings.Contains(label, "containerd.io/gc.ref.content") {
+						return nil // Found the required label
+					}
+				}
+				return fmt.Errorf("digest %s does not have associated gc label", digest)
+			}
+			return fmt.Errorf("digest %s not found in digestMap", digest)
+		}
+
+		// Check all layer digests
+		for _, layer := range manifest.Layers {
+			if err := checkDigestLabels(layer.Digest.String()); err != nil {
+				return err
+			}
+		}
+
+		// Check config digest
+		if err := checkDigestLabels(manifest.Config.Digest.String()); err != nil {
+			return err
+		}
+
+		return nil
+	}
+
 	if originalDigest == convertedDigest {
 		t.Fatalf("conversion did not change the digest: %s", originalDigest)
 	}
@@ -197,7 +248,21 @@ func validateConversion(t *testing.T, sh *shell.Shell, originalDigest, converted
 		if dg, ok := manifest.Annotations[soci.ImageAnnotationSociIndexDigest]; !ok || dg != sociIndexDesc.Digest.String() {
 			t.Errorf("manifest %v does not contain expected soci index digest %v", manifestDesc, sociIndexDesc.Digest)
 		}
+
+		// get labels associated with the manifest and extract the digest
+		digestLabelMap, err := getDigestLabelMapping(manifestDesc.Digest.String())
+		if err != nil {
+			t.Errorf("failed to get config digest from manifest: %v", err)
+			continue
+		}
+
+		err = verifyImageLayersContainsLabel(digestLabelMap, manifest)
+		if err != nil {
+			t.Errorf("error verifying layers: %v", err)
+			continue
+		}
 	}
+
 }
 
 func TestConvert(t *testing.T) {
--- a/soci/soci_convert.go
+++ b/soci/soci_convert.go
@@ -307,6 +307,18 @@ func (b *IndexBuilder) annotateImages(ctx context.Context, ociIndex *ocispec.Ind
 			}
 			indexWithMetadata.Desc.Annotations[IndexAnnotationImageManifestDigest] = manifestDesc.Digest.String()
 		}
+
+		err = store.LabelGCRefContent(ctx, b.blobStore, *manifestDesc, "config", manifest.Config.Digest.String())
+		if err != nil {
+			return err
+		}
+
+		for i, layer := range manifest.Layers {
+			err = store.LabelGCRefContent(ctx, b.blobStore, *manifestDesc, fmt.Sprintf("l.%d", i), layer.Digest.String())
+			if err != nil {
+				return err
+			}
+		}
 	}
 	return nil
 }


integration/convert_test.go
```diff
@@ -198,6 +198,28 @@ func validateConversion(t *testing.T, sh *shell.Shell, originalDigest, converted
...
+
+	if len(convertedLabelsOutput) == 0 {
+		convertedLabels = []string{}
+	}
+	if len(originalLabelsOutput) == 0 {
+		originalLabels = []string{}
+	}
+
+	// Validate that conversion added GC labels
+	if len(convertedLabels) < len(originalLabels) {
```
Swapnanil-Gupta: This can fail. 
For example, in alpine image:

Original GC labels:
```
containerd.io/distribution.source.docker.io=library/alpine,
containerd.io/gc.ref.content.config=sha256:7acffee03fe864cd6b88219a1028855d6c912e7cf6fac633aa4307529fd0cc08,
containerd.io/gc.ref.content.l.0=sha256:014e56e613968f73cce0858124ca5fbc601d7888099969a4eea69f31dcd71a53
```

Converted GC labels:
```
containerd.io/gc.ref.content.config=sha256:7acffee03fe864cd6b88219a1028855d6c912e7cf6fac633aa4307529fd0cc08,
containerd.io/gc.ref.content.l.0=sha256:014e56e613968f73cce0858124ca5fbc601d7888099969a4eea69f31dcd71a53
```

integration/convert_test.go
```diff
@@ -198,6 +198,28 @@ func validateConversion(t *testing.T, sh *shell.Shell, originalDigest, converted
 			t.Errorf("manifest %v does not contain expected soci index digest %v", manifestDesc, sociIndexDesc.Digest)
 		}
 	}
+
```
Swapnanil-Gupta: Maybe something like this based on Kern's comment?

``` go
// verify config exists in the content store
var config ocispec.ImageConfig
configBytes := sh.O("ctr", "content", "get", manifest.Config.Digest.String())
if err := json.Unmarshal(configBytes, &config); err != nil {
	t.Errorf("failed to decode manifest config: %v", err)
}

// verify config gc label
configGCLabel := "containerd.io/gc.ref.content.config"
if originalManifestLabels[configGCLabel] != convertedManifestLabels[configGCLabel] {
	t.Errorf("converted image has different gc config label")
}

// verify layers exist in the content store
for i, layer := range manifest.Layers {
	o, err := sh.CombinedOLog("ctr", "content", "get", layer.Digest.String())
	if err != nil || strings.Contains(string(o), "not found") {
		t.Errorf("failed to get layer %v", layer.Digest)
	}

	// verify layer gc label
	layerGCLabel := fmt.Sprintf("containerd.io/gc.ref.content.l.%d", i)
	if originalManifestLabels[layerGCLabel] != convertedManifestLabels[layerGCLabel] {
		t.Errorf("converted image has different gc layer label")
	}
}
```

integration/convert_test.go
```diff
@@ -198,6 +198,28 @@ func validateConversion(t *testing.T, sh *shell.Shell, originalDigest, converted
...
+	// Get GC labels
+	convertedLabelsOutput := sh.O("ctr", "content", "label", convertedDigest)
```
Swapnanil-Gupta: This gets the labels of the image index. Is it testing the newly added gc labels of the manifests?

integration/convert_test.go
coderbirju: The issue here is content get does not give us the label information associated with a digest, while running `ctr labels` only gives raw label information.

integration/convert_test.go
coderbirju: fixed this

soci/soci_convert.go
```diff
@@ -307,6 +307,18 @@ func (b *IndexBuilder) annotateImages(ctx context.Context, ociIndex *ocispec.Ind
...
 			indexWithMetadata.Desc.Annotations[IndexAnnotationImageManifestDigest] = manifestDesc.Digest.String()
 		}
+
+		err = store.LabelGCRefContent(ctx, b.blobStore, *manifestDesc, "config", manifest.Config.Digest.String())
+		if err != nil {
+			return err
+		}
+
+		for i, layer := range manifest.Layers {
+			err = store.LabelGCRefContent(ctx, b.blobStore, *manifestDesc, fmt.Sprintf("l.%d", i), layer.Digest.String())
```
sondavidb, line 317: nit: does this match the format of containerd's GC labeling? It might be good to make sure it does (though not strictly needed since all that matters is the key)

integration/convert_test.go
```diff
@@ -197,7 +197,45 @@ func validateConversion(t *testing.T, sh *shell.Shell, originalDigest, converted
...
+		if err != nil {
+			t.Errorf("failed to get config digest from manifest: %v", err)
+			continue
+		}
+		// use the digest to get the image
+		configBytes := sh.O("ctr", "content", "get", configDigest)
+
+		var configInfo ocispec.Image
+		if err := json.Unmarshal(configBytes, &configInfo); err != nil {
+			t.Errorf("failed to decode config: %v", err)
```
sondavidb: `continue` after this line?

integration/convert_test.go
```diff
@@ -197,7 +197,45 @@ func validateConversion(t *testing.T, sh *shell.Shell, originalDigest, converted
 		if dg, ok := manifest.Annotations[soci.ImageAnnotationSociIndexDigest]; !ok || dg != sociIndexDesc.Digest.String() {
 			t.Errorf("manifest %v does not contain expected soci index digest %v", manifestDesc, sociIndexDesc.Digest)
 		}
+
+		// get labels associated with the manifest and extract the digest
+		configDigest, err := getManifestLabelsAndConfigDigest(sh, manifestDesc.Digest.String())
```
sondavidb: Is there value in having this as a standalone function? Just feels a little strange given how specific the func is for this test. If the logic is too complex a local func would make more sense IMO (unless you think the logic is extensible enough for future/existing tests)

integration/convert_test.go
Kern--: The labels are ordered so I think we could do what Swapnanil is suggesting, but maybe that's a little too brittle.

I think we could do the inverse, though: Parse the output of `ctr labels` into a map of value -> list of labels, then for each image layer, ensure there is a unique key that has the gc content label prefix.

Ultimately, our goal is to ensure that content doesn't get garbage collected. Verifying that the labels are set correctly on the manifest is a proxy for that which could be wrong if, for example, we get leases wrong and content gets gc'd before we label everything. Could we measure directly by forcing a gc run then verifying all the content that we expect still exists?

coderbirju: [Discussed Offline] 
We now follow the following steps to ensure all the digests are referrenced in the labels
- make a map of digest(key) -> label(list) 
- iterate the layer digests and config digest from the manifest
- verify that there is atleast one label associated with the digests

Swapnanil-Gupta: Should we not verify that the manifest config and layers are still present in the content store after the `convert and replace` process, which would have caught the issue? Or is that not necessary? @Kern--

Kern--: I think that would be ideal, but we need to be sure a GC run completed after the replace and before the check. The only way to do that is to delete and image or lease and specify that GC should happen synchronously (e.g. `ctr lease delete --sync`).

I think checking labels is a reasonable alternative that's definitely deterministic.

integration/convert_test.go
```diff
@@ -101,6 +101,57 @@ func TestConvertWithForceRecreateZtocs(t *testing.T) {
...
+			if len(keyVal) != 2 {
+				continue
+			}
+			digestMap[keyVal[1]] = append(digestMap[keyVal[1]], keyVal[0])
+		}
+		return digestMap, nil
+	}
+
+	// Helper function to verify image layers contain required labels
+	verifyImageLayersContainsLabel := func(digestMap map[string][]string, manifest ocispec.Manifest) (bool, error) {
```
sondavidb: `bool, err` seems redundant, return one or the other IMO

integration/convert_test.go
```diff
@@ -197,7 +248,21 @@ func validateConversion(t *testing.T, sh *shell.Shell, originalDigest, converted
...
+		if err != nil {
+			t.Errorf("failed to get config digest from manifest: %v", err)
+			continue
+		}
+
+		err = verifyImageLayersContainsLabel(digestLabelMap, manifest)
+		if err != nil {
+			t.Errorf("error verifying layers: %v", err)
+			continue
+		}
```
sondavidb, line 263: Can we add one more check to ensure that the content exists via `ctr content get [digest]`? (It might not be necessary if we feel the label existing is enough proof that it's in the content store but good to check IMO.)

Swapnanil-Gupta: I asked the same question. Based on Kern's reply this check will make sense when we have ensured that a GC run was completed after the convert and replace process (which we currently don't have a deterministic way of doing?).

sondavidb: Oh yeah I missed that conversation above. I guess it makes sense.

PR #1812, opened 2025-12-09
Ignore runtime-spec in dependency bump script

sondavidb:
**Issue #, if available:**

**Description of changes:**
Our dependency bump script keeps failing due to runtime-spec changes. Once this issue is fixed can revert this: https://github.com/containerd/containerd/issues/12493

**Testing performed:**


Kern-- (approved):
Can you create an issue to track this?

It doesn't look like the runtime spec bump was backported, so we will also need to update to containerd v2. I don't think there's any reason not to at this point since we'll still test against containerd 1.7.

Swapnanil-Gupta (approved):

PR #1844, opened 2026-01-14
Honor registry mirrors when fetching SOCI artifacts

### LINKED ISSUE #1013: [Bug] Artifact fetcher does not honor mirrors
### Description

For every unique image manifest digest, we attempt to fetch SOCi artifacts. We take the source information of a blob, most importantly the [original image reference,](https://github.com/awslabs/soci-snapshotter/blob/91ce7a8fcb74deb01a950b891b7c8ad5d0dcb2e1/fs/fs.go#L238) and create a [new ORAS `*remote.Repository`](https://github.com/awslabs/soci-snapshotter/blob/91ce7a8fcb74deb01a950b891b7c8ad5d0dcb2e1/fs/fs.go#L227) that we use to fetch artifacts. This means that even if we had a mirror configured for a host, we we would be fetching SOCI artifacts from the original host. We should support fetching from multiple endpoints, just like we do in our [span fetcher](https://github.com/awslabs/soci-snapshotter/blob/91ce7a8fcb74deb01a950b891b7c8ad5d0dcb2e1/fs/remote/resolver.go#L197).

_No response_

We could expand [`sociContext`](https://github.com/awslabs/soci-snapshotter/blob/91ce7a8fcb74deb01a950b891b7c8ad5d0dcb2e1/fs/fs.go#L217-L225) to include the host configurations, so that during [`Init`](https://github.com/awslabs/soci-snapshotter/blob/91ce7a8fcb74deb01a950b891b7c8ad5d0dcb2e1/fs/fs.go#L227) we can try to fetch artifacts from each host. 

### Host information

1. OS: 
2. Snapshotter Version: 
3. Containerd Version: 

_No response_

milosgajdos:
**Issue #, if available:**
Fixes #1013

**Description of changes:**
SOCI artifact discovery/fetch (index + zTOCs) was constructing an ORAS repository from the original image locator (e.g. docker.io/...), causing the snapshotter to ignore configured registry mirrors and talk to the origin registry for SOCI artifacts.

This changes the SOCI index fetch path to use the resolved docker.RegistryHost (mirror host + client) when constructing the ORAS repository locator. This makes SOCI artifact requests follow the same mirror configuration used for layer pulls.

**Testing performed:**
✅ 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,6 +56,7 @@ FROM public.ecr.aws/amazonlinux/amazonlinux:2023 AS rapidgzip-builder
 
 ARG RAPIDGZIP_VERSION
 ARG TARGETARCH
+ENV SOCI_SNAPSHOTTER_SRC=/src/github.com/awslabs/soci-snapshotter
 
 RUN dnf update -y && dnf install -y \
     binutils \
@@ -90,7 +91,7 @@ RUN git clone https://github.com/mxmlnkn/rapidgzip.git && \
           -DCMAKE_C_FLAGS="-O2 -fPIC" \
           ${ISAL_FLAGS} \
           -DCMAKE_BUILD_TYPE=Release .. && \
-    make -j$(nproc) && \
+    cmake --build . --target rapidgzip --parallel "$(nproc)" && \
     cp src/tools/rapidgzip /opt/rapidgzip/usr/local/bin/ && \
     chmod +x /opt/rapidgzip/usr/local/bin/rapidgzip && \
     cd ../.. && \
@@ -104,9 +105,10 @@ ARG NERDCTL_VERSION
 ARG TARGETARCH
 ENV GOPROXY=direct
 ENV GOCOVERDIR=/test_coverage
+ENV SOCI_SNAPSHOTTER_SRC=/src/github.com/awslabs/soci-snapshotter
 
 COPY ./integ_entrypoint.sh /integ_entrypoint.sh
-COPY . $GOPATH/src/github.com/awslabs/soci-snapshotter
+COPY . ${SOCI_SNAPSHOTTER_SRC}
 RUN dnf update && dnf upgrade && dnf install -y \
     diffutils \
     findutils \
@@ -123,14 +125,14 @@ RUN dnf update && dnf upgrade && dnf install -y \
 COPY --from=igzip-builder /opt/igzip/usr /usr/local
 COPY --from=rapidgzip-builder /opt/rapidgzip/usr/local /usr/local
 
-RUN cp $GOPATH/src/github.com/awslabs/soci-snapshotter/out/soci /usr/local/bin/ \
-    && cp $GOPATH/src/github.com/awslabs/soci-snapshotter/out/soci-snapshotter-grpc /usr/local/bin/ \
+RUN cp ${SOCI_SNAPSHOTTER_SRC}/out/soci /usr/local/bin/ \
+    && cp ${SOCI_SNAPSHOTTER_SRC}/out/soci-snapshotter-grpc /usr/local/bin/ \
     && mkdir /etc/soci-snapshotter-grpc \
     && mkdir /etc/containerd/ \
-    && cp $GOPATH/src/github.com/awslabs/soci-snapshotter/integration/config/etc/soci-snapshotter-grpc/config.toml /etc/soci-snapshotter-grpc/ \
-    && cp $GOPATH/src/github.com/awslabs/soci-snapshotter/integration/config/etc/containerd/config.toml /etc/containerd/ \
-    && cp $GOPATH/src/github.com/awslabs/soci-snapshotter/soci-snapshotter.service /etc/systemd/system \
-    && cp $GOPATH/src/github.com/awslabs/soci-snapshotter/soci-snapshotter.socket /etc/systemd/system
+    && cp ${SOCI_SNAPSHOTTER_SRC}/integration/config/etc/soci-snapshotter-grpc/config.toml /etc/soci-snapshotter-grpc/ \
+    && cp ${SOCI_SNAPSHOTTER_SRC}/integration/config/etc/containerd/config.toml /etc/containerd/ \
+    && cp ${SOCI_SNAPSHOTTER_SRC}/soci-snapshotter.service /etc/systemd/system \
+    && cp ${SOCI_SNAPSHOTTER_SRC}/soci-snapshotter.socket /etc/systemd/system
 RUN curl -sSL --output /tmp/containerd.tgz https://github.com/containerd/containerd/releases/download/v${CONTAINERD_VERSION}/containerd-${CONTAINERD_VERSION}-linux-${TARGETARCH:-amd64}.tar.gz \
     && tar zxvf /tmp/containerd.tgz -C /usr/local/ \
     && rm -f /tmp/containerd.tgz
--- a/Makefile
+++ b/Makefile
@@ -178,7 +178,12 @@ $(COVDIR)/unit: $(COVDIR)
 integration: build
 	@echo "$@"
 	@echo "SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT)"
-	@GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true gotestsum --rerun-fails --format standard-verbose --packages ./integration -- -timeout 0 $(GO_TEST_FLAGS)
+	@if command -v gotestsum >/dev/null 2>&1; then \
+		GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true gotestsum --rerun-fails --format standard-verbose --packages ./integration -- -timeout 0 $(GO_TEST_FLAGS); \
+	else \
+		echo "gotestsum not found; falling back to go test ./integration"; \
+		GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true go test ./integration -timeout 0 $(GO_TEST_FLAGS); \
+	fi
 
 show-integration-coverage: $(COVDIR)/unit
 	go tool covdata percent -i $(COVDIR)/integration
--- a/Makefile
+++ b/Makefile
@@ -182,7 +182,7 @@ integration: build
 		GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true gotestsum --rerun-fails --format standard-verbose --packages ./integration -- -timeout 0 $(GO_TEST_FLAGS); \
 	else \
 		echo "gotestsum not found; falling back to go test ./integration"; \
-		GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true go test ./integration -timeout 0 $(GO_TEST_FLAGS); \
+		GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true go test -v ./integration -timeout 0 $(GO_TEST_FLAGS); \
 	fi
 
 show-integration-coverage: $(COVDIR)/unit
--- a/fs/fetch_soci_index_test.go
+++ b/fs/fetch_soci_index_test.go
@@ -1,0 +1,166 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package fs
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strconv"
+	"sync/atomic"
+	"testing"
+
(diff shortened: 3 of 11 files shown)


fs/fs.go
```diff
@@ -447,11 +447,18 @@ func (fs *filesystem) MountParallel(ctx context.Context, mountpoint string, labe
...
 	s := src[0]
+	if len(s.Hosts) == 0 {
+		return fmt.Errorf("no registry hosts available for %q", imageRef)
+	}
 	client := s.Hosts[0].Client
 	refspec, err := reference.Parse(imageRef)
 	if err != nil {
 		return fmt.Errorf("cannot parse image ref (%s): %w", imageRef, err)
 	}
+	refspec, err = withLocatorHost(refspec, s.Hosts[0].Host)
```
sondavidb: Doesn't this just use the first locator host? Seems if there are mirrors this would fail. We would have to add logic outside of this to retry the whole flow without mirrors I believe. (Any changes here would have to be done in MountLocal too.)

sondavidb (commented):
Hey, thanks for the contribution and sorry for leaving this hanging for so long. I left a comment to make sure the feature is working. Would also love an integration test as well if possible. After changes are made happy to run them through CI.

milosgajdos:
Hey! Sorry, I was away on hols, but will be picking this up today. Thanks for your patience

milosgajdos:
> Would also love an integration test as well if possible.

Added some integration tests - not sure if that's what you are after but figured it could be something to at least start a conversation

sondavidb:
Ah, I was looking more at testing in the `integration/` folder. I think something along the lines of mirroring the image to a local registry, having the SOCI config primarily pull from some bogus source, and then have the mirror point to the local registry. Is that a sensical workflow?

milosgajdos:
Oh I see! I misunderstood 😬 yeah I can look into that; it seems sensible yeah

milosgajdos:
@sondavidb I smacked in an integration test case in https://github.com/awslabs/soci-snapshotter/pull/1844/commits/76fff9b7efe66681a1d62ed2e392dddb6f3c124b

My thinking was adding a case where we test a fallback to good mirror when trying a dead mirror (bogus) first. Basically that should verify that we recover from a bad mirror endpoint and still pull via SOCI. LMK if this is along the lines you're thinking too.

sondavidb:
Yeah that should be fine by me. I'll look closer at it tomorrow but will run our CI for now. Thanks!

Dockerfile
```diff
@@ -90,7 +91,7 @@ RUN git clone https://github.com/mxmlnkn/rapidgzip.git && \
           -DCMAKE_C_FLAGS="-O2 -fPIC" \
           ${ISAL_FLAGS} \
           -DCMAKE_BUILD_TYPE=Release .. && \
-    make -j$(nproc) && \
+    cmake --build . --target rapidgzip --parallel "$(nproc)" && \
```
milosgajdos, line 94: this stage only needs one artifact: the `rapidgzip` binary, So building the whole default CMake project was wasted work: this asks `CMake` for exactly the target we actually copy into the final image, while still letting `CMake` choose the right underlying build tool and dependency graph. It cut this stage on my build machine from about 279s to about 94s.

milosgajdos:
I've added some additional build instructions to docs and some improvements for [the speed of] CI.

Dockerfile
sondavidb: Great find, thanks for this, this always took a while for me to compile too. I did notice that rapidgzip tended to be the step that failed when our Dockerfile failed to build. Wonder if this should help to alleviate this.

fs/fetch_soci_index_test.go
sondavidb, line 1: We should leave our integration tests within the integration test framework. We definitely shouldn't be pulling whole images for it as there's a very real throttling issue that can happen.

I'm ok deleting this entirely or moving the residuals to the integration tests.

fs/fs.go
```diff
@@ -447,19 +447,29 @@ func (fs *filesystem) MountParallel(ctx context.Context, mountpoint string, labe
 	}
 	// download the target layer
 	s := src[0]
-	client := s.Hosts[0].Client
+	if len(s.Hosts) == 0 {
+		return fmt.Errorf("no registry hosts available for %q", imageRef)
+	}
+	host := s.Hosts[0]
```
sondavidb: Hm, maybe I'm missing something but I don't think this works since it only gets the first host from the list of hosts. Unless there's something I'm missing here?

Same in MountLocal

Makefile
```diff
@@ -178,7 +178,12 @@ $(COVDIR)/unit: $(COVDIR)
 integration: build
 	@echo "$@"
 	@echo "SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT)"
-	@GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true gotestsum --rerun-fails --format standard-verbose --packages ./integration -- -timeout 0 $(GO_TEST_FLAGS)
+	@if command -v gotestsum >/dev/null 2>&1; then \
+		GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true gotestsum --rerun-fails --format standard-verbose --packages ./integration -- -timeout 0 $(GO_TEST_FLAGS); \
+	else \
+		echo "gotestsum not found; falling back to go test ./integration"; \
+		GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true go test ./integration -timeout 0 $(GO_TEST_FLAGS); \
```
sondavidb: For consistency with above

```suggestion
		GO111MODULE=$(GO111MODULE_VALUE) SOCI_SNAPSHOTTER_PROJECT_ROOT=$(SOCI_SNAPSHOTTER_PROJECT_ROOT) ENABLE_INTEGRATION_TEST=true go test -v ./integration -timeout 0 $(GO_TEST_FLAGS); \
```

fs/fetch_soci_index_test.go
milosgajdos: I can bin this file if you want me to, yeah

fs/fs.go
milosgajdos: Hmmmm, I *thought* I fixed this...ok let me dig in a bit

fs/fetch_soci_index_test.go
sondavidb: I'd like to get rid of the HTTP calls to the upstream repos. We could probably recreate similar enough integration testing if the testing doesn't already exist

sondavidb:
Just FYI for whenever this gets merged, we do rebase and merge so you will have to squash the commits accordingly and have each commit DCO signed. Doesn't have to be now of course, just a heads up 🙂

fs/fs.go
milosgajdos: Ok, spent some time on this today and added bunch regression tests: https://github.com/awslabs/soci-snapshotter/pull/1844/changes/9315f827d274b45862a1854c6cacd4a78bf5a17f

milosgajdos: Because each retry reuses the same mountpoint, and `Unpack` writes into that directory incrementally, `resetMountpoint(...)` tries to prevent the next mirror host retry from running on top of partially extracted files from the previous failed attempt - that'd be bad 😓  basically, `resetMountpoint` makes each host retry start from a clean extraction directory; that _should_ be fine. I've done some local experiments with this latest commits, but I will be doing more. I'll report back if I discover anything before you get around reviewing this

fs/fetch_soci_index_test.go
milosgajdos: I dont think we do any HTTP calls to upstream repos. We use `httptest.NewServer` only, but the mention of `docker.io` does seem confusing there I agree; though it's only used as an input string for ref parsing...I can change it to some other ref string if that works?

milosgajdos:
CI fail is on the accepted commit suggestion 😬

sondavidb:
CI fail just looks like DCO. Unless I'm misunderstanding, which, apologies if so

milosgajdos:
> CI fail just looks like DCO. Unless I'm misunderstanding, which, apologies if so

yeah, GH [platform] fails to reconcile commit signatures when you accept suggestions on a fork; that failing DCO is my accepting your Makefile suggestion - itll go away after squash

milosgajdos:
@sondavidb anything else you want me to address here?

sondavidb:
Nope! Just need to take a look a little closer at the changes before approving. But at a glance they look good.

Thanks again!

fs/fs.go
```diff
@@ -447,32 +447,37 @@ func (fs *filesystem) MountParallel(ctx context.Context, mountpoint string, labe
...
+	for _, hostRef := range hostRefs {
+		if !fs.inProgressImageUnpacks.ImageExists(imageDigest) {
+			err := fs.preloadAllLayers(ctx, desc, imageDigest, hostRef.refspec, hostRef.host, fs.contentStore)
+			if err != nil {
+				hostErr = errors.Join(hostErr, fmt.Errorf("registry host %q: %w", hostRef.host.Host, err))
+				continue
+			}
+		}
+
+		err := fs.rebase(ctx, desc.Digest, imageDigest, mountpoint)
```
sondavidb, line 466: Rebasing can probably be put outside of this loop as it's a fully offline-operation so we don't need to try it more than once on a bad registry.

sondavidb (commented):
LGTM, just a small thing and squash and I'll approve it. I'll also see if I can get someone else on my team to take a look.

Awesome work on this, we've been wanting this for a while :)

fs/fs.go
```diff
@@ -800,50 +778,218 @@ func (fs *filesystem) MountLocal(ctx context.Context, mountpoint string, labels
...
+	}
+	return hostRefs, hostErr
+}
+
+func (fs *filesystem) unpackLocalLayerFromRegistryHosts(ctx context.Context, mountpoint string, mounts []mount.Mount, desc ocispec.Descriptor, uncompressedDigest digest.Digest, hostRefs []registryHostRef) error {
+	var hostErr error
+	for _, hostRef := range hostRefs {
+		if err := resetMountpoint(mountpoint); err != nil {
+			hostErr = errors.Join(hostErr, fmt.Errorf("registry host %q: %w", hostRef.host.Host, err))
+			continue
```
sondavidb, line 842: nit: I see a lot of `continue`s being used throughout this PR. Which is fine, obviously, but I wonder if having traditional returns and then having each call be a separate function would be easier to work with (example below)

```go
func unpackLocalLayerFromRegistryHosts() error {
    var hostErr error
    for _, hostRef := range hostRefs {
        err := otherFunc(vars)
        if err == nil {
            return nil
        }
        hostErr = errors.Join(hostErr, err)
    }
}

func otherFunc(var vars) error {
    // core logic
}
```

Not needed by any means. Just food for thought if that was something you think would be more usable.

fs/fs.go
milosgajdos: So, I'm wondering... shouldnt `rebase` actually remain in the loop because it is where the current host’s async premount failure becomes visible...`preloadAllLayers` only starts premount work. The actual host-specific failure can surface later when `rebase` waits on the layer job...wouldnt we lose the retry if we moved it outside the loop? 🤔

fs/fs.go
milosgajdos: So, in my Go code I use `continue` a lot, but Im happy to rework this - old habits die hard 😬 lmk

fs/fs.go
milosgajdos: the rename/stat portion of `rebase` is offline, but `rebase` currently also waits for the host-specific `premount` job via `layerJob.errCh` https://github.com/milosgajdos/soci-snapshotter/blob/9315f827d274b45862a1854c6cacd4a78bf5a17f/fs/fs.go#L629

fs/fs.go
sondavidb: Different styles I suppose :P no worries, I'm not super opinionated on this.

fs/fs.go
sondavidb: The jobs are keyed via digest and not by ref, and if preloading a layer fails then it will cancel the entire image job and return, so we won't be accidentally getting another registry host's image job.

Though, I wonder now if that's a problem. Should fallbacks for the parallel pull mode be per-layer or per-image? I think per-layer would be more consistent with other snapshotter behaviors. If so, we should maybe move this to the preloadAllLayers and rotate through registry hosts per layer.

But I'm also OK taking this as a followup. We've iterated on this plenty already. (You're more than welcome to change this now if you wish, though.)

sondavidb (dismissed):

milosgajdos:
@sondavidb mind poking some maintainers so we can move forward with this? 🙏

sondavidb:
Sure, sorry for the drag @milosgajdos. I've gone ahead and pinged someone but as he's not around right now I don't expect him to look at this immediately.

fs/fs.go
```diff
@@ -447,32 +447,37 @@ func (fs *filesystem) MountParallel(ctx context.Context, mountpoint string, labe
...
+				hostErr = errors.Join(hostErr, fmt.Errorf("registry host %q: %w", hostRef.host.Host, err))
+				continue
+			}
+		}
+
+		err := fs.rebase(ctx, desc.Digest, imageDigest, mountpoint)
 		if err != nil {
-			return fmt.Errorf("failed to preload layers for image manifest digest %s: %w", imageDigest, err)
+			hostErr = errors.Join(hostErr, fmt.Errorf("registry host %q: failed to rebase layer %s: %w", hostRef.host.Host, desc.Digest, err))
+			continue
```
henry118, line 469: rebase failures should have nothing to do with registry mirrors. i don't think we need to try the next host.

fs/fs.go
```diff
@@ -447,32 +447,37 @@ func (fs *filesystem) MountParallel(ctx context.Context, mountpoint string, labe
...
-	if err != nil {
-		return fmt.Errorf("failed to rebase layer %s: %w", desc.Digest, err)
+		return nil
 	}
-	return nil
+	if hostErr != nil {
+		return fmt.Errorf("failed to preload layers for image manifest digest %s: %w", imageDigest, hostErr)
+	}
+	return fmt.Errorf("failed to preload layers for image manifest digest %s", imageDigest)
 }
```
henry118, line 478: Feel like these can be simplified. esp the last return somehow swallowed the internal errors.

henry118 (commented):
@milosgajdos Thanks for this contribution. 

At high level, it occurs to me that the retry granularity here is a bit coarse. I think the only errors that worth retrying the next host are the host I/O related errors. i.e, failing to resolve the host. Other types of errors such as post-processing of downloaded artifacts (decompression/unpacking) should result an immediate failure.

So in other words, I think the logic of iterating the hosts should be implemented at lower level where the actual HTTP requests are initiated.

YQ-Wang:
Interested user here :) I’ve been following this because I’d like to use the feature.

The remaining blocker seems to be retry scope rather than the mirror-awareness behavior itself. The maintainer comments make sense to me: retrying the next host for actual registry-host / network-facing failures seems reasonable, while local post-processing failures like unpack/apply/rebase should fail immediately. It also seems important to preserve the underlying error rather than returning a generic fallback error.

Would a narrow fix along those lines be enough to unblock this PR, with any finer-grained retry redesign handled as follow-up? Happy to test an updated branch or create a PR to this current branch if useful .

fs/fs.go
milosgajdos: I completely missed this comment; work has kept me busy for a few weeks, but I'll be picking this up today

Issue #1851, opened 2026-01-26
[Bug] snapshotter will always try to build index for platofrm it is running on even if the image for the platform does not exist


mitja-hofer:
### Description

`soci_converter.go` function `Convert` correctly handles single-architecture images to build for the architecture specified in the image. If the received is not an image, but an image manifest, the index builder will try to build for the architecture of the system it is running on, which is not necessarily referenced in the image manifest.

### Steps to reproduce the bug

If you have ARM architecture only, and create an SBOM as well, you end up with an image manifest referencing ARM architecture only. As https://github.com/awslabs/cfn-ecr-aws-soci-index-builder builds only for AMD64, the snapshotter runs on AMD64 and will always try to create an index for AMD64 even the image does not exist. This will fail the snapshotter with `failed to convert OCI index: image manifest for linux/amd64: not found`.

### Describe the results you expected

I would expect to build SOCI index only for architectures referenced by the image manifest.

### Host information

1. OS: Amazon Linux 2
2. Snapshotter Version: 
3. Containerd Version:

sondavidb:
Linked PR is closed. Will close this too. Feel free to re-open if you feel further changes need to be made in this package.

Thanks!

PR #1856, opened 2026-01-30
fix:  default to all image platforms for OCI Index conversion

Shubhranshu153:
**Issue #, if available:**
https://github.com/awslabs/soci-snapshotter/issues/1851

**Description of changes:**
When converting an OCI Index the Convert function incorrectly defaulted to the system platform instead of all available platforms in the image. This caused failures when converting cross-architecture images.

**Testing performed:**

### DIFF
--- a/soci/soci_convert.go
+++ b/soci/soci_convert.go
@@ -84,7 +84,7 @@ func (b *IndexBuilder) Convert(ctx context.Context, img images.Image, opts ...Co
 	if len(allPlatforms) == 0 {
 		return nil, errors.New("image does not support any platforms")
 	}
-	defaultPlatform := platforms.DefaultSpec()
+	var defaultPlatforms []ocispec.Platform
 	if images.IsManifestType(img.Target.MediaType) {
 		// If the image's target descriptor is a single manifest, then it will not
 		// contain a platform because that information is stored in the image config instead.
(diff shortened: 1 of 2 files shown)


sondavidb:
Actually, can we double check that the test case you added is not already covered here? https://github.com/awslabs/soci-snapshotter/blob/main/integration/convert_test.go#L18

soci/soci_convert.go
```diff
@@ -168,6 +178,35 @@ func (b *IndexBuilder) Convert(ctx context.Context, img images.Image, opts ...Co
 	return &ociIndexDesc, nil
 }
 
+// filterAvailablePlatforms filters the given platforms to only include those whose
+// manifests are actually available in the content store. This is necessary because
+// when pulling a multi-platform image without --all-platforms, the OCI Index still
+// lists all platforms, but only some manifests are actually downloaded.
+func (b *IndexBuilder) filterAvailablePlatforms(ctx context.Context, target ocispec.Descriptor, allPlatforms []ocispec.Platform) []ocispec.Platform {
```
Swapnanil-Gupta, line 185: How will this work with the `--all-platforms` flag? Does it become redundant?

soci/soci_convert.go
```diff
@@ -168,6 +178,35 @@ func (b *IndexBuilder) Convert(ctx context.Context, img images.Image, opts ...Co
...
+	var availablePlatforms []ocispec.Platform
+	for _, platform := range allPlatforms {
+		matcher := platforms.OnlyStrict(platform)
+		desc, err := GetImageManifestDescriptor(ctx, b.contentStore, target, matcher)
+		if err != nil {
+			// Platform manifest not found in index, skip
+			continue
+		}
+		// Check if the manifest content is actually available in the content store
+		_, err = b.contentStore.ReaderAt(ctx, *desc)
```
henry118, line 200: [Exists](https://github.com/containerd/containerd/blob/release/1.7/content/helpers.go#L334) is probably better func to call here. otherwise we need to close the returned readers.

soci/soci_convert.go
henry118: +1. 

I think we are basically replicating the `--all-platforms` behavior here.

Fundamentally `soci` CLI does provide a option to convert for platforms but the CFN stack is [not enabling](https://github.com/awslabs/cfn-ecr-aws-soci-index-builder/blob/main/functions/source/soci-index-generator-lambda/handler.go#L339-L346) this behavior. Should we fix the CFN instead?

Shubhranshu153: hmm. But i think depending on the host platform is a fundamental error as we are defaulting to a platform that may not exist. We can probably add a check to see if the  manifest with a valid platform is 1 and if yes it would default to that.

Shubhranshu153:
Closing this one we will move the logic to soci index builder

PR #1862, opened 2026-02-10
chore: upgrade go version to 1.25

coderbirju:
**Issue #, if available:**
Move to newer go version now that 1.26 is released

**Description of changes:**

**Testing performed:**


cmd/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/awslabs/soci-snapshotter/cmd
 
-go 1.24.0
```
Swapnanil-Gupta, line 3: Why do we need to update minimum golang version?

coderbirju: SInce we're not testing with go 1.24 anymore I don't see a reason to have 1.24 in the go mod

Shubhranshu153: Although not required upgrading it is helpful as a marker whats the min desired version i think

Shubhranshu153:
we should have CI run both on 1.25 and 1.26. May not be everything but main test suite should atleast

PR #1864, opened 2026-02-16
[Closed] Add standalone convert mode for SOCI index creation without containerd

prafgup:
New PR - https://github.com/awslabs/soci-snapshotter/pull/1870

### DIFF
--- a/cmd/soci/commands/convert.go
+++ b/cmd/soci/commands/convert.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/internal"
@@ -61,9 +62,9 @@ var ConvertCommand = &cli.Command{
 	Name:      "convert",
 	Usage:     "convert an OCI image to a SOCI enabled image",
 	ArgsUsage: "[flags] <image_ref> <dest_ref>",
-	Flags: append(
-		append(
-			internal.PlatformFlags,
+	Flags: slices.Concat(
+		internal.PlatformFlags,
+		[]cli.Flag{
 			&cli.Int64Flag{
 				Name:  spanSizeFlag,
 				Usage: "Span size that soci index uses to segment layer data. Default is 4 MiB",
@@ -84,8 +85,18 @@ var ConvertCommand = &cli.Command{
 				Value:   false,
 				Aliases: []string{forceRecreateZtocsFlagShort},
 			},
-		),
-		internal.PrefetchFlags()...,
+			&cli.BoolFlag{
+				Name:  "standalone",
+				Usage: "Run in standalone mode without containerd runtime. In this mode, the command will download the source image, perform the conversion, and push the converted image back to the registry without requiring a running containerd instance.",
+			},
+			&cli.BoolFlag{
+				Name:    "verbose",
+				Aliases: []string{"v"},
+				Usage:   "Show detailed progress output",
+			},
+		},
+		internal.RegistryFlags,
+		internal.PrefetchFlags(),
 	),
 	Action: func(ctx context.Context, cmd *cli.Command) error {
 		srcRef := cmd.Args().Get(0)
@@ -103,13 +114,8 @@ var ConvertCommand = &cli.Command{
 			return fmt.Errorf("%w: %w", ErrInvalidDestRef, err)
 		}
 
-		var optimizations []soci.Optimization
-		for _, o := range cmd.StringSlice(optimizationFlag) {
-			optimization, err := soci.ParseOptimization(o)
-			if err != nil {
-				return err
-			}
-			optimizations = append(optimizations, optimization)
+		if cmd.Bool("standalone") {
+			return runStandaloneConvert(ctx, cmd, srcRef, dstRef)
 		}
 
 		client, ctx, cancel, err := internal.NewClient(ctx, cmd)
@@ -125,10 +131,6 @@ var ConvertCommand = &cli.Command{
 			return err
 		}
 
-		spanSize := cmd.Int64(spanSizeFlag)
-		minLayerSize := cmd.Int64(minLayerSizeFlag)
-		forceRecreateZtocs := cmd.Bool(forceRecreateZtocsFlag)
-
 		blobStore, err := store.NewContentStore(internal.ContentStoreOptions(ctx, cmd)...)
 		if err != nil {
 			return err
@@ -139,25 +141,13 @@ var ConvertCommand = &cli.Command{
 			return err
 		}
 
-		builderOpts := []soci.BuilderOption{
-			soci.WithMinLayerSize(minLayerSize),
-			soci.WithSpanSize(spanSize),
-			soci.WithBuildToolIdentifier(buildToolIdentifier),
-			soci.WithOptimizations(optimizations),
-			soci.WithArtifactsDb(artifactsDb),
-			soci.WithForceRecreateZtocs(forceRecreateZtocs),
-		}
-
-		allPrefetchFiles, err := internal.ParsePrefetchFiles(cmd)
+		builderOpts, err := parseBuilderOptions(cmd)
 		if err != nil {
-			return fmt.Errorf("failed to parse prefetch files: %w", err)
+			return err
 		}
-		if len(allPrefetchFiles) > 0 {
-			builderOpts = append(builderOpts, soci.WithPrefetchPaths(allPrefetchFiles))
-		}
+		builderOpts = append(builderOpts, soci.WithArtifactsDb(artifactsDb))
 
 		builder, err := soci.NewIndexBuilder(cs, blobStore, builderOpts...)
-
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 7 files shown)


cmd/soci/commands/internal/standalone.go
```diff
@@ -0,0 +1,382 @@
...
+
+// PushConvertedImage pushes a converted OCI Index (containing image + SOCI indexes) to the registry
+func PushConvertedImage(ctx context.Context, cmd *cli.Command, info *StandaloneImageInfo, convertedDesc ocispec.Descriptor, outputRefs []string) error {
+	username, password, err := ResolveCredentials(ctx, cmd, info.RefSpec.Hostname())
+	if err != nil {
+		return fmt.Errorf("failed to resolve credentials: %w", err)
+	}
+
+	for _, ref := range outputRefs {
+		if err := pushDescriptorToReference(ctx, cmd, info, convertedDesc, ref, username, password, "SOCI-enabled image"); err != nil {
```
VVKot: This current assumes that username / password for the source registry is going to work for destination one, which isn't necessarily the case?

cmd/soci/commands/internal/standalone.go
```diff
@@ -0,0 +1,382 @@
...
+
+	authClient := setupAuthClient(cmd, username, password)
+	repo.Client = authClient
+	repo.PlainHTTP = cmd.Bool("plain-http")
+
+	return repo, nil
+}
+
+// PushConvertedImage pushes a converted OCI Index (containing image + SOCI indexes) to the registry
+func PushConvertedImage(ctx context.Context, cmd *cli.Command, info *StandaloneImageInfo, convertedDesc ocispec.Descriptor, outputRefs []string) error {
```
VVKot: Taking a step back, maybe we can reuse `PushCommand`, or just fully rely on it - instead making pushing part of the new standalone command?

cmd/soci/commands/internal/standalone.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+// Cleanup removes temporary files created during standalone operation
+func (info *StandaloneImageInfo) Cleanup() error {
+	if info.TempDir != "" {
+		return os.RemoveAll(info.TempDir)
+	}
+	return nil
+}
+func GetManifestDescriptorForPlatform(ctx context.Context, cs content.Store, imageTarget ocispec.Descriptor, platform ocispec.Platform) (*ocispec.Descriptor, error) {
```
VVKot: I believe this is unused?

soci/store/oras_content_store.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	if !exists {
+		return nil, fmt.Errorf("%s: %w", desc.Digest, errdefs.ErrNotFound)
+	}
+
+	// Fetch the content
+	rc, err := s.orasStore.Fetch(ctx, desc)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch content: %w", err)
+	}
```
VVKot: Maybe I misunderstand, but didn't oraslib.CopyGraph already download the image for us earlier?

cmd/soci/commands/internal/standalone.go
```diff
@@ -0,0 +1,382 @@
...
+			return ac.Username, ac.Password, nil
+		}
+	}
+
+	// Return empty credentials for public registries
+	return "", "", nil
+}
+
+// GetPlatformsFromCommand extracts platform specifications from CLI flags
+func GetPlatformsFromCommand(cmd *cli.Command) ([]ocispec.Platform, error) {
```
VVKot: Is this consciously different from `GetPlatforms`, or should we just reuse that one?

soci/store/oras_content_store.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/errdefs"
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2/content/oci"
+)
+
+// OrasContentStore is an adapter that implements containerd's content.Store interface
+// using an ORAS OCI store. This allows the SOCI IndexBuilder to work with images
+// downloaded via ORAS without requiring a containerd daemon.
```
VVKot: Do we even need this class? Any reason not to use `local.NewStore(...)` like the other indexers do?

cmd/soci/commands/internal/standalone.go
prafgup: Yes, seems like a leftover from previous testing, removing it !

cmd/soci/commands/internal/standalone.go
prafgup: I did had a look at this initially, `func GetPlatforms(ctx context.Context, cmd *cli.Command, img images.Image, cs content.Store)` uses the `img.Target` which we don't have yet available as we need to parse it from the image that we have on disk. For normal `convert` the manifest would be provided by Containerd and we can easily deduce the platforms in the manifest.

Thus I kept it separate for this implementation where `GetPlatformsFromCommand` helps us resolve the manifest targets. The other way would be to get all the manifests and then filter only the ones that we need  by wrapping GetPlatforms. I'll check if there is a better way for doing this.

cmd/soci/commands/internal/standalone.go
prafgup: Yes, the current assumption is that Auth would be same in case provided as `--user` else we would look for auth creds in DockerConfigFIle. 
We can extend this to `--user-pull` and `--user-push` if we are strongly against using same creds for pulling and pushing using `--user`, that can come as a future extention as well if needed as we just need to change the order of resolving creds.

Usually for most use-cases one pulls the image from the same repository (or public repository) and pushed to the same repository (non-public).

I'll update to resolve credentials before pushing everytime from docker config (so correct creds are picked up from DockerConfigFile for respective repository).

soci/store/oras_content_store.go
prafgup: Created a parallel store to adhere to the SOCI Store's specs - https://github.com/awslabs/soci-snapshotter/blob/main/soci/store/store.go#L40-L56

I'll check if I can simplify this with `local.NewStore` without creating a full store of type `SociStore`

cmd/soci/commands/internal/standalone.go
prafgup: I checked this previously and seemed like the pre existing Push command is overly complex with logic for V1 and V2 soci indexes alongside the containerd store, we might only be able to reuse a very small part of the whole push command. Thus refactoring it does not make sense here. The convert command only supports V2 soci indexes.

https://github.com/awslabs/soci-snapshotter/blob/bf86675cc81f4b52c5203dc8f08e539b1e3e6b72/cmd/soci/commands/push.go#L169-L203

and 

https://github.com/awslabs/soci-snapshotter/blob/bf86675cc81f4b52c5203dc8f08e539b1e3e6b72/cmd/soci/commands/push.go#L275-L279

The only extra flag in push is `--max-concurrent-uploads` which when not specified would default to 3

soci/store/oras_content_store.go
prafgup: `orasStore.Fetch` does not download the image/blobs but just returns a reader to where its already downloaded.

cmd/soci/commands/internal/standalone.go
prafgup: I see that `soci.ConvertWithPlatforms` would do the magic here and only convert for the requested platforms for the image. Thus we don't need to pre-filter the manifests per platform, thus removing this logic of pre filtering and relying entirely on internal IndexBuilder to filter out the platforms. The only side effect I see is that in case of failures (asking for converting a platform which does not exist in images manifest) would be deferred to convert time instead of image load time which is fine and acceptable as per the current Convert as well.

cmd/soci/commands/push.go
```diff
@@ -120,34 +117,21 @@ if they are available in the snapshotter's local content store.
 			return err
 		}
 
-		dst, err := remote.NewRepository(refspec.Locator)
+		username, password, err := internal.ResolveCredentials(cmd, refspec.Hostname(), "username")
```
prafgup: Using "username" instead of "user" which is a pre-existing bug - https://github.com/awslabs/soci-snapshotter/issues/1865

prafgup: Will change to `--user` in case the fix is merged before this PR - https://github.com/awslabs/soci-snapshotter/pull/1866

prafgup:
https://github.com/awslabs/soci-snapshotter/pull/1867 https://github.com/awslabs/soci-snapshotter/pull/1867#issuecomment-3924031277 would lead to merge conflicts as its touching the input flags behaviors, I would rebase with main in case that PR is merged soon. Also would refactor flags such as `"standalone"` as part of rebase later.

cmd/soci/commands/convert.go
```diff
@@ -84,8 +85,18 @@ var ConvertCommand = &cli.Command{
...
-		internal.PrefetchFlags()...,
+			&cli.BoolFlag{
+				Name:  "standalone",
+				Usage: "Run in standalone mode without containerd runtime. In this mode, the command will download the source image, perform the conversion, and push the converted image back to the registry without requiring a running containerd instance.",
+			},
+			&cli.BoolFlag{
+				Name:    "verbose",
+				Aliases: []string{"v"},
+				Usage:   "Show detailed progress output",
+			},
```
prafgup, line 96: I was also thinking of implementing something like a `quiet` flag instead which would be the opposite of verbose (used in push command) which would default to "always log" and when quiet is set it would not log these extra info logs. 

Currently the convert command (default) anyways always outputs something like - 
```
ztoc skipped - layer sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1 (application/vnd.oci.image.layer.v1.tar+gzip) size 32 is less than min-layer-size  10485760
...
```

thus I kept extra logging as verbose instead of making it a opt-out. Let me know if quiet makes more sense here than verbose.

integration/standalone_convert_test.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/containerd/platforms"
+)
+
+func TestStandaloneConvertBasic(t *testing.T) {
+	t.Parallel()
+	regConfig := newRegistryConfig()
+	sh, done := newShellWithRegistry(t, regConfig)
+	defer done()
+
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t))
```
VVKot, line 33: Can this and below instead be `disableContainerd`? So that we definitely test that standalone mode work as expected

cmd/soci/commands/convert.go
```diff
@@ -202,3 +193,95 @@ var ConvertCommand = &cli.Command{
...
+// runStandaloneConvert runs the convert command in standalone mode (without containerd)
+func runStandaloneConvert(ctx context.Context, cmd *cli.Command, srcRef string, dstRef string) error {
+	verbose := cmd.Bool("verbose")
+	if verbose {
+		fmt.Printf("Standalone mode: downloading image %s\n", srcRef)
+	}
+
+	imageInfo, err := internal.DownloadImageFromRegistry(ctx, cmd, srcRef)
+	if err != nil {
+		return fmt.Errorf("failed to download image: %w", err)
```
VVKot: nit: the errors from the above call are already wrapped, you can just return them

cmd/soci/commands/convert.go
```diff
@@ -202,3 +193,95 @@ var ConvertCommand = &cli.Command{
...
+		}
+		optimizations = append(optimizations, optimization)
+	}
+
+	builderOpts := []soci.BuilderOption{
+		soci.WithMinLayerSize(cmd.Int64("min-layer-size")),
+		soci.WithSpanSize(cmd.Int64("span-size")),
+		soci.WithBuildToolIdentifier(buildToolIdentifier),
+		soci.WithOptimizations(optimizations),
+		soci.WithForceRecreateZtocs(cmd.Bool("force")),
```
VVKot: nit: those strings already exist in create.go, maybe you can reuse them somehow?

cmd/soci/commands/convert.go
```diff
@@ -202,3 +193,95 @@ var ConvertCommand = &cli.Command{
 		return err
 	},
 }
+
+// runStandaloneConvert runs the convert command in standalone mode (without containerd)
+func runStandaloneConvert(ctx context.Context, cmd *cli.Command, srcRef string, dstRef string) error {
```
VVKot, line 197: I think somewhere in there we should be doing:

```
artifactsDbPath := path.Join(dataDir, artifactsDbName)
      artifactsDb, err := soci.NewDB(artifactsDbPath)
```

similar to https://github.com/awslabs/cfn-ecr-aws-soci-index-builder/blob/4c9ce0a57e49c5bfacc0967392f1d2bf8b65f022/functions/source/soci-index-generator-lambda/handler.go#L305 / https://github.com/CloudSnorkel/standalone-soci-indexer/blob/8e7e875b23218c71b5ff45ad19d1803c437db7bd/handler.go#L149 ?

integration/standalone_convert_test.go
prafgup: Apparently we do require containerd for the test setup `copyImage` (that downloads a image to mirror. And later when verifying that soci snapshotter works `"nerdctl", "pull", "-q", "--snapshotter=soci", "--all-platforms", dstRef`.

But we can stop containerd just before the sonvert standalone conversion and restart it after conversion. I'll do that in new revision :)

cmd/soci/commands/convert.go
prafgup: Although not required, as we without it we would anyways create a default DB which would anyways be cleaned up and not reused in CI jobs. But just for consistency I'll add a /tmp location for this DB (as by default it would create in /var soci's path

Issue #1865, opened 2026-02-18
[Bug] Push command checking wrong "--username" flag instead of "--user" flag


prafgup:
### Description

According to the docs as well as Push integration tests, Push command should be using the `"--user"` flag for creds lookup

#### Documentation - https://github.com/awslabs/soci-snapshotter/blob/53346dda740518d886735e17a0c17e4c2ba7fd61/docs/registry-authentication.md#username--password-parameter
> The SOCI CLI can also receive credentials via the --user parameter which accepts a USERNAME:PASSWORD string.

#### Integration Tests - 
- https://github.com/awslabs/soci-snapshotter/blob/53346dda740518d886735e17a0c17e4c2ba7fd61/integration/push_test.go#L66-L70
- https://github.com/awslabs/soci-snapshotter/blob/53346dda740518d886735e17a0c17e4c2ba7fd61/integration/push_test.go#L127-L130
- etc...

#### Push credential logic - 
https://github.com/awslabs/soci-snapshotter/blob/53346dda740518d886735e17a0c17e4c2ba7fd61/cmd/soci/commands/push.go#L131-L136

Integration tests seem to be working because we explicitly set up the creds in the docker config here for each integration test - 

- https://github.com/awslabs/soci-snapshotter/blob/53346dda740518d886735e17a0c17e4c2ba7fd61/integration/util_test.go#L702-L706

and then try to use the same creds via `regConfig.creds()` thus is entirly skips the `--username` check and just uses the fallback docker credentials from docker config file.

---

### Describe the results you expected

`--user` to be respected instead of `--username`

### Host information

1. OS: Any
2. Snapshotter Version: 0.12.1 and below
3. Containerd Version: Any

prafgup:
Created a PR with the fix - https://github.com/awslabs/soci-snapshotter/pull/1866 (would be a breaking change in case someone is using `--username`). I believe another way here would be to support both `--user` and `--username` to not break existing users.

prafgup:
Fixed by - https://github.com/awslabs/soci-snapshotter/pull/1866

PR #1866, opened 2026-02-18
Change wrong "username" flag to "user" for creds lookup

prafgup:
**Issue #, if available:**
Fixes - https://github.com/awslabs/soci-snapshotter/issues/1865

**Description of changes:**

According to the docs as well as Push integration tests, Push command should be using the `"--user"` flag for creds lookup

Previously integration tests seem to be working because we explicitly set up the creds in the docker config here for each integration test - 

- https://github.com/awslabs/soci-snapshotter/blob/53346dda740518d886735e17a0c17e4c2ba7fd61/integration/util_test.go#L702-L706

and then try to use the same creds via `regConfig.creds()` thus is entirly skips the `--username` check and just uses the fallback docker credentials from docker config file.

**Testing performed:**
Ran Push integration tests
```
sudo GO_TEST_FLAGS="-run TestPushWithUserFlag -count=1" make integration

cd cmd/ ; GO111MODULE=auto go build -o /Volumes/git/prafulg/soci-snapshotter/out/soci-snapshotter-grpc  -ldflags '-X github.com/awslabs/soci-snapshotter/version.Version=e85e45bc.m -X github.com/awslabs/soci-snapshotter/version.Revision=e85e45bc85578616af7543fb37e35e4a476eb40f.m  -s -w '  ./soci-snapshotter-grpc
cd cmd/ ; GO111MODULE=auto go build -o /Volumes/git/prafulg/soci-snapshotter/out/soci  -ldflags '-X github.com/awslabs/soci-snapshotter/version.Version=e85e45bc.m -X github.com/awslabs/soci-snapshotter/version.Revision=e85e45bc85578616af7543fb37e35e4a476eb40f.m  -s -w '  ./soci
integration
=== RUN   TestPushWithUserFlag
--- PASS: TestPushWithUserFlag (20.66s)
PASS
ok      github.com/awslabs/soci-snapshotter/integration 130.860s
```
and 
```
sudo GO_TEST_FLAGS="-run TestSociArtifactsPushAndPull -count=1" make 
integration
cd cmd/ ; GO111MODULE=auto go build -o /Volumes/git/prafulg/soci-snapshotter/out/soci-snapshotter-grpc  -ldflags '-X github.com/awslabs/soci-snapshotter/version.Version=e85e45bc.m -X github.com/awslabs/soci-snapshotter/version.Revision=e85e45bc85578616af7543fb37e35e4a476eb40f.m  -s -w '  ./soci-snapshotter-grpc
cd cmd/ ; GO111MODULE=auto go build -o /Volumes/git/prafulg/soci-snapshotter/out/soci  -ldflags '-X github.com/awslabs/soci-snapshotter/version.Version=e85e45bc.m -X github.com/awslabs/soci-snapshotter/version.Revision=e85e45bc85578616af7543fb37e35e4a476eb40f.m  -s -w '  ./soci
integration
=== RUN   TestSociArtifactsPushAndPull
=== RUN   TestSociArtifactsPushAndPull/amd64
=== RUN   TestSociArtifactsPushAndPull/arm64
--- PASS: TestSociArtifactsPushAndPull (38.07s)
    --- PASS: TestSociArtifactsPushAndPull/amd64 (17.02s)
    --- PASS: TestSociArtifactsPushAndPull/arm64 (19.98s)
PASS
ok      github.com/awslabs/soci-snapshotter/integration 40.036s
```

### DIFF
--- a/cmd/soci/commands/internal/registry.go
+++ b/cmd/soci/commands/internal/registry.go
@@ -33,6 +33,10 @@
 package internal
 
 import (
+	"io"
+	"strings"
+
+	dockercliconfig "github.com/docker/cli/cli/config"
 	"github.com/urfave/cli/v3"
 )
 
@@ -80,4 +84,27 @@ var RegistryFlags = []cli.Flag{
 		Name:  "http-trace",
 		Usage: "Enable HTTP tracing for registry interactions",
 	},
+}
+
+// ResolveCredentials resolves registry credentials from multiple sources:
+// 1. CLI --user/-u flag (username:password)
+// 2. Docker config file (~/.docker/config.json)
+// 3. Empty credentials for public registries
+func ResolveCredentials(cmd *cli.Command, hostname string) (username, password string) {
+	if cmd.IsSet("user") {
+		user := cmd.String("user")
+		if i := strings.IndexByte(user, ':'); i > 0 {
+			return user[:i], user[i+1:]
+		}
+		return user, ""
+	}
+
+	cf := dockercliconfig.LoadDefaultConfigFile(io.Discard)
+	if cf.ContainsAuth() {
+		if ac, err := cf.GetAuthConfig(hostname); err == nil {
+			return ac.Username, ac.Password
+		}
+	}
+
+	return "", ""
 }
--- a/integration/push_test.go
+++ b/integration/push_test.go
@@ -65,7 +65,7 @@ func TestSociArtifactsPushAndPull(t *testing.T) {
 				t.Fatalf("could not get digest of local content store: %v", err)
 			}
 
-			sh.X("soci", "push", "--user", regConfig.creds(), "--platform", tt.Platform, regConfig.mirror(imageName).ref)
+			sh.X("soci", "push", "--platform", tt.Platform, regConfig.mirror(imageName).ref)
 			sh.X("rm", "-rf", filepath.Join(store.DefaultSociContentStorePath, "blobs", "sha256"))
 			sh.X(append(imagePullCmd, "--soci-index-digest", indexDigest, "--platform", tt.Platform, regConfig.mirror(imageName).ref)...)
 
@@ -78,6 +78,37 @@ func TestSociArtifactsPushAndPull(t *testing.T) {
 				t.Fatalf("unexpected digests before and after rpull; before = %v, after = %v", artifactsStoreContentDigest, artifactsStoreContentDigestAfterRPull)
 			}
 		})
+	}
+}
+
+func TestPushWithUserFlag(t *testing.T) {
+	regConfig := newRegistryConfig()
+	sh, done := newShellWithRegistry(t, regConfig)
+	defer done()
+
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t))
+
+	imageName := ubuntuImage
+	copyImage(sh, dockerhub(imageName), regConfig.mirror(imageName))
+	buildIndex(sh, regConfig.mirror(imageName), withMinLayerSize(0))
+
+	// Remove Docker config so there is no credential fallback.
+	// This forces authentication to rely solely on the --user/-u flag.
+	sh.X("rm", "-f", "/root/.docker/config.json")
+
+	_, err := sh.OLog("soci", "push", "--user", regConfig.creds(), regConfig.mirror(imageName).ref)
+	if err != nil {
+		t.Fatalf("push with --user flag should succeed, but got error: %v", err)
+	}
+
+	_, err = sh.OLog("soci", "push", "-u", regConfig.creds(), regConfig.mirror(imageName).ref)
+	if err != nil {
+		t.Fatalf("push with -u flag should succeed, but got error: %v", err)
+	}
+
+	_, err = sh.OLog("soci", "push", regConfig.mirror(imageName).ref)
+	if err == nil {
+		t.Fatal("push without --user flag should fail when Docker config is removed, but it succeeded")
 	}
 }
 
--- a/cmd/soci/commands/push.go
+++ b/cmd/soci/commands/push.go
@@ -20,12 +20,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"math"
 	"net/http"
 	"slices"
 	"sort"
-	"strings"
 
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/internal"
 	"github.com/awslabs/soci-snapshotter/fs"
@@ -33,7 +31,6 @@ import (
 	"github.com/awslabs/soci-snapshotter/soci/store"
 	"github.com/containerd/containerd/reference"
 	"github.com/containerd/platforms"
-	dockercliconfig "github.com/docker/cli/cli/config"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/urfave/cli/v3"
 	oraslib "oras.land/oras-go/v2"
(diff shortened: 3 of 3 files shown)


sondavidb:
Good catch, thanks for this!

I think the preferred change here might be to change our docs/integration tests to use `--username` (or just shorthand `-u`) as changing the flag entirely is a backwards-incompatible change which might break existing workflows. This would also mimic nerdctl's [login](https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-nerdctl-login) flags. Is this an acceptable solution for you?

prafgup:
@sondavidb Sounds like a acceptable solution to update the docs to maintain backwards compatibility. As well as add `-u` flag for the same.

My only minor concern is that usually we have something like `nerdctl login --username="${USERNAME}" --password="${PASSWORD}"` where `username` just signifies the username and not the password. 

But the current implementation for soci is `soci X --username <username>:<password>` which is more of a curl's authentication format - `curl --user <username>:<password>` or containerd's `ctr-remote's` - `ctr-remote image rpull --user <username>:<password> X`
- https://github.com/containerd/stargz-snapshotter/blob/main/docs/ctr-remote.md#optimizing-an-image-with-custom-configuration

I believe an acceptable solution here is to have both `--user` and `--username`  both pointing to `<username>:<password>` this would keep us consistent with containerd's and stargz's authentication flow as well as be backwards compatible.
- https://github.com/containerd/stargz-snapshotter/blob/main/docs/overview.md#dockerconfig-based-authentication


What do you think about supporting both `--user` (+ `-u`)  and `--username`?

sondavidb:
Huh, didn't know that `ctr` also used `--user`. Strange then that nerdctl seems to be the different one...

In any case, that seems acceptable to me. We already allow for shorthand `-u` so I think we would only have to alias `--user`.

prafgup:
Ack, I will extent the current solution to support `--user` `-u` and `--username`.

Will refactor the auth logic into a separate function as well, as it would help in my upcoming PR :) 
- https://github.com/awslabs/soci-snapshotter/pull/1864/changes#diff-33710459084d8fa6e66cf3997cc1f3dd06dbc16a8fb434e281608585e42c0b94

prafgup:
> We already allow for shorthand -u so I think we would only have to alias --user

~I dont see any reference where we allow `-u` :) But I'll add that too.~

I see that we actually allow `--user` and `-u` but never use it, instead we used `--username`  😅 - https://github.com/awslabs/soci-snapshotter/blob/53346dda740518d886735e17a0c17e4c2ba7fd61/cmd/soci/commands/internal/registry.go#L49-L53


That means `--username` was never being populated? 🤔  as that is not in allowed flags. Or maybe the Flags list is just for CLI help rendering and allows any arbitrary flags.

Swapnanil-Gupta:
Thanks you for the contribution. Does it also make sense to add a testcase to actually read the username & password from the CLI flags instead of from the Docker config - which would have caught the issue?

prafgup:
Iterating over my previous comment - https://github.com/awslabs/soci-snapshotter/pull/1866#issuecomment-3922905294

Seems like no one was using `--username` as its not one of the allowed flags and the CLI would error out if used.
```
sudo ./out/soci push --username A:B

Incorrect Usage: flag provided but not defined: -username
soci: flag provided but not defined: -username
```

while this works (does not error out)- 
```
sudo ./out/soci push -u A:B
sudo ./out/soci push --user A:B
```

Thus I think the better solution here is to keep only `--user` and alias `-u` :) (and remove unused reference to `--username`) 

I will refactor the same soon and see if its feasible to add a test.  

What do you think @sondavidb  @Swapnanil-Gupta

prafgup:
@sondavidb  @Swapnanil-Gupta Updated the PR and added/updated Integration tests as well to verify behavior with and without `--user` flag.
Feel free to review 😃

sondavidb:
Yeah seems you're right. Guess we were just never populating this field and it was never caught because of what you mentioned about us pre-logging into the registry. So I'm fine with not adding the `--username` alias and just sticking to `--user`, and just amending the flag changes.

~~I would like an integration test for this as well (we can just setup the registry then call `nerdctl logout` on the registry) but I'm also happy to take that up as a followup instead if you'd prefer that.~~ Oops, just saw the newest changes.

integration/push_test.go
```diff
@@ -65,7 +65,7 @@ func TestSociArtifactsPushAndPull(t *testing.T) {
 				t.Fatalf("could not get digest of local content store: %v", err)
 			}
 
-			sh.X("soci", "push", "--user", regConfig.creds(), "--platform", tt.Platform, regConfig.mirror(imageName).ref)
```
sondavidb, line 68: For consistency's sake I'd like to keep this in for now. We really don't need the user flag at all I agree, but I think excluding it in this specific scenario might be confusing, so I'd rather just remove all the `--user` flags at once in a separate PR.

integration/push_test.go
```diff
@@ -81,6 +81,37 @@ func TestSociArtifactsPushAndPull(t *testing.T) {
...
 }
 
+func TestPushWithUserFlag(t *testing.T) {
+	regConfig := newRegistryConfig()
+	sh, done := newShellWithRegistry(t, regConfig)
+	defer done()
+
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t))
+
+	imageName := ubuntuImage
```
sondavidb, line 91: nit: Alpine image would be smaller so might be better here

sondavidb (approved):
Thanks! LGTM on green CI

integration/push_test.go
prafgup: I was thinking the same, to remove `--user` flags from every other test. I'd do it as a follow-ups in another PR :) (have a few more PRs incoming)

integration/push_test.go
prafgup: Will do in a follow-up PR :)

sondavidb:
CI is unrelated, a rerun should do the trick.

We should really add retries on failures. Our testing suite flakiness has reduced considerably IMO but these network issues get irritating at times..

cmd/soci/commands/internal/registry.go
```diff
@@ -81,3 +85,26 @@ var RegistryFlags = []cli.Flag{
 		Usage: "Enable HTTP tracing for registry interactions",
 	},
 }
+
+// ResolveCredentials resolves registry credentials from multiple sources:
+// 1. CLI --user/-u flag (username:password)
+// 2. Docker config file (~/.docker/config.json)
+// 3. Empty credentials for public registries
+func ResolveCredentials(cmd *cli.Command, hostname string) (username, password string) {
+	if cmd.IsSet("user") {
```
Shubhranshu153, line 94: will this break backward compatibility with username?

sondavidb: We discussed above, `--username` was never actually added as a flag. If you try it it will say that `--username` is not a valid flag, while `--user` works. So we are not breaking backwards compatibility, and really we were never taking the credentials to begin with, and never noticed because we were using `nerdctl login` in our testing suite before running any tests.

prafgup:
> CI is unrelated, a rerun should do the trick.
> 
> We should really add retries on failures. Our testing suite flakiness has reduced considerably IMO but these network issues get irritating at times..

Seems fair, I'll add `sh.Retry(3 ..)` for all `sh.X("nerdctl", "pull", ...)` in the followup :)

sondavidb:
That seems acceptable, though of course there are some test cases we want to fail so we should make sure to avoid that.

prafgup:
Again `read tcp 172.19.0.2:41728->3.163.245.13:443: read: connection reset by peer"` 🫠 
@sondavidb can you restart one last time?  If it fails again, I'll add the retries with this PR itself.

sondavidb:
I'm also OK with just pushing this with failing tests FWIW. The rest of the environments are passing without issue.

PR #1867, opened 2026-02-19
Use variables for all SOCI CLI flag names

sondavidb:
**Issue #, if available:**

**Description of changes:**
Should be a no-op in practice, just chores :)

We identified a problem in #1866 where we were looking for the value of `--username` when pushing an image instead of using `--user`. This PR should hopefully help us avoid this mistake in the future.

This should also help us to better identify dead code/flags that are no longer used in the CLI, such as the LabelFlag, which was removed in a separate commit. There's still quite a few pieces of code that are gangling (e.g. many of the registry flags) but some of these might be more desirable in the future.

Added a "global" package as well so CLI packages can refer to globally set variables as needed.

There's probably other housekeeping to be done here (internal package is a bit messy IMO) but this is a good start.

**Testing performed:**
Ensured the changes built via `make`

### DIFF
--- a/cmd/soci/commands/create.go
+++ b/cmd/soci/commands/create.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"slices"
 
+	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/global"
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/internal"
 	"github.com/awslabs/soci-snapshotter/soci"
 	"github.com/awslabs/soci-snapshotter/soci/store"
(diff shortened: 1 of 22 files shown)


sondavidb:
@prafgup just an FYI since I know you're doing a large PR on the CLI. Don't want you to be blindsided by a huge rebase 🙂

prafgup:
> @prafgup just an FYI since I know you're doing a large PR on the CLI. Don't want you to be blindsided by a huge rebase 🙂

Thanks for notifying!! I'll rebase once this is merged 😓

cmd/soci/commands/create.go
```diff
@@ -48,31 +72,10 @@ var CreateCommand = &cli.Command{
...
-			},
-			&cli.BoolFlag{
-				Name:    forceRecreateZtocsFlag,
-				Usage:   "Force recreate zTOCs for layers even if they already exist. Defaults to false.",
-				Value:   false,
-				Aliases: []string{forceRecreateZtocsFlagShort},
-			},
-		),
-		internal.PrefetchFlags()...,
+	Flags: append(append(
```
prafgup: Instead of chaining `append`s we can simply use - `slices.Concat(internal.PlatformFlags, createZtocFlags, X)`

Eg - 
- https://github.com/prafgup/soci-snapshotter/blob/17a2cc67a53dd2a7c0f4b37aec83247c054da173/cmd/soci/commands/convert.go#L65-L68

cmd/soci/commands/ztoc/list.go
```diff
@@ -33,34 +34,40 @@ import (
 	"github.com/urfave/cli/v3"
 )
 
+const (
+	ztocDigestFlag = "ztoc-digest"
+	imageRefFlag   = "image-ref"
+	quietFlag      = "quiet"
```
prafgup, line 40: nit; imo `"quiet"` could be in global.go flags file, I see it being repeted in, index, push and ztoc.

cmd/soci/commands/push.go
```diff
@@ -63,18 +65,18 @@ If multiple soci indices exist for the given image, the most recent one will be
 After pushing the soci artifacts, they should be available in the registry. Soci artifacts will be pushed only
 if they are available in the snapshotter's local content store.
 `,
-	Flags: append(append(append(
+	Flags: append(append(append(append(
```
prafgup: Similar comment to https://github.com/awslabs/soci-snapshotter/pull/1867#discussion_r2827584150
(as well as I think we can refactor other commands as well to use `slices.Concat` instead of `append(append(append(append(.....`)

cmd/soci/commands/ztoc/list.go
prafgup: Also off-topic, I was wondering what would be appropriate for `convert` command, should logs be opt-in (using `verbose`) or opt-out (using `quiet`)
More info - https://github.com/awslabs/soci-snapshotter/pull/1864/

sondavidb: Good point, I do think we can do that but I'd like to keep this PR a no-op. But a global quiet flag would be a little nicer I think.

> should logs be opt-in (using verbose) or opt-out (using quiet)

Tricky to answer as a catch-all, I generally err on the side of less verbose logs if I can help it as certain customers are sensitive on output size for storage reasons. But we should be using `containerd/log` anyway which combined with the `log-level` flag should allow a minimal useful output while providing more verbose logs for debugging. Hope that answers your question 🙂

cmd/soci/commands/push.go
sondavidb: Thanks good catch. I pushed a new commit to do this.

Issue #1057, opened 2024-01-30
[FEATURE] Create a SOCI Index without a Container Runtime


ollypom:
### Description

Today the `soci create` command can only create a SOCI index from a container image that is stored locally to the `soci` client in a containerd image store.

https://github.com/awslabs/soci-snapshotter/blob/db7df3ab2a840bc927417a84448263885b3e21ff/cmd/soci/commands/create.go#L63-L74

However the [SOCI Index Builder](https://github.com/aws-ia/cfn-ecr-aws-soci-index-builder/) is able to generate a SOCI Index by first downloading and extracting a container image to a local arbitrary directory on the host using the [ORAS library](https://github.com/aws-ia/cfn-ecr-aws-soci-index-builder/blob/0de73cb216048b033df31fa304d846937d89fe8d/functions/source/soci-index-generator-lambda/handler.go#L94). Then it imports the [logic](https://github.com/aws-ia/cfn-ecr-aws-soci-index-builder/blob/0de73cb216048b033df31fa304d846937d89fe8d/functions/source/soci-index-generator-lambda/handler.go#L295) from this repository to create the SOCI Index. Showing that its possible to generate a SOCI Index without a container runtime.

For CI/CD use cases it is often challenging to run a container runtime in a build job without adding additional privileges to the runner. So there is appetite to generate Indexes without any dependency on a container runtime, this has been shown by @kichik who has taken the logic out of the SOCI Index Builder into a standalone library https://github.com/CloudSnorkel/standalone-soci-indexer. 

This feature request is to take the same logic as the SOCI Index Builder and put it behind a flag in the `soci create` CLI.

Kern--:
Related to https://github.com/awslabs/soci-snapshotter/issues/760

smoehrle:
Are you willing to accept a PR for this?

I'm not really proficient with Go but might give it a try, since this is something I would like to have/use.

sondavidb:
> Are you willing to accept a PR for this?

PRs are always welcome, this in particular is a longstanding feature request that has been pushed off for a while.

A lot of this existing logic exists in [aws-ia/cfn-ecr-aws-soci-index-builder](https://github.com/aws-ia/cfn-ecr-aws-soci-index-builder) and as someone mentioned above in [CloudSnorkel/standalone-soci-indexer](https://github.com/CloudSnorkel/standalone-soci-indexer), so if your main inspiration will be from either of these it should be fine. We just need to make sure that the file management remains the same.

[...]

Issue #1069, opened 2024-02-06
[FEATURE] Implement `Delete` for SOCI content store


austinvazquez:
### Description

Now that https://github.com/oras-project/oras-go/releases/tag/v2.4.0 has been released, support for `store.Delete` can be added for the SOCI content store. This would help with the orphaned zTOCs issue observed by other issue(s).

### Describe the solution you'd like

Passthrough implementation of `Delete` from SOCI content store to underlying content store.

tuananh:
@austinvazquez oras memory store doesn't seem to add Delete yet so I guess we can't add store.Delete yet?

prafgup:
I see significant amount of memory is being leaked when we delete the images - 

Clean SOCI State - 
```
prafulg@lima-default:/Users/prafulg/Desktop/Work/SOCI-EXPLORE$ sudo du -h --max-depth=1  /var/lib/soci-snapshotter-grpc
4.0K    /var/lib/soci-snapshotter-grpc/soci
4.0K    /var/lib/soci-snapshotter-grpc/unpack
24K     /var/lib/soci-snapshotter-grpc/snapshotter
16K     /var/lib/soci-snapshotter-grpc/content
68K     /var/lib/soci-snapshotter-grpc
```

A single image pull - 
```
prafulg@lima-default:/Users/prafulg/Desktop/Work/SOCI-EXPLORE$ sudo du -h --max-depth=1  /var/lib/soci-snapshotter-grpc
168M    /var/lib/soci-snapshotter-grpc/soci
4.0K    /var/lib/soci-snapshotter-grpc/unpack
757M    /var/lib/soci-snapshotter-grpc/snapshotter
19M     /var/lib/soci-snapshotter-grpc/content
954M    /var/lib/soci-snapshotter-grpc
```

Pulled image is deleted - 
```
prafulg@lima-default:/Volumes/git/prafulg/soci-snapshotter$ sudo nerdctl rmi 2e434b1ccaa5

prafulg@lima-default:/Users/prafulg/Desktop/Work/SOCI-EXPLORE$ sudo du -h --max-depth=1  /var/lib/soci-snapshotter-grpc
8.0K    /var/lib/soci-snapshotter-grpc/soci
4.0K    /var/lib/soci-snapshotter-grpc/unpack
120K    /var/lib/soci-snapshotter-grpc/snapshotter
19M     /var/lib/soci-snapshotter-grpc/content
30M     /var/lib/soci-snapshotter-grpc
```

[...]

PR #1870, opened 2026-02-20
Implement Standalone SOCI Image Convert mode for SOCI index creation without containerd

prafgup:
### NEW PR - https://github.com/awslabs/soci-snapshotter/pull/1881

**Issue #, if available:**
Closes https://github.com/awslabs/soci-snapshotter/issues/1057

**Description of changes:**

This PR adds standalone mode to the `soci convert` command, enabling SOCI index creation without requiring a running containerd daemon. This is particularly useful for CI/CD pipelines and environments where running containerd is impractical without privileged mode.

  ### Key Features

  - **Standalone Mode**: Download images directly from registries using ORAS, create SOCI indexes, and push converted images back to registries
  - **Authentication**: Full support for registry authentication via `--user` flag, Docker config, and environment variables
  - **TLS Options**: Support for `--skip-verify` and `--plain-http` flags
  - **Platform Selection**: Compatible with `--platform` and `--all-platforms` flags
  - **Verbose Output**: Optional `--verbose` flag for detailed progress information

 ### Usage Example

  ```
  soci convert --standalone \
    --user myuser:mypass \
    --verbose \
    source-registry.com/myimage:latest \
    dest-registry.com/myimage:latest-soci  
```

**Testing performed:**

  -  Added new Integration test and all integration tests pass (TestStandaloneConvert*)
    - Basic conversion with validation
    - Specific platform selection (--platform)
    - Authentication with --user flag
    - Error handling (nonexistent images, missing arguments)
    - Idempotency verification
  - Tested on ARM64 architecture (linux/arm64/v8) (lima on macOs)
     - `lima sudo GO_TEST_FLAGS="-run TestStandaloneConvert" make integration`
     - `lima sudo make test`
  - Manual verification of image download, conversion, and push workflow
  
```
sudo ./out/soci convert xxx.xxx.com/xxx:full xxx.xxx.com/xxx:standalone-test-99 --all-platforms --standalone --user $REPOSITORY:$TOKEN  --verbose

Standalone mode: downloading image xxx.xxx.com/xxx:full
Downloading image xxx.xxx.com/xxx:full from registry...
  Downloading fd398e9fa269 (application/vnd.oci.image.layer.v1.tar+gzip, 5616874 bytes)
  Downloading 12bb71ceae34 (application/vnd.oci.image.layer.v1.tar+gzip, 221568 bytes)
  Downloading d28cb90eb7da (application/vnd.oci.image.config.v1+json, 17958 bytes)
  Downloading ef6465a1674b (application/vnd.oci.image.layer.v1.tar+gzip, 78 bytes)
  Downloading 3e9c02b40ec1 (application/vnd.oci.image.layer.v1.tar+gzip, 14489277 bytes)
  Downloading 7511aa994dfe (application/vnd.oci.image.layer.v1.tar+gzip, 3222 bytes)
  Downloading 5b91cf883cde (application/vnd.oci.image.layer.v1.tar+gzip, 10392231 bytes)
  Downloading f6a26d16afc7 (application/vnd.oci.image.layer.v1.tar+gzip, 10395600 bytes)
  Downloading 4f4fb700ef54 (application/vnd.oci.image.layer.v1.tar+gzip, 32 bytes)
  Downloading d8c6b361b623 (application/vnd.oci.image.layer.v1.tar+gzip, 271302478 bytes)
  Downloading faf5416d559a (application/vnd.oci.image.manifest.v1+json, 2383 bytes)
Successfully downloaded image
Converting image to SOCI-enabled format with 1 platform(s)
ztoc skipped - layer sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1 (application/vnd.oci.image.layer.v1.tar+gzip) size 32 is less than min-layer-size 10485760
ztoc skipped - layer sha256:5b91cf883cde9e49909af81cdce1cc89948745e38796395b0a90d859d7db97d2 (application/vnd.oci.image.layer.v1.tar+gzip) size 10392231 is less than min-layer-size 10485760
ztoc skipped - layer sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1 (application/vnd.oci.image.layer.v1.tar+gzip) size 32 is less than min-layer-size 10485760
ztoc skipped - layer sha256:ef6465a1674b258144bd7eac86347823926436a5d53f100cadfe098c82e4c4d4 (application/vnd.oci.image.layer.v1.tar+gzip) size 78 is less than min-layer-size 10485760
ztoc skipped - layer sha256:f6a26d16afc76b03f36290b56770ae0db2ab11aab759d2c7bd49bdfc865b3b87 (application/vnd.oci.image.layer.v1.tar+gzip) size 10395600 is less than min-layer-size 10485760
ztoc skipped - layer sha256:fd398e9fa269e6877ec10f14d61837fc23f05d757530fed2946073018dc07705 (application/vnd.oci.image.layer.v1.tar+gzip) size 5616874 is less than min-layer-size 10485760
ztoc skipped - layer sha256:12bb71ceae3401038ce93aa7ce3a115f4a5cb3e9a85277ef4a97718c0fc9298e (application/vnd.oci.image.layer.v1.tar+gzip) size 221568 is less than min-layer-size 10485760
ztoc skipped - layer sha256:ef6465a1674b258144bd7eac86347823926436a5d53f100cadfe098c82e4c4d4 (application/vnd.oci.image.layer.v1.tar+gzip) size 78 is less than min-layer-size 10485760
ztoc skipped - layer sha256:7511aa994dfed5f54c697fcec6db7176449e7d87317b3f6ad093185182ce01c7 (application/vnd.oci.image.layer.v1.tar+gzip) size 3222 is less than min-layer-size 10485760
layer sha256:d8c6b361b623cabb0a89fad4a7d491d7d543f0834ca0c581e0176e1b76a68a15 -> ztoc sha256:ea8f91d2a57139b053186c8a1ae1901684bc06fda5b7d00aef0bd3d6d4e9e4c1
Successfully created SOCI-enabled image: sha256:f265a1fddffc035eb949ac3ff03a597ce4c823b8bde2ae6cb63a0c85f69d0de9
Pushing SOCI-enabled image to xxx.xxx.com/xxx:standalone-test-99...
Successfully pushed SOCI-enabled image to xxx.xxx.com/xxx:standalone-test-99 (digest: sha256:f265a1fddffc035eb949ac3ff03a597ce4c823b8bde2ae6cb63a0c85f69d0de9)
```

Integration tests -

```
sudo GO_TEST_FLAGS="-run TestStandalone -count=1" make integration
cd cmd/ ; GO111MODULE=auto go build -o /Volumes/git/prafulg/soci-snapshotter/out/soci-snapshotter-grpc  -ldflags '-X github.com/awslabs/soci-snapshotter/version.Version=be11c954.m -X github.com/awslabs/soci-snapshotter/version.Revision=be11c954b73ee597b504985deea72eebdefce2f2.m  -s -w '  ./soci-snapshotter-grpc
cd cmd/ ; GO111MODULE=auto go build -o /Volumes/git/prafulg/soci-snapshotter/out/soci  -ldflags '-X github.com/awslabs/soci-snapshotter/version.Version=be11c954.m -X github.com/awslabs/soci-snapshotter/version.Revision=be11c954b73ee597b504985deea72eebdefce2f2.m  -s -w '  ./soci
integration
=== RUN   TestStandaloneConvertBasic
=== PAUSE TestStandaloneConvertBasic
=== RUN   TestStandaloneConvertSpecificPlatform
=== PAUSE TestStandaloneConvertSpecificPlatform
=== RUN   TestStandaloneConvertWithUserAuth
=== PAUSE TestStandaloneConvertWithUserAuth
=== RUN   TestStandaloneInvalidConversion
=== PAUSE TestStandaloneInvalidConversion
=== RUN   TestStandaloneConvertIdempotent
=== PAUSE TestStandaloneConvertIdempotent
=== CONT  TestStandaloneConvertBasic
=== CONT  TestStandaloneInvalidConversion
=== CONT  TestStandaloneConvertSpecificPlatform
=== CONT  TestStandaloneConvertWithUserAuth
=== CONT  TestStandaloneConvertIdempotent
=== RUN   TestStandaloneInvalidConversion/nonexistent_image
=== RUN   TestStandaloneInvalidConversion/missing_destination_ref
--- PASS: TestStandaloneInvalidConversion (1.38s)
    --- PASS: TestStandaloneInvalidConversion/nonexistent_image (0.11s)
    --- PASS: TestStandaloneInvalidConversion/missing_destination_ref (0.03s)
--- PASS: TestStandaloneConvertSpecificPlatform (21.16s)
--- PASS: TestStandaloneConvertWithUserAuth (21.28s)
--- PASS: TestStandaloneConvertIdempotent (22.26s)
--- PASS: TestStandaloneConvertBasic (22.51s)
PASS
ok      github.com/awslabs/soci-snapshotter/integration 132.773s
```

### DIFF
--- a/cmd/soci/commands/convert.go
+++ b/cmd/soci/commands/convert.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/awslabs/soci-snapshotter/cmd/soci/commands/internal"
@@ -61,9 +62,9 @@ var ConvertCommand = &cli.Command{
 	Name:      "convert",
 	Usage:     "convert an OCI image to a SOCI enabled image",
 	ArgsUsage: "[flags] <image_ref> <dest_ref>",
-	Flags: append(
-		append(
-			internal.PlatformFlags,
+	Flags: slices.Concat(
+		internal.PlatformFlags,
+		[]cli.Flag{
 			&cli.Int64Flag{
 				Name:  spanSizeFlag,
 				Usage: "Span size that soci index uses to segment layer data. Default is 4 MiB",
@@ -84,8 +85,18 @@ var ConvertCommand = &cli.Command{
 				Value:   false,
 				Aliases: []string{forceRecreateZtocsFlagShort},
 			},
-		),
-		internal.PrefetchFlags()...,
+			&cli.BoolFlag{
+				Name:  "standalone",
+				Usage: "Run in standalone mode without containerd runtime. In this mode, the command will download the source image, perform the conversion, and push the converted image back to the registry without requiring a running containerd instance.",
+			},
+			&cli.BoolFlag{
+				Name:    "verbose",
+				Aliases: []string{"v"},
+				Usage:   "Show detailed progress output",
+			},
+		},
+		internal.RegistryFlags,
+		internal.PrefetchFlags(),
 	),
 	Action: func(ctx context.Context, cmd *cli.Command) error {
 		srcRef := cmd.Args().Get(0)
@@ -103,13 +114,8 @@ var ConvertCommand = &cli.Command{
 			return fmt.Errorf("%w: %w", ErrInvalidDestRef, err)
 		}
 
-		var optimizations []soci.Optimization
-		for _, o := range cmd.StringSlice(optimizationFlag) {
-			optimization, err := soci.ParseOptimization(o)
-			if err != nil {
-				return err
-			}
-			optimizations = append(optimizations, optimization)
+		if cmd.Bool("standalone") {
+			return runStandaloneConvert(ctx, cmd, srcRef, dstRef)
 		}
 
 		client, ctx, cancel, err := internal.NewClient(ctx, cmd)
@@ -125,10 +131,6 @@ var ConvertCommand = &cli.Command{
 			return err
 		}
 
-		spanSize := cmd.Int64(spanSizeFlag)
-		minLayerSize := cmd.Int64(minLayerSizeFlag)
-		forceRecreateZtocs := cmd.Bool(forceRecreateZtocsFlag)
-
 		blobStore, err := store.NewContentStore(internal.ContentStoreOptions(ctx, cmd)...)
 		if err != nil {
 			return err
@@ -139,25 +141,13 @@ var ConvertCommand = &cli.Command{
 			return err
 		}
 
-		builderOpts := []soci.BuilderOption{
-			soci.WithMinLayerSize(minLayerSize),
-			soci.WithSpanSize(spanSize),
-			soci.WithBuildToolIdentifier(buildToolIdentifier),
-			soci.WithOptimizations(optimizations),
-			soci.WithArtifactsDb(artifactsDb),
-			soci.WithForceRecreateZtocs(forceRecreateZtocs),
-		}
-
-		allPrefetchFiles, err := internal.ParsePrefetchFiles(cmd)
+		builderOpts, err := parseBuilderOptions(cmd)
 		if err != nil {
-			return fmt.Errorf("failed to parse prefetch files: %w", err)
+			return err
 		}
-		if len(allPrefetchFiles) > 0 {
-			builderOpts = append(builderOpts, soci.WithPrefetchPaths(allPrefetchFiles))
-		}
+		builderOpts = append(builderOpts, soci.WithArtifactsDb(artifactsDb))
 
 		builder, err := soci.NewIndexBuilder(cs, blobStore, builderOpts...)
-
 		if err != nil {
 			return err
 		}
@@ -201,4 +191,98 @@ var ConvertCommand = &cli.Command{
 
 		return err
 	},
+}
+
+// runStandaloneConvert runs the convert command in standalone mode (without containerd)
+func runStandaloneConvert(ctx context.Context, cmd *cli.Command, srcRef string, dstRef string) error {
+	verbose := cmd.Bool("verbose")
+	if verbose {
+		fmt.Printf("Standalone mode: downloading image %s\n", srcRef)
+	}
+
+	imageInfo, err := internal.DownloadImageFromRegistry(ctx, cmd, srcRef)
+	if err != nil {
+		return err
+	}
+	defer imageInfo.Cleanup()
+
+	builderOpts, err := parseBuilderOptions(cmd)
+	if err != nil {
+		return err
+	}
+
+	artifactsDb, err := soci.NewDB(soci.ArtifactsDbPath(imageInfo.TempDir))
+	if err != nil {
+		return fmt.Errorf("failed to create artifacts database: %w", err)
+	}
+	builderOpts = append(builderOpts, soci.WithArtifactsDb(artifactsDb))
+
+	sociStore := &store.SociStore{Store: imageInfo.OrasStore}
+
+	builder, err := soci.NewIndexBuilder(imageInfo.ContentStore, sociStore, builderOpts...)
+	if err != nil {
+		return err
+	}
+
+	requestedPlatforms, err := internal.GetPlatforms(ctx, cmd, imageInfo.Image, imageInfo.ContentStore)
+	if err != nil {
+		return err
+	}
+
+	batchCtx, done, err := sociStore.BatchOpen(ctx)
+	if err != nil {
+		return err
+	}
+	defer done(ctx)
+
+	if verbose {
+		fmt.Printf("Converting image to SOCI-enabled format with %d platform(s)\n", len(requestedPlatforms))
+	}
+
+	convertedDesc, err := builder.Convert(batchCtx, imageInfo.Image, soci.ConvertWithPlatforms(requestedPlatforms...))
+	if err != nil {
+		return err
+	}
+
+	if verbose {
+		fmt.Printf("Successfully created SOCI-enabled image: %s\n", convertedDesc.Digest)
+	}
+
+	if err := internal.PushConvertedImage(ctx, cmd, imageInfo, *convertedDesc, dstRef); err != nil {
+		return err
+	}
+
+	if verbose {
+		fmt.Println("Standalone mode: SOCI conversion completed successfully")
+	}
+	return nil
+}
+
+func parseBuilderOptions(cmd *cli.Command) ([]soci.BuilderOption, error) {
+	var optimizations []soci.Optimization
+	for _, o := range cmd.StringSlice("optimizations") {
+		optimization, err := soci.ParseOptimization(o)
+		if err != nil {
+			return nil, err
+		}
+		optimizations = append(optimizations, optimization)
+	}
+
+	builderOpts := []soci.BuilderOption{
+		soci.WithMinLayerSize(cmd.Int64(minLayerSizeFlag)),
+		soci.WithSpanSize(cmd.Int64(spanSizeFlag)),
+		soci.WithBuildToolIdentifier(buildToolIdentifier),
+		soci.WithOptimizations(optimizations),
+		soci.WithForceRecreateZtocs(cmd.Bool(forceRecreateZtocsFlag)),
+	}
+
+	allPrefetchFiles, err := internal.ParsePrefetchFiles(cmd)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse prefetch files: %w", err)
+	}
+	if len(allPrefetchFiles) > 0 {
+		builderOpts = append(builderOpts, soci.WithPrefetchPaths(allPrefetchFiles))
+	}
+
+	return builderOpts, nil
 }
--- a/cmd/soci/commands/internal/standalone.go
+++ b/cmd/soci/commands/internal/standalone.go
@@ -1,0 +1,205 @@
+/*
+   Copyright The Soci Snapshotter Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package internal
+
+import (
+	"cmp"
+	"context"
+	"crypto/tls"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/awslabs/soci-snapshotter/soci"
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/content/local"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/reference"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/urfave/cli/v3"
+	oraslib "oras.land/oras-go/v2"
+	"oras.land/oras-go/v2/content/oci"
+	"oras.land/oras-go/v2/registry/remote"
+	"oras.land/oras-go/v2/registry/remote/auth"
+)
+
+type StandaloneImageInfo struct {
+	ContentStore content.Store
+	Image        images.Image
+	TempDir      string
+	OrasStore    *oci.Store
+}
+
+func (info *StandaloneImageInfo) Cleanup() {
+	if info.TempDir != "" {
+		cleanupTempDir(info.TempDir)
+	}
+}
+
+// DownloadImageFromRegistry downloads an image from a registry using ORAS
+// and returns a content store and image metadata that can be used by IndexBuilder.
+func DownloadImageFromRegistry(ctx context.Context, cmd *cli.Command, imageRef string) (*StandaloneImageInfo, error) {
+	refspec, err := reference.Parse(imageRef)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse image reference: %w", err)
+	}
+
+	tmpDir, err := os.MkdirTemp("", "soci-*")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temp directory: %w", err)
+	}
+
+	orasStore, err := oci.New(tmpDir)
+	if err != nil {
+		cleanupTempDir(tmpDir)
+		return nil, fmt.Errorf("failed to create ORAS store: %w", err)
+	}
+
+	repo, err := newRemoteRepo(cmd, refspec)
+	if err != nil {
+		cleanupTempDir(tmpDir)
+		return nil, fmt.Errorf("failed to create remote repository: %w", err)
+	}
+
+	verbose := cmd.Bool("verbose")
+	if verbose {
+		fmt.Printf("Downloading image %s from registry...\n", imageRef)
+	}
+
+	manifestDesc, err := repo.Resolve(ctx, refspec.Object)
+	if err != nil {
+		cleanupTempDir(tmpDir)
+		return nil, fmt.Errorf("failed to resolve image %s: %w", imageRef, err)
+	}
+
+	copyOpts := oraslib.DefaultCopyGraphOptions
+	if verbose {
+		copyOpts.PreCopy = func(_ context.Context, desc ocispec.Descriptor) error {
+			mediaType := cmp.Or(desc.MediaType, soci.SociLayerMediaType)
+			fmt.Printf("  Downloading %s (%s, %d bytes)\n", desc.Digest.Encoded()[:12], mediaType, desc.Size)
+			return nil
+		}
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by prafgup** [comment]
Hey SOCI Devs 👋 
Would appreciate a review here.
cc @Shubhranshu153 @sondavidb @Swapnanil-Gupta

Also happy to followup with any changes (in this or a followup PR) 😄

File: cmd/soci/commands/internal/standalone.go [hunk 1]
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	tmpDir := filepath.Join(os.TempDir(), fmt.Sprintf("soci-%s", uuid.New().String()))
+	if err := os.MkdirAll(tmpDir, 0755); err != nil {
+		return nil, fmt.Errorf("failed to create temp directory: %w", err)
+	}
+
+	orasStore, err := oci.New(tmpDir)
+	if err != nil {
+		cleanupTempDir(tmpDir)
```
VVKot, line 68: Can we `defer` this instead of doing it in every error branch? You probably want to cleanup on success as well?

prafgup: I had a look earlier into refactoring this and figured that it would require us having named returns and then checking error in the defer func. 
Overall using named returns make things somewhat complex to read, thus I let it be like this where we would cleanup on specific errors.

We do always cleanup in success scenarios here upstream - https://github.com/prafgup/soci-snapshotter/blob/83db223b6161ecde510d4a437d432eebb81baf35/cmd/soci/commands/convert.go#L207 lemme check if we can remove this manual cleanup then in case of errors and just move `imageInfo.Cleanup()` to be before error.

prafgup: So in case of error upstream `runStandaloneConvert` wont have `imageInfo` to trigger a cleanup, thus I think its fine to keep it as is. Upstream always makes sure to cleanup in case `DownloadImageFromRegistry` [is successful](https://github.com/prafgup/soci-snapshotter/blob/83db223b6161ecde510d4a437d432eebb81baf35/cmd/soci/commands/convert.go#L207), and in case of failure in `DownloadImageFromRegistry` we would manually cleanup before returning a error.

cmd/soci/commands/internal/standalone.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, fmt.Errorf("failed to create ORAS store: %w", err)
+	}
+
+	repo, err := newRemoteRepo(cmd, refspec)
+	if err != nil {
+		cleanupTempDir(tmpDir)
+		return nil, fmt.Errorf("failed to create remote repository: %w", err)
+	}
+
+	verbose := cmd.Bool("verbose")
```
Shubhranshu153, line 78: nit: Can this be from the global debug flags.

cmd/soci/commands/internal/standalone.go
```diff
@@ -0,0 +1,207 @@
...
+	if verbose {
+		fmt.Printf("Downloading image %s from registry...\n", imageRef)
```
Shubhranshu153, line 80: suggest using logging

cmd/soci/commands/internal/standalone.go
```diff
@@ -0,0 +1,207 @@
...
+// DownloadImageFromRegistry downloads an image from a registry using ORAS
+// and returns a content store and image metadata that can be used by IndexBuilder.
+func DownloadImageFromRegistry(ctx context.Context, cmd *cli.Command, imageRef string) (*StandaloneImageInfo, error) {
+	refspec, err := reference.Parse(imageRef)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse image reference: %w", err)
+	}
+
+	tmpDir := filepath.Join(os.TempDir(), fmt.Sprintf("soci-%s", uuid.New().String()))
+	if err := os.MkdirAll(tmpDir, 0755); err != nil {
```
Shubhranshu153: Does it create a risk where non root actors have access to the image layers albiet temporary. Seems like a potential risk here.

cmd/soci/commands/internal/standalone.go
prafgup: I saw other commands such as `push` was using `fmt.Printf` as well for logging, thus kept it consistent. Should I move the current Printf logging to use context logger instead?

- https://github.com/awslabs/soci-snapshotter/blob/b4baddbf466ba5b8a520ee7539b0d9fd051563d5/cmd/soci/commands/push.go#L253

cmd/soci/commands/internal/standalone.go
prafgup: Thanks, I think it makes sense to use debug for logging instead of `verbose` I will change :)

cmd/soci/commands/internal/standalone.go
prafgup: I will try to restrict the permissions to `0700` instead here, or maybe replace with `os.MkdirTemp` as it defaults to `0700` permissions.

cmd/soci/commands/internal/standalone.go
Shubhranshu153: i am not sure the historical context for it. let me get back to this.

Shubhranshu153: ok it seems any user specific cli outputs are via printf via convention so its correct here to use printf. Apologies for the confusion.
As a result we can keep the verbose

cmd/soci/commands/convert.go
```diff
@@ -139,25 +141,13 @@ var ConvertCommand = &cli.Command{
...
-			soci.WithMinLayerSize(minLayerSize),
-			soci.WithSpanSize(spanSize),
-			soci.WithBuildToolIdentifier(buildToolIdentifier),
-			soci.WithOptimizations(optimizations),
-			soci.WithArtifactsDb(artifactsDb),
-			soci.WithForceRecreateZtocs(forceRecreateZtocs),
-		}
-
-		allPrefetchFiles, err := internal.ParsePrefetchFiles(cmd)
+		builderOpts, err := parseBuilderOptions(cmd)
```
Shubhranshu153, line 144: nit: This seems like a refactor, ideally we create a separate commit for it as the purpose is different.

cmd/soci/commands/internal/standalone.go
prafgup: @Shubhranshu153 Updated to use `os.MkdirTemp` - https://github.com/awslabs/soci-snapshotter/compare/83db223b6161ecde510d4a437d432eebb81baf35..ddb23f1ea17a214f33294a076b72fd07769e6440

Also, just to note all the previous checks (builds/integration tests) had succeeded.

cmd/soci/commands/convert.go
Shubhranshu153: i see its for the purpose of deduplication of logic introduced

prafgup: I had to use the same logic in my new `runStandaloneConvert` function, thus thought to refactor from the existing convert command.

Will make note to refactor in separate PR from now on :)

cmd/soci/commands/convert.go
```diff
@@ -202,3 +192,97 @@ var ConvertCommand = &cli.Command{
...
 }
+
+// runStandaloneConvert runs the convert command in standalone mode (without containerd)
+func runStandaloneConvert(ctx context.Context, cmd *cli.Command, srcRef string, dstRef string) error {
+	verbose := cmd.Bool("verbose")
+	if verbose {
+		fmt.Printf("Standalone mode: downloading image %s\n", srcRef)
+	}
+
+	imageInfo, err := internal.DownloadImageFromRegistry(ctx, cmd, srcRef)
```
Shubhranshu153, line 203: seems we are adding a operation which is not strictly done by convert. Soci doesnt have a pull command probably because it assumes ctr/ other cli are providing the image is there. 

Overall seems like architecturally different to the purpose of convert. Any reason we cannot pull image to soci content store and do the convert and then push. Probably adds complexity vs being architecturally consistent.

prafgup: Right, if we were to think of this feature as of multiple steps (pull + convert + push) it would be a bit overly complex where - 
- We would have to maintain a persistent local content store across command invocations, essentially creating a replacement for containerd store.
- Let a lot of abstractions from `StandaloneImageInfo`  be handled by user across commands.
- This feature is usually going to be used by CI/CD jobs, where having a single command is much simpler for automation.

I believe having a bit of tradeoff of consistency for simplicity would be the better choice here, as we don't see much extension of standalone convert in future atleast architecturally. 

On a side note we do currently have deprecated `push` and `create` command which does not support V2 Manifests - https://github.com/awslabs/soci-snapshotter/blob/main/docs/cli-usage.md#soci-create and seems like we focused more on the convert command for V2 Manifests, thus I believe its the right place for standalone convert too.

What do you think? 😄

Shubhranshu153: Yeah i agree with the cost but doing a push and pull in convert seems a bit hacky.  Having a single command is ok as long as it is fundamentally doing a specific operation it was designed for. Still think having a pull/convert and push command here makes better architectural sense. For daemon less services we can use the soci content store.

@sondavidb what do you think?

prafgup: Hey @Shubhranshu153 @sondavidb :)
Requesting an update here regarding this thread for the path forward ^ (this is blocking some development on our end 😓)

One of the other possible option that I can think of apart from keeping this as a flag in `convert` instead, is that we can have another independent command something like `standalone-convert` that would do this flow of `convert --standalone`. 
But I do believe that with proper documentation `convert --standalone` does seem to be the right place for this logic (of converting a image upstream) instead of expanding it across multiple pull/convert/push commands (where push has already been deprecated for V2 use).

Shubhranshu153: if we really dont want to architect it entirely one way can be to introduce to introduce a --pull and use the soci store but push i think still needs to be a separate cli command. And push is not deprecated rather its not extended to v2 migration. Having pull/convert/push in a single cli is inherently changing the cmd purpose.

prafgup: Ack, so I'd assume this would be the flow - 

```
soci convert --pull --standalone IMG_A IMG_B
soci push --standalone IMG_B
```

- In future integrate the `--pull` command to also pull for the normal containerd convert flow as well (for this PR i'll keep it out for now).
- For push we would use the same SOCI store location (location used to store the IMG_B) and assume the image `IMG_B` to be present (else error out).
  - ps. We would not cleanup the docker image and let it be there ig? (or maybe in future we can extend this to something like `soci push --standalone IMG_B --cleanup` or something, for now I'll keep it simple and not delete the downloaded tar image.

What are your thoughts? Does it sound about right? If so I'd implement this flow 😃 
cc @Shubhranshu153 @sondavidb

Shubhranshu153:
```
soci convert --pull IMG_A IMG_B
soci push IMG_B
```

Also we probably need a remove command to clean the oci store
```
soci image rm <> or
soci rmi <>
```

For any of the above operation if the content store is not soci we return "not implemented". Before we proceed though i want David to read thru it, have messaged to take a look.

sondavidb: > push is not deprecated rather its not extended to v2 migration

It's not deprecated correct, however it's also not the right thing to do as pushing a SOCI index standalone is extremely unlikely to be what you want. That's kind of the idea here. You want to pull with your CLI command of choice, convert it to a SOCI v2 image, then push the image, which will in turn push the SOCI index with it. `push` just doesn't really fit in here because you can't just change a SOCI index, since doing so updates the digest which means the image needs to be reconverted.

Apologies for the delay, been caught up in other things but I am reviewing the rest of it now.

sondavidb: > Also we probably need a remove command to clean the oci store

See https://github.com/awslabs/soci-snapshotter/issues/1069, we can't do this in the SOCI content store with our abstraction. But if we have the root in a tmpdir it should be ok?

cmd/soci/commands/convert.go
```diff
@@ -202,3 +192,97 @@ var ConvertCommand = &cli.Command{
...
+		return err
+	}
+
+	if verbose {
+		fmt.Println("Standalone mode: SOCI conversion completed successfully")
+	}
+	return nil
+}
+
+func parseBuilderOptions(cmd *cli.Command) ([]soci.BuilderOption, error) {
```
sondavidb, line 261: This is also used for `create`, we should probably do some deduping here. Happy to take that as a followup though.

cmd/soci/commands/internal/standalone.go
sondavidb: I'm not really a fan of this pattern since we have to be sure every single error out handles this error which is not very Go-like.

I've seen a couple of patterns that I think would flow a little nicer:
1. Do initial setup for the tempdir, then have a separate function (e.g. `downloadImageFromRegistry`) that returns an error. Call that func from `DownloadImageFromRegistry` and on error call the cleanup func. This is probably the least error-prone way to do this.
2. Declare error var after setting up the dir, then defer a func that calls the cleanup func if `err != nil`. I generally prefer this option less as if you declare an error within a nested function and forget to refer to the global err var you can inadvertently not call cleanup.

cmd/soci/commands/internal/standalone.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	if verbose {
+		fmt.Printf("Successfully pushed to %s (digest: %s)\n", dstRef, convertedDesc.Digest)
+	}
+
+	return nil
+}
+
+func newRemoteRepo(cmd *cli.Command, refspec reference.Spec) (*remote.Repository, error) {
```
sondavidb, line 167: Is it possible to reuse anything from the `fs` package? (I haven't checked but I'm wary of re-duplicating logic, especially when it comes to registry authentication.)

cmd/soci/commands/internal/standalone.go
```diff
@@ -0,0 +1,205 @@
...
+	authClient := &auth.Client{
+		Credential: func(_ context.Context, host string) (auth.Credential, error) {
+			return auth.Credential{
+				Username: username,
+				Password: password,
+			}, nil
+		},
+	}
+
+	if cmd.Bool("skip-verify") {
```
sondavidb, line 194: This seems inverted?
```suggestion
	if !cmd.Bool("skip-verify") {
```

integration/standalone_convert_test.go
```diff
@@ -0,0 +1,220 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/awslabs/soci-snapshotter/config"
+	"github.com/containerd/platforms"
+)
+
+func TestStandaloneConvertBasic(t *testing.T) {
+	t.Parallel()
```
sondavidb, line 28: nit: Little torn on this but OK with it. Our integration tests run a separate container for each test (`newShellWithRegistry` I believe composes a container) so parallelizing too much might be a lot for the runner.

cmd/soci/commands/convert.go
prafgup: @sondavidb thanks for the review, I glanced over other commends and would address them later. But just want to confirm this main flow. 

>You want to pull with your CLI command of choice, convert it to a SOCI v2 image, then push the image, which will in turn push the SOCI index with it.

Just to confirm the wording, are you suggesting something like - 
```
crane pull nginx:latest IMG_TAR_PATH

soci convert --standalone IMG_TAR_PATH IMG_OUTPUT_TAR_PATH

crane push IMG_OUTPUT_TAR_PATH my-reg/my-image:my-tag
```

I believe this makes sense as well. A user can pull the image as tar in some path and we would load that up and index it and store as output. Also in this case we won't have to do any registry authentications as we are neither pulling nor pushing.

Would you prefer this logic to be made explicit by user sending in `--standalone` flag in create command? Or should we auto detect that the path given (input and output) is from disk and if a tar file exist there proceed with conversion without containerd?

Shubhranshu153: @david for v2 don't we push the entire image with the index together. Here standalone means not using containerd rather doing everything using the soci cli

Also the delete option needs to be created but if soci store is used to pull convert and push images as we need to clean it up also. Or we create tmp store abstraction. (Still I think a delete is needed)

Shubhranshu153: I really like the idea of using some other cli to pull and push images if that's an option @prafgup

sondavidb: Yeah, what you said would be my ideal scenario @prafgup. It is honestly probably the least intrusive solution 🙂

prafgup: Hey @sondavidb  @Shubhranshu153 ,

I have created another PR here - https://github.com/awslabs/soci-snapshotter/pull/1881

Fundamentally its a bit different from this one thus thought of closing this PR and creating a new clean one.

Would appreciate a Review on the same 😄 (I'll update the docs in a followup once that is merged. Also happy to resolve any small nits and refactoring as followups there as well as Update the docs in followups)

PR #1871, opened 2026-02-23
Support concatenated gzip (mgzip/pigz) layers in ztoc

prafgup:
**Issue #, if available:**
Fixes #1834. The C-level zinfo code only processed the first gzip member, causing all ztoc entries to map to span 0 for mgzip-compressed layers. This adds handling for multiple concatenated gzip members in all three functions: index generation, file-based extraction, and buffer-based extraction.

**Description of changes:**

- Handle concatenated gzip members in `generate_zinfo_from_fp` by detecting additional data after Z_STREAM_END and resetting inflate state for the next member
  - Ref - https://github.com/madler/zlib/blob/master/examples/zran.c#L221-L228
- Handle concatenated gzip members in both extraction functions (`extract_data_from_fp`, `extract_data_from_buffer`) by manually skipping the 8-byte gzip trailer on the first member boundary (raw inflate mode), then switching to gzip mode (windowBits=47) which auto-consumes subsequent trailers
  -  Ref - https://github.com/madler/zlib/blob/master/examples/zran.c#L418-L436

**Testing performed:**
- Added unit test - 
```
lima sudo go test -v -count=1 -run TestDecompressWithPigz ./ztoc/...
=== RUN   TestDecompressWithPigz
=== RUN   TestDecompressWithPigz/span_size_64KB
=== RUN   TestDecompressWithPigz/span_size_256KB
=== RUN   TestDecompressWithPigz/span_size_1MB
--- PASS: TestDecompressWithPigz (0.53s)
    --- PASS: TestDecompressWithPigz/span_size_64KB (0.17s)
    --- PASS: TestDecompressWithPigz/span_size_256KB (0.17s)
    --- PASS: TestDecompressWithPigz/span_size_1MB (0.18s)
PASS
ok      github.com/awslabs/soci-snapshotter/ztoc        0.540s
PASS
ok      github.com/awslabs/soci-snapshotter/ztoc/compression    0.002s [no tests to run]
?       github.com/awslabs/soci-snapshotter/ztoc/compression/fbs/zinfo  [no test files]
?       github.com/awslabs/soci-snapshotter/ztoc/fbs/ztoc       [no test files]
```

```
lima sudo GO_TEST_FLAGS="-run TestSociZtocWithEstargzLayers -count=1 -timeout 10m" make integration


 SOCI_SNAPSHOTTER_PROJECT_ROOT=/Volumes/git/prafulg/soci-snapshotter
 === RUN   TestSociZtocWithEstargzLayers
[...]

PR #1874, opened 2026-02-25
Port GzipZinfo to Go

kzys:
GzipZinfo was implemented in C because;

1. The original implementation, zran.c was in C
2. compress/flate didn't expose the necessary APIs to implement the logic in Go

This commit ports GzipZinfo to pure Go by forking compress/flate and exposing the APIs we need.

**Issue #, if available:**

**Description of changes:**

**Testing performed:**

### DIFF
--- a/ztoc/compression/flate/dict_decoder.go
+++ b/ztoc/compression/flate/dict_decoder.go
@@ -1,0 +1,149 @@
+// Copyright 2016 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Forked from Go stdlib compress/flate/dict_decoder.go.
+// Added Window() method to export sliding window state for zinfo checkpoints.
+
+package flate
+
+const WindowSize = 32768
+
+// dictDecoder implements the LZ77 sliding dictionary as used in decompression.
+type dictDecoder struct {
+	hist []byte // Sliding window history
+
+	// Invariant: 0 <= rdPos <= wrPos <= len(hist)
+	wrPos int  // Current output position in buffer
+	rdPos int  // Have emitted hist[:rdPos] already
+	full  bool // Has a full window length been written yet?
+
+	totalWritten int64 // Total uncompressed bytes written (across all wraps)
+}
+
+func (dd *dictDecoder) init(size int, dict []byte) {
+	*dd = dictDecoder{hist: dd.hist}
+
+	if cap(dd.hist) < size {
+		dd.hist = make([]byte, size)
+	}
+	dd.hist = dd.hist[:size]
+
+	if len(dict) > len(dd.hist) {
+		dict = dict[len(dict)-len(dd.hist):]
+	}
+	dd.wrPos = copy(dd.hist, dict)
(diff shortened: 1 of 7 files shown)


kzys:
Before connecting the component to the rest of the snapshotter, I'd like to discuss about the following topics;

- Whether we are fine forking compless/flate or not.
- How to switch the implementations. I personally want to switch them in runtime, but it is possible to have build flags.
- Package structure

prafgup:
A couple of general queries - 
- Will this go based implementation be as performant as C based existing implementations? Did you benchmark the same? would be interesting to see the performance gain/degradation.
- Will this support pigz based parallel gzip compressions? I was working to add it here - https://github.com/awslabs/soci-snapshotter/pull/1871

sondavidb:
Hey Kaz. Great to see you again 🙂 

> Whether we are fine forking compless/flate or not.

I don't like it but it also seems to be the best way to achieve the API we want. Based on the [change history](https://cs.opensource.google/go/go/+/master:src/compress/flate/;bpv=1) the package seems to get a few modifications a year so I'm a little concerned about that rate of change, however I don't think drifting is a huge issue for a library that serves a pretty static function. So it's not my favorite but probably the best choice to achieve this.

There is also the question of licensing, though. From my understanding of BSD-3 Clause we need to display the [license](https://pkg.go.dev/compress/flate?tab=licenses) somewhere in the distribution. In the package is OK by me but I hope I'm not missing anything from a legal standpoint here.

> How to switch the implementations. I personally want to switch them in runtime, but it is possible to have build flags.

I don't see any reason to use the C version outside of performance reasons. I'd much rather just choose one or the other. (Unless I'm misunderstanding, which if so, my apologies.)

> Package structure

This is something I've been trying to figure out for a while, I would very much like to have a folder of packages that are not actually our code (e.g. we use some id-mapping shenanigans in https://github.com/awslabs/soci-snapshotter/tree/main/idtools which is just copied code from https://github.com/containerd/containerd/commit/83aaa89). I'm not opinionated on where it goes for now, but if it would be easier I can try and get a PR out on the sooner end to consolidate it, so that this work is easier for you.

kzys:
Sorry for the late reply.

> Will this go based implementation be as performant as C based existing implementations? Did you benchmark the same? would be interesting to see the performance gain/degradation.

No I haven't. Let me add benchmarks in this PR.


> Will this support pigz based parallel gzip compressions?

Yes. It should be possible.

kzys:
> There is also the question of licensing, though. From my understanding of BSD-3 Clause we need to display the [license](https://pkg.go.dev/compress/flate?tab=licenses) somewhere in the distribution. In the package is OK by me but I hope I'm not missing anything from a legal standpoint here.

Ah right. I probably need to change the header since it is pointing the `LICENSE` file, which is Apache 2.0 in this repository.

> I'm not opinionated on where it goes for now, but if it would be easier I can try and get a PR out on the sooner end to consolidate it, so that this work is easier for you.

Yes. But Go's packages are not hierarchical (`foo/bar` can't access stuff inside `foo` and vice versa). Moving packages to another directory wouldn't break stuff much.

PR #1875, opened 2026-02-27
Run unit tests that require root perms in CI

### LINKED ISSUE #892: [Bug] Snapshotter unit tests which require root permissions are skipped in CI
### Description

7 snapshotter unit tests require root permissions; however, the current CI setup is not running unit tests as root. This means these tests are being skipped. Moreover when running the tests locally with root permissions, the tests are still skipped.

Run unit tests with verbose output to see the skipped tests:
```
GO_TEST_FLAGS="-v -count=1" make test | grep "SKIP:"
```

Output:
```
--- SKIP: TestRemotePrepare (0.00s)
--- SKIP: TestRemoteView (0.00s)
--- SKIP: TestRemoteOverlay (0.00s)
--- SKIP: TestRemoteCommit (0.00s)
--- SKIP: TestFailureDetection (0.00s)
--- SKIP: TestOverlay (0.00s)
--- SKIP: TestOverlayOverlayRead (0.00s)
```

Run unit tests with root permissions results in the same behavior:
```
GO_TEST_FLAGS="-v -count=1" sudo make test | grep "SKIP:"
```

Output:
```
--- SKIP: TestRemotePrepare (0.00s)
--- SKIP: TestRemoteView (0.00s)
--- SKIP: TestRemoteOverlay (0.00s)
--- SKIP: TestRemoteCommit (0.00s)
--- SKIP: TestFailureDetection (0.00s)
--- SKIP: TestOverlay (0.00s)
--- SKIP: TestOverlayOverlayRead (0.00s)
```

1. All of the snapshotter's unit tests should be ran in CI.
2. Running unit tests locally with root permissions should run all unit tests.

### Host information

1. OS: Amazon Linux
2. Snapshotter Version: 
```
soci version fe3fc93.m fe3fc9339a3fa5a3f602cfcbc21842bb8d3b9028.m
```
3. Containerd Version: 
```
containerd github.com/containerd/containerd v1.6.24-36-gbf6f562cc bf6f562ccf2734f6f5e5723d043c45bb7d9620ac
```

The unit tests are taking dependency on `testutil.RequiresRoot` from `github.com/containerd/containerd/pkg/testutil`.

sondavidb:
**Issue #, if available:**
Closes #892 

**Description of changes:**
Added the necessary flags to the Makefile to pass `-test.root` to the testing binaries. We can't just use `GO_TEST_ARGS` as passing `-test.root` there will cause an unrecognized flag error in all of our other packages, as only the snapshot package uses the `-test.root` flag, so passing it to the test binary allows all the other tests to just ignore it.

Really I don't think we should be using containerd's `testutil.RequiresRoot` at all. But this is a decent bandaid fix.

This also added a separate commit to fix our failing remote tests as we did not have a namespace attached to the contexts passed in which would cause them to fail.

**Testing performed:**
`GO_TEST_BINARY_FLAGS='-test.root'` make test failed locally but seems like issues with my local setup. Opening as draft to see if runners will have the same issue. If not, will take out of draft state.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,6 +47,23 @@ jobs:
       - name: Show test coverage
         run: make show-test-coverage
 
+  # This is a temporary workaround to ensure we are testing our packages
+  # that require root (currently only the snapshot package).
+  # ref: https://github.com/awslabs/soci-snapshotter/issues/1878
+  test-root:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        go-version: ['1.25.7', '1.26.0']
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v6
+        with:
+          go-version: ${{ matrix.go-version }}
+      # Only snapshot package has tests that require root
+      - run: sudo -E env PATH=$PATH go test ./snapshot -test.root
+
   integration:
     needs: setup
     runs-on: ${{ fromJSON(needs.setup.outputs.runner-labels)[matrix.os] }}
--- a/snapshot/snapshot_test.go
+++ b/snapshot/snapshot_test.go
@@ -43,6 +43,7 @@ import (
 
 	"github.com/awslabs/soci-snapshotter/idtools"
 	"github.com/containerd/containerd/mount"
+	"github.com/containerd/containerd/namespaces"
 	"github.com/containerd/containerd/pkg/testutil"
 	"github.com/containerd/containerd/snapshots"
 	"github.com/containerd/containerd/snapshots/overlay/overlayutils"
@@ -57,9 +58,7 @@ const (
 	brokenLabel              = "containerd.io/snapshot/broken"
 )
 
-func prepareWithTarget(t *testing.T, sn snapshots.Snapshotter, target, key, parent string, labels map[string]string) string {
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+func prepareWithTarget(ctx context.Context, t *testing.T, sn snapshots.Snapshotter, target, key, parent string, labels map[string]string) string {
 	if labels == nil {
 		labels = make(map[string]string)
 	}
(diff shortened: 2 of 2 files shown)


sondavidb:
Been testing with `go test ./snapshot -test.root` with root perms.

Two issues here:

1. It seems to not be happy with the permissions I'm adding to the Mount syscalls. That's annoying and I can't figure that out...
2. The last known commit that I could get passing with these changes is ed7a741d027e91b007908dee3d243839783b344d all the way back in March 2023 😬 b64bf2fb763a273d16c8bbcd08f08e73ceeaa71f seems to be the breaking change, which is the containerd 1.7 upgrade.

What bothers me is that stargz runs this exact test but it does not fail. I attempted to run it locally and it did fail there as well. Wonder if there's some setup we're missing here...

In any case, it's an option to skip the test entirely if it's causing so much trouble. But I would like to figure it out if possible.

sondavidb:
I found the culprit. Our CodeBuild runners are on 4.14 which is causing issues with mounting. Newest Ubuntu runners are using 6.14 which doesn't have these issues. We still have issues on CodeBuild with cross-device linking as well which I haven't figured out the reason for, but CI on regular GitHub Ubuntu runners seems to pass.

As a bandaid solution, I'll create a separate workflow that only runs root tests on an Ubuntu runner. I'll create an issue to track this as well.

Shubhranshu153 (approved):

.github/workflows/build.yml
```diff
@@ -47,6 +47,23 @@ jobs:
       - name: Show test coverage
         run: make show-test-coverage
 
+  # This is a temporary workaround to ensure we are testing our packages
```
Shubhranshu153, line 50: why do we consider it as a temporary work around?

sondavidb: Our CodeBuild runners cannot run these tests, likely due to an old kernel. So we are using a whole new GitHub-provisioned runner just for the tests, which is not ideal. We need our root tests to pass on CodeBuild to consolidate our tests.

Shubhranshu153: What is the primary reason to use Codebuild?

sondavidb: ARM support on Linux is our primary reason. We can't really get rid of that IMO since we have seen ARM-specific failures previously.

coderbirju (approved):

PR #1879, opened 2026-02-28
Add retry mechanism for failing integration tests

sondavidb:
**Issue #, if available:**

**Description of changes:**
Our integration tests have become flakier for networking reasons. I decided to add a wrapper script to catch failing tests and retry only the ones that have failed.

This isn't perfect as the `-run` flag in `go test` takes regex, so if the failing test name happens to be the prefix for another test, it will run more tests than necessary (e.g. `-run TestConvert` would also run `TestConvertWithForceRecreateZtocs` and `TestConvertAndPush`). However, this is still better than what we have currently, so I guess it's fine for now.

**Testing performed:**
Ran `./scripts/integration-with-retries 3` locally and ensured it passed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,8 @@ on:
       - 'ztoc/**' # Includes flatbuf + C files
       - '!benchmark/**'
 
+env:
+  GOTESTSUM_VERSION: 1.13.0
 
 jobs:
   setup:
@@ -64,6 +66,8 @@ jobs:
       - uses: actions/setup-go@v6
         with:
           go-version: ${{ matrix.go-version }}
+      - name: Install gotestsum
+        run: go install gotest.tools/gotestsum@v${{ env.GOTESTSUM_VERSION }}
       - name: Install zlib static on AL2 ARM instances
         if: matrix.os == 'al2-arm'
         run: dnf install zlib-static.aarch64 -y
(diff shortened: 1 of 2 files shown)


sondavidb:
Looks like failed tests retry: https://github.com/awslabs/soci-snapshotter/actions/runs/22537189541/job/65286713612?pr=1879#step:5:865

scripts/integration-with-retries.sh
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bash
```
Shubhranshu153: can we not incorporate retries in the test suite. Looks hacky to do test parsing for retries. Also have we investigated why tests are flaky?

Swapnanil-Gupta:
> e.g. `-run TestConvert` would also run `TestConvertWithForceRecreateZtocs` and `TestConvertAndPush`

Can we use `-run '^TestConvert$'` to match only `TestConvert`?

scripts/integration-with-retries.sh
sondavidb: Tests are flaky for network reasons. Pulling images sometimes loses connection which I always imagined was due to hitting rate limits but it's possible it's for other reasons. It happens sporadically across different tests, though TestConvert seems to get it the most.

Shubhranshu153: Cant we integrate the retries into the test suite. Parsing and doing retries can be flaky if outputs change.

sondavidb: Used `gotestsum` instead.

Shubhranshu153:
Test seem to fail.

sondavidb:
> Test seem to fail.

Wrong version 🤦 correct one should run them now

sondavidb:
Nice, we got a failure that got caught: https://github.com/awslabs/soci-snapshotter/actions/runs/22596642133/job/65468271427?pr=1879

Wish we could do something about these network errors though. Would be nice only to fail on those.

PR #1881, opened 2026-03-01
Add Standalone SOCI Image Convert mode for SOCI index creation using disk based OCI Layouts without containerd

prafgup:
**Issue #, if available:**
Closes https://github.com/awslabs/soci-snapshotter/issues/1057

**Description of changes:**
This PR adds standalone mode to the `soci convert` command, enabling SOCI index creation without requiring a running containerd daemon. This is particularly useful for CI/CD pipelines and environments where running containerd is impractical without privileged mode.

  ### Key Features

  - **Standalone Mode (`--standalone`)**: Reads an OCI image layout (tar or directory) from disk, creates SOCI indexes, and writes the converted image back to disk
  - **Output Format (`--format`)**: Supports `oci-archive` (tar, default) and `oci-dir` (directory) output formats
     -  This follows the same convention used by tools like [Podman](https://docs.podman.io/en/latest/markdown/podman-save.1.html) (`--format oci-archive` / `oci-dir`), [Skopeo](https://github.com/containers/skopeo) (`copy oci-archive:` / `oci:`), and [crane](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_pull.md) using `--format oci` for directory output.

  ### Usage Example

  **Tar output (oci-archive, default):**
```
# Pull an image as OCI layout
crane pull --format oci myregistry.com/myimage:latest /tmp/my-image

# Convert to SOCI-enabled image (tar output)
soci convert --standalone /tmp/my-image /tmp/my-soci-image.tar

# Import into containerd
ctr images import --no-unpack --index-name myregistry.com/myimage:latest-soci /tmp/my-soci-image.tar

--- or push with crane like - (crane only supports dir for oci layout pushing)

mkdir -p /tmp/my-soci-image-dir
tar -xf /tmp/my-soci-image.tar -C /tmp/my-soci-image-dir
crane push /tmp/my-soci-image-dir myregistry.com/myimage:latest-soci
```

  **Directory output (oci-dir):**
  ```
# Pull an image as OCI layout
crane pull --format oci myregistry.com/myimage:latest /tmp/my-image

# Convert to SOCI-enabled image (directory output)
soci convert --standalone --format oci-dir /tmp/my-image /tmp/my-soci-image

# Push back to registry
crane push /tmp/my-soci-image myregistry.com/myimage:latest-soc

```

**Testing performed:**

  -  Added new Integration test and all integration tests pass (TestStandaloneConvert*)
    - Basic conversion with validation (all permutation of input (tar/dir) and output (tar/dir)
    - Specific platform selection (--platform)
    - Error handling (nonexistent images, missing arguments, wrong output format)
    - Idempotency verification
  - Tested on ARM64 architecture (linux/arm64/v8) (lima on macOs)
     - `lima sudo GO_TEST_FLAGS="-run TestStandalone -count=1" make integration`
     - `lima sudo make test`
  - Manual verification of image download, conversion, and push workflow

Standalone Convert with oci-archive output - 
```
crane auth login source-registry.com --username $REPOSITORY --password $TOKEN
crane pull --format oci source-registry.com/myimage:latest /tmp/my-crane-image-dir

sudo ./out/soci convert --standalone /tmp/my-crane-image-dir /tmp/my-soci-image.tar

ztoc skipped - layer sha256:7511aa994dfed5f54c697fcec6db7176449e7d87317b3f6ad093185182ce01c7 (application/vnd.oci.image.layer.v1.tar+gzip) size 3222 is less than min-layer-size 10485760
ztoc skipped - layer sha256:12bb71ceae3401038ce93aa7ce3a115f4a5cb3e9a85277ef4a97718c0fc9298e (application/vnd.oci.image.layer.v1.tar+gzip) size 221568 is less than min-layer-size 10485760
....
layer sha256:d8c6b361b623cabb0a89fad4a7d491d7d543f0834ca0c581e0176e1b76a68a15 -> ztoc sha256:ea8f91d2a57139b053186c8a1ae1901684bc06fda5b7d00aef0bd3d6d4e9e4c1



mkdir -p /tmp/my-soci-image-dir
tar -xf /tmp/my-soci-image.tar -C /tmp/my-soci-image-dir
crane push /tmp/my-soci-image-dir dest-registry.com/myimage:latest-soci 

2026/03/01 17:51:41 existing manifest: sha256:fe64b32c84403bd29758342ac0f78a384961bf6788d5de22b9fea1e7c7f1c84a
2026/03/01 17:51:41 existing manifest: sha256:c60d93f19b8a3dd886faa3b455ea809d0055173aa0e501fb0926f196d736a984
2026/03/01 17:51:41 dest-registry.com/myimage:latest-soci : digest: sha256:f265a1fddffc035eb949ac3ff03a597ce4c823b8bde2ae6cb63a0c85f69d0de9 size: 803
dest-registry.com/myimage@sha256:f265a1fddffc035eb949ac3ff03a597ce4c823b8bde2ae6cb63a0c85f69d0de9

```

Standalone Convert with oci-dir output - 

```
sudo ./out/soci convert --standalone /tmp/my-crane-image-dir /tmp/my-soci-image --format oci-dir

ztoc skipped - layer sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1 (application/vnd.oci.image.layer.v1.tar+gzip) size 32 is less than min-layer-size 10485760
ztoc skipped - layer sha256:7511aa994dfed5f54c697fcec6db7176449e7d87317b3f6ad093185182ce01c7 (application/vnd.oci.image.layer.v1.tar+gzip) size 3222 is less than min-layer-size 10485760
...
ztoc skipped - layer sha256:f6a26d16afc76b03f36290b56770ae0db2ab11aab759d2c7bd49bdfc865b3b87 (application/vnd.oci.image.layer.v1.tar+gzip) size 10395600 is less than min-layer-size 10485760
ztoc skipped - layer sha256:ef6465a1674b258144bd7eac86347823926436a5d53f100cadfe098c82e4c4d4 (application/vnd.oci.image.layer.v1.tar+gzip) size 78 is less than min-layer-size 10485760


crane push /tmp/my-soci-image dest-registry.com/myimage:latest-soci 
2026/03/01 20:50:51 existing manifest: sha256:fe64b32c84403bd29758342ac0f78a384961bf6788d5de22b9fea1e7c7f1c84a
2026/03/01 20:50:51 existing manifest: sha256:c60d93f19b8a3dd886faa3b455ea809d0055173aa0e501fb0926f196d736a984
2026/03/01 20:50:51 dest-registry.com/myimage:latest-soci : digest: sha256:f265a1fddffc035eb949ac3ff03a597ce4c823b8bde2ae6cb63a0c85f69d0de9 size: 803
dest-registry.com/myimage:@sha256:f265a1fddffc035eb949ac3ff03a597ce4c823b8bde2ae6cb63a0c85f69d0de9
```

Integration tests - 

```
sudo GO_TEST_FLAGS="-run TestStandalone -count=1" make integration

=== RUN   TestStandaloneConvertBasic
=== RUN   TestStandaloneConvertBasic/dir-to-tar
=== RUN   TestStandaloneConvertBasic/tar-to-tar
=== RUN   TestStandaloneConvertBasic/dir-to-dir
=== RUN   TestStandaloneConvertBasic/tar-to-dir
--- PASS: TestStandaloneConvertBasic (31.63s)
    --- PASS: TestStandaloneConvertBasic/dir-to-tar (2.92s)
    --- PASS: TestStandaloneConvertBasic/tar-to-tar (2.65s)
    --- PASS: TestStandaloneConvertBasic/dir-to-dir (2.71s)
    --- PASS: TestStandaloneConvertBasic/tar-to-dir (2.69s)
=== RUN   TestStandaloneConvertSpecificPlatform
--- PASS: TestStandaloneConvertSpecificPlatform (18.89s)
=== RUN   TestStandaloneInvalidConversion
=== RUN   TestStandaloneInvalidConversion/nonexistent_input
=== RUN   TestStandaloneInvalidConversion/invalid_format
=== RUN   TestStandaloneInvalidConversion/missing_destination
--- PASS: TestStandaloneInvalidConversion (1.19s)
    --- PASS: TestStandaloneInvalidConversion/nonexistent_input (0.02s)
    --- PASS: TestStandaloneInvalidConversion/invalid_format (0.03s)
    --- PASS: TestStandaloneInvalidConversion/missing_destination (0.03s)
=== RUN   TestStandaloneConvertIdempotent
--- PASS: TestStandaloneConvertIdempotent (19.88s)
PASS
ok      github.com/awslabs/soci-snapshotter/integration 193.187s
```

### DIFF
--- a/cmd/soci/commands/convert.go
+++ b/cmd/soci/commands/convert.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"slices"
 	"strings"
 
@@ -31,6 +32,14 @@ import (
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/reference"
 	"github.com/urfave/cli/v3"
+)
+
+const (
+	standaloneFlag   = "standalone"
+	outputFormatFlag = "format"
+
+	outputFormatOCIArchive = "oci-archive"
+	outputFormatOCIDir     = "oci-dir"
 )
 
 var ErrInvalidDestRef = errors.New(`the destination image must be a tagged ref of the form "registry/repository:tag"`)
@@ -59,34 +68,52 @@ func verifyRef(r string) error {
 // ConvertCommand converts an image into a SOCI enabled image.
 // The new image is added to the containerd content store and can
 // be pushed and deployed like a normal image.
+//
+// In standalone mode, the command reads an OCI image layout (tar or directory)
+// and writes a converted OCI image layout (tar or directory) without requiring containerd.
 var ConvertCommand = &cli.Command{
 	Name:      "convert",
 	Usage:     "convert an OCI image to a SOCI enabled image",
 	ArgsUsage: "[flags] <image_ref> <dest_ref>",
-	Flags:     slices.Concat(internal.PlatformFlags, createZtocFlags, internal.PrefetchFlags),
+	Flags: slices.Concat(
+		internal.PlatformFlags,
+		createZtocFlags,
+		internal.PrefetchFlags,
+		[]cli.Flag{
+			&cli.BoolFlag{
+				Name:  standaloneFlag,
+				Usage: "Run in standalone mode without containerd runtime. In this mode, the command reads an OCI image layout (tar or directory) and writes a converted OCI image layout without requiring a running containerd instance.",
+			},
+			&cli.StringFlag{
+				Name:  outputFormatFlag,
+				Usage: "Output format for standalone mode: oci-archive (tar) or oci-dir (directory).",
+				Value: outputFormatOCIArchive,
+				Validator: func(s string) error {
+					if s != outputFormatOCIArchive && s != outputFormatOCIDir {
+						return fmt.Errorf("unsupported output format %q: must be %q or %q", s, outputFormatOCIArchive, outputFormatOCIDir)
+					}
+					return nil
+				},
+			},
+		}),
 	Action: func(ctx context.Context, cmd *cli.Command) error {
-		srcRef := cmd.Args().Get(0)
-		if srcRef == "" {
+		src := cmd.Args().Get(0)
+		if src == "" {
 			return errors.New("source image needs to be specified")
 		}
 
-		dstRef := cmd.Args().Get(1)
-		if dstRef == "" {
-			return errors.New("destination image needs to be specified")
+		dst := cmd.Args().Get(1)
+		if dst == "" {
+			return errors.New("destination needs to be specified")
 		}
 
-		err := verifyRef(dstRef)
+		if cmd.Bool(standaloneFlag) {
+			return runStandaloneConvert(ctx, cmd, src, dst)
+		}
+
+		err := verifyRef(dst)
 		if err != nil {
 			return fmt.Errorf("%w: %w", ErrInvalidDestRef, err)
-		}
-
-		var optimizations []soci.Optimization
-		for _, o := range cmd.StringSlice(optimizationFlag) {
-			optimization, err := soci.ParseOptimization(o)
-			if err != nil {
-				return err
-			}
-			optimizations = append(optimizations, optimization)
 		}
 
 		client, ctx, cancel, err := internal.NewClient(ctx, cmd)
@@ -97,14 +124,10 @@ var ConvertCommand = &cli.Command{
 
 		cs := client.ContentStore()
 		is := client.ImageService()
-		srcImg, err := is.Get(ctx, srcRef)
+		srcImg, err := is.Get(ctx, src)
 		if err != nil {
 			return err
 		}
-
-		spanSize := cmd.Int64(spanSizeFlag)
-		minLayerSize := cmd.Int64(minLayerSizeFlag)
-		forceRecreateZtocs := cmd.Bool(forceRecreateZtocsFlag)
 
 		blobStore, err := store.NewContentStore(internal.ContentStoreOptions(ctx, cmd)...)
 		if err != nil {
@@ -116,25 +139,13 @@ var ConvertCommand = &cli.Command{
 			return err
 		}
 
-		builderOpts := []soci.BuilderOption{
-			soci.WithMinLayerSize(minLayerSize),
-			soci.WithSpanSize(spanSize),
-			soci.WithBuildToolIdentifier(buildToolIdentifier),
-			soci.WithOptimizations(optimizations),
-			soci.WithArtifactsDb(artifactsDb),
-			soci.WithForceRecreateZtocs(forceRecreateZtocs),
-		}
-
-		allPrefetchFiles, err := internal.ParsePrefetchFiles(cmd)
+		builderOpts, err := parseBuilderOptions(cmd)
 		if err != nil {
-			return fmt.Errorf("failed to parse prefetch files: %w", err)
+			return err
 		}
-		if len(allPrefetchFiles) > 0 {
-			builderOpts = append(builderOpts, soci.WithPrefetchPaths(allPrefetchFiles))
-		}
+		builderOpts = append(builderOpts, soci.WithArtifactsDb(artifactsDb))
 
 		builder, err := soci.NewIndexBuilder(cs, blobStore, builderOpts...)
-
 		if err != nil {
 			return err
 		}
@@ -162,10 +173,10 @@ var ConvertCommand = &cli.Command{
 		}
 
 		im := images.Image{
-			Name:   dstRef,
+			Name:   dst,
 			Target: *desc,
 		}
-		img, err := is.Get(ctx, dstRef)
+		img, err := is.Get(ctx, dst)
 		if err != nil {
 			if !errors.Is(err, errdefs.ErrNotFound) {
 				return err
@@ -178,4 +189,95 @@ var ConvertCommand = &cli.Command{
 
 		return err
 	},
+}
+
+// runStandaloneConvert runs the convert command in standalone mode (without containerd).
+// It reads an OCI image layout (tar or directory), performs the SOCI conversion, and
+// writes the result as an OCI image layout tar or directory based on the --format flag.
+func runStandaloneConvert(ctx context.Context, cmd *cli.Command, inputPath string, outputPath string) error {
+	format := cmd.String(outputFormatFlag)
+
+	ociLayoutDir, err := os.MkdirTemp("", "soci-oci-*")
+	if err != nil {
+		return fmt.Errorf("failed to create temp directory: %w", err)
+	}
+	defer os.RemoveAll(ociLayoutDir)
+
+	imageInfo, err := internal.LoadImage(ctx, inputPath, ociLayoutDir)
+	if err != nil {
+		return err
+	}
+
+	artifactsDir, err := os.MkdirTemp("", "soci-artifacts-*")
+	if err != nil {
+		return fmt.Errorf("failed to create artifacts temp directory: %w", err)
+	}
+	defer os.RemoveAll(artifactsDir)
+
+	artifactsDb, err := soci.NewDB(soci.ArtifactsDbPath(artifactsDir))
+	if err != nil {
+		return fmt.Errorf("failed to create artifacts database: %w", err)
+	}
+
+	builderOpts, err := parseBuilderOptions(cmd)
+	if err != nil {
+		return err
+	}
+	builderOpts = append(builderOpts, soci.WithArtifactsDb(artifactsDb))
+
+	sociStore := &store.SociStore{Store: imageInfo.OrasStore}
+	builder, err := soci.NewIndexBuilder(imageInfo.ContentStore, sociStore, builderOpts...)
+	if err != nil {
+		return err
+	}
+
+	requestedPlatforms, err := internal.GetPlatforms(ctx, cmd, imageInfo.Image, imageInfo.ContentStore)
+	if err != nil {
+		return err
+	}
+
+	batchCtx, done, err := sociStore.BatchOpen(ctx)
+	if err != nil {
+		return err
+	}
+	defer done(ctx)
+
+	convertedDesc, err := builder.Convert(batchCtx, imageInfo.Image, soci.ConvertWithPlatforms(requestedPlatforms...))
+	if err != nil {
+		return err
+	}
+
+	if format == outputFormatOCIDir {
+		return internal.SaveImageToDir(ociLayoutDir, *convertedDesc, outputPath)
+	}
+	return internal.SaveImageToTar(ctx, imageInfo.ContentStore, *convertedDesc, outputPath)
+}
+
+func parseBuilderOptions(cmd *cli.Command) ([]soci.BuilderOption, error) {
+	var optimizations []soci.Optimization
+	for _, o := range cmd.StringSlice(optimizationFlag) {
+		optimization, err := soci.ParseOptimization(o)
+		if err != nil {
+			return nil, err
+		}
+		optimizations = append(optimizations, optimization)
+	}
+
+	builderOpts := []soci.BuilderOption{
+		soci.WithMinLayerSize(cmd.Int64(minLayerSizeFlag)),
+		soci.WithSpanSize(cmd.Int64(spanSizeFlag)),
+		soci.WithBuildToolIdentifier(buildToolIdentifier),
+		soci.WithOptimizations(optimizations),
+		soci.WithForceRecreateZtocs(cmd.Bool(forceRecreateZtocsFlag)),
+	}
+
+	allPrefetchFiles, err := internal.ParsePrefetchFiles(cmd)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse prefetch files: %w", err)
+	}
+	if len(allPrefetchFiles) > 0 {
+		builderOpts = append(builderOpts, soci.WithPrefetchPaths(allPrefetchFiles))
+	}
+
+	return builderOpts, nil
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: integration/convert_standalone_test.go
Comment (sondavidb) [inline integration/convert_standalone_test.go:1]: very nit: would prefer `convert_standalone_test.go` just so it's next to the convert tests alphabetically 🙂

File: Dockerfile [hunk 1]
```diff
@@ -141,3 +143,6 @@ RUN curl -sSL --output /tmp/runc https://github.com/opencontainers/runc/releases
 RUN curl -sSL --output /tmp/nerdctl.tgz https://github.com/containerd/nerdctl/releases/download/v${NERDCTL_VERSION}/nerdctl-${NERDCTL_VERSION}-linux-${TARGETARCH:-amd64}.tar.gz \
     && tar zxvf /tmp/nerdctl.tgz -C /usr/local/bin/ \
     && rm -f /tmp/nerdctl.tgz
+RUN curl -sSL --output /tmp/crane.tgz https://github.com/google/go-containerregistry/releases/download/v${CRANE_VERSION}/go-containerregistry_Linux_$(echo ${TARGETARCH:-amd64} | sed s/amd64/x86_64/).tar.gz \
```
sondavidb: If this is the only way to pull images to an OCI dir this is fine, but I'd heavily prefer not pulling a new dependency as I think the many requests being made might be what's causing issues with our testing suite (unfounded speculation, not certain). Would prefer if we can find a way to use nerdctl/ctr/anything else that is already installed.

Though, maybe the real answer is that we need to figure out how to stop these network blips.

integration/standalone_convert_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+			sociV2Enabled := withPullModes(config.PullModes{
+				SOCIv1: config.V1{Enable: false},
+				SOCIv2: config.V2{Enable: true},
+			})
+			rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t, sociV2Enabled))
+			sh.X("nerdctl", "rmi", "-f", dstRef)
+			sh.X("nerdctl", "pull", "-q", "--snapshotter=soci", dstRef)
+
+			sh.X("nerdctl", "run", "--rm", "--net", "none", "--snapshotter=soci", dstRef, "echo", "success")
```
sondavidb: Can we use `runSociCmd` instead?

Also wondering why we are testing with a container run here but using `validateConversion` for the other tests.

integration/standalone_convert_test.go
```diff
@@ -0,0 +1,247 @@
...
+	defer done()
+
+	rebootContainerd(t, sh, getContainerdConfigToml(t, false), getSnapshotterConfigToml(t))
+
+	imageRef := nginxImage
+	mirrorImg := regConfig.mirror(imageRef)
+	srcRef := mirrorImg.ref
+	srcDigest := getImageDigest(sh, srcRef)
+	inputDir := "/tmp/standalone-basic-input"
+	inputTar := "/tmp/standalone-basic-input.tar"
```
sondavidb: I can see concurrent runs of the tests being problematic. Can we just use tempdirs for these, and for other test cases?

integration/convert_standalone_test.go
```diff
@@ -0,0 +1,247 @@
...
+			"--standalone",
+			"/tmp/nonexistent-input",
+			"/tmp/output.tar",
+		)
+		if err == nil {
+			t.Fatal("expected error for nonexistent input")
+		}
+
+		outputStr := string(output)
+		if !strings.Contains(outputStr, "failed to access input") {
```
sondavidb, line 161: nit: Not a fan of string checking here. But since we can't import the variable from the `cmd` package this is probably the next best thing.

integration/convert_standalone_test.go
```diff
@@ -0,0 +1,247 @@
...
+			t.Fatalf("expected error about failed input access, got: %s", outputStr)
+		}
+	})
+
+	t.Run("invalid format", func(t *testing.T) {
```
sondavidb, line 166: Wonder if this and the test below are better suited as unit tests?

sondavidb (commented):
Generally LGTM, some small things I'd like looked at before approving.

Thanks!

Dockerfile
prafgup: > Though, maybe the real answer is that we need to figure out how to stop these network blips.

True, I saw most of the failures were because of rapidgzip cloning failures probably due to git clone issues which is not related to these changes. Eg - https://github.com/awslabs/soci-snapshotter/actions/runs/22553066639/job/65605782655?pr=1881

https://github.com/awslabs/soci-snapshotter/blob/c76cb8b10785e30ea55a534f61478e7510e6c2be/Dockerfile#L74-L79

I believe curl should be relatively stable.


> I think the many requests being made might be what's causing issues with our testing suite (unfounded speculation, not certain)

Igzip and rapidgzip does seem to be parallel so that might be the rootcause here.

The curl downloading is sequential and should not be affected much - 
- https://github.com/awslabs/soci-snapshotter/blob/c76cb8b10785e30ea55a534f61478e7510e6c2be/Dockerfile#L137-L143



> If this is the only way to pull images to an OCI dir this is fine

I checked it with `ctr images export` (exports tar) as well, which works, but then that would defeat the e2e purpose of using a containerdless tool to create OCI dir or tar and using them. Thus I kept this simply with crane. I believe adding this curl would not be creating any issues for CI builds as its sequential downloading. I can monitor some failures in future and if they are due to crane I can change back to ctr export, but for now I believe crane would be the right choice for actual testing of this flow. Let me know if you still think we should change back to ctr export (containerd required tool) for this in integration testing.

integration/convert_standalone_test.go
prafgup: I don't think we have unit tests testing the command executions 🤔 
Overall these tests would fail pretty fast and take few ms to execute, so i think we can keep them here for now. What do you think?

```
--- PASS: TestStandaloneInvalidConversion (1.19s)
    --- PASS: TestStandaloneInvalidConversion/nonexistent_input (0.02s)
    --- PASS: TestStandaloneInvalidConversion/invalid_format (0.03s)
    --- PASS: TestStandaloneInvalidConversion/missing_destination (0.03s)
```

integration/standalone_convert_test.go
prafgup: Yes, I'll add `runSociCmd` here instead.

> Also wondering why we are testing with a container run here but using validateConversion for the other tests.


I see in convert_test we just `validateConversion` in each test and assuming the run would succeed just relying on correctness of `validateConversion` helper function.
- https://github.com/awslabs/soci-snapshotter/blob/c76cb8b10785e30ea55a534f61478e7510e6c2be/integration/convert_test.go#L357-L360

I believe its a good idea to run the image as well for a basic test apart from doing a `validateConversion` to be sure it works e2e. I'll add `validateConversion` here as well. 
What do you think, should we should remove the image run and rely fully on `validateConversion` here or is it fine to have like a e2e with image run in a basic test here?

Dockerfile
sondavidb: >  probably due to git clone issues which is not related to these changes

I had an issue building locally for a very long time until I cleaned out some cache somewhere and that suddenly made it start working again. For me it would always fail on the cmake/make step (forget which one). But not sure if that's the issue here.

If the issue is due to `git clone` then it's likely a network issue from downloading the needed artifacts which is why I was a little concerned about adding another external download here. With that said, it also never fails when acquiring all the dependencies from the package manager. So not sure what to make of that.

Anyway, if we need it we need it. That's fine.

> but then that would defeat the e2e purpose of using a containerdless tool to create OCI dir or tar and using them

I see your point but if we'd be getting the same input from ctr/nerdctl compared to crane without any super hacky shenanigans I would prefer ctr (and would prefer nerdctl over ctr if possible — seems they have an export command as well). Do you foresee any discrepancies between pulling with crane and exporting with nerdctl when it comes to putting either into the standalone convert?

prafgup: > Do you foresee any discrepancies between pulling with crane and exporting with nerdctl when it comes to putting either into the standalone convert?

Not really, I'd assume they should do the same thing. Just that Crane would actually download and create the dir/tar in 1 pass where as we would have to download image and and then export when using nerdctl. Overall I am fine with any approach, if we deem that exporting would be the right choice here I will change the implementation to use nerdctl/crt instead of using crane.

integration/standalone_convert_test.go
sondavidb: I just want consistency more than anything. If I were to look at this code at a glance I'd wonder why this specific command uses a run to validate while the others don't. (Really we have these discrepancies everywhere, but I'd like to keep them to a minimum when possible.)

With that said, I'd prefer to stick to only using `validateConversion` but I also see your point on wanting to do a quick run to ensure it starts up properly. Maybe we can add a `TestRunAfterConvert` or something in `run_test.go`. We don't currently have an integration test for running with convert so maybe it would be useful to add there?

prafgup: Ack, I'll change it to `validateConversion` and have a followup PR to add convert tests like `TestRunAfterConvert` you mentioned.

integration/convert_standalone_test.go
sondavidb: > I don't think we have unit tests testing the command executions 

Hm, yeah. I was thinking that I would like simple unit tests to ensure cmd outputs are what we expect. But I don't think we have a good structure for that, so I'm fine not changing this bit.

Dockerfile
sondavidb: Sure. Let's just use nerdctl export then

prafgup: Seems like nerdctl exports with a manifest list where the root is a docker manifest list and not a OCI list. Moreover the Docker manifest list references ALL platforms (amd64, arm64, etc.) even when using `--platform arm64`, but only arm64 blobs are exported. 
```
{"manifests":[{"mediaType":"application/vnd.docker.distribution.manifest.list.v2+json","digest":"sha256:f4e3b6..."}]}      
```                                
                                                                                                                                
When soci convert processes this, annotateImages correctly skips missing platforms but doesn't remove their descriptors from the OCI index. Leading to export issues using archive.Export - 

https://github.com/awslabs/soci-snapshotter/blob/c76cb8b10785e30ea55a534f61478e7510e6c2be/soci/soci_convert.go#L268-L272

Trying to filter out the manifests whose blobs are not present in case the root descriptor is a manifest list.

prafgup:
Hey @sondavidb ,
Resolved all your comments 😄 feel free to review again. (Also fixed the issue I mentioned here https://github.com/awslabs/soci-snapshotter/pull/1881#discussion_r2886762536)

sondavidb:
Sure. I'll take a look tomorrow. Will run CI in the meantime.

cmd/soci/commands/convert.go
```diff
@@ -59,34 +68,52 @@ func verifyRef(r string) error {
...
 	Name:      "convert",
 	Usage:     "convert an OCI image to a SOCI enabled image",
 	ArgsUsage: "[flags] <image_ref> <dest_ref>",
-	Flags:     slices.Concat(internal.PlatformFlags, createZtocFlags, internal.PrefetchFlags),
+	Flags: slices.Concat(
+		internal.PlatformFlags,
+		createZtocFlags,
+		internal.PrefetchFlags,
+		[]cli.Flag{
+			&cli.BoolFlag{
```
Shubhranshu153, line 83: nit: Here does a flag --input --output works as per the convention in docker load and docker save. That makes the functionality more extendable.

prafgup: For this feature the input is auto detected (input can be tar or dir only), and we have a flag for output (to either output as oci-dir or tar)

This follows the same convention used by tools like [Podman](https://docs.podman.io/en/latest/markdown/podman-save.1.html) (--format oci-archive / oci-dir), [Skopeo](https://github.com/containers/skopeo) (copy oci-archive: / oci:), and [crane](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_pull.md) using `--format oci` for directory output (defaults to tar). Thus here we are adhering to the convention of other tools as well.

cmd/soci/commands/internal/standalone.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// Multiple (but not all) platforms available: write a filtered manifest list
+	manifestList.Manifests = available
+	filteredData, err := json.Marshal(manifestList)
+	if err != nil {
+		return ocispec.Descriptor{}, fmt.Errorf("failed to marshal filtered manifest list: %w", err)
+	}
+	filteredDigest := digest.FromBytes(filteredData)
+	filteredPath := filepath.Join(layoutDir, "blobs", filteredDigest.Algorithm().String(), filteredDigest.Encoded())
+	if err := os.WriteFile(filteredPath, filteredData, 0644); err != nil {
```
sondavidb, line 203: Wonder if 644 perms should be hardcoded or if we should allow any sort of user config. Technically none of the operations here require root perms, but other SOCI CLI commands must be run as root anyway so it's probably fine. Just food for thought.

cmd/soci/commands/convert.go
Shubhranshu153: Yeah it was more in the context of  decoupling the input and output, but this works too for the use case.

Shubhranshu153:
LGTM. Thanks for working on it.

cmd/soci/commands/internal/standalone.go
prafgup: This new created file (in our own dir) should only be readable and should not be modified my anyone thus I think 644 should be fine here and we can leave it as is :)

prafgup:
Thanks for the review 🚀 

I'll followup with the doc update PR once this is merged 😄  (also would appreciate a release for this feature so we can start consuming it internally)

sondavidb:
> (also would appreciate a release for this feature so we can start consuming it internally)

Sure, I can bring this up and see when we can get a release out.

Issue #1924, opened 2026-04-02
[Bug] SOCI 2nd Image pull not working with layer unavailable error


prafgup:
### Description

Hey Devs,
(I have used AI for some analysis and writeup, but have confirmed that things are correct)

When using soci-snapshotter-grpc with a private registry that uses `Basic auth` (username and token) and short-lived credentials (via CRI keychain, we use k8s `Secrets` and remove them once pod has finished), layers that were successfully pulled/partially pulled become unavailable hours later. The checkAvailability health check fails because the snapshotter cannot re-authenticate with the original credentials expire.

#### Environment                                                     
  - soci-snapshotter version: 0.12.1
  - Registry auth: Basic (WWW-Authenticate: Basic realm="XXXX")
  - Credential source: CRI keychain (cri_keychain.enable_keychain = true)                                                                                                  
  - Container runtime: containerd                                                                                                                                          
                                                                                                                                                                           
#### Behavior                                                                                                                                                                 
                                                                  
First pull (works): Kubelet issues PullImage() → CRI keychain receives fresh credentials → SOCI artifacts fetched and layers mounted via FUSE successfully.              

Hours later (fails): A new pod pulls the same image. checkAvailability runs on the existing FUSE-mounted layers, attempts to verify blob connectivity, receives 401, and  tries to re-authenticate. Re-authentication fails with "failed to find supported auth scheme: not implemented", causing layers to be marked unavailable.



When a pod is launched and kubelet tries to pull the same image hours after on the same node I see this error -

```
failed to create containerd container: error unpacking image: failed to prepare extraction snapshot 
"extract-86525.... sha256:c49c6f8114925036148a3596d382b44b770483093594ba21be9923ea3d51a71c": 
layer "177" unavailable: unavailable
```

My soci config - 
```
[cri_keychain]
enable_keychain = true
image_service_path = "/run/containerd/containerd.sock"
```


Heres a rough timeline of the error - 

```
  ┌─────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │    Time     │                                                              Event                                                               │
  ├─────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 15:04:51    │ Image pull starts, snapshots prepared. Credentials in CRI keychain.                                                              │
  ├─────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
[...]

PR #1925, opened 2026-04-02
Invalidate host creds in case of failure and retry

prafgup:
**Issue #, if available:**

Related to https://github.com/awslabs/soci-snapshotter/issues/1924
Disclaimer - Coded by AI

**Description of changes:**

**Testing performed:**

 Proof:
  - InvalidateRegistryHosts is real → integration test PASSES (container runs after credential rotation)
  - InvalidateRegistryHosts is no-op → integration test FAILS with layer "5" unavailable: unavailable (exact same error as your production issue)

The mechanism: Short-lived Bearer tokens expire → docker.Authorizer's cached handler tries fetchToken with old password baked in TokenOptions → token server rejects → error inside doBearerAuth (before AddResponses) → handler never deleted → Authorizer stuck → our fix invalidates the registryHostMap → fresh Authorizer with empty handlers → re-authenticates with new password from docker config → works.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -164,6 +164,7 @@ type Option func(*options)
 
 type options struct {
 	getSources        source.GetSources
+	invalidateHosts   source.InvalidateHosts
 	resolveHandlers   map[string]remote.Handler
 	metadataStore     metadata.Store
 	overlayOpaqueType layer.OverlayOpaqueType
@@ -174,6 +175,12 @@ type options struct {
 func WithGetSources(s source.GetSources) Option {
 	return func(opts *options) {
 		opts.getSources = s
+	}
+}
+
+func WithInvalidateHosts(f source.InvalidateHosts) Option {
+	return func(opts *options) {
+		opts.invalidateHosts = f
 	}
 }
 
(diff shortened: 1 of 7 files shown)


prafgup:
@sondavidb I (+AI) was able to replicate the `layer "5" unavailable: unavailable` issue with the integration test here (very hacky, basically expire the custom token with TTL and fetch again to trigger check()). This fails without the change in this PR but actually re-auths (refreshes) when the auth is expired.
- The main logic change is in fs.go where we invalidate the cache in case of error when refreshing and refresh again.

This in no way is ready for review. I just want your views on the FS cache invalidation logic :) (I will refine the PR once I have your views aligned, tho integration test is not really viable in this case without a lot of custom changes like in this PR)

sondavidb:
Idea LGTM, I'm happy to move forward with this. I do have some concerns in terms of credentials on a per-layer basis, as I believe we cache the credentials per layer as well. But I don't think this has to be comprehensive by any means — if we can refresh credentials on a snapshotter level that is already helping a ton of pain points for Kubernetes use cases.

prafgup:
Thanks, I will go ahead and refine the PR (by Monday, afk for some days) 😄

Note: Integration test is gonna be very complex with the current testing setup thus would not have any for now. I'll research a bit in future if theres any easier way of implementing a integration test.

Issue #522, opened 2023-03-22
[FEATURE] Support kubernetes credential rotation


Kern--:
### Description

The SOCI snapshotter supports getting credentials from kubernetes, however it [caches these credentials on the initial image pull and stores them forever](https://github.com/awslabs/soci-snapshotter/blob/main/service/keychain/cri/cri.go#L129). It's possible for these credentials to rotate during the life of a container workload which would result in further remote requests failing.

### Describe the solution you'd like

The snapshotter has some mechanism for refreshing kubernetes credentials.

PR #1927, opened 2026-04-06
Invalidate cached registry hosts to re-auth on layer check failure

prafgup:
**Issue #, if available:**

Addresses - https://github.com/awslabs/soci-snapshotter/issues/1924 and help with https://github.com/awslabs/soci-snapshotter/issues/522

**Description of changes:**

When a layer check fails and refresh with cached sources also fails, invalidate the cached registry host configurations and retry with fresh credentials. This handles cases where cached auth tokens have expired between the initial pull and a subsequent layer access as seen in the issue https://github.com/awslabs/soci-snapshotter/issues/1924

Prev discussions - 
- https://github.com/awslabs/soci-snapshotter/pull/1925#issuecomment-4185372117
- https://github.com/awslabs/soci-snapshotter/issues/1924

**Testing performed:**

Added Unit tests, and performed a exhaustive integration test as part of the draft testing via https://github.com/awslabs/soci-snapshotter/pull/1925

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -165,6 +165,7 @@ type Option func(*options)
 type options struct {
 	getSources        source.GetSources
 	resolveHandlers   map[string]remote.Handler
+	invalidateHosts   source.InvalidateHosts
 	metadataStore     metadata.Store
 	overlayOpaqueType layer.OverlayOpaqueType
 	maxConcurrency    int64
@@ -174,6 +175,12 @@ type options struct {
 func WithGetSources(s source.GetSources) Option {
 	return func(opts *options) {
 		opts.getSources = s
+	}
+}
+
+func WithInvalidateHosts(invalidateHosts source.InvalidateHosts) Option {
+	return func(opts *options) {
+		opts.invalidateHosts = invalidateHosts
 	}
 }
 
@@ -320,6 +327,7 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 		ctx:                         ctx,
 		resolver:                    r,
 		getSources:                  getSources,
+		invalidateHosts:             fsOpts.invalidateHosts,
 		debug:                       cfg.Debug,
 		layer:                       make(map[string]layer.Layer),
 		disableVerification:         cfg.DisableVerification,
@@ -414,6 +422,7 @@ type filesystem struct {
 	layerMu                     sync.Mutex
 	disableVerification         bool
 	getSources                  source.GetSources
+	invalidateHosts             source.InvalidateHosts
 	metricsController           *layermetrics.Controller
 	attrTimeout                 time.Duration
 	entryTimeout                time.Duration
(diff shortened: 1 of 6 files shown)


prafgup:
Hey @sondavidb as per our discussion (https://github.com/awslabs/soci-snapshotter/pull/1925#issuecomment-4185372117) I have made the minimal change to help us with credential rotation in case of Layer check failures to help with https://github.com/awslabs/soci-snapshotter/issues/1924. Can you please review the same (and maybe @coderbirju as he has some previous context recently 😄)?

ps. Would really appreciate if this could be released as a new tagged minor version once merged. (We have a setup that uses the tagged version's binaries when building AMIs and its not viable to build custom SOCI from commit (we do build soci from commit for the standalone convert path but these are very different workflows), thus a tagged release would really help us here to unblock our development) (also thanks for the prev v0.13.0 release 🙏 )

In the meantime I am also exploring other paths if invalidating the cache there as well might be useful (eg. Mount, MountParallel, MountLocal etc..) will followup later if I find anything useful there (draft experimentation in https://github.com/prafgup/soci-snapshotter/pull/2 , but it seems it would improve reliability even more).

prafgup:
Note: Although this would help a bit, but I am seeing the original unavailable issue again. 
- One of the other issue I identified is, when a layer is shared among 2 images (A and B) but A was pulled first and the creds were invalidated, when B is pulled it tries to use A's invalid creds to pull/check the shared layer while it should be using B's creds now.

  - Here we walk the parent when calling Prepare which calls checkAvailability (which would only see the old snapshot stored label of Image A) -
    - https://github.com/awslabs/soci-snapshotter/blob/804dd7b93c96ec33604d50d67070055d86928d33/snapshot/snapshot.go#L1070-L1091
   - Also seems like our registry is not returning `Www-Authenticate` header on 401 (with correct cached username login) leading to failure. We might have to fix this on our registry end (or have some hack like this) - 
     -   Note: I have fixed this issue for our registry to always return `Www-Authenticate` header on 401 and now the `"failed to find supported auth scheme: not implemented"` is no longer seen but instead `unexpected status code on redirect 401` which is a auth issue of using old creds.
     -  https://github.com/awslabs/soci-snapshotter/blob/804dd7b93c96ec33604d50d67070055d86928d33/service/resolver/client.go#L249-L259

prafgup:
Currently would be testing this which is built on top of the current PR - https://github.com/prafgup/soci-snapshotter/pull/3 (detailed explanation in the description)
- It might resolve the issue mentioned in the above comment - https://github.com/awslabs/soci-snapshotter/pull/1927#issuecomment-4199212435

sondavidb:
> when a layer is shared among 2 images (A and B) but A was pulled first and the creds were invalidated, when B is pulled it tries to use A's invalid creds to pull/check the shared layer while it should be using B's creds now.

Yeah, this is a side effect of reusing content from shasum. Works great for ahead-of-time pulls but not lazy-loading pulls, as you're seeing now.

I don't know if there's an elegant solution here — the crux of the issue is that we need to be using separate sets of credentials for the same content. We could probably store a set of creds for each ref, and we can probably reuse some form of the mirroring logic where we just try different sets of credentials and invalidate/refresh nonexistant ones.

prafgup:
> We could probably store a set of creds for each ref, and we can probably reuse some form of the mirroring logic where we just try different sets of credentials and invalidate/refresh nonexistant ones.

I believe this would get a bit complex overall 🤔 and a bit hard to test/maintain. But definitely a solution to consider. (I would explore that solution once we have a good enough patch for this, I think the mirror solution might help us eliminate race conditions and issues due to churn)

@sondavidb Just to update I am currently testing https://github.com/prafgup/soci-snapshotter/pull/3 and it seems promising atleast till now (I am able to launch pods with same images on the same node, as well as diff images sharing layers). Testing it further and would let it bake for sometime to confirm 😄 

Can you maybe check if https://github.com/prafgup/soci-snapshotter/pull/3 seems alright to you? (note: its built on top of this PR, we would need both)

prafgup:
I let the https://github.com/prafgup/soci-snapshotter/pull/3 change bake for multiple hours and I haven't seen any `layer "XXX" unavailable: unavailable` errors 😄  or any failing/stuck soci enabled pod launches.

Seems like a decent patch overall for k8s credential refreshing for now.

Update: Seeing same unavailable issue again, seems like Prepare call is skipped in case snapshot already exists, instead Mount is called directly without Prepare. I am researching if we can easily make a credential rotation mechanism (list of creds from cri keychain, cached per ref on PullImage call)

Issue #1843, opened 2026-01-14
[FEATURE] Allow both parallel pull and lazy pull


han-steve:
### Description

When `parallel_pull_unpack.enable=false` (lazy loading mode) and no SOCI index exists for an image, the snapshotter defers to containerd's standard pull. This results in very slow sequential downloads.

We can't just enable parallel pull for everything because we want lazy loading for images that do have indexes.

### Describe the solution you'd like

Add a config option to use SOCI's parallel pull as a fallback when no index is found, instead of deferring to containerd.

Something like:
```toml
[pull_modes.parallel_pull_unpack]
  enable = false
  fallback_on_no_index = true
```

### Describe any alternative solutions/features you've considered

As a stretch goal, for indexed images, it would be great to combine lazy loading with parallel background prefetch. Container starts instantly via FUSE mount, while the rest of the image downloads in parallel in the background. This would give us both fast startup and fast library imports.

sondavidb:
Hey, thanks for opening this issue!

TL;DR It is technically possible but we are not confident in enabling this feature quite yet due to the gaps in our containerd content store interactions.

I think this is a valid use case, and in theory the codepath to enable this isn't difficult. However I think there's implications here that go beyond the code change that makes me hesitant to immediately enable it.

The biggest issue is the technical side, where the lazy-loading pull mode prefers the SOCI content store, while parallel pulling requires containerd content store. The obvious solution here is to use the containerd content store, but we don't have confidence that we have full compatibility with lazy-loading and the containerd content store just yet. This is mostly due to the garbage collector. When pulling content we should link back to the manifest, which ties to the image, so that all resources can be removed when the image is removed. However since we link to the root this isn't done automatically, and I'm not sure the snapshotter should call back to the manifest (we probably could but it might not be best practices — still, it might be the best thing to do for a seamless workflow? This is TBD). We also probably need more rigorous testing, in particular with CRI/general Kubernetes workflows.

There's also the product side where we want to see what it would look like while maintaining backwards compatibility, but your suggestion seems to be a good compromise and I like it, so at least we have options there.

One thing that's a possibility is having the flag be experimental while we work to bridge the gap. I can't make a guarantee we will want to do that but at least it would provide users with the option to do so with the implication that it doesn't come with the same stability promises we try to deliver with the rest of SOCI.

Feel free to submit a PR for any of the technical parts — if we can close the technical gap between the SOCI content store + containerd content store it could give us more confidence in enabling this feature down the road.

[...]

PR #1938, opened 2026-04-12
feat: add parallel_pull_as_fallback config option

deeppcs:
## Summary

Adds `parallel_pull_as_fallback` option under `[pull_modes.parallel_pull_unpack]` that enables parallel-pull as an automatic fallback when lazy-load finds no SOCI index - instead of deferring to the container runtime's slower sequential pull.

## Problem

Lazy-load and parallel-pull are mutually exclusive today. When lazy-load is enabled and no SOCI index exists, the snapshotter defers to containerd's sequential pull (5-40% slower than Docker). This forces operators to choose one mode for all images.

## Solution

With `parallel_pull_as_fallback = true` (and `enable = false`):
- Images **with** a SOCI index → lazy-load
- Images **without** a SOCI index → parallel-pull fallback
- No image hits the slow sequential path

Defaults to `false` - zero behavior change for existing users.

With parallel_pull_as_fallback = true (and enable = false), the snapshotter first attempts lazy-load for every image. Only when no SOCI index is found does it fall back to parallel-pull instead of the slow sequential path. This gives optimal performance for both cases - images with and without SOCI index.

The option defaults to false, preserving existing behavior for all current users. When enable = true, the fallback is a no-op since parallel-pull is already the primary mode.

## Config Example

```toml
[pull_modes.soci_v2]
  enable = true

  enable = false
  parallel_pull_as_fallback = true
  max_concurrent_downloads_per_image = 10
  concurrent_download_chunk_size = "16mb"
  max_concurrent_unpacks_per_image = 10
  discard_unpacked_layers = true
```

## Testing
Tested on AL2 and AL2023 instances with both small (nginx) and large (20GB+) container images. All existing unit tests pass.

| `enable` | `parallel_pull_as_fallback` | Result |
|---|---|---|
| `false` | `false` (default) | Default behavior - lazy-load, sequential fallback on no index |
| `false` | `true` | **New behavior** - lazy-load, parallel-pull fallback on no index |
| `true` | `false` (default) | Parallel-pull primary, skips lazy-load entirely (existing) |
| `true` | `true` | Same as above - `enable` wins, fallback flag is ignored |

### DIFF
--- a/config/pull_modes.go
+++ b/config/pull_modes.go
@@ -44,6 +44,19 @@ type V2 struct {
 type Parallel struct {
 	ParallelConfig
 	Enable bool `toml:"enable"`
+
+	// ParallelPullAsFallback enables parallel-pull as an automatic fallback
+	// when lazy-load is the primary mode but no SOCI index is found for an image.
+	// When true (and Enable is false), the snapshotter will first attempt lazy-load;
+	// if no SOCI index exists, it falls back to parallel-pull instead of deferring
+	// to the container runtime's slower sequential pull.
+	// If Enable is true, this option is a no-op (parallel-pull is already the primary mode).
+	//
+	// EXPERIMENTAL: This requires the containerd content store for both lazy-load
+	// and parallel-pull (unless discard_unpacked_layers = true).
+	// Lazy-load with the containerd content store may have
+	// garbage collection edge cases. See https://github.com/awslabs/soci-snapshotter/issues/1843
+	ExperimentalParallelPullAsFallback bool `toml:"experimental_parallel_pull_as_fallback"`
 }
 
 func defaultPullModes(cfg *Config) error {
--- a/docs/config.md
+++ b/docs/config.md
@@ -97,6 +97,24 @@ This set of variables must be at the top of your TOML file due to not belonging 
 
 ## config/service.go
 
+## config/pull_modes.go
+
+### [pull_modes.soci_v1]
+- `enable` (bool) — Enables SOCI v1 index discovery via the OCI Referrers API. Default: false.
+
+### [pull_modes.soci_v2]
+- `enable` (bool) — Enables SOCI v2 index discovery via image manifest annotations. Default: true.
+
+### [pull_modes.parallel_pull_unpack]
+- `enable` (bool) — Enables parallel pull and unpack as the primary pull mode. When true, lazy-load is skipped entirely. Default: false.
+- `experimental_parallel_pull_as_fallback` (bool) — **[EXPERIMENTAL]** When true (and `enable` is false), uses parallel-pull as an automatic fallback when lazy-load is the primary mode but no SOCI index is found for an image. Requires containerd content store (unless `discard_unpacked_layers = true`). Lazy-load with the containerd content store may have garbage collection edge cases. See [#1843](https://github.com/awslabs/soci-snapshotter/issues/1843). Ignored when `enable` is true. Default: false.
+- `max_concurrent_downloads` (int) — Max concurrent downloads across all images. -1 for unlimited. Default: -1.
+- `max_concurrent_downloads_per_image` (int) — Max concurrent downloads per image. Default: 3.
+- `concurrent_download_chunk_size` (string) — Size of each download chunk (e.g. "8mb", "16mb"). Empty means full layer. Default: "".
+- `max_concurrent_unpacks` (int) — Max concurrent unpacks across all images. -1 for unlimited. Default: -1.
+- `max_concurrent_unpacks_per_image` (int) — Max concurrent unpacks per image. Default: 1.
+- `discard_unpacked_layers` (bool) — Discard layer blobs after unpacking to save disk space. Default: false.
+
 ### [snapshotter]
 - `min_layer_size` (int) — Sets the minimum threshold for lazy loading a layer. Any layer smaller than this value will ignore the zTOC for the layer and pull the entire layer ahead of time. We generally recommend setting it to 10MiB (10000000). Default: 0.
 - `allow_invalid_mounts_on_restart` (bool) — Allows the snapshotter to start even if preexisting snapshots cannot connect to their data source on startup. Useful on unexpected daemon crashes/corruption. Default: false.
--- a/docs/parallel-mode.md
+++ b/docs/parallel-mode.md
@@ -94,6 +94,32 @@ If you have any questions or need further assistance, please don't hesitate to r
 
 ## Known Limitations
 
+### Parallel Pull as Fallback for Lazy-Load
+
+When using lazy-load as the primary pull mode (`soci_v2.enable = true` or `soci_v1.enable = true`), images without a SOCI index normally fall back to the container runtime's sequential pull, which can be slower than a standard image pull. To avoid this behavior, you can enable `experimental_parallel_pull_as_fallback`:
+
+```toml
+[pull_modes.soci_v2]
+  enable = true
+
+[pull_modes.parallel_pull_unpack]
+  enable = false
+  experimental_parallel_pull_as_fallback = true
+  max_concurrent_downloads_per_image = 10
+  concurrent_download_chunk_size = "16mb"
+  max_concurrent_unpacks_per_image = 10
+  discard_unpacked_layers = true
+```
+
+With this configuration:
+- Images **with** a SOCI index use lazy-load
+- Images **without** a SOCI index use parallel-pull
+- No image falls through to the slow sequential containerd pull
+
+> **EXPERIMENTAL**: This option requires the containerd content store (`type = "containerd"` under `[content_store]`) for both lazy-load and parallel-pull. Lazy-load with the containerd content store may have garbage collection edge cases and does not carry the same stability guarantees as using either mode independently. See [#1843](https://github.com/awslabs/soci-snapshotter/issues/1843) for details.
+
+Note: `experimental_parallel_pull_as_fallback` is ignored when `enable = true`, since parallel-pull is already the primary mode in that case.
+
 ### Registries
 
 Any registry that supports ranged GET requests and has sufficient request limits should work with parallel pull mode. If a registry is rate limiting image pull requests, users can attempt to lower `max_concurrent_downloads` or `max_concurrent_downloads_per_image` and see if it alleviates the issue, however this will result in less of a performance benefit compared to regular pulling.
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -242,7 +242,12 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 	if pullModes.Parallel.Enable &&
 		cfg.ContentStoreConfig.Type != config.ContainerdContentStoreType &&
 		!pullModes.Parallel.DiscardUnpackedLayers {
-		return nil, errors.New("parallel_pull_unpack mode requires containerd content store (type=\"containerd\" under [content_store])")
+		return nil, errors.New("parallel_pull_unpack mode requires containerd content store (type=\"containerd\" under [content_store] or discard_unpacked_layers = true)")
+	}
+	if pullModes.Parallel.ExperimentalParallelPullAsFallback &&
+		cfg.ContentStoreConfig.Type != config.ContainerdContentStoreType &&
+		!pullModes.Parallel.DiscardUnpackedLayers {
+		return nil, errors.New("experimental_parallel_pull_as_fallback requires containerd content store (type=\"containerd\" under [content_store] or discard_unpacked_layers = true)")
 	}
 	client := store.NewContainerdClient(cfg.ContentStoreConfig.ContainerdAddress)
 
@@ -339,7 +344,7 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 }
 
 func createParallelPullStructs(ctx context.Context, storage LayerUnpackJobStorage, parallelConfig *config.Parallel) (*unpackJobs, error) {
-	if !parallelConfig.Enable {
+	if !parallelConfig.Enable && !parallelConfig.ExperimentalParallelPullAsFallback {
 		return nil, nil
 	}
 
@@ -430,7 +435,7 @@ type filesystem struct {
 }
 
 func (fs *filesystem) MountParallel(ctx context.Context, mountpoint string, labels map[string]string, mounts []mount.Mount) error {
-	if !fs.pullModes.Parallel.Enable {
+	if !fs.pullModes.Parallel.Enable && !fs.pullModes.Parallel.ExperimentalParallelPullAsFallback {
 		return ErrParallelPullIsDisabled
 	}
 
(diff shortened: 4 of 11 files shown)


snapshot/snapshot.go
```diff
@@ -435,7 +447,24 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 	// pulling and unpacking all the layers in the image.
 	// The exception is if we are using parallel pull and unpack,
 	// in which case we want to handle all snapshots ourselves.
+	// When parallelPullAsFallback is enabled, we use parallel pull instead of
+	// deferring to the container runtime's slower sequential pull.
 	if deferToContainerRuntime {
```
ayush-panta, line 453: The changes here duplicate error handling below (if/else + error handling starting line 472. I think you can just fallback in this block. Consider something like this:
```
	// Lazy loading did not succeed. If no SOCI index was found and neither
	// parallel pull nor parallel-pull-as-fallback is enabled, defer to the
	// container runtime for a sequential pull. Otherwise, fall through to
	// parallel pull.
	if deferToContainerRuntime {
		if o.parallelPullUnpack || o.parallelPullAsFallback {
			log.G(lCtx).WithField("layerDigest", base.Labels[ctdsnapshotters.TargetLayerDigestLabel]).
				Info("no SOCI index found; falling back to parallel pull/unpack")
		} else {
			log.G(lCtx).WithField(deferredSnapshotLogKey, prepareSucceeded).WithError(err).Warnf("%v; %v", ErrNoIndex, ErrDeferToContainerRuntime)
			return mounts, nil
		}
	}

	// Lazy loading did not succeed — either no SOCI index was found
	// (parallel pull enabled, fell through above) or a SOCI index exists
	// but this layer has no ztoc (ErrNoZtoc). Use parallel pull if
	// enabled (either as primary or fallback), otherwise SOCI handles
	// the sequential unpack itself.
	if o.parallelPullUnpack || o.parallelPullAsFallback {
		log.G(ctx).WithField("layerDigest", base.Labels[ctdsnapshotters.TargetLayerDigestLabel]).Info("preparing snapshot with parallel pull/unpack")
		err = o.prepareParallelPullSnapshot(lCtx, key, base.Labels, mounts)
	} else {
		log.G(ctx).WithField("layerDigest", base.Labels[ctdsnapshotters.TargetLayerDigestLabel]).Info("preparing snapshot as local snapshot")
		err = o.prepareLocalSnapshot(lCtx, key, base.Labels, mounts)
		if err == nil {
			base.Labels[source.HasSociIndexDigest] = "true"
		}
	}
```

sondavidb:
Please see #1843, I've given my thoughts on this already there, but TL;DR I am not comfortable with putting out this change because we cannot give the same stability guarantees. If we can make it very clear that this can and will break at times, I'd be okay with this.

docs/design-docs/parallel-pull-fallback.md
sondavidb, line 1: Not necessary, would prefer to delete unless you have a strong reason to keep this. The CLI command documentation is already more than sufficient.

docs/parallel-mode.md
```diff
@@ -94,6 +94,30 @@ If you have any questions or need further assistance, please don't hesitate to r
...
+[pull_modes.parallel_pull_unpack]
+  enable = false
+  parallel_pull_as_fallback = true
+  max_concurrent_downloads_per_image = 10
+  concurrent_download_chunk_size = "16mb"
+  max_concurrent_unpacks_per_image = 10
+  discard_unpacked_layers = true
+```
+
+With this configuration:
```
sondavidb, line 114: Please remove the performance numbers as they are not applicable to every setup and are highly workflow-dependent. (And it's generally bad practice to advertise numbers for a general-use tool such as SOCI where environments vary wildly.)

The numbers for lazy-loading are also not accurate as there is no real pulling step. In exchange, every FS operation during the workload will take a little more time as it needs to be fetched on demand and go through FUSE. Hence how much faster it is is highly dependent on your workflow.

docs/config.md
```diff
@@ -97,6 +97,24 @@ This set of variables must be at the top of your TOML file due to not belonging
...
+
+### [pull_modes.parallel_pull_unpack]
+- `enable` (bool) — Enables parallel pull and unpack as the primary pull mode. When true, lazy-load is skipped entirely. Default: false.
+- `parallel_pull_as_fallback` (bool) — When true (and `enable` is false), uses parallel-pull as an automatic fallback when lazy-load is the primary mode but no SOCI index is found for an image. This avoids the slow sequential containerd pull that occurs when lazy-load is enabled and no SOCI index exists. Ignored when `enable` is true. Default: false.
+- `max_concurrent_downloads` (int) — Max concurrent downloads across all images. -1 for unlimited. Default: -1.
+- `max_concurrent_downloads_per_image` (int) — Max concurrent downloads per image. Default: 3.
+- `concurrent_download_chunk_size` (string) — Size of each download chunk (e.g. "8mb", "16mb"). Empty means full layer. Default: "".
+- `max_concurrent_unpacks` (int) — Max concurrent unpacks across all images. -1 for unlimited. Default: -1.
+- `max_concurrent_unpacks_per_image` (int) — Max concurrent unpacks per image. Default: 1.
+- `discard_unpacked_layers` (bool) — Discard layer blobs after unpacking to save disk space. Default: false.
```
sondavidb, line 116: Worth noting that turning it on can also help performance for environments that are throttled by disk throughput

docs/config.md
```diff
@@ -97,6 +97,24 @@ This set of variables must be at the top of your TOML file due to not belonging
...
+
+### [pull_modes.soci_v1]
+- `enable` (bool) — Enables SOCI v1 index discovery via the OCI Referrers API. Default: false.
+
+### [pull_modes.soci_v2]
+- `enable` (bool) — Enables SOCI v2 index discovery via image manifest annotations. Default: true.
+
+### [pull_modes.parallel_pull_unpack]
+- `enable` (bool) — Enables parallel pull and unpack as the primary pull mode. When true, lazy-load is skipped entirely. Default: false.
+- `experimental_parallel_pull_as_fallback` (bool) — **[EXPERIMENTAL]** When true (and `enable` is false), uses parallel-pull as an automatic fallback when lazy-load is the primary mode but no SOCI index is found for an image. Requires containerd content store. Can also help performance for environments that are throttled by disk throughput. Lazy-load with the containerd content store may have garbage collection edge cases. See [#1843](https://github.com/awslabs/soci-snapshotter/issues/1843). Ignored when `enable` is true. Default: false.
```
sondavidb: How does this help in environments throttled by disk throughput?

docs/parallel-mode.md
```diff
@@ -94,6 +94,32 @@ If you have any questions or need further assistance, please don't hesitate to r
 
 ## Known Limitations
 
+### Parallel Pull as Fallback for Lazy-Load
+
+When using lazy-load as the primary pull mode (`soci_v2.enable = true`), images without a SOCI index normally fall back to the container runtime's sequential pull, which can be slower than a standard Docker pull. To avoid this regression, you can enable `experimental_parallel_pull_as_fallback`:
```
sondavidb: Can also work with SOCIv1 (at least with the changes that have been made). Also some minor wording change suggestions.

```suggestion
When using lazy-load as the primary pull mode (`soci_v2.enable = true` or `soci_v1.enable = true`), images without a SOCI index normally fall back to the container runtime's sequential pull, which can be slower than a standard image pull. To avoid this behavior, you can enable `experimental_parallel_pull_as_fallback`:
```

integration/pull_modes_test.go
```diff
@@ -364,3 +364,141 @@ func testDanglingV2Annotation(t *testing.T, imgName string) {
...
+	noIndexRef, err := reference.Parse(noIndexInfo.ref)
+	if err != nil {
+		t.Fatal(err)
+	}
+	noIndexRef.Object = ":no-soci-index"
+	sh.X("nerdctl", "pull", "-q", srcInfo.ref)
+	sh.X("nerdctl", "image", "tag", srcInfo.ref, noIndexRef.String())
+	sh.X("nerdctl", "push", noIndexRef.String())
+
+	tests := []struct {
```
sondavidb, line 401: Should we also add tests to check if we disallow snapshotter startup with containerd content store? (I forget if our testing suite has this flexibility, but if not you can also check if you can load this with `discard_unpacked_layers = true` and make sure that passes.)

config/pull_modes.go
```diff
@@ -44,6 +44,18 @@ type V2 struct {
...
 	Enable bool `toml:"enable"`
+
+	// ParallelPullAsFallback enables parallel-pull as an automatic fallback
+	// when lazy-load is the primary mode but no SOCI index is found for an image.
+	// When true (and Enable is false), the snapshotter will first attempt lazy-load;
+	// if no SOCI index exists, it falls back to parallel-pull instead of deferring
+	// to the container runtime's slower sequential pull.
+	// If Enable is true, this option is a no-op (parallel-pull is already the primary mode).
+	//
+	// EXPERIMENTAL: This requires the containerd content store for both lazy-load
```
sondavidb, line 55: Technically with `discard_unpacked_layers = true` this isn't necessarily true, we can use the SOCI content store. I missed mentioning this before, sorry about that.

snapshot/snapshot.go
```diff
@@ -433,14 +445,28 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 	// If the underlying FileSystem deems that the image is unable to be lazy loaded,
 	// then we should completely fallback to the container runtime to handle
 	// pulling and unpacking all the layers in the image.
-	// The exception is if we are using parallel pull and unpack,
+	// The exception is if we are using parallel pull and unpack (or parallel-pull-as-fallback),
 	// in which case we want to handle all snapshots ourselves.
+	// If no SOCI index was found and neither parallel pull nor parallel-pull-as-fallback
+	// is enabled, defer to the container runtime for a sequential pull.
+	// Otherwise, fall through to parallel pull.
 	if deferToContainerRuntime {
```
sondavidb, line 453: nit (and possibly out of scope): wonder if we should rename this since it's clear that it does not signify that we are deferring to container runtime, but rather that there was no SOCI index detected.

config/pull_modes.go
```diff
@@ -44,6 +44,19 @@ type V2 struct {
...
+	// and parallel-pull (or the SOCI content store with discard_unpacked_layers = true).
```
sondavidb:
```suggestion
	// and parallel-pull (unless discard_unpacked_layers = true).
```

docs/config.md
```diff
@@ -97,6 +97,24 @@ This set of variables must be at the top of your TOML file due to not belonging
...
+
+### [pull_modes.soci_v1]
+- `enable` (bool) — Enables SOCI v1 index discovery via the OCI Referrers API. Default: false.
+
+### [pull_modes.soci_v2]
+- `enable` (bool) — Enables SOCI v2 index discovery via image manifest annotations. Default: true.
+
+### [pull_modes.parallel_pull_unpack]
+- `enable` (bool) — Enables parallel pull and unpack as the primary pull mode. When true, lazy-load is skipped entirely. Default: false.
+- `experimental_parallel_pull_as_fallback` (bool) — **[EXPERIMENTAL]** When true (and `enable` is false), uses parallel-pull as an automatic fallback when lazy-load is the primary mode but no SOCI index is found for an image. Requires containerd content store (or SOCI content store with `discard_unpacked_layers = true`). Lazy-load with the containerd content store may have garbage collection edge cases. See [#1843](https://github.com/awslabs/soci-snapshotter/issues/1843). Ignored when `enable` is true. Default: false.
```
sondavidb:
```suggestion
- `experimental_parallel_pull_as_fallback` (bool) — **[EXPERIMENTAL]** When true (and `enable` is false), uses parallel-pull as an automatic fallback when lazy-load is the primary mode but no SOCI index is found for an image. Requires containerd content store (unless `discard_unpacked_layers = true`). Lazy-load with the containerd content store may have garbage collection edge cases. See [#1843](https://github.com/awslabs/soci-snapshotter/issues/1843). Ignored when `enable` is true. Default: false.
```

docs/config.md
sondavidb: Bump

fs/fs.go
```diff
@@ -242,7 +242,12 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
 	if pullModes.Parallel.Enable &&
 		cfg.ContentStoreConfig.Type != config.ContainerdContentStoreType &&
 		!pullModes.Parallel.DiscardUnpackedLayers {
-		return nil, errors.New("parallel_pull_unpack mode requires containerd content store (type=\"containerd\" under [content_store])")
+		return nil, errors.New("parallel_pull_unpack mode requires containerd content store (type=\"containerd\" under [content_store]) or discard_unpacked_layers = true")
```
sondavidb:
```suggestion
		return nil, errors.New("parallel_pull_unpack mode requires containerd content store (type=\"containerd\" under [content_store] or discard_unpacked_layers = true)")
```

fs/fs.go
```diff
@@ -242,7 +242,12 @@ func NewFilesystem(ctx context.Context, root string, cfg config.FSConfig, opts .
...
+	}
+	if pullModes.Parallel.ExperimentalParallelPullAsFallback &&
+		cfg.ContentStoreConfig.Type != config.ContainerdContentStoreType &&
+		!pullModes.Parallel.DiscardUnpackedLayers {
+		return nil, errors.New("experimental_parallel_pull_as_fallback requires containerd content store (type=\"containerd\" under [content_store]) or discard_unpacked_layers = true")
```
sondavidb:
```suggestion
		return nil, errors.New("experimental_parallel_pull_as_fallback requires containerd content store (type=\"containerd\" under [content_store] or discard_unpacked_layers = true"))
```

sondavidb (commented):
LGTM just minor formatting stuff. Can run CI soon as well

sondavidb (dismissed):

ayush-panta (dismissed):

integration/pull_modes_test.go
```diff
@@ -364,3 +364,159 @@ func testDanglingV2Annotation(t *testing.T, imgName string) {
...
+	v2IndexDigest := createAndPushV2Index(t, sh, srcInfo, dstInfo)
+
+	// Also push the original (non-indexed) image to a separate tag for fallback tests
+	rebootContainerd(t, sh, "", "")
+	noIndexInfo := regConfig.mirror(imgName)
+	noIndexRef, err := reference.Parse(noIndexInfo.ref)
+	if err != nil {
+		t.Fatal(err)
+	}
+	noIndexRef.Object = ":no-soci-index"
```
sondavidb:
```suggestion
	noIndexRef.Object = "no-soci-index"
```

sondavidb (commented):
CI failing, this should hopefully make the test case pass

ayush-panta (approved):

sondavidb (approved):

PR #1953, opened 2026-05-01
Upgrade to go 1.25.9

prafgup:
**Issue #, if available:**

**Description of changes:**

Upgrade go version to latest 1.25.9 as it fixes a lot of CVEs (this was being flagged by our systems as this old version of go had a lot of critical CVEs)

Few CVEs - 
- https://nvd.nist.gov/vuln/detail/CVE-2026-27143
- https://nvd.nist.gov/vuln/detail/CVE-2026-32281
- https://nvd.nist.gov/vuln/detail/CVE-2026-32283
-  \+ 3 more

**Testing performed:**

```
prafulg@lima-default:/Volumes/git/prafulg/soci-snapshotter$ sudo make test
test-lib
ok      github.com/awslabs/soci-snapshotter/cache       1.016s
ok      github.com/awslabs/soci-snapshotter/config      1.011s
ok      github.com/awslabs/soci-snapshotter/fs  2.763s
ok      github.com/awslabs/soci-snapshotter/fs/backgroundfetcher        22.558s
ok      github.com/awslabs/soci-snapshotter/fs/layer    9.042s
?       github.com/awslabs/soci-snapshotter/fs/metrics/common   [no test files]
?       github.com/awslabs/soci-snapshotter/fs/metrics/layer    [no test files]
ok      github.com/awslabs/soci-snapshotter/fs/reader   7.354s
ok      github.com/awslabs/soci-snapshotter/fs/remote   8.140s
?       github.com/awslabs/soci-snapshotter/fs/source   [no test files]
ok      github.com/awslabs/soci-snapshotter/fs/span-manager     19.149s
ok      github.com/awslabs/soci-snapshotter/idtools     1.006s
ok      github.com/awslabs/soci-snapshotter/integration 1.019s
ok      github.com/awslabs/soci-snapshotter/internal/archive/compression        1.012s
ok      github.com/awslabs/soci-snapshotter/internal/http       1.007s
ok      github.com/awslabs/soci-snapshotter/internal/os 1.007s
ok      github.com/awslabs/soci-snapshotter/metadata    4.932s
?       github.com/awslabs/soci-snapshotter/service     [no test files]
?       github.com/awslabs/soci-snapshotter/service/keychain/cri/v1     [no test files]
?       github.com/awslabs/soci-snapshotter/service/keychain/cri/v1alpha        [no test files]
?       github.com/awslabs/soci-snapshotter/service/keychain/dockerconfig       [no test files]
[...]

PR #1957, opened 2026-05-08
update go toolchain version to 1.26.3

ayush-panta:
**Description of changes:** Update all references of golang to 1.26.3. Also updated `GOLANGCI_LINT_VERSION` to 2.12.2 and fixed a few minor linter issues.

**Testing performed:** Testing with CI. May need to test performance due to new Green Tea GC.


sondavidb:
Please see my comment [here](https://github.com/awslabs/soci-snapshotter/pull/1953#issuecomment-4360699896) for bumping the go.mod version (we really shouldn't touch the version unless necessary, though we can use the toolchain directive if we really wanted to), if this is to patch CVE's let's do Go 1.25.10.

Also, if we are fixing linting issues, can we separate that commit from the Golang upgrade commit?

PR #1973, opened 2026-05-20
Properly find preload files with absolute path

### LINKED ISSUE #1944: [Bug] prefetch-file does not work when having "/" as prefix for files
### Description

Looking at the docs - https://github.com/awslabs/soci-snapshotter/blob/main/docs/prefetch.md#cli-flags
 
Seems like matching logic is broken here - https://github.com/awslabs/soci-snapshotter/blob/f1fcd05ffd3c99eae1e6d6c69c992c428d54b425/soci/soci_index.go#L777-L779

 ### Leading slash (as docs say) - fails                                                                                                                          
```
sudo ./out/soci convert --platform linux/amd64 --min-layer-size 0 \
--prefetch-file /etc/alpine-release --prefetch-file /etc/os-release \
docker.io/library/alpine:3.19 docker.io/library/alpine:3.19-soci        
                                                                                   
layer sha256:17a39c0ba978cc27001e9c56a480f98106e1ab74bd56eb302f9fd4cf758ea43f -> ztoc sha256:82f0be14270614df7c360a289796b1e75b0b17bf9313fffcb72f0cad3451d264 (already exists)
WARN[0000] prefetch file /etc/alpine-release not found in any layer 
WARN[0000] prefetch file /etc/os-release not found in any layer 
```                                                                                     
                                                                                                                                                                 
 ### No leading slash - works      
```                                
sudo ./out/soci convert --platform linux/amd64 --min-layer-size 0 \
--prefetch-file etc/alpine-release --prefetch-file etc/os-release \
docker.io/library/alpine:3.19 docker.io/library/alpine:3.19-soci2

layer sha256:17a39c0ba978cc27001e9c56a480f98106e1ab74bd56eb302f9fd4cf758ea43f -> ztoc sha256:82f0be14270614df7c360a289796b1e75b0b17bf9313fffcb72f0cad3451d264 (already exists)
Created prefetch artifact for layer sha256:17a39c0ba978cc27001e9c56a480f98106e1ab74bd56eb302f9fd4cf758ea43f: sha256:520900f4a24d9f67823edbe61838eec1b7bd8c64667f0a577a264715d4e9a37b
```

Same with Standalone convert - 

 ### Leading slash - fails    
```
sudo ./out/soci convert --standalone --min-layer-size 0 --prefetch-file /etc/alpine-release --prefetch-file /etc/os-release --format oci-dir /tmp/alpine.tar /tmp/alpine-soci

layer sha256:17a39c0ba978cc27001e9c56a480f98106e1ab74bd56eb302f9fd4cf758ea43f -> ztoc sha256:82f0be14270614df7c360a289796b1e75b0b17bf9313fffcb72f0cad3451d264

WARN[0000] prefetch file /etc/alpine-release not found in any layer 
WARN[0000] prefetch file /etc/os-release not found in any layer
```

 ### No leading slash - works      
```
sudo ./out/soci convert \
--standalone --min-layer-size 0 \
--prefetch-file etc/alpine-release \
--prefetch-file etc/os-release \
--format oci-dir \
/tmp/alpine.tar /tmp/alpine-soci

layer sha256:17a39c0ba978cc27001e9c56a480f98106e1ab74bd56eb302f9fd4cf758ea43f -> ztoc sha256:82f0be14270614df7c360a289796b1e75b0b17bf9313fffcb72f0cad3451d264
Created prefetch artifact for layer sha256:17a39c0ba978cc27001e9c56a480f98106e1ab74bd56eb302f9fd4cf758ea43f: sha256:520900f4a24d9f67823edbe61838eec1b7bd8c64667f0a577a264715d4e9a37b
```

See description ^

Prefetching index creating should work when file have leading /

### Host information

1. OS: Lima on MacOs
2. Snapshotter Version: latest (0.13.x)
3. Containerd Version: not-required


_No response_

sondavidb:
**Issue #, if available:**
Closes #1944

**Description of changes:**
`filepath.Clean` does not remove the leading slash, so we would previously fail to add files to preloading if an absolute path was specified during the index creation process. Removing the leading slash manually should fix this behavior.

Added integration testing to vet this behavior as well. For completion's sake I also added unit testing but it is frankly not a very useful test as it always passes. We should allow the tests to catch and test these errors more rigorously, but this should at least make sure that in the future if we choose to do so, that the code is in place to accept these changes.

**Testing performed:**
Checked that the changes to `TestSociCreateWithPrefetchArtifacts` fails without the other changes and passes with them.

### DIFF
--- a/soci/soci_index.go
+++ b/soci/soci_index.go
@@ -590,7 +590,10 @@ func (b *IndexBuilder) build(ctx context.Context, img images.Image, buildCfg bui
 
 	var prefetchDescs []*ocispec.Descriptor
 	if len(b.config.prefetchPaths) > 0 && len(builtZtocs) > 0 {
-		prefetchDescs = b.buildPrefetchLayer(ctx, builtZtocs, manifest.Layers)
+		prefetchDescs, err = b.buildPrefetchLayer(ctx, builtZtocs, manifest.Layers)
+		if err != nil {
+			return nil, fmt.Errorf("failed to build prefetch layers: %w", err)
+		}
 	}
 
 	ztocsDesc := make([]ocispec.Descriptor, 0, len(sociLayersDesc)+len(prefetchDescs))
@@ -752,9 +755,9 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 }
 
 // buildPrefetchLayer builds and stores prefetch artifacts for layers that contain prefetch files
-func (b *IndexBuilder) buildPrefetchLayer(ctx context.Context, builtZtocs []*ztocWithLayer, layers []ocispec.Descriptor) []*ocispec.Descriptor {
+func (b *IndexBuilder) buildPrefetchLayer(ctx context.Context, builtZtocs []*ztocWithLayer, layers []ocispec.Descriptor) ([]*ocispec.Descriptor, error) {
 	if len(b.config.prefetchPaths) == 0 {
-		return nil
+		return nil, nil
 	}
 
 	layerZtocMap := make(map[string]*ztocWithLayer, len(builtZtocs))
@@ -765,6 +768,9 @@ func (b *IndexBuilder) buildPrefetchLayer(ctx context.Context, builtZtocs []*zto
 	layerPrefetchSpansMap := make(map[string][]PrefetchSpan)
 	for _, prefetchPath := range b.config.prefetchPaths {
 		found := false
+		// Strip leading slash if absolute path is given in prefetchPaths
+		// Copied from cleanEntryPath in metadata/reader.go
+		prefetchPath = strings.TrimPrefix(filepath.Clean(string(os.PathSeparator)+prefetchPath), string(os.PathSeparator))
 	PrefetchPathLoop:
 		for i := len(layers) - 1; i >= 0; i-- {
 			layer := layers[i]
@@ -775,9 +781,12 @@ func (b *IndexBuilder) buildPrefetchLayer(ctx context.Context, builtZtocs []*zto
 			}
 
 			for _, metadata := range zwl.ztoc.TOC.FileMetadata {
-				if filepath.Clean(metadata.Name) == prefetchPath {
+				// metadata name should already be cleaned when stored so we don't need to re-clean it
+				if metadata.Name == prefetchPath {
 					gz, err := zwl.ztoc.Zinfo()
 					if err != nil {
+						// The layer has issues but the prefetech path is potentially still findable.
+						// Proceed to search the next layer for the file.
 						log.G(ctx).WithError(err).Warnf("failed to get zinfo for layer %s", layerDigest)
 						break PrefetchPathLoop
 					}
@@ -806,7 +815,7 @@ func (b *IndexBuilder) buildPrefetchLayer(ctx context.Context, builtZtocs []*zto
 	}
 
 	if len(layerPrefetchSpansMap) == 0 {
-		return nil
+		return nil, nil
 	}
 
 	prefetchDescs := make([]*ocispec.Descriptor, 0, len(layerPrefetchSpansMap))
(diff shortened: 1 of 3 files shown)


sondavidb:
cc @ChengyuZhu6 @prafgup since you two were previously interested in this

soci/soci_index.go
```diff
@@ -752,9 +755,9 @@ func (b *IndexBuilder) buildSociLayer(ctx context.Context, desc ocispec.Descript
 }
 
 // buildPrefetchLayer builds and stores prefetch artifacts for layers that contain prefetch files
-func (b *IndexBuilder) buildPrefetchLayer(ctx context.Context, builtZtocs []*ztocWithLayer, layers []ocispec.Descriptor) []*ocispec.Descriptor {
+func (b *IndexBuilder) buildPrefetchLayer(ctx context.Context, builtZtocs []*ztocWithLayer, layers []ocispec.Descriptor) ([]*ocispec.Descriptor, error) {
 	if len(b.config.prefetchPaths) == 0 {
-		return nil
+		return nil, nil
 	}
```
prafgup, line 761: nit; logically I don't think this condition does anything as we are already doing a `len(b.config.prefetchPaths) > 0` check before calling `buildPrefetchLayer`

soci/soci_index.go
```diff
@@ -765,6 +768,10 @@ func (b *IndexBuilder) buildPrefetchLayer(ctx context.Context, builtZtocs []*zto
 	layerPrefetchSpansMap := make(map[string][]PrefetchSpan)
 	for _, prefetchPath := range b.config.prefetchPaths {
 		found := false
+		// Strip leading slash if absolute path is given in prefetchPaths
+		if prefetchPath[0] == '/' {
+			prefetchPath = prefetchPath[1:]
+		}
```
prafgup: nit; we can simply use `prefetchPath = strings.TrimLeft(prefetchPath, "/")` to trim all leading shashes or `strings.TrimPrefix(prefetchPath, "/")` to trim one slash.

prafgup (approved):
Overall LGTM, just one concern (haven't tested) but iirc the base image can have files rooted with `/` and not be relative (possibly possible for normal layers as well, so there could be a file `/a/b` in a layer), so now as we are trimming the prefetch path (from input, so `/a/b` becomes `a/b`) before matching with existing files, will this lead to skipping some files which were actually rooted with `/`? 🤔

sondavidb:
> will this lead to skipping some files which were actually rooted with /? 🤔

We don't store it with the metadata anyway, see [here](https://github.com/awslabs/soci-snapshotter/blob/cffa6253ee59b61683037adc9a57b466624c11f6/metadata/reader.go#L391) where we trim all files prefixed with `/`. I don't think this is an issue then?

soci/soci_index.go
sondavidb: Went a little further and decided to copy the code from [cleanEntryPath](https://github.com/awslabs/soci-snapshotter/blob/cffa6253ee59b61683037adc9a57b466624c11f6/metadata/reader.go#L391). I could have made the function public but given how simple it is I figured it would be best to just copy the logic over to avoid having too much of a dependency web. (I don't think it would have been a big deal but just to be safe and future-proof it a little.)

soci/soci_index.go
sondavidb: It's useful for unit testing since we call it directly [here](https://github.com/awslabs/soci-snapshotter/pull/1973/changes#diff-70654a93ca0493bfb90c25d2bb217aebdfb92c1929909a35e9b73f503476afedR554), and I think regardless of redundancy it's nice to have it near the top for clarity's sake. But maybe we can do a better job at being clear of the division of responsibility here.

prafgup:
Yeah, I think if metadata is stored with trimmed prefix and we are just matching both prefix trimmed paths, it should work fine :)

TIL - 
- Within a single tar, you can physically write two records with paths that normalize to the same name. tar doesn't forbid duplicates. The applier just processes them in order and the last write wins. containerd's archive.Apply doesn't treat this as an error.
- Across layers, even if layer N has a/b and layer N+1 has /a/b (pre-normalization), after the upper layer is applied it's just a/b shadowing a/b. No coexistence.


ps. was thinking what happens if 2 layers have same file paths `L1 -> a/b`, `L2 -> a/b` and the FS requests for `a/b` what would soci fetch in this case, L1 or L2 🤔 (or maybe even the same path `a/b` and `a/b` in L1 itself) maybe will deep dive sometime, but not related to this change.

sondavidb:
> was thinking what happens if 2 layers have same file paths L1 -> a/b, L2 -> a/b and the FS requests for a/b what would soci fetch in this case, L1 or L2 🤔

L2 will be fetched since we start from the [topmost layer](https://github.com/awslabs/soci-snapshotter/blob/cffa6253ee59b61683037adc9a57b466624c11f6/soci/soci_index.go#L769) (which makes sense, given that we probably don't want to load an old version of a file).

ChengyuZhu6 (approved):
LGTM

ayush-panta (approved):
lgtm

coderbirju (approved):

PR #1980, opened 2026-05-21
Enforce monotonic checkpoint offsets

sondavidb:
**Issue #, if available:**

**Description of changes:**
Our zTOC logic assumes monotonic checkpoint offsets for its logic to change but never enforced it. This change enforces this logic by forcing strictly increasing start and end of checkpoints.

There's technically no reason a checkpoint can't be zero length so we can maybe make the change to allow that but it's probably more sensical to not allow it to begin with.

**Testing performed:**

### DIFF
--- a/fs/span-manager/span_manager.go
+++ b/fs/span-manager/span_manager.go
@@ -36,10 +36,11 @@ import (
 
 // Specific error types raised by SpanManager.
 var (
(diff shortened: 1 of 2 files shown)


ayush-panta (approved):
LGTM.

coderbirju (approved):

PR #1248, opened 2024-05-23
Add create release branch workflow

austinvazquez:
**Issue #, if available:**
Follow-up to #1238 

On each release, there is a manual set of steps to generate the `THIRD_PARTY_LICENSES` file and set the version in the getting start guide on the new release branch. It would be nice to have some automation for creating the release branch and generating the changes needed so maintainers only have to approve and merge.

**Description of changes:**
This change adds a GitHub Actions workflow for creating a release branch and generating the pull request with the changes needed for `THIRD_PARTY_LICENSES` file and updating the getting started guide.

**Testing performed:**

**Alternatives considered:**
1. A variation where the changes are committed and pushed by automation instead of creating a pull request. This approach doesn't give maintainers a manual gate for approving the changes.

**Limitations:**
1. Similar to other workflows which create pull requests, the PR must be manually closed and reopened for workflow checks to run.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1992, opened 2026-05-29
Add workflow for creating release branches

sondavidb:
**Issue #, if available:**
Alterantive to #1248

**Description of changes:**
Building off of #1248, I wanted to add some additional automation to the whole process to make the artifact generation as smooth as possible for new minor version bumps, which the original PR did well. However, I also wanted to make sure that patch releases could be just as smooth as minor version releases. To achieve this, I ended up separating out the original workflow into three separate workflows.

The first workflow creates the branch and subsequently calls the other two workflows. One workflow automatically generates the THIRD_PARTY_LICENSES file and creates an automatic PR with the changes. It takes a version tag and puts a PR against the appropriate release branch. The other workflow updates our docs with the new version and creates an automatic PR with the changes, taking the same input and making the PR against the same branch.

The latter two workflows run on workflow_call to be called from the create branch, but they also run on workflow_dispatch and push actions as well. The former is to allow for reruns in case of any errors. The latter is to automatically run every time there is a new push to a release branch, which generally doesn't happen unless a new release is happening. In the case of the latter, it will auto-increment the semver's patch version by finding the previous tag and incrementing the patch version.

I also renamed our `update-getting-started-guide.yml` to `update-version-in-docs.yml` which is a little less smooth but more generic sounding, and updated our script to `update-version-in-docs.sh` as well.

The end result is a workflow that runs as follows for a new minor release:

- Create branch from workflow. Workflow takes a major.minor (i.e. if you want to create release/0.15 you would input 0.15), and an optional commit hash (if empty will take from tip of main)
- Run workflow
- Automation will create the workflow branch and additionally run the PR creation workflows

And for a patch release:
- Push changes to patch branch
- Automation will run the PR creation workflows

After this all that is needed is the tag push, which can maybe be automated down the line. Hopefully this makes updating for minor + patch releases smoother.

Thanks to @austinvazquez for working on this first!

**Testing performed:**
I wanted to test both a minor release and a patch release to see if it worked correctly.

On my [fork](https://github.com/sondavidb/soci-snapshotter) I did the following steps:
- Use the [workflow](https://github.com/sondavidb/soci-snapshotter/actions/runs/26664869737) to create a release branch `release/0.15`
- Merged the THIRD_PARTY_LICENSES [PR](https://github.com/sondavidb/soci-snapshotter/pull/113) and update docs [PR](https://github.com/sondavidb/soci-snapshotter/pull/112) into the release branch
- Cut a tag `v0.15.0` and published a new [release](https://github.com/sondavidb/soci-snapshotter/releases/tag/v0.15.0) for it
- Checked that the [workflow](https://github.com/sondavidb/soci-snapshotter/actions/runs/26665029607) created the correct [PR](https://github.com/sondavidb/soci-snapshotter/pull/114)
- Downgraded a random Golang package to force the third party licenses to need to change and pushed my [commit](https://github.com/sondavidb/soci-snapshotter/commit/8607d81a62071bc226c6fc30dc6b1ce82cd40c8a)
- Saw the licenses [workflow](https://github.com/sondavidb/soci-snapshotter/actions/runs/26665067481) created the correct [PR](https://github.com/sondavidb/soci-snapshotter/pull/116) and the docs [workflow](https://github.com/sondavidb/soci-snapshotter/actions/runs/26665067475) created the correct [PR](https://github.com/sondavidb/soci-snapshotter/pull/115)

EDIT: Had to move around some permissions to be safe to avoid PRs having write access. Seems to work even with changing this so that's good: http://github.com/sondavidb/soci-snapshotter/actions/runs/26666786897


austinvazquez:
Funny timing. I referenced some of this work recently. 😄

sondavidb:
> Funny timing. I referenced some of this work recently. 😄

Coincidence! Been quite a while so that's really funny. I was just bugged about making our release automation being smoother again and I remembered you did some work on automating it. Glad you had a good base so I could build off it instead of having to do it from scratch

.github/workflows/create-release-branch.yml
```diff
@@ -0,0 +1,108 @@
...
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-22.04
+
+    env:
+      VERSION: ''
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Mock workflow inputs on pull request
```
coderbirju, line 39: Can we use intermediate env variable following the [github guidelines](https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable) for preventing script injection attacks?

We did a similar thing in finch a few months ago - https://github.com/runfinch/finch-core/pull/686

.github/workflows/update-version-in-docs.yml
```diff
@@ -0,0 +1,128 @@
...
+      - 'scripts/update-version-in-docs.sh'
+      # Run workflow on changes to the getting started guide to validate
+      # changes to the documentation do not break the update workflow.
+      - 'docs/getting-started.md'
+      - 'docs/eks.md'
+  # Run workflow on all pushes to a release branch
+  push:
+    branches: ['release/**']
+
+permissions:
```
coderbirju: can we move this to the `update-version` block? 
Does `test-update-version` job need these permissions?

.github/workflows/create-release-branch.yml
sondavidb: I don't think it's useful here since all of the inputs come from GitHub webhooks. But I've added them in a couple other spots where we take user input (you would need write perms to even run the workflows anyway but always good to have extra layers of protection). I think it's not 100% the same way but our own trusted input (i.e. GH actions workflow content) should do it

.github/workflows/update-version-in-docs.yml
sondavidb: Yeah good catch thanks. No we don't need them on PR. I forgot to remove this block when moving around perms

coderbirju (approved):

ayush-panta (approved):

Issue #1016, opened 2023-12-29
[Bug] Fix synchronous fetch count metric


turan18:
### Description

We increment a metric(`SynchronousReadRegistryFetchCount`) every time we [need to fetch](https://github.com/awslabs/soci-snapshotter/blob/00b3501b309d13c8e00e8c244c55476943cdfbb6/fs/reader/reader.go#L243) bytes of content synchronously. This isn't really indicative of _registry fetches_ as we could have returned the content from our local cache. What we really need to know is the number of synchronous fetch operations that led to cache misses where we actually had to go over the network to get the requested bytes. 

In a perfect scenario, this number should always be 0, meaning we should always have the content fetched and cached before the container needs it. In practice, this is almost always never the case, so looking at the metric should give us an idea of how well our background fetcher is able to fetch the _needed_ content and whether or not we should pursue optimizations like prioritized pre-fetching (eg: Load Order Documents #749). 

_Note: This isn't a perfect heuristic for determining the benefit of LOD as there could be several reasons that are unrelated to the ordering of span fetches of why the background fetcher couldn't fetch and cache the request content before hand. (eg: Maybe there were many sequential spans that the container needed and the background fetcher couldn't get to them in time or some transient failure occurred when trying to fetch/cache the content)._

### Describe the results you expected

Ideally, we should increment this metric when we have to synchronously fetch the content over the network. We can move the metric to the end of [`getSpanContent`](https://github.com/awslabs/soci-snapshotter/blob/00b3501b309d13c8e00e8c244c55476943cdfbb6/fs/span-manager/span_manager.go#L291). The metric itself is at the layer level and so it also requires the layer digest dimension, so we'll have to figure out a way to pass the layer digest into the span manager or just omit it entirely, although I think it may be helpful to be able to aggregate at a layer level.

### Host information

1. OS: N/A
2. Snapshotter Version: v0.4.1
3. Containerd Version: N/A

### Any additional context or information about the bug

Related to: #749

PR #2015, opened 2026-06-12
fix(metrics): Increment SynchronousReadRegistryFetchCount only on registry fetch

ayush-panta:
**Issue #, if available:** #1016

**Description of changes:** Move `SynchronousReadRegistryFetchCount` metric increment from `reader.go` (fired on every `ReadAt`, including cache hits) to `span_manager.go`'s `getSpanContent` (where it fires only when a span is in `unrequested` state and an actual network fetch to the registry is issued). To preserve the layer digest dimension on the metric, the layer digest is now passed into `SpanManager` at construction time.

**Testing performed:** Added `TestSynchronousFetchMetricOnlyFiresOnNetworkFetch` unit test that verifies:
- Metric increments when `getSpanContent` hits an unrequested span (network fetch)
- Metric does NOT increment on a subsequent read of the same span (cache hit)
- Metric does NOT increment when the background fetcher has already fetched the span

### DIFF
--- a/fs/metrics/common/metrics.go
+++ b/fs/metrics/common/metrics.go
@@ -39,6 +39,7 @@ import (
 
 	digest "github.com/opencontainers/go-digest"
 	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
 )
 
 const (
@@ -86,7 +87,7 @@ const (
 	BackgroundFetch   = "background_fetch"
 
 	SynchronousReadCount              = "synchronous_read_count"
-	SynchronousReadRegistryFetchCount = "synchronous_read_remote_registry_fetch_count" // TODO revisit (wrong place)
+	SynchronousReadRegistryFetchCount = "synchronous_read_remote_registry_fetch_count"
 	SynchronousBytesServed            = "synchronous_bytes_served"
 
 	// Global fuse failure state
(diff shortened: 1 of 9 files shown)


fs/metrics/common/metrics.go
```diff
@@ -238,6 +239,15 @@ func IncOperationCount(operation string, layer digest.Digest) {
 	operationCount.WithLabelValues(operation, layer.String()).Inc()
 }
 
+// GetOperationCount returns the current value of an operation count metric for a given layer.
+func GetOperationCount(operation string, layer digest.Digest) float64 {
```
sondavidb, line 243: Seems this is only used for testing, should this be moved to a testing func?

fs/span-manager/span_manager_test.go
```diff
@@ -503,3 +505,55 @@ type readerFn func([]byte, int64) (int, error)
...
+		t.Fatalf("failed to create ztoc: %v", err)
+	}
+	c := cache.NewMemoryCache()
+	defer c.Close()
+
+	layerDigest := digest.FromString("metric-test-layer")
+	m, err := New(toc, r, c, 0, layerDigest)
+	assert.Nil(t, err)
+
+	getBefore := func() float64 {
```
sondavidb: nit: `getBefore` -> `get` or similar? Seems we use it to get the metric count before and after an operation.

sondavidb (commented):
Generally LGTM some small things

fs/metrics/common/metrics.go
ayush-panta: It depends on the un-exported `operationCount`var. I think we would have to export `OperationCount` and move the current func to `getOperationCount()` in span `fs/span-manager/span_manager_test.go`. I don't know if we should export the metric, though.

sondavidb: Hm, yeah ok I see what you're saying, since it would prevent anyone using our library from accidentally incrementing metrics on their own via just exporting a variable (which may not even be accurate given we have no way to provide a synchronous read outside of this function). This is fine by me then.

Go is just a little special in that everything is exportable and in theory any exported code can be called by other functions unless it's in the `internal` package. I'd definitely like us to be more intentional on what we do and do not expose to external users, but that's a discussion well well beyond the scope of this PR.

sondavidb (approved):
Thanks for the work on this, LGTM. Definitely need to make sure that we call this out in our release notes since it could be confusing for folks who have relied on this metric.

Swapnanil-Gupta (approved):

PR #2019, opened 2026-06-18
Add checks before generating a release

sondavidb:
**Issue #, if available:**

**Description of changes:**
Ideally we would want to run checks before pushing a tag, however the infra required for the GPG key does not currently exist and would require a revamp of our existing architecture. So this is the stop-gap solution I've arrived at, which will let maintainers push a tag but not allow the release to be created if the requisite checks don't pass.

This isn't an amazing solution but having a dead tag floating around for a couple of minutes is definitely better than missing release artifacts.

**Testing performed:**
Workflow run with correct artifacts in place: https://github.com/sondavidb/soci-snapshotter/actions/runs/27792753716
Workflow cut off main (missing THIRD_PARTY_LICENSES): https://github.com/sondavidb/soci-snapshotter/actions/runs/27792764860
Workflow run with correct assets in place but bad THIRD_PARTY_LICENSES: https://github.com/sondavidb/soci-snapshotter/actions/runs/27793042063

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -19,7 +19,68 @@ permissions:
   deployments: write
 
 jobs:
+  # Ensure all the correct artifacts are in the release branch
+  # setup needs this on PR to test the release workflow on PR
+  # and there's no easy way to skip all steps while allowing
+  # setup workflow to run so just skip all the steps that can fail.
+  check-release-branch-state:
+    runs-on: ubuntu-22.04
+
+    env:
+      CHECKOUT_REF: '' # Filled in first steps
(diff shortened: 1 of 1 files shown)


coderbirju (approved):

coderbirju (commented):
> Workflow run with correct artifacts in place: https://github.com/sondavidb/soci-snapshotter/actions/runs/27792753716
Workflow cut off main (missing THIRD_PARTY_LICENSES): https://github.com/sondavidb/soci-snapshotter/actions/runs/27792753716

Seems like both these are pointing to the same run

ayush-panta (approved):

sondavidb:
Thanks @coderbirju edited

PR #2028, opened 2026-07-09
feat(metrics): Add pull/mount critical-path observability metrics

erezzarum:
**Issue #, if available:**

**Description of changes:**

Adds nine new operation types to the existing `soci_fs_operation_duration_milliseconds` histogram and `soci_fs_operation_count` counter, instrumenting the parts of the image-pull / layer-mount critical path that were previously unmeasured.

The existing metrics attribute overall `mount` latency but do not break it down into its sub-operations, and several synchronous steps on the resolve/mount critical path (registry round-trips, zTOC fetch/parse, FUSE mount) had no metric at all. When diagnosing per-layer and parallel-pull latency, this made it impossible to tell *where* time was going, the `mount` total was a black box, and the pre-existing `remote_registry_get` metric only covers span-content reads (the FUSE read path), not resolve-time round-trips.

These metrics make the full mount critical path attributable: `mount ≈ soci_index_fetch (amortized, once per image) + blob_redirect + get_layer_size + ztoc_unmarshal + init_metadata_store + fuse_mount + overhead` and expose pre-resolution effectiveness and fetch-fan-out contention.

## New metrics

Mount critical-path sub-operations (histogram, labeled by `operation_type`, `layer`):

| Metric | Measures |
|--------|----------|
| `ztoc_unmarshal` | Reading + deserializing a layer's zTOC flatbuffer during resolve |
| `fuse_mount` | `fuse.NewServer` + `WaitMount` (the kernel FUSE mount) |
| `blob_redirect` | The registry → storage (e.g. ECR → S3) blob-URL redirect resolution |
| `get_layer_size` | The extra registry round-trip to discover a layer's size when absent from the descriptor |
| `soci_index_fetch` | Total time of `FetchSociArtifacts` (index + all zTOC blobs), once per image |
| `ztoc_fetch` | Fetching + storing a single zTOC blob, keyed by the image layer digest |

Pre-resolution / cache effectiveness (counter, labeled by `operation_type`):

| Metric | Measures |
|--------|----------|
| `resolve_cache_hit` | Layer resolves served from the resolver LRU (pre-resolution landed) |
| `resolve_cache_miss` | Layer resolves that hit the critical path (not cached / evicted before use) |
| `preresolve_queue_drop` | Pre-resolve requests dropped because the preresolver work queue was full |

The hit/miss ratio directly shows whether pre-resolution is landing or whether
resolved layers are being evicted before use (i.e. `resolve_result_entry` too
small for the concurrent working set). `preresolve_queue_drop > 0` indicates the
shared preresolver queue saturated under heavy concurrent-mount load.

**Testing performed:**

- `go build ./fs/...` and `go vet ./fs/... ./fs/metrics/...` pass.
- `go test ./fs/` passes.

### DIFF
--- a/fs/artifact_fetcher.go
+++ b/fs/artifact_fetcher.go
@@ -26,7 +26,9 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"time"
 
+	commonmetrics "github.com/awslabs/soci-snapshotter/fs/metrics/common"
 	sociremote "github.com/awslabs/soci-snapshotter/fs/remote"
 	socihttp "github.com/awslabs/soci-snapshotter/internal/http"
 	"github.com/awslabs/soci-snapshotter/soci"
@@ -35,6 +37,7 @@ import (
 	"github.com/containerd/containerd/v2/core/remotes/docker"
 	"github.com/containerd/containerd/v2/pkg/reference"
 	"github.com/containerd/log"
+	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"golang.org/x/sync/errgroup"
 	"oras.land/oras-go/v2/content"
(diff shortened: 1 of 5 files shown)


fs/artifact_fetcher.go
```diff
@@ -370,17 +378,27 @@ func FetchSociArtifacts(ctx context.Context, refspec reference.Spec, indexDesc o
...
+				layerDigest = blob.Digest
+			}
+			ztocFetchStart := time.Now()
 			rc, local, err := fetcher.Fetch(ctx, blob)
 			if err != nil {
 				return fmt.Errorf("cannot fetch artifact: %w", err)
 			}
 			defer rc.Close()
 			if local {
+				commonmetrics.MeasureLatencyInMilliseconds(commonmetrics.ZtocFetch, layerDigest, ztocFetchStart)
```
sondavidb: Hm, I wonder if we should even include the metric when it's locally stored?

sondavidb (dismissed):
One small thing but LGTM

fs/artifact_fetcher.go
erezzarum: Correct, we shouldn't, it will just report 0ms, we care only about remote fetching.
Fixed.

sondavidb (approved):

KlwntSingh (approved):

PR #2030, opened 2026-07-09
feat(metadata): Add "db-multi" per-layer metadata store to avoid the shared bbolt writer lock

erezzarum:
**Issue #, if available:**

**Description of changes:**

Adds a new metadata store, `metadata_store = "db-multi"`, that gives each layer its own bbolt database instead of the single shared database used by the default `"db"` store.

The default `"db"` store writes all layers' metadata to one `metadata.db`. bbolt is single-writer, so concurrent layer initializations (multi-layer images, or many images pulled at once) serialize on that one writer lock. `"db-multi"` removes the contention by writing each layer to an independent database.

**Testing performed:**

### DIFF
--- a/cmd/soci-snapshotter-grpc/main.go
+++ b/cmd/soci-snapshotter-grpc/main.go
@@ -324,7 +324,8 @@ func serve(ctx context.Context, rpc *grpc.Server, addr string, rs snapshots.Snap
 }
 
 const (
-	dbMetadataType = "db"
+	dbMetadataType      = "db"
+	dbMultiMetadataType = "db-multi"
 )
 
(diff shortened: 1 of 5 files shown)


sondavidb (dismissed):

ayush-panta (approved):

sondavidb:
CI failures are totally unrelated, we've run into this issue before but I could never figure out the root cause 🙁 will try and see how I can get CI to go

sondavidb:
Force-pushed this to just rebase from main which has our CI fixes, no other changes

sondavidb (approved):

Issue #1854, opened 2026-01-29
[FEATURE] CRI-O runtime support


prafgup:
### Description

Hi maintainers,

I'd like to request support for CRI-O runtime, particularly for running SOCI on OpenShift clusters.
Currently, SOCI only supports containerd as the runtime. However, OpenShift (and other Kubernetes distributions) use CRI-O as their default container runtime. This limits SOCI adoption for organizations running on these platforms.
For reference, the parent project [stargz-snapshotter already supports multiple runtimes](https://github.com/containerd/stargz-snapshotter/blob/main/docs/integration.md#cri-runtimes) including CRI-O, demonstrating that this architecture is feasible for lazy-loading snapshotters.

I do understand that the main usecase for SOCI was containerd (primary for AWS Fargate) https://github.com/awslabs/soci-snapshotter/issues/486#issuecomment-1464665690 but I think it would really benefit organizations if support for other runtimes were added as well, atleast trying and matching the parent project stargz.

Use case:
- Running lazy-pulling in OpenShift environments
- Organizations standardized on CRI-O who want to benefit from SOCI's lazy loading capabilities

Questions:
- Is runtime-agnostic support (or CRI-O specifically) on the roadmap?
- Has the idea of separating the core SOCI FS from the snapshotter been explored further

Shubhranshu153:
Hi @prafgup,

Thank you for the feature request. At the moment we dont have a contributor for the feature and haven't prioritized in the roadmap. We are happy to take contributions for the feature. 

For now i have added to our backlog to triage.

sondavidb:
> Has the idea of separating the core SOCI FS from the snapshotter been explored further

This has been explored for a while, in an ideal world we would do it but given the current state of the FS it's gotten to be a messy problem trying to separate it. Not off the table though.

PR #2034, opened 2026-07-13
Support CRI-O and Podman

prafgup:
**Issue #, if available:**

Implements https://github.com/awslabs/soci-snapshotter/issues/1854

**Description of changes:**

Still working on this but have verified its working end-to-end. Adapted largely from Stargz.

Stargz Store - 
- https://github.com/containerd/stargz-snapshotter/blob/main/cmd/stargz-store/main.go
- https://github.com/containerd/stargz-snapshotter/blob/main/store/fs.go

container/storage Annotation - 
- https://github.com/containers/storage/blob/main/pkg/chunked/toc/toc.go

**Testing performed:**

Tested end-to-end by pulling a soci image using cri-o runtime.

### DIFF
--- a/cmd/soci/commands/convert.go
+++ b/cmd/soci/commands/convert.go
@@ -38,6 +38,7 @@ import (
 const (
 	StandaloneFlag   = "standalone"
 	outputFormatFlag = "format"
+	crioFlag         = "crio"
 
 	outputFormatOCIArchive = "oci-archive"
 	outputFormatOCIDir     = "oci-dir"
@@ -95,6 +96,10 @@ var ConvertCommand = &cli.Command{
 					}
 					return nil
 				},
+			},
+			&cli.BoolFlag{
+				Name:  crioFlag,
+				Usage: "Prepare the converted image for lazy loading under CRI-O/Podman: annotate each indexed layer with its ztoc (TOC) digest (containerd.io/snapshot/stargz/toc.digest) so containers/storage consults a soci-store additional layer store.",
 			},
 		}),
 	Action: func(ctx context.Context, cmd *cli.Command) error {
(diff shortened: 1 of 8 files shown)


soci/soci_index.go
```diff
@@ -78,6 +78,12 @@ const (
 	// a SOCI index digest for the image.
 	ImageAnnotationSociIndexDigest = "com.amazon.soci.index-digest"
 
+	// TOCDigestAnnotation on an image layer descriptor tells containers/storage
+	// (CRI-O and Podman) to resolve that layer through an additional layer store
+	// (soci-store). We reuse eStargz's annotation key; the value is the digest of
+	// the layer's ztoc blob.
+	TOCDigestAnnotation = "containerd.io/snapshot/stargz/toc.digest"
```
prafgup, line 85: @sondavidb do you think we can get our own soci toc annotation key merged in containers/storage? or are we fine with using stargz digest annotation for soci in future as well?

https://github.com/containers/storage/blob/83cf57466529353aced8f1803f2302698e0b5cb7/pkg/chunked/toc/toc.go#L10-L14

cmd/soci/commands/convert.go
```diff
@@ -96,6 +97,10 @@ var ConvertCommand = &cli.Command{
 					return nil
 				},
 			},
+			&cli.BoolFlag{
+				Name:  crioFlag,
+				Usage: "Prepare the converted image for lazy loading under CRI-O/Podman: annotate each indexed layer with its ztoc (TOC) digest (containerd.io/snapshot/stargz/toc.digest) so containers/storage consults a soci-store additional layer store.",
+			},
```
prafgup, line 103: we can remove this flag and make it a default to annotate all ztoc index layers with the key

soci/soci_index.go
sondavidb: Yeah this seems feasible. I can submit a PR for this and put it here if we decide to go through with this, the change should be straightforward enough. (Though, it seems the change should be [here](https://github.com/podman-container-tools/container-libs/blob/b17b0a4c55bb953de8d783ea26534df22fdc9f2b/storage/pkg/chunked/toc/toc.go#L24-L28) instead based on the readme?)

With that said, we could probably reuse the stargz digest annotation, it would just depend on how much the workflow would differ between the two types of workflows. Which could be only minor deviations, or could be larger ones. Not sure yet, I'd have to see.

prafgup: Ah yes you are right we should be using `podman-container-tools/container-libs`

first time seeing a github link https://github.com/containers/container-libs redirecting to another link https://github.com/podman-container-tools/container-libs 

I think keeping stargz annotation would restrict us ever having a stargz optimized soci image or soci optimized stargz image as the annotation would clash. Thus I think we should for long term think of having a new tag. What do you think?

But again I think for experimentation we can keep using this stargz flag under --crio-experimentation or some flag and later on once this feature needs to be GA we can decide to create our own tag and default this tag for all our index layers.

That being said I would be passively working on this addition but for starters I have verified that soci pulling does work on cri-o with this PR changes :)

PR #2038, opened 2026-07-14
fix(backgroundfetcher): Background fetcher queue blocks layer mounts during silence period

erezzarum:
**Issue #, if available:**

**Description of changes:**

`BackgroundFetcher.Add` does a blocking channel send, and it's called from `Resolve` on the mount critical path. When more layers mount than the queue can hold (`max_queue_size`, default 100), mounts block waiting for space.
But the fetcher sleeps for `silence_period_msec` (default 30s) on each new image mount before draining anything, so during a multi-image launch, mounts stall for the full silence period.

When launching 10 images (~160 layers) at once: pull time was ~30s with the default config, dropping to ~3s when we set `silence_period_msec=100`.
The correlation made the root cause obvious.

 Switching the work queue from a fixed-capacity channel to a mutex-guarded slice.
`Add` now appends instantly, never blocks, never drops. Every layer still gets background-fetched once the silence period ends.

Also made `Pause()` non-blocking (capacity-1 channel with select/default) so it can't pile up or stall the mount path either.

`max_queue_size` stays in the config as a logging threshold, exceeding it emits a debug log but doesn't cap the queue. Existing configs keep working unchanged.

Why not just increase max_queue_size? A bigger number just moves the cliff. Any workload that exceeds it silently re-triggers the stall with no error or log to point at. Making `Add` structurally non-blocking removes the failure mode entirely.

**Testing performed:**

### DIFF
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -66,10 +66,14 @@ const (
 	// The background fetcher will fetch a single span every `defaultFetchPeriod`.
 	defaultBgFetchPeriodMsec = 500
 
-	// defaultBgMaxQueueSize specifies the maximum size of the bg-fetcher work queue i.e., the maximum number
-	// of span managers that can be queued. In case of overflow, the `Add` call
-	// will block until a span manager is removed from the workqueue.
-	defaultBgMaxQueueSize = 100
+	// defaultBgMaxQueueSize specifies the maximum size of the bg-fetcher work queue,
+	// i.e., the maximum number of span managers that can be queued. When full,
+	// the entry selected by drop_policy is evicted; Add never blocks.
+	defaultBgMaxQueueSize = 300
+
+	// defaultBgDropPolicy specifies which work-queue entry is evicted when
+	// max_queue_size is reached. Valid values: "oldest", "newest".
+	defaultBgDropPolicy = "newest"
 
 	// defaultBgMetricEmitPeriodSec is the default amount of interval at which the background fetcher emits metrics
 	defaultBgMetricEmitPeriodSec = 10
--- a/fs/backgroundfetcher/background_fetcher.go
+++ b/fs/backgroundfetcher/background_fetcher.go
@@ -19,6 +19,7 @@ package backgroundfetcher
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	commonmetrics "github.com/awslabs/soci-snapshotter/fs/metrics/common"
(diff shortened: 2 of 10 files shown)


fs/backgroundfetcher/background_fetcher.go
```diff
@@ -110,9 +120,40 @@ func NewBackgroundFetcher(opts ...Option) (*BackgroundFetcher, error) {
...
+// mounts (previously observed as pulls stalling for a full silence period),
+// and dropping would permanently skip background fetching for the layer.
 func (bf *BackgroundFetcher) Add(resolver Resolver) {
-	bf.workQueue <- resolver
+	bf.workQueueMu.Lock()
+	bf.workQueue = append(bf.workQueue, resolver)
+	n := len(bf.workQueue)
+	bf.workQueueMu.Unlock()
+	if bf.maxQueueSize > 0 && n > bf.maxQueueSize {
+		log.L.WithField("queueSize", n).WithField("maxQueueSize", bf.maxQueueSize).
```
sondavidb: Few things here
- As is, this would just spam logs, maybe this should only print when n = maxSize to avoid doing so?
- It should be fine since I think storing a function would be a tiny amount of space, however I do wonder if there's a potential for an image to queue so many jobs that it OOM-kills the snapshotter? I don't think it's really feasible but I do feel a safeguard against this via some max queue size would be good.
- Can the queue size actually drop the pre-resolve function? I don't believe it's critical to preresolve so it's not the worst if it drops given our queue size is reasonable. (In which case the concern above is not valid)

erezzarum: 1. Agree on the logs, i can just add one print when we reach it.
2. It's a really small footprint, but i can put a max queue size, though one already exists for a different reason
3. If adding a queue size, what will happen here is that any pre-resolve function will get dropped and there will be no background fetching for it at all for this function, this is why i created the spill over, without it, it's just block everything every pause.

erezzarum: I'll use the `max_queue_size` we already have and make it configurable as `unlimited (-1)` with eviction policy, `oldest` or `newest`, i will create a metric for tracking evictions as well.

erezzarum:
Added a configurable bounded queue with drop policy

config/defaults.go
```diff
@@ -66,11 +66,15 @@ const (
...
-	// of span managers that can be queued. In case of overflow, the `Add` call
-	// will block until a span manager is removed from the workqueue.
+	// defaultBgMaxQueueSize specifies the maximum size of the bg-fetcher work queue,
+	// i.e., the maximum number of span managers that can be queued. When full,
+	// the entry selected by drop_policy is evicted; Add never blocks.
 	defaultBgMaxQueueSize = 100
 
+	// defaultBgDropPolicy specifies which work-queue entry is evicted when
+	// max_queue_size is reached. Valid values: "oldest", "newest".
+	defaultBgDropPolicy = "oldest"
```
sondavidb: Existing behavior would be to drop the newest one. Would that be a safer default policy? (Should we even allow a difference between the drop policy? Will the end user understand/care what difference this makes? How will they know which one to use?)

sondavidb (commented):
One thing + need merge conflicts resolved, everything outside of the drop policy looks great to me

config/defaults.go
erezzarum: I will set defaults to drop the newest one and defaults `max_queue_size` to `300`.
Most end users don't need to tweak it and setting `max_queue_size` to `300` (instead of unlimited) is pretty safe, considering the max layer a container can have is 127 and it's just store a function, it should be safe.

sondavidb (approved):

erezzarum:
@sondavidb are there any pending changes needed here? It looks like there are no merge conflicts with main

sondavidb:
@erezzarum let me bump internally to see if we can get one more, sorry for the delay

ayush-panta (approved):
LGTM

PR #2039, opened 2026-07-14
feat(metadata): Add metadata_db_no_sync and metadata_insertion_chunk_size config

erezzarum:
**Issue #, if available:**

**Description of changes:**

Two new configuration values for tuning the bbolt metadata store initialization:

- **`metadata_db_no_sync`** (bool, default false): disables bbolt's per-transaction fsync. The metadata DB is derived data rebuilt from zTOCs on every daemon restart, so fsync durability is unnecessary. Skipping it removes synchronous disk flushes from layer metadata initialization.
- **`metadata_insertion_chunk_size`** (int, default 5000): max node insertions per bbolt transaction during metadata init. Larger chunks amortize per-commit overhead (B-tree rebalance, page COW, and fsync when enabled) over more insertions.

Under parallel image mounts, the metadata store's synchronous fsyncs and frequent small commits add measurable I/O latency per layer. These configuration values trade durabilit the snapshotter never needs (the DB is ephemeral) for faster initialization.

**Testing performed:**

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -72,6 +72,18 @@ type Config struct {
 	// MetadataStore is the type of the metadata store to use.
 	MetadataStore string `toml:"metadata_store"`
 
+	// MetadataDBNoSync disables bbolt's per-transaction fsync for the metadata
+	// store. The metadata DB is derived data rebuilt from zTOCs on every daemon
+	// restart, so fsync durability is unnecessary. Skipping it removes
+	// synchronous disk flushes from layer metadata initialization.
+	MetadataDBNoSync bool `toml:"metadata_db_no_sync"`
+
+	// MetadataInsertionChunkSize is the maximum number of node insertions per
+	// bbolt transaction during metadata store initialization. Larger chunks
+	// amortize per-commit overhead over more insertions. 0 means use the
+	// default (5000).
+	MetadataInsertionChunkSize int `toml:"metadata_insertion_chunk_size"`
+
 	// SkipCheckSnapshotterSupported is a flag to skip check for overlayfs support needed to confirm if SOCI can work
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -72,6 +72,18 @@ type Config struct {
 	// MetadataStore is the type of the metadata store to use.
 	MetadataStore string `toml:"metadata_store"`
 
+	// MetadataDBNoSync disables bbolt's per-transaction fsync for the metadata
+	// store. The metadata DB is derived data rebuilt from zTOCs on every daemon
+	// restart, so fsync durability is unnecessary. Skipping it removes
```
sondavidb, line 77: I wonder if the real answer here is to have the ability to use an existing db? (Not pertinent to this PR, per se, just food for thought)

sondavidb (approved):

config/config.go
erezzarum: WDYM by using an existing db?

sondavidb: Sorry, to be clearer, I think that this being "unnecessary" is only because it's not working properly. On paper if we persist a boltdb on disk we should be able to reuse the boltdb data to not have to re-fetch a ton of data, but we don't. I wonder if this is something worth fixing, or if we should consider a snapshot restart an effective hard reset.

(Again, not something to cover in this PR I think. Just something I've been thinking about related to our boltdb)

erezzarum: If i understand, it rebuilds from the zTOC on every layer mount.
If i understand, after a restart, because the cache is empty (resolver) so every layer that containerd re-mounts is a cache miss, so it's being rebuilt from the zTOC.
This also explains why sometimes we will see pull time for layers that are already cached on disk, that pull time is indicating the time it takes to rebuild the metadata after it was evicted from the resolver cache.

ayush-panta (approved):

Issue #1965, opened 2026-05-13
[Bug] CraftBlobURL passes full image reference to DefaultScheme, causing HTTPS for localhost registries in parallel pull


abdullahsumsum:
### Description

`CraftBlobURL` in `fs/remote/resolver.go:462` passes the full image reference string to `resolver.DefaultScheme()` instead of just the hostname. This causes `docker.IsLocalhost()` to fail to recognize localhost, so the scheme defaults to https even though the registry is plain HTTP on localhost.

```go
// current (broken)
func CraftBlobURL(reference string, ref registry.Reference) string {
    return fmt.Sprintf("%s://%s/v2/%s/blobs/%s", resolver.DefaultScheme(reference), ref.Host(), ref.Repository, ref.Reference)
}
```

`DefaultScheme` calls `docker.IsLocalhost(reference)` where `reference` is something like `localhost:30980/my-repo/my-image@sha256:abcd....` `net.SplitHostPort` fails on this (it's not a valid host:port pair), so `IsLocalhost` returns `false` and the scheme becomes `https`.
Meanwhile, `newRemoteStore` correctly sets `repo.PlainHTTP = true` via `docker.MatchLocalhost(refspec.Hostname())`, but `CraftBlobURL` never uses the repo's `PlainHTTP` flag — it independently re-derives the scheme and gets it wrong.
The fix should be straightforward — use `ref.Host()` instead of `reference`:

```go
func CraftBlobURL(reference string, ref registry.Reference) string {
    return fmt.Sprintf("%s://%s/v2/%s/blobs/%s", resolver.DefaultScheme(ref.Host()), ref.Host(), ref.Repository, ref.Reference)
}
```

This only affects the parallel pull path (`doInitialFetch` → `CraftBlobURL`). Lazy pull and standard pull work fine because they go through the ORAS repo which has PlainHTTP set correctly.

### Steps to reproduce the bug

1. Run a plain HTTP registry on localhost (e.g. `localhost:30980`)
2. Configure containerd's `hosts.toml` for the registry with `skip_verify = true` and `http` scheme
3. Enable SOCI parallel pull (e.g. via EKS `FastImagePull: true` feature gate, or `pull_modes.parallel.enable = true` in SOCI config)
4. Pull an image from the localhost registry

Then pulls fail with:

```bash
error getting header info: Get "https://localhost:30980/v2/.../blobs/sha256:...": 
[...]

PR #2040, opened 2026-07-14
fix: Honor [resolver] insecure flag for SOCI index and blob fetch paths

erezzarum:
**Issue #, if available:** 
#1965 

**Description of changes:**

The `[resolver.host.<host>.mirrors] insecure = true` config only controlled the layer-blob resolver in `service/resolver`.
The SOCI index fetch, zTOC blob fetches, and parallel-pull blob operations each derived their URL scheme independently via `docker.MatchLocalhost` (which only treats literal localhost as plain HTTP). 
Registries configured as insecure mirrors (e.g. an internal registry on a non-localhost IP over plain HTTP), got HTTPS requests that failed.

**Config example:**
```toml
[resolver]
  [resolver.host."100.65.30.95:32624"]
    [[resolver.host."100.65.30.95:32624".mirrors]]
      host = "http://100.65.30.95:32624"
      insecure = true
```

**Testing performed:**

### DIFF
--- a/fs/artifact_fetcher.go
+++ b/fs/artifact_fetcher.go
@@ -75,8 +75,8 @@ type orasBlobStore struct {
 	*remote.Repository
 }
 
-func newRemoteBlobStore(refspec reference.Spec, client *http.Client) (*orasBlobStore, error) {
-	repo, err := newRemoteStore(refspec, client)
+func newRemoteBlobStore(refspec reference.Spec, client *http.Client, plainHTTP bool) (*orasBlobStore, error) {
+	repo, err := newRemoteStore(refspec, client, plainHTTP)
 	if err != nil {
 		return nil, fmt.Errorf("cannot create remote store: %w", err)
 	}
@@ -96,7 +96,7 @@ func (r *orasBlobStore) Resolve(ctx context.Context, reference string) (ocispec.
 	}
 
 	tr := &clientWrapper{r.Client}
-	url := sociremote.CraftBlobURL(reference, ref)
+	url := sociremote.CraftBlobURL(reference, ref, r.PlainHTTP)
 	resp, err := sociremote.GetHeader(ctx, url, tr)
 	if err != nil {
 		return ocispec.Descriptor{}, err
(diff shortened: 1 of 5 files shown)


sondavidb (approved):
LGTM, I'd normally ask for a test but not sure if our current testing suite is capable of allowing regular HTTP requests as is, and even if we configured the registry created by the testing suite to do so it probably isn't the safest to do so...

ayush-panta (approved):

prafgup:
Not related to this, but I was randomly checking some benchmark metrics (might be a bit random) and I am seeing significant improvements over lazy times (more than 2x faster!)

- Old 0.14.1 release - https://github.com/awslabs/soci-snapshotter/actions/runs/27714426756/job/81985725195
- Current master - https://github.com/awslabs/soci-snapshotter/actions/runs/30129545681/job/89602241544

Just wanted to point out that it might be due to the recent perf improvements :) (or just noise over benchmarks)

cc @erezzarum

erezzarum:
> Not related to this, but I was randomly checking some benchmark metrics (might be a bit random) and I am seeing significant improvements over lazy times (more than 2x faster!)
> 
> * Old 0.14.1 release - https://github.com/awslabs/soci-snapshotter/actions/runs/27714426756/job/81985725195
> * Current master - https://github.com/awslabs/soci-snapshotter/actions/runs/30129545681/job/89602241544
> 
> Just wanted to point out that it might be due to the recent perf improvements :) (or just noise over benchmarks)
> 
> cc @erezzarum

Yes, all the recent PRs introduced performance improvements, you will see major performance improvements especially under high concurrency (multiple concurrent image pulls)

Can you try this configuration and see if you see any more improvement? it will slightly increase memory usage.

```
metadata_db_no_sync = true
resolve_result_entry = 60
[background_fetch]
max_queue_size = -1
```

prafgup:
Sure, its a bit hard with our setup currently to experiment changing soci config but I will try it our in the upcoming weeks!

Also would be nice to also get a new release out !

prafgup:
One shot vibe coded this (checks benchmark results for the past X commits on main), I do feel we have more lower lows now than before but have a lot of ci flakiness run variance to tell exactly regarding the performance improvements.

image

Eg - high latency commit - https://github.com/awslabs/soci-snapshotter/actions/runs/30497741177/job/90732721823

image

image

prafgup:
Published the above application here - https://github.com/prafgup/soci-bench/tree/main 
To run - 
```
npm install
npm run dev
```

PR #2041, opened 2026-07-15
Use CodeBuild runners for tests running as root

### LINKED ISSUE #1878: [Bug] Unit tests requiring root in CI cannot run on CodeBuild runners
ref: #1875 

**TL;DR We have to test unit tests requiring root on GitHub actions images, which reduces coverage and is inefficient to spin up a whole instance just for one package of tests. We should consolidate once the appropriate conditions are met.**

Our current CI for testing runs on CodeBuild runners and kernel 4.14. Newest Ubuntu runners provided by GitHub as of this issue are using Ubuntu 24-04 on kernel 6.14.

CodeBuild's old kernel causes issues when doing mount syscalls in root unit tests. CodeBuild moving to 5.10 should fix this specific issue. However I was having some trouble locally getting tests to pass, as the overlay testing suite's `Rename` call seems to attempt to rename across devices which is not supported with `os.Rename`. I was able to consistently reproduce it across EC2 instances but for some reason GitHub runners do not seem to have this issue. I suspect there's a setup difference here causing the failure in one but not the other, but I'm not certain as I haven't yet gotten CodeBuild runners to run this test successfully due to aforementioned kernel issues.

#1875 provides a workaround for this by just running root unit tests in a separate runner, however this is obviously suboptimal. On top of provisioning an entire runner just for the few unit tests, we also do not get proper test coverage.

When CodeBuild moves to 5.10, we should see if we can run root tests on CodeBuild. If so, we should combine the two tests. Example of how this could be done: https://github.com/sondavidb/soci-snapshotter/commit/8d13c20270d7efcd59f2ca91cecc773ca986ae16

Once we no longer have to rely on a separate runner specifically for root unit tests, this issue can be closed.

sondavidb:
**Issue #, if available:**
Closes #1878 

**Description of changes:**
CodeBuild updated its kernel so our workaround in #1875 shouldn't be needed anymore. Added some targets to make testing easier.

**Testing performed:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,25 +46,9 @@ jobs:
         run: dnf install zlib-static.aarch64 -y
       - run: make
       - run: make test-with-coverage
+      - run: sudo -E env PATH=$PATH make test-root-with-coverage
       - name: Show test coverage
         run: make show-test-coverage
-
-  # This is a temporary workaround to ensure we are testing our packages
-  # that require root (currently only the snapshot package).
-  # ref: https://github.com/awslabs/soci-snapshotter/issues/1878
-  test-root:
-    runs-on: ubuntu-24.04
-    strategy:
-      matrix:
-        go-version: ['1.25.11', '1.26.4']
-    timeout-minutes: 15
-    steps:
-      - uses: actions/checkout@v7
-      - uses: actions/setup-go@v6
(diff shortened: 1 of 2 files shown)


sondavidb:
We may have to change something in CodeBuild, so I'll leave this in draft for now

PR #2042, opened 2026-07-15
feat(resolver): Add opt-in registry auth client sharing to reduce token exchanges

erezzarum:
**Description of changes:**

Every image reference gets its own `AuthClient` with its own `docker.Authorizer`, so N images pulled from the same registry with the same credentials perform N separate auth token exchanges (401 → token endpoint GET → retried request, ~2 extra round-trips each). During multi-image launches this adds seconds of serial auth latency.

- When many layers of one image resolve concurrently, they can all hit a 401 before the first token lands, each independently fetching the same token (observed ~6 redundant exchanges for a single 8-layer image).
- The per-image-ref `registryHostMap` cache grows without bound on long-lived daemons.

**1. Opt-in auth client sharing** (`enable_auth_client_sharing`, default `false` - default behavior is unchanged). When enabled, AuthClients are shared between image references keyed by `(host, credential fingerprint)`:

- Credentials are resolved up front through the same providers the client uses at request time and SHA-256 hashed into the cache key, so two images share a client (and its token) **only** when their credentials are byte-identical. Images with different credentials (e.g. different K8s pull secrets) get distinct clients and never share tokens.
- Credential providers can be scoped per image ref (the CRI keychain stores creds by image ref and removes them after pull), so the shared client's credential function tries all image refs that joined it, newest-first, rather than binding to a single ref.

**2. Concurrent 401 challenge deduplication** (always on). When multiple goroutines hit a 401 simultaneously, only the first fetches the token; the rest wait for the in-flight result. Reduces the ~6 redundant exchanges per image to ~1.

**3. TTL-bounded caches** (`auth_client_ttl_sec`, default 3600). Cached auth clients and resolved registry-host configurations are lazily evicted on access and periodically swept. This also bounds the pre-existing unbounded growth of `registryHostMap`. Negative disables expiry.

```toml
[resolver]
  enable_auth_client_sharing = true   # opt in; default false (per-image clients)
  auth_client_ttl_sec = 3600          # negative = never expire
```

With sharing enabled, a 10-image batch from one registry performs 1 token exchange instead of 10. The challenge dedup alone (no opt-in needed) removes the redundant concurrent exchanges within each image.

**Testing performed:**

### DIFF
--- a/service/resolver/registry.go
+++ b/service/resolver/registry.go
@@ -33,6 +33,7 @@
 package resolver
 
 import (
+	"crypto/sha256"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -40,6 +41,7 @@ import (
 	"time"
 
 	"github.com/awslabs/soci-snapshotter/config"
+	socihttp "github.com/awslabs/soci-snapshotter/internal/http"
 	rhttp "github.com/hashicorp/go-retryablehttp"
 
 	"github.com/containerd/containerd/v2/core/remotes/docker"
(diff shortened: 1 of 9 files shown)


service/resolver/registry.go
```diff
@@ -66,16 +68,147 @@ type RegistryManager struct {
...
 		registryConfig:  registryConfig,
 		creds:           credsFuncs,
 		registryHostMap: &sync.Map{},
+		authClientMap:   &sync.Map{},
+		authClientTTL:   time.Duration(registryConfig.AuthClientTTLSec) * time.Second,
+	}
+	if rm.authClientTTL > 0 {
+		// Periodic sweep so never-again-accessed entries are also reclaimed.
+		go func() {
+			ticker := time.NewTicker(rm.authClientTTL)
```
sondavidb, line 142: We should probably also tie this to the context of the snapshotter service, i.e. if the snapshotter context is killed then this should also stop. Will probably need to mutate the function header for this.

sondavidb (commented):
1 thing LGTM otherwose

service/resolver/registry.go
ayush-panta: @erezzarum Could you address this? We want to get a release out and we'd like to include this PR. Thanks!

erezzarum: Fixed, also added tests for this use case.

sondavidb (approved):

coderbirju (approved):

Issue #2037, opened 2026-07-14
[Bug] content_store=containerd: cross-process bbolt/lease deadlock between soci-snapshotter and containerd on first boot


vincer:
## Description

With `content_store.type = "containerd"`, soci-snapshotter and containerd can deadlock each other across the gRPC boundary when lazy mounts run concurrently with unindexed-image unpacks. Once wedged, **every** subsequent pull on the node hangs (we observed a wedge persist 53+ minutes until manual intervention); restarting the snapshotter clears it. We believe this is the root cause of the symptom class reported in #1826 ("containers stuck in creating/terminating states with containerd content store").

The deadlock strongly favors *first boot* of a machine whose disk is a lazily-hydrating volume (e.g. EC2 EBS from an AMI snapshot): cold-block latency stretches bbolt read-transaction windows from milliseconds to seconds, wide enough to interleave with the other three edges. In our environment: ~2/3 of fresh boots wedged; 0/4 reboots of the same node ever did, even after wiping the stores — which makes this near-invisible in ordinary testing.

## The cycle (four edges, each captured in a goroutine dump)

We captured goroutine dumps from **both** daemons on a wedged node and matched the waiting edges:

1. soci's `prepareRemoteSnapshot` holds a bbolt metastore **read transaction across the entire `fs.Mount`** (`snapshot/snapshot.go:1066` at v0.14.0), inside which it synchronously creates a containerd lease — `ContainerdStore.BatchOpen` calls `client.WithLease` (`soci/store/store.go:415`).
2. containerd's lease handler needs its own meta.db **writer**.
3. That writer is held by containerd snapshot Prepare/Commit transactions whose proxied RPCs call back **into soci** (unindexed image unpack routed through the proxy snapshotter).
4. Those RPCs queue behind soci's bbolt **writer**, which is blocked in `DB.mmap` growing a freshly-created `metadata.db` — and mmap growth waits for all open read transactions to close. Back to edge 1.

No deadline exists anywhere in the cycle, so the wedge is permanent.

## Bisect evidence

- Single config-bit A/B via boot-time config injection on identical machines in the same hour: stock configuration 5/5 wedged; one flag flipped, 2/2 clean.
- Two-image-variant A/B validating the structural fix: 5/5 clean vs 0/5.
- Serializing our own boot-time pulls was **insufficient** (still wedged 2/3) — the concurrency that completes the cycle came from ordinary system-container pulls, so any realistic node hits it.

## Workaround

[...]

Issue #1826, opened 2025-12-19
[Bug] Containers getting stuck in creating and terminating states with containerd content store


chiragjn:
### Description

We recently rolled out snapshotter version 0.12.1 with following config to multiple clusters

```yaml
[cri_keychain]
  enable_keychain = true
  image_service_path = "/run/containerd/containerd.sock"
[content_store]
  type = "containerd"
[pull_modes]
  [pull_modes.soci_v1]
    enable = true
  [pull_modes.soci_v2]
    enable = true
  [pull_modes.parallel_pull_unpack]
    enable = false
    max_concurrent_downloads = -1
    max_concurrent_downloads_per_image = 3
    concurrent_download_chunk_size = "16mb"
    max_concurrent_unpacks = -1
    max_concurrent_unpacks_per_image = 1
    discard_unpacked_layers = true
  [pull_modes.parallel_pull_unpack.decompress_streams."gzip"]
    path = "/usr/bin/unpigz"
    args = ["-d", "-c"]
```

Across multiple clusters we noticed the same issue - lot of containers get stuck in Container Creating state. There is no clear pattern yet but few nodes work fine
I do not have logs to share right away, but I will try to get containerd and snaphotter logs from a test cluster

Our guess right now is 

```
[content_store]
  type = "containerd"
```

is causing the issue

We have reverted this to soci and testing it out on a test cluster. Is content store containerd safe to use without parallel pull?

### Describe the results you expected

Containers should not get stuck in Container Creating / Terminating

### Host information

1. OS: EKS AL2023
2. Snapshotter Version: 0.12.1
3. Containerd Version: 2.1.5

Kern--:
We haven't enabled the containerd content store for the snapshotter by default yet because we haven't done enough testing to verify the integration.

If you can supply logs, we can help debug this. For now, I would recommend using the SOCI content store in production.

chiragjn:
[containerd-soci.logs.txt](https://github.com/user-attachments/files/24271699/containerd-soci.logs.txt)

This is what I got from a node, not sure if useful. Anyway we have switched to soci content store and things look fine

Kern--:
I didn't find much in the logs and I wasn't able to reproduce the issue with crictl - we may need to investigate with a full k8s setup. Can you share more about your setup? How are you configuring the snapshotter in the EKS AMI?

Is falling back to the SOCI store working for you? Is there anything you want from using the containerd content store?

vincer:
[...]

PR #2043, opened 2026-07-15
fs/reader: fix spancache fd leak in file Verify

NahumLitvin:
`file.Verify` never closes the reader returned by `spanManager.GetContents`, leaking one spancache file descriptor per verified file. The sibling call site in `ReadAt` (~45 lines up) closes its reader with `defer r.Close()` — `Verify` is the only `GetContents` caller that doesn't.

`Verify` runs on first access of every file in a lazily mounted layer, so pod-startup waves that touch many files across several images leak FDs in bursts. In production we drove the snapshotter past its `LimitNOFILE` (65535) during concurrent pod starts — hundreds to thousands of `too many open files` errors on spancache blob opens per storm (worst node: 15k in one day) — and the resulting stuck state escalated into the containers-stuck-in-creating symptom class reported in #1826 / #2037 (in our case a containerd-side GC hang; we've submitted a resilience fix there as containerd/containerd#13799). Steady-state FD counts look deceptively healthy afterwards because Go's `os.File` finalizers eventually reclaim leaked descriptors once the readers are GC'd.

One-line fix: `defer tarHeaderReader.Close()`.

Happy to share (lightly redacted) goroutine dumps and journal evidence from the production incident if useful.

### DIFF
--- a/fs/reader/reader.go
+++ b/fs/reader/reader.go
@@ -220,6 +220,7 @@ func (sf *file) Verify() (retErr error) {
 	if err != nil {
 		return err
 	}
+	defer tarHeaderReader.Close()
 	counterReader := ioutils.NewPositionTrackerReader(tarHeaderReader)
 	tarReader := tar.NewReader(counterReader)
 	tarHeader, err := tarReader.Next()


sondavidb (approved):
LGTM pending CI, thanks for this!

ayush-panta (approved):

sondavidb (approved):

PR #2049, opened 2026-07-26
fix(metadata): remove stale metadata DB on startup

ayush-panta:
**Description of changes:** The metadata DB is ephemeral and always rebuilt from zTOCs on layer mount. After an unclean shutdown, orphaned buckets from the previous run persist in the file and are never reused. Remove the file on startup to prevent unbounded growth from repeated crashes.

**Testing performed:** Added unit tests. More testing via CI.

### DIFF
--- a/cmd/soci-snapshotter-grpc/main.go
+++ b/cmd/soci-snapshotter-grpc/main.go
@@ -335,13 +335,20 @@ func getMetadataStore(ctx context.Context, rootDir string, config config.Config)
 			"store_type": config.MetadataStore,
 		}).Debug("initializing metadata store")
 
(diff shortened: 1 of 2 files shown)


sondavidb:
> Remove the file on startup to prevent unbounded growth from repeated crashes.

I think technically so long as the same images continue to be used I don't think the boltdb should grow since I believe we key off layers. Still a valid point to be clear, it's not good that we are keeping so much unused garbage.

I do wonder how we should be caring about state though. On restart should we totally wipe the slate clean? Is it intuitive that we remove all the snapshot mounts on a restart? containerd restart keeps all images, why shouldn't we? Or is our current slate not intuitive? I'm really not sure and I think we should be clear about the design choice we'd be making here before going forward.

PR #2050, opened 2026-07-29
Dockerfile: skip unnecessary builds

sondavidb:
**Description of changes:** Fix CI failures in the rapidgzip Docker build stage:
1. Initialize git submodules after checkout, as rapidgzip's submodule build dependencies weren't being fetched. Exclude bzip2-tests (hosted on sourceware.org, unreachable from CI, only test data).
2. Build only the rapidgzip binary target instead of the full project.

**Testing performed:** CI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,6 +75,7 @@ RUN mkdir -p /opt/rapidgzip/usr/local/bin
 RUN git clone https://github.com/mxmlnkn/rapidgzip.git && \
     cd rapidgzip && \
     git checkout "rapidgzip-v${RAPIDGZIP_VERSION}" && \
+    git -c submodule."external/bzip2-tests".update=none submodule update --init --recursive && \
     if [ "$TARGETARCH" = "arm64" ] || [ "$(uname -m)" = "aarch64" ]; then \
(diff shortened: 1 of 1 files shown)


ayush-panta (approved):
Looks good, I updated the PR description with a bit more detail.

coderbirju (approved):

sondavidb:
Adding some more context, I opened a test PR that doesn't use `-q` when using docker compose, which spawned [this job](https://github.com/awslabs/soci-snapshotter/actions/runs/30493055823/job/90715386017?pr=2051), and the logs show [this line](https://github.com/awslabs/soci-snapshotter/actions/runs/30493055823/job/90715386017?pr=2051#step:6:5289) where we are getting a 403 trying to download tests for a submodule we don't even need.

I think this points to a couple of things here:
1. We are importing more submodules than needed. Maybe this is the only thing we shouldn't import, but I think we should make sure we aren't importing more than necessary
2. Our `docker build` eating errors is bad. We should probably try to preserve existing behavior for the happy path.

sondavidb:
#2051 should update 2 from above. 1 I think we still need to vet. I'll open an issue

PR #2063, opened 2026-08-05
Make soci.NewDB return a new database instance per call

kzys:
## Issue #, if available

n/a

## Description of changes

`soci.NewDB` guarded its initialization with a package-level `sync.Once`, which latched both the first path and the first failure: a second call with a different path silently returned the database opened at the first path, and a failed first open left every subsequent call failing regardless of path, with the real error swallowed into a log line.

Nothing in this repository needs the singleton — the daemon never touches `ArtifactsDb`, each CLI command calls `NewDB` once, and the package's own tests already bypass it (`newTestableDb`). But it bites library consumers that call `NewDB` with more than one path in a process.

This change:

- Makes `NewDB` return a fresh `*ArtifactsDb` on every call, propagating open errors to the caller.
- Adds `(*ArtifactsDb).Close()` to release bolt's file lock, deferred at every CLI call site.
- Adds `(*IndexBuilder).Close()`, which closes the artifacts database only if the builder opened it itself; a database injected via `WithArtifactsDb` is left for its owner to close.

Behavior note: `NewDB` calls are no longer deduplicated, so a process constructing multiple `IndexBuilder`s should share a database via `WithArtifactsDb` — bolt blocks when opening a file that is already locked.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/soci/commands/convert.go
+++ b/cmd/soci/commands/convert.go
@@ -139,6 +139,7 @@ var ConvertCommand = &cli.Command{
 		if err != nil {
 			return err
 		}
+		defer artifactsDb.Close()
 
 		builderOpts, err := parseBuilderOptions(cmd)
 		if err != nil {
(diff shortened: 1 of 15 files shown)


kzys:
Hi there! Sorry I couldn't finish #1874 (and seems #2027 will do the same). I'm going to use SOCI for my project and will upstream a few changes in August/September!

Issue #2035, opened 2026-07-14
[Bug] parallel_pull_unpack with discard_unpacked_layers=true fails pulls with "compressed digests did not match" when a layer is already in the shared containerd content store


vincer:
## Description

With `parallel_pull_unpack` enabled, `discard_unpacked_layers = true`, and the containerd content store (`content_store.type = "containerd"`), pulls of images that share a layer with previously-pulled content fail deterministically with `compressed digests did not match` — even though the layer bytes are provably correct (the blob downloads cleanly and its sha256 matches the manifest digest). The failure occurs in ~100 ms, far too fast to be a transfer problem, which is the tell that it is a local verification defect rather than corruption.

## Root cause

The compressed-digest verifier is created per layer when `discard_unpacked_layers` is set (`fs/fs.go:583` at v0.14.0), but only the **remote-download** path ever starts it (`fs/parallel_artifact_fetcher.go:195-198`). When the layer already exists in the local content store, the fetch short-circuits at `fs/parallel_artifact_fetcher.go:68-73` and the verifier is never started. `Verified()` on a never-started verifier returns `false` (`fs/unpacker.go:83-89`), which is then reported as the digest mismatch at `fs/unpacker.go:220-221`.

So: a *correct* local cache hit is indistinguishable from a corrupt download.

## Reproduction

1. Pull image A containing layer L, so L lands in the shared containerd content store. Any layer shared across images works; deterministic-build systems (e.g. Bazel `rules_oci`) commonly produce a byte-identical empty gzip layer (~49 bytes) shared by many images, which makes this trigger fleet-wide.
2. Pull unindexed image B that also contains L, routing through `parallel_pull_unpack`.
3. B fails 100% of the time with `compressed digests did not match`, blaming L.

In our fleet this became deterministic on every node because boot-time monitoring images pre-seeded the shared empty-gzip layer before any workload image was pulled. Validation on fresh nodes (empty stores) passes, which is how it escaped pre-production testing.

## Workaround

`discard_unpacked_layers = false`. Verified A/B/A on live nodes: fails in ~2 s → clean 3m23s pull with the flag → fails again on revert. Cost: ~17 GiB retained compressed blobs per large image.

## Versions

- v0.14.0; code identical at v0.14.1; functionally identical on `main` (only containerd v2 import-path drift).
- Related but distinct: #1915 covers `discard_unpacked_layers` content-completeness semantics; this issue is a verification failure on a correct local hit.

## Suggested fix

[...]

PR #2073, opened 2026-09-03
fix: Mark the compressed verifier satisfied for local content store hits

mukul-atomicwork:
**Issue #, if available:**
#2035

**Description of changes:**

With `discard_unpacked_layers` set, a layer that is already present in the content store fails the pull with `compressed digests did not match`, even though the bytes are correct.

The compressed verifier is only constructed when that flag is set, and it is only ever started on the remote download path. When `parallelArtifactFetcher.Fetch` short-circuits on a local content store hit, the bytes never pass through it, so `Verified()` returns false for never having been started, and `Apply` reports a mismatch.

This change marks the verifier as satisfied on that branch.

**On what this does and does not give up**

The compressed verifier is a read-time check, and ingest verification is a historical fact about the write, so these are not the same guarantee. Removing the read-time check on this path does mean post-ingest corruption of the blob file, bitrot or something writing into the content store directory, is no longer caught by *that* check.

What still catches it is the uncompressed verifier. It is constructed unconditionally, started inline in `Apply` on every path, and enforced against the diffID from the manifest, so corruption that reaches the filesystem is still rejected.

The residual gap is corruption that changes the compressed bytes but decompresses to identical output, for instance rewritten gzip header fields or a re-encoded deflate stream. That no longer surfaces. The correct bytes still land on disk, so it is not a container integrity problem, but the store would hold a blob that no longer matches its descriptor digest and SOCI would not notice. I judged that acceptable against a bug that fails pulls outright, but calling it out so the tradeoff is explicit.

Worth noting for the local hit case specifically: in discard mode SOCI never writes layers to the content store at all (`parallel_unpacker.go`, the `!local && !lu.discardUnpackedLayers` guard), so a local hit under this flag came from containerd's own pull or from an earlier non-discard configuration. Both verify on ingest.

**On the implementation**

The verifier is shared: `fs.go` passes the same `*asyncVerifier` to both `NewLayerArchive` and `newParallelArtifactFetcher`, so marking it in the fetcher is visible to the archive. The flag is written in `Fetch` and read in `Apply`, which run in that order on the same goroutine in `parallelLayerUnpacker.Unpack`. I have noted that ordering requirement on the field.

I did not clear the verifier reference instead, since the archive holds its own pointer to the same object and the fetcher cannot reach it.

`asyncVerifier` is unexported, so `MarkLocallyVerified` is unreachable outside the `fs` package and misuse is bounded to one package.

I did not start the verifier over the local content, since that is a full re-read of every cached layer to recompute something the content store already guaranteed, on the path this feature exists to make fast.

**Not addressed here**

`Verified()` returns the same false for "never started", "context cancelled" and a genuine mismatch, so a cancelled pull also surfaces as a digest mismatch. Fixing that changes the signature and touches every caller, so it is out of scope here. Filing separately.

**Testing performed:**

Run on Linux in a `golang:1.26` container, since `fs` transitively imports containerd's `overlayutils` and will not build on macOS.

Two tests in `fs/adaptive_fetch_image_layers_test.go`, sharing a fixture helper:

- `TestLocalContentStoreHitPassesDigestValidation` seeds a layer into the local store and asserts it unpacks. Removing the `MarkLocallyVerified` call makes it fail with `compressed digests did not match`, so it gates on the fix.
- `TestRemoteFetchStillVerifiesCompressedDigest` serves the same contents from the remote at a different gzip level, so the diffID matches but the compressed digest does not, and asserts `Apply` returns `ErrCompressedDigestMismatch`. This shows compressed verification still works end to end after the change. Hoisting the call out of the local hit branch makes it fail.

  Worth noting since it is the reason this one earns its place: `TestFailureModes` covers a failing `Apply` through a fake archive that returns a canned error, so nothing in the package currently exercises real compressed digest rejection. This is the first test that does.

`go test ./fs/...`, `go vet ./fs/` and `gofmt` are clean.

For context on how often this bites: we logged 1717 occurrences of this error in seven days across one EKS fleet, on 0.15.0 as shipped in Bottlerocket. The `discard_unpacked_layers = false` workaround costs roughly 28s on a cold 8.5 GB pull, 105s against 77s, measured on the same instance type.


mukul-atomicwork:
cc: @sondavidb pls take a look into this. TIA

fs/local_hit_discard_repro_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"oras.land/oras-go/v2/content/memory"
+)
+
+// buildLayer returns a real gzipped tar containing one file, plus its
+// compressed digest and its diffID.
+func buildLayer(t *testing.T) (gz []byte, compressed, uncompressed digest.Digest) {
+	t.Helper()
+	var raw bytes.Buffer
+	tw := tar.NewWriter(&raw)
+	body := []byte("soci regression fixture\n")
```
sondavidb: Can we use our [TestRand](https://github.com/awslabs/soci-snapshotter/blob/3049d2b43ba8f08dff2dcb7e8413643b76ad5396/util/testutil/random.go#L30-L38) instead to have different layers?

fs/local_hit_discard_repro_test.go
```diff
@@ -0,0 +1,238 @@
...
+	if _, err := zw.Write(raw.Bytes()); err != nil {
+		t.Fatal(err)
+	}
+	if err := zw.Close(); err != nil {
+		t.Fatal(err)
+	}
+	return zipped.Bytes(), digest.FromBytes(zipped.Bytes()), digest.FromBytes(raw.Bytes())
+}
+
+// TestLocalContentStoreHitWithDiscardUnpackedLayers reproduces awslabs/soci-snapshotter#2035.
```
sondavidb: I don't think the linked PR is useful to link here, just explaining what it does is plenty. Maybe `TestLocalContentStoreHitWithDiscardUnpackedLayers ensures digest validation passes for content already present in the content store` would be more succinct.

fs/local_hit_discard_repro_test.go
```diff
@@ -0,0 +1,238 @@
...
+)
+
+// buildLayer returns a real gzipped tar containing one file, plus its
+// compressed digest and its diffID.
+func buildLayer(t *testing.T) (gz []byte, compressed, uncompressed digest.Digest) {
+	t.Helper()
+	var raw bytes.Buffer
+	tw := tar.NewWriter(&raw)
+	body := []byte("soci regression fixture\n")
+	if err := tw.WriteHeader(&tar.Header{Name: "hello.txt", Mode: 0o644, Size: int64(len(body))}); err != nil {
```
sondavidb: nit: Separate lines would help make this a little clearer and Go-esque.

fs/local_hit_discard_repro_test.go
```diff
@@ -0,0 +1,238 @@
...
+	// The layer is already in the content store, as it would be if any other
+	// image on the node had pulled it. This is the case #2035 describes.
+	local := &fakeLocalStore{Store: memory.New()}
+	if err := local.Push(ctx, desc, bytes.NewReader(gz)); err != nil {
+		t.Fatalf("seeding the local store: %v", err)
+	}
+
+	parallelCfg := &config.Parallel{
+		Enable: true,
+		ParallelConfig: config.ParallelConfig{
```
sondavidb: Don't we only need `DiscardUnpackedLayers` here?

fs/local_hit_discard_repro_test.go
sondavidb, line 1: We don't need a separate file for this, we can probably use [adaptive_fetch_image_layers_test.go](https://github.com/awslabs/soci-snapshotter/blob/main/fs/adaptive_fetch_image_layers_test.go).

fs/local_hit_discard_repro_test.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// TestLocalContentStoreHitWithDiscardUnpackedLayers reproduces awslabs/soci-snapshotter#2035.
+//
+// With discard_unpacked_layers set, fs.go constructs a compressed verifier. When the
+// layer is already present in the shared content store, parallelArtifactFetcher.Fetch
+// short-circuits and never starts that verifier, so Verified() reports false for a
+// layer whose bytes are provably correct, and Apply rejects it as a digest mismatch.
+func TestLocalContentStoreHitWithDiscardUnpackedLayers(t *testing.T) {
+	ctx := context.Background()
```
sondavidb: Can we tie this to a CancelCause context to pass to `GetOrAddImageJob` and defer the cancel here?

fs/local_hit_discard_repro_test.go
```diff
@@ -0,0 +1,238 @@
...
+		t.Fatalf("Fetch: %v", err)
+	}
+	defer rc.Close()
+	if !fromLocal {
+		t.Fatal("expected the layer to be served from the local content store")
+	}
+
+	root := t.TempDir()
+	if _, err := layerArchive.Apply(ctx, root, rc); err != nil {
+		if strings.Contains(err.Error(), "compressed digests did not match") {
```
sondavidb: Instead of this, let's separate out the error [here](https://github.com/awslabs/soci-snapshotter/blob/31bfe86b42c96b6d829ad079567666e1a818cb1f/fs/unpacker.go#L221) into a variable and then use `errors.Is(err, NewErrorName)` to ensure this doesn't change down the line.

fs/local_hit_discard_repro_test.go
```diff
@@ -0,0 +1,238 @@
...
+			t.Fatalf("#2035: correct local content rejected as a digest mismatch: %v", err)
```
sondavidb: Why are we referring to the PR number here?

fs/local_hit_discard_repro_test.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+		t.Fatalf("Apply: %v", err)
+	}
+	// Prove the layer was really unpacked, not just that Apply returned nil.
+	if got, err := os.ReadFile(filepath.Join(root, "hello.txt")); err != nil {
+		t.Fatalf("layer was not applied: %v", err)
+	} else if string(got) != "soci regression fixture\n" {
```
sondavidb: Make sure to change this to also match the string above

fs/local_hit_discard_repro_test.go
```diff
@@ -0,0 +1,238 @@
...
+		t.Fatalf("unexpected file contents: %q", got)
+	}
+}
+
+// corruptLocalStore reports a local hit but hands back bytes that do not match
```
sondavidb: How is this check distinct from checking if Apply fails on a bad reader [here](https://github.com/awslabs/soci-snapshotter/blob/31bfe86b42c96b6d829ad079567666e1a818cb1f/fs/unpacker_test.go#L34)?

fs/unpacker.go
```diff
@@ -80,10 +83,25 @@ func (av *asyncVerifier) AsyncVerify(reader io.ReadCloser) {
 	av.started = true
 }
 
+// MarkExternallyVerified records that the content this verifier guards was
+// validated elsewhere, for example by the content store when the blob was
+// ingested. Without this, a layer served from the local content store is never
+// passed through the verifier, and Verified reports false for content that is
+// provably correct.
+func (av *asyncVerifier) MarkExternallyVerified() {
+	if av == nil || av.v == nil {
```
sondavidb: We don't need to check if the Verifier is nil or not

fs/local_hit_discard_repro_test.go
sondavidb: Also, if it's in the content store already, doesn't that mean it already passed digest verification? What use case are you hoping to cover here?

fs/parallel_artifact_fetcher.go
```diff
@@ -69,6 +69,10 @@ func (f *parallelArtifactFetcher) Fetch(ctx context.Context, desc ocispec.Descri
 	// Check local store first
 	rc, err := f.localStore.Fetch(ctx, desc)
 	if err == nil {
+		// The content store verified this blob's digest when it was ingested, and
```
sondavidb: nit: Every comment related to this function is far too verbose. The function name conveys perfectly well what this does.

Alternatively, can we just set `v.verifier == nil` or something similar? Since `Verified` calls will just return true on nil, it would remove further if cases. But maybe it's bound to lead to unexpected consequences? Not really sure, either way I'm ok here.

fs/unpacker.go
```diff
@@ -57,6 +57,9 @@ type asyncVerifier struct {
 	v       digest.Verifier
 	waitCh  chan struct{}
 	started bool
+	// externallyVerified records that the content was already validated by
+	// another component, so this verifier has nothing left to check.
+	externallyVerified bool
```
sondavidb: nit: `externallyVerified` -> `contentExistsLocally`, `locallyVerified`, etc. might be a little clearer? I don't think there are many other use cases if any where it would be verified. We can always adjust this later if we need it to cover more use cases.

fs/local_hit_discard_repro_test.go
mukul-atomicwork: Done, using testutil.NewTestRand and BuildTar/File for the layer.

fs/local_hit_discard_repro_test.go
mukul-atomicwork: Used your wording verbatim and dropped the link.

fs/local_hit_discard_repro_test.go
mukul-atomicwork: Done.

fs/local_hit_discard_repro_test.go
mukul-atomicwork: Trimmed to Enable and DiscardUnpackedLayers. checkParallelPullUnpack only compares the per-image limits against the globals when the globals are set, so leaving them at zero validates fine.

fs/local_hit_discard_repro_test.go
mukul-atomicwork: Moved there and deleted the separate file.

fs/local_hit_discard_repro_test.go
mukul-atomicwork: Done.

fs/local_hit_discard_repro_test.go
mukul-atomicwork: Done.

fs/local_hit_discard_repro_test.go
mukul-atomicwork: removed

fs/local_hit_discard_repro_test.go
mukul-atomicwork: Now the test asserts with errors.Is so we don't have worry about this

fs/local_hit_discard_repro_test.go
mukul-atomicwork: Removed, I'd added that to make sure skipping the compressed verifier couldn't quietly turn into no verification at all, if someone reworked this path later.

fs/unpacker.go
mukul-atomicwork: Removed, had added it for safety.

fs/parallel_artifact_fetcher.go
mukul-atomicwork: Made comment shorter.

fs/unpacker.go
mukul-atomicwork: Went with locallyVerified, and renamed the method to MarkLocallyVerified to match.

fs/parallel_artifact_fetcher.go
mukul-atomicwork: On setting the verifier to nil, I tried it and it doesn't work: swapping the call for f.verifier = nil leaves the test failing with compressed digests did not match. fs.go hands the same *asyncVerifier to both NewLayerArchive and newParallelArtifactFetcher, so clearing the fetcher's field leaves layerArchive.compressed pointing at the same object. The fetcher has no handle on the archive to clear the other reference.

mukul-atomicwork:
@sondavidb addressed all the comments. Pls add if you have any additional requirements.

sondavidb (dismissed):
LGTM pending CI, thanks!

sondavidb:
Looks like linter issue

mukul-atomicwork:
> Looks like linter issue

@sondavidb I have fixed this. Tested builds of all version. Please review.

fs/adaptive_fetch_image_layers_test.go
```diff
@@ -618,3 +627,146 @@ func TestParallelStructCreation(t *testing.T) {
...
+		t.Fatalf("failed to create layer unpack storage: %v", err)
+	}
+	jobs, err := newUnpackJobs(ctx, &config.Parallel{
+		Enable:         true,
+		ParallelConfig: config.ParallelConfig{DiscardUnpackedLayers: true},
+	}, storage)
+	if err != nil {
+		t.Fatalf("failed to create unpack jobs: %v", err)
+	}
+	layerJob, err := jobs.AddLayerJob(desc.Digest.String(), desc.Digest.String(), cancel)
```
sondavidb, line 683: nit: I think we should be keying off of some dummy image key, but technically it doesn't matter

sondavidb (approved):
Thanks!

sondavidb:
@mukul-atomicwork could you please squash your commits?

shashank-boyapally (approved):

sondavidb:
Oh I forgot I can do it myself. Hopefully this doesn't cause issues

PR #2075, opened 2026-09-06
Fix resolveSpan passing an absolute offset as a span-relative one

prafgup:
**Issue #, if available:**

Checks failing on main after the PR https://github.com/awslabs/soci-snapshotter/pull/2072 was merged to upgrade go version - https://github.com/awslabs/soci-snapshotter/actions/runs/33809525300/job/100837517543

```
--- FAIL: TestSpanManager (1.49s)
    --- FAIL: TestSpanManager/a_file_from_20_spans (1.49s)
panic: runtime error: slice bounds out of range [:1312256] with capacity 0 [recovered, repanicked]

```

All the other places in this test file uses relative offset like this - `m.getSpanContent(0, 0, s.endUncompOffset-s.startUncompOffset)` apart from this specific call inside `resolveSpan` helper function.

The current issue we caught is also mentioned here - https://github.com/golang/go/commit/2747d887eb683f4fca0ed1350b3cc1aec860cf70

**Description of changes:**

Disclaimer I used AI to figure out this issue and fix it -

`TestSpanManager/a_file_from_20_spans` panics on Go 1.27 with `slice bounds out of range [:1312256] with capacity 0`.

The test-only `resolveSpan` helper passed the span's *absolute* `endUncompOffset` to `getSpanContent`, which expects offsets *relative* to the start of the span. Every other caller correctly passes `endUncompOffset - startUncompOffset`.

The bug has been there since e193fe82 but was invisible: by the time the test calls `resolveSpan`, every span the file touches is already cached and uncompressed, so `getSpanContent` takes the cache path and builds a lazy `io.SectionReader` -  an oversized size there just EOFs early and is never read.

Go 1.27 didn't break anything; it just produced the first span layout that exercises the bug. Its `compress/flate` emits a different block layout, so the ztoc's checkpoints land differently:

| | spans | last span |
|---|---|---|
| Go 1.25 | MaxSpanID=16, 81920 B/span | `uncomp=[1294793, 1312256)`, size 17463 |
| Go 1.27.1 | MaxSpanID=14, 98304 B/span | `uncomp=[1312256, 1312256)`, **size 0** |

Under 1.27 the final checkpoint lands exactly at the end of the uncompressed stream, creating an empty trailing span (legal - `buildAllSpans` skips the monotonicity check for `i == MaxSpanID`). It sits past the file's content range, so `GetContents` never touches it and it is still `unrequested` when `resolveSpan` reaches it. That takes the fetch path, `uncompressSpan` returns an empty buffer for a zero-size span, and slicing it to the archive size panics.

Production code is unaffected: the exported `ResolveSpan` calls `fetchAndCacheSpan` directly, and `GetContents` derives relative offsets via `getSpanInfo`.

**Testing performed:**

```
`./fs/span-manager/...` passes under both Go 1.25 and Go 1.27.1 (it panics
under 1.27.1 without this change). Full `make test` passes under 1.27.1.
```

### DIFF
--- a/fs/span-manager/span_manager.go
+++ b/fs/span-manager/span_manager.go
@@ -210,8 +210,11 @@ func (m *SpanManager) resolveSpan(spanID compression.SpanID) error {
 		return ErrExceedMaxSpan
 	}
 
-	// this func itself doesn't use the returned span data
-	_, err := m.getSpanContent(spanID, 0, m.spans[spanID].endUncompOffset)
+	// this func itself doesn't use the returned span data.
+	// `getSpanContent` takes offsets relative to the start of the span, so the
+	// whole span is [0, uncompressed span size).
+	s := m.spans[spanID]
+	_, err := m.getSpanContent(spanID, 0, s.endUncompOffset-s.startUncompOffset)
 	return err
 }
 


prafgup:
Some more explanation by AI - 

What's actually happening

Two independent things had to line up. The bug was already there; Go 1.27 supplied the input shape that exposed it.

1. Go 1.27 doubled flate's stored-block size. The test data is random, so flate can't compress it and emits stored (raw) DEFLATE blocks. I parsed the block headers directly on 1 MiB of random data:

go1.25:   LEN=16384  16384  16387  16387  16384 ...   gzip total 1048919
go1.27.1: LEN=32768  32768  32774  32768  32774 ...   gzip total 1048756

16 KiB blocks → 32 KiB blocks. That's also the whole reason the archive got 152 bytes smaller: fewer block headers.

2. soci's checkpoints land on block boundaries. gzip_zinfo.c:280 creates a checkpoint at a DEFLATE block boundary once more than span_size (64 KiB) of uncompressed data has passed since the last one:

- 1.25: 4×16 KiB = 65536, not > 65536, so it takes 5 blocks → spans of 81920 bytes
- 1.27: 2×32 KiB = 65536, not > 65536, so it takes 3 blocks → spans of 98304 bytes

3. With 98304-byte spans, a checkpoint lands exactly at the end of the data. All 1312256 uncompressed bytes have been emitted, and the only thing left in the stream is flate's final BFINAL block, which carries no data. Result: span 14 uncomp=[1312256, 1312256) — zero bytes long. Legal, just degenerate. Under 1.25 the last checkpoint fell at 1294793, leaving 17463 real bytes, so no empty span.

4. Nothing requests that span. The test file occupies [512, 1311232); the empty span sits past it (in the tar's 1024-byte trailer region), so GetContents never touches it and it stays unrequested.

5. The panic. The test loop calls resolveSpan(i) for every span, passing the absolute endUncompOffset (1312256) where a span-relative size was expected:

- Spans 0–13 were already cached and uncompressed by the earlier GetContents, so they took the cache branch. That builds a lazy io.SectionReader — an oversized size just EOFs early and is never read. The bug is invisible here. This is why it hid since e193fe82.
- Span 14 is unrequested, so it took the fetch branch: uncompressSpan returns []byte{} for a zero-size span, then uncompBuf[0 : 0+1312256] slices an empty buffer → slice bounds out of range [:1312256] with capacity 0.

That's also why the bound in the panic message was 1312256 — it's the whole UncompressedArchiveSize, which is the tell that an absolute offset was being used as a length.

prafgup:
cc @sondavidb

coderbirju (approved):

sondavidb (approved):
Maybe a good sign we need some extra randomness in our testing...


Thanks!

PR #2076, opened 2026-09-07
fix(fs): prevent SIGSEGV crash and ImagePullBackOff hang on cancelled parallel pulls (#2036)

### LINKED ISSUE #2036: [Bug] parallel_pull_unpack: concurrent duplicate pull of the same image removes the in-flight image job; nil dereference can crash the daemon
## Description

In `parallel_pull_unpack` mode, two concurrent pulls of the same (unindexed) image race on the shared image-job map. The second pull's failure/cleanup path calls `RemoveImageWithError`, deleting the image job out from under the first pull. Best case, both pulls fail. Worst case, the snapshotter daemon crashes with a nil-pointer SIGSEGV — and a snapshotter crash takes every running container on the node down with it.

## Evidence

Over a 3-day window on a large fleet we observed 8 daemon panics, every one at the identical program counter, with fault address `0x50` matching the offset of the `.layers` field, and the stack:

```
fs.preloadAllLayers -> unpackJobs.AddLayerJob -> fs.MountParallel
```

Telemetry at the crash instant showed two snapshot prepares for the same image logging "no SOCI index found; falling back to parallel pull/unpack" **in the same millisecond**, ~5 ms before the SIGSEGV. The trigger in production is a scheduler placing many identical jobs on one node simultaneously; crashes clustered exactly during those launch bursts.

## Root cause (v0.14.0 source)

- `MountParallel` does a check-then-act on the image job with no atomic claim/join (`fs/fs.go:466`).
- `RemoveImageWithError` `delete`s the shared image job (`fs/adaptive_fetch_image_layers.go:446`).
- `AddLayerJob` then dereferences the now-nil `.layers` map (`fs/adaptive_fetch_image_layers.go:213-214`).

There is no guard that makes a second concurrent pull of the same image *join* the in-flight job; creation and removal race freely. Note PR #1638 (in v0.14.0) widened the `RemoveImageWithError` call sites for failure cleanup but did not add a duplicate-pull guard — it arguably widened this race's surface.

## Reproduction

~8 concurrent pulls of the same unindexed multi-layer image reproduce the job-deletion failure on demand; the SIGSEGV is the probabilistic worst case of the same race (deterministic crash site, probabilistic trigger).

## Workaround

Disable the parallel pull fallback for unindexed images so they route through containerd's sequential puller. Validated with an A/B/A harness on live nodes: stock config crashes, fallback-disabled config survives 16 consecutive rounds of forced concurrent pulls.

## Versions

v0.14.0; code identical at v0.14.1; functionally identical on `main` (containerd v2 import drift only).

## Suggested fix

Reference-count or single-flight the image job: a duplicate pull should join the in-flight job (or wait on it), and `RemoveImageWithError` should only remove the job when the last referent releases it.

coderbirju:
## Summary

Fixes failure modes of `parallel-pull-unpack` mode reported in #2036:

- **Daemon crash (SIGSEGV):** `panic: nil pointer dereference [signal SIGSEGV addr=0x50]` at `unpackJobs.AddLayerJob` (the `.layers` field offset), taking every container on the node down.
- **Silent node poisoning / ImagePullBackOff hang:** a `(node, image)` pair wedges — every pull retry fails with `failed to rebase layer …: context canceled`, the node stays `Ready=True`, and it never recovers.

## Root cause

`fs.MountParallel` runs once per layer. The first layer's `preloadAllLayers` creates a single shared `imageUnpackJob` and spawns premount goroutines for every layer on a shared premount context. Crucially, each layer job's cancel func **is** the shared image cancel (`withImageUnpackJob`: `luj.cancel = image.cancel`), and the premount/rebase failure paths called `RemoveImageWithError` (whole-image cancel + delete) and `layerJob.Cancel()`.

So when a pull is cancelled or a layer fails, the shared image job is torn down. If a **concurrent duplicate pull of the same image** is in flight (e.g. several pods of the same image scheduled in the same second):
- `RemoveImageWithError` deletes the map entry between another Prepare's `GetOrAddImageJob` and `AddLayerJob` → `AddLayerJob` dereferences a nil entry → **SIGSEGV**; or
- the shared premount context is cancelled mid-fetch → every sibling layer dies together (`context canceled` / `compressed digests did not match`), the pull can't finish, the 5-minute CRI `image_pull_progress_timeout` cancels it, kubelet retries ~5 min later, and the cycle repeats indefinitely → **the field hang**.

Both are the same root cause: a cancelled/failed pull tears down a shared image job that other in-flight pulls of the same image still depend on.

## Fix

- `AddLayerJob` looks the image job up under the lock and returns `ErrImageUnpackJobNotFound` instead of dereferencing a nil entry (kills the crash).
- New `layerUnpackJob.markCancelled()` sets only that layer's status to cancelled **without** invoking the shared image cancel.
- The premount and rebase failure paths now remove **only the failing layer job** (`markCancelled()` + `Remove(layerJob)`) instead of the whole image. The image job is evicted normally once its last layer job is removed, so a cancelled sibling no longer poisons a concurrent duplicate pull.

Two regression tests added: `TestAddLayerJobAfterImageRemovedDoesNotPanic` (crash) and `TestRemovingOneLayerJobKeepsSharedImageJob` (shared-job survival).

## Verification

Reproduced and verified on a single host running containerd 2.2.5 + soci-snapshotter (built from source) + CRI.

Same host/harness, this change vs stock:

| | stock v0.15.0 | this change |
|---|---|---|
| SIGSEGV panics | 8–12 | **0** |
| daemon crashes | 6+ | **0** |
| burst rounds that self-heal on retry | — | **5/5** |
| hang-style duplicate+cancel rounds that converge | — | **5/5** |
| steady-state pulls | — | **8/8** |

`go test ./fs/` (full package, including GC tests) passes. There are some flaky tests that fail, being adderessed by this PR #2075 

Fixes #2036

---
AI Assisted PR

### DIFF
--- a/fs/adaptive_fetch_image_layers.go
+++ b/fs/adaptive_fetch_image_layers.go
@@ -179,7 +179,7 @@ func checkParallelPullUnpack(cfg *config.Parallel) error {
 }
 
 // GetOrAddImageJob adds the requisite image job to unpackJobs.
-// If the job already exists, return nil.
+// If the job already exists, return it.
 // Else, return the newly created imageUnpackJob.
 func (jobs *unpackJobs) GetOrAddImageJob(imageDigest string, cancel context.CancelCauseFunc) *imageUnpackJob {
 	jobs.mu.Lock()
@@ -205,13 +205,21 @@ func (jobs *unpackJobs) AddLayerJob(imageJob *imageUnpackJob, layerDigest string
 	jobs.mu.Lock()
 	defer jobs.mu.Unlock()
 
+	// The image job may have been removed by a concurrent cancelled/failed pull
+	// of the same image between GetOrAddImageJob and here. Look it up under the
+	// lock and fail gracefully instead of dereferencing a nil map entry, which
+	// previously crashed the daemon with a SIGSEGV (see awslabs/soci-snapshotter#2036).
+	image, ok := jobs.images[imageJob.imageDigest]
+	if !ok {
+		return nil, fmt.Errorf("%w: %s", ErrImageUnpackJobNotFound, imageJob.imageDigest)
+	}
+
 	layerJob, err := newLayerUnpackJob(layerDigest, jobs.storage, withImageUnpackJob(imageJob))
 	if err != nil {
 		return nil, err
 	}
 
-	jobs.images[imageJob.imageDigest].layers[layerDigest] =
-		append(jobs.images[imageJob.imageDigest].layers[layerDigest], layerJob)
+	image.layers[layerDigest] = append(image.layers[layerDigest], layerJob)
 
 	return layerJob, nil
 }
@@ -453,6 +461,10 @@ func (jobs *unpackJobs) RemoveImageWithError(imageDigest string, cause error) er
 	}
 
 	// Cancelling an image will cancel all layer unpack jobs.
+	// The job is then fully evicted from the map so a subsequent pull of the same
+	// image rebuilds a fresh job rather than joining this cancelled one (whose
+	// premount context is dead) and failing forever with "context canceled" —
+	// the permanent (node, image) poisoning reported in this defect.
 	imageJob.Cancel(cause)
 	delete(jobs.images, imageDigest)
 	return nil
@@ -700,6 +712,17 @@ func (job *layerUnpackJob) AcquireUnpackLease(ctx context.Context) (func(), erro
 
 func (job *layerUnpackJob) Cancel(cause error) {
 	job.cancel(cause)
+	job.status.Store(LayerUnpackJobCancelled)
+}
+
+// markCancelled marks this layer job cancelled WITHOUT invoking the shared image
+// cancel function. In the current design every layer job's cancel is the image
+// job's cancel (see withImageUnpackJob), so calling Cancel() on one layer tears
+// down the shared premount context and every sibling layer with it. When a single
+// layer/pull fails but a concurrent duplicate pull of the same image is still in
+// flight, we mark just this job cancelled (so Remove can reclaim it) and leave the
+// shared context intact for the siblings.
+func (job *layerUnpackJob) markCancelled() {
 	job.status.Store(LayerUnpackJobCancelled)
 }
 
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -601,7 +601,14 @@ func (fs *filesystem) premount(ctx context.Context, desc ocispec.Descriptor, ref
 			err = cErr
 		}
 		if err != nil {
-			fs.inProgressImageUnpacks.RemoveImageWithError(layerJob.imageDigest, err)
+			// Remove ONLY this layer's job, marking it cancelled WITHOUT invoking
+			// the shared image cancel; otherwise a single layer/pull failure tears
+			// down the shared premount context and every concurrent duplicate pull
+			// of the same image with it (the #2036 SIGSEGV and the field
+			// ImagePullBackOff hang). The image job is evicted once its last layer
+			// job is removed.
+			layerJob.markCancelled()
+			fs.inProgressImageUnpacks.Remove(layerJob, err)
 		}
 		layerJob.errCh <- err
 		close(layerJob.errCh)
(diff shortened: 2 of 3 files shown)


sondavidb (commented):
IMO, the crux of the problem is the race condition between `GetOrAddImageJob` and `AddLayerJob`. If we can somehow mark an image job as continuing to be used, we can avoid the timeout for the most part. The cancel context would probably still propagate, but maybe we can send a new context?

This is just an idea but LMK what you think

fs/adaptive_fetch_image_layers.go
```diff
@@ -453,6 +461,10 @@ func (jobs *unpackJobs) RemoveImageWithError(imageDigest string, cause error) er
 	}
 
 	// Cancelling an image will cancel all layer unpack jobs.
+	// The job is then fully evicted from the map so a subsequent pull of the same
+	// image rebuilds a fresh job rather than joining this cancelled one (whose
+	// premount context is dead) and failing forever with "context canceled" —
+	// the permanent (node, image) poisoning reported in this defect.
```
sondavidb, line 467: Why is there an extra comment without any extra code? I think it's clear that cancelling it will cancel all the layer unpack jobs, this extra context isn't helpful IMO. (The phrase "this defect" lacks context anyway, nor do I think it's helpful to refer to this specific issue for this.)

fs/fs.go
```diff
@@ -601,7 +601,14 @@ func (fs *filesystem) premount(ctx context.Context, desc ocispec.Descriptor, ref
 			err = cErr
 		}
 		if err != nil {
-			fs.inProgressImageUnpacks.RemoveImageWithError(layerJob.imageDigest, err)
+			// Remove ONLY this layer's job, marking it cancelled WITHOUT invoking
+			// the shared image cancel; otherwise a single layer/pull failure tears
+			// down the shared premount context and every concurrent duplicate pull
+			// of the same image with it (the #2036 SIGSEGV and the field
+			// ImagePullBackOff hang). The image job is evicted once its last layer
+			// job is removed.
```
sondavidb, line 609: Can we condense this comment to something like `// Mark as cancelled to remove it in Remove()`? It's a lot more succinct.

fs/fs.go
```diff
@@ -648,8 +655,13 @@ func (fs *filesystem) rebase(ctx context.Context, dgst digest.Digest, imageDiges
...
 	defer func() {
 		if err != nil {
-			layerJob.Cancel(err)
-			fs.inProgressImageUnpacks.RemoveImageWithError(layerJob.imageDigest, err)
+			// Remove ONLY this layer's job on failure and mark it cancelled without
+			// invoking the shared image cancel, so a cancelled/failed pull cannot
+			// tear down the shared image job (and its premount context) that a
+			// concurrent duplicate pull of the same image still needs (see
+			// awslabs/soci-snapshotter#2036 and the field ImagePullBackOff hang).
+			layerJob.markCancelled()
```
sondavidb, line 663: Same here

fs/adaptive_fetch_image_layers.go
```diff
@@ -703,6 +715,17 @@ func (job *layerUnpackJob) Cancel(cause error) {
...
 }
 
+// markCancelled marks this layer job cancelled WITHOUT invoking the shared image
+// cancel function. In the current design every layer job's cancel is the image
+// job's cancel (see withImageUnpackJob), so calling Cancel() on one layer tears
+// down the shared premount context and every sibling layer with it. When a single
+// layer/pull fails but a concurrent duplicate pull of the same image is still in
+// flight, we mark just this job cancelled (so Remove can reclaim it) and leave the
+// shared context intact for the siblings.
+func (job *layerUnpackJob) markCancelled() {
```
sondavidb, line 725: Is there a reason we are separating this out into a separate function? Why can't we just call `job.status.Store(LayerUnpackJobCancelled)`?

fs/adaptive_fetch_image_layers.go
```diff
@@ -205,13 +205,21 @@ func (jobs *unpackJobs) AddLayerJob(imageJob *imageUnpackJob, layerDigest string
 	jobs.mu.Lock()
 	defer jobs.mu.Unlock()
 
+	// The image job may have been removed by a concurrent cancelled/failed pull
+	// of the same image between GetOrAddImageJob and here. Look it up under the
+	// lock and fail gracefully instead of dereferencing a nil map entry, which
+	// previously crashed the daemon with a SIGSEGV (see awslabs/soci-snapshotter#2036).
```
sondavidb, line 211: Can we not refer to the specific issue here? A comment shouldn't have to refer to external links unless there's a lot of context. The issue being solved here is just ensuring we don't panic, which I don't think needs further justification

sondavidb (changes requested):
I don't think this is correct. If we cancel a job in-flight this means we will no longer be cancelling in-flight image operations that get cancelled for other reasons, so we will be leaking resources.

I'll think a little more on this, I added some small comments but I do think the implementation as a whole needs to be thought through some more.

PR #2078, opened 2026-09-08
[Feat]: Add support for custom HTTP request headers via snapshot labels

raduConstantinescu:
# Add support for custom HTTP request headers via snapshot labels

## Summary

Adding support for custom HTTP request headers via snapshot label `soci.header.<name> = <value>` which adds the header `<name>: <value>` to the registry requests that the snapshotter
makes. 

This covers byte-range fetches, liveness checks, URL refreshes, and
index/zTOC fetches.

Example: 

`soci.header.x-request-id=1234 `
would turn into:
 `x-request-id` :`1234`

## Why

Example use case: Callers of the snapshotter may want to identify or correlate their requests. 
The snapshotter fetches layer content on its own connection, separate from the
containerd pull. A caller cannot tag that traffic today. For example, a caller
cannot attach a request id. This change adds that ability.

## How

`source.HeadersFromLabels` reads the labels. `fs.fs` puts the headers on the
context with `socihttp.WithCustomHeaders` at each mount entry point:
`Mount`, `MountParallel`, and `MountLocal`. `Mount` also seeds them in the
neighboring-layer pre-resolution loop. `preloadAllLayers` copies them from the
parent context into the premount context.

Each `httpFetcher` reads the headers once when the snapshotter creates it. It
re-attaches them on every `fetch`, `check`, and `refreshURL`. `AuthClient.Do`
sets them on the request. The SOCI index and zTOC fetches use the same
`AuthClient`, so they carry the headers too.

The fetcher reads the headers once, at creation. The fetcher cache key is
ref+digest and holds no header. A cached fetcher reuses the first caller's
headers. So a per-request header cannot identify a single caller after two pulls
share a layer.

## RFC basis

- **Validation** — `golang.org/x/net/http/httpguts` checks each name as a
  `token` (RFC 9110 §5.6.2) and each value as a `field-value` (§5.5). The value
  check rejects CR and LF. This blocks header injection and request splitting
  (§11.1). The snapshotter drops an invalid name or value with a WARN.
- **Reserved headers** — the snapshotter or a shared library sets Range,
  Accept-Encoding, Accept, Content-Type, Content-Length, Authorization,
  User-Agent, and Referer. The snapshotter drops a label that names one of
  these with a WARN. The parser and the sink both drop it. The match is
  case-insensitive (§5.1).

## Testing

Unit tests: `internal/http/headers_test.go`, `fs/source/source_test.go`, and
`fs/remote/resolver_test.go`. They cover attach, absent, reserved dropped,
invalid dropped, multiple values, concurrent isolation, and a real httptest
server.

```
Before (no label):  PENDING  PASTE
After  (label set): PENDING  PASTE
Reserved/invalid:   PENDING PASTE
```

The snapshotter behaves the same as before when no `soci.header.` label is
present.


internal/http/headers.go
```diff
@@ -0,0 +1,71 @@
...
+// customHeadersKey is the context key for the custom request headers.
+type customHeadersKey struct{}
+
+// WithCustomHeaders returns a context that adds h to every request the AuthClient
+// sends under it.
+func WithCustomHeaders(ctx context.Context, h http.Header) context.Context {
+	if len(h) == 0 {
+		return ctx
+	}
+	return context.WithValue(ctx, customHeadersKey{}, h)
```
sondavidb, line 67: Should we also check if the context value is already set?

fs/source/source.go
```diff
@@ -90,8 +94,50 @@ const (
...
+// HeadersFromLabels skips a label and logs a warning when:
+//   - the header name is reserved (see [socihttp.ReservedHeaders]), or
+//   - the name or value is not a valid HTTP header field.
+//
+// A skipped label never fails the mount. The value check rejects CR and LF, so a
+// label cannot inject a second header or split the request.
+func HeadersFromLabels(ctx context.Context, labels map[string]string) http.Header {
+	h := http.Header{}
+	for k, v := range labels {
+		if v == "" {
```
sondavidb: Is there ever a use case where we just want the key to exist and have a blank value? Per a quick Google search this is technically valid but causes some bugs. Not opinionated either way though.

fs/source/source.go
```diff
@@ -90,8 +94,50 @@ const (
...
+			continue
+		}
+		name := strings.TrimPrefix(k, CustomHeaderLabelPrefix)
+		if name == k || name == "" {
```
sondavidb: if `name == ""` do we want to warn that there's an empty label?

internal/http/headers.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"context"
+	"net/http"
+)
+
+// These constants name the HTTP headers that the snapshotter or a shared library
+// sets on its own requests. A caller must not override them through a custom
+// header. ReservedHeaders lists them.
+const (
```
sondavidb, line 27: Really nice improvement here, thanks for this 🙂

internal/http/headers.go
sondavidb, line 1: Can we use the variable names to refer to headers in our test files as well? e.g. [here](https://github.com/sondavidb/soci-snapshotter/blob/27a4f87e50c3ee6040d798f86b5bce5924f58de4/fs/remote/blob_test.go#L444) and in the new tests you made.

fs/remote/resolver.go
```diff
@@ -276,6 +281,7 @@ func newHTTPFetcher(ctx context.Context, fc *fetcherConfig) (*httpFetcher, error
 
 func (f *httpFetcher) fetch(ctx context.Context, rs []region, retry bool) (multipartReadCloser, error) {
 	ctx = docker.WithScope(ctx, f.scope)
+	ctx = socihttp.WithCustomHeaders(ctx, f.customHeaders)
```
sondavidb, line 284: Noticing this is something we keep adding. Is there a more elegant way to propagate the headers aside from the context? (Is this the most elegant solution?)

sondavidb (commented):
First pass-through of it, overall looks good, need to look at tests and a couple of implementation stuff.

Thanks for this! Will also run tests soon

fs/source/source.go
raduConstantinescu: yes I think we should, actually in both cases with different warnings, name == k means passing labels without the prefix as well.

internal/http/headers.go
raduConstantinescu: Thanks! Do you think this would restrict some sort of functionality? My reasoning was introducing custom headers with overriding capabilities should 1. not break the snapshotter, 2. have clear warnings in the source code.

internal/http/headers.go
raduConstantinescu: Will think a bit about this, I believe in general if we add these constants of the headers, in case they somehow get changed in another PR, ideally tests would fail and not get "updated" because of their value, if it makes sense.

I think ideally implementation detail such as the Header Value should be decoupled from the testing, so that we can catch regressions/bugs easier.

fs/remote/resolver.go
raduConstantinescu: As far as I am aware propagating the ctx from the first entry-point all the way down is the way to do it in go, and re-freshing the ctx (prepending this value) at it's start is the best practice, for example for a request id, in this case we want the custom struct so that it does not clash ever with any other implementation

See: [These constants name the HTTP headers that the snapshotter or a shared library](https://goperf.dev/01-common-patterns/context/#propagating-request-ids-for-distributed-tracing)

> ctx := context.WithValue(context.Background(), "requestID", "12345") 

in the code examples

fs/source/source.go
sondavidb: `name == k` is the case where a label is just not related to this feature. I don't think we need a warning for that since there are a lot of other labels that are passed to this request.

internal/http/headers.go
sondavidb: Wonder if we can just use this? https://github.com/go-http-utils/headers/blob/master/headers.go (or similar)

sondavidb: > I think ideally implementation detail such as the Header Value should be decoupled from the testing, so that we can catch regressions/bugs easier.

I think that's fair, in theory this list should never change.

internal/http/headers.go
raduConstantinescu: it is a good idea, added in REV2

internal/http/headers.go
raduConstantinescu: Decided not to use https://github.com/go-http-utils/headers/blob/master/headers.go  as it is not frequently maintained, also I do not think go std lib has any constants for them, so decided to hand craft them.

fs/source/source.go
raduConstantinescu: +

fs/source/source.go
raduConstantinescu: i kept it as not allowing empty headers, for the same reason I think it is usually a "client" option which we can do

fs/remote/resolver_test.go
```diff
@@ -402,6 +404,210 @@ func (u *userAgentRoundTripper) RoundTrip(req *http.Request) (*http.Response, er
...
+	}
+	if got := rt.got.Get("x-request-id"); got != "caller-123" {
+		t.Fatalf("custom header not on the wire: got %q", got)
+	}
+	if got := rt.got.Get("Range"); got != "bytes=0-1" {
+		t.Fatalf("reserved custom Range overrode the fetcher's own range: got %q", got)
+	}
+}
+
+func TestMultipleCustomHeadersReachWireOnFetch(t *testing.T) {
```
sondavidb: We are doing a lot of redundant work like the above, wonder if we should condense into two separate test cases that go through the same run (we achieve this in a few tests, see [here](https://github.com/awslabs/soci-snapshotter/blob/be07ef0ad13821a33e3b5a830acaf12bfef3e8a3/integration/run_test.go#L57-L115) for an example)

fs/remote/resolver_test.go
```diff
@@ -402,6 +404,210 @@ func (u *userAgentRoundTripper) RoundTrip(req *http.Request) (*http.Response, er
...
+	}
+	if err := f.check(); err != nil {
+		t.Fatalf("check failed: %v", err)
+	}
+	if got := rt.got.Get("x-request-id"); got != "caller-123" {
+		t.Fatalf("custom header lost on check(): got %q", got)
+	}
+}
+
+func TestCustomHeaderCapturedFromContext(t *testing.T) {
```
sondavidb: Sorry if I'm not understanding, but what does this do that isn't covered in the above two test cases?

fs/remote/resolver_test.go
```diff
@@ -402,6 +404,210 @@ func (u *userAgentRoundTripper) RoundTrip(req *http.Request) (*http.Response, er
...
+	}
+	if _, err := f.fetch(context.Background(), []region{{b: 0, e: 1}}, true); err != nil {
+		t.Fatalf("fetch: %v", err)
+	}
+	if got := rt.got.Get("x-request-id"); got != "caller-123" {
+		t.Fatalf("header not captured from context and sent on the wire: got %q", got)
+	}
+}
+
+func TestCustomHeaderOverRealServer(t *testing.T) {
```
sondavidb: This is a little heavyweight for a unit test IMO. Is there any way we can put this into an integration test? If not this is probably acceptable?

fs/source/source_test.go
```diff
@@ -112,3 +118,170 @@ func TestNeighboringLayersSizeLengthMismatch(t *testing.T) {
...
+	if got := h.Get("Range"); got != "" {
+		t.Fatalf("reserved header Range should be dropped, got %q", got)
+	}
+	// Only the two valid headers should be present.
+	if len(h) != 2 {
+		t.Fatalf("expected exactly 2 headers, got %d: %v", len(h), h)
+	}
+}
+
+func TestHeadersFromLabels_None(t *testing.T) {
```
sondavidb: Can this not be consolidated with the above test?

fs/source/source_test.go
```diff
@@ -112,3 +118,170 @@ func TestNeighboringLayersSizeLengthMismatch(t *testing.T) {
...
+	for _, name := range invalid {
+		if !warned[name] {
+			t.Fatalf("expected a WARN for dropped invalid header %q", name)
+		}
+	}
+}
+
+// TestHeadersFromLabels_ReservedDropped verifies every reserved header name is
+// dropped (case-insensitively) and a WARN is logged for each.
+func TestHeadersFromLabels_ReservedDropped(t *testing.T) {
```
sondavidb: All of these tests feel like they are testing extremely similar behavior. Consolidating will probably make them easier to parse?

internal/http/auth.go
```diff
@@ -145,6 +145,13 @@ func (ac *AuthClient) Do(req *http.Request) (*http.Response, error) {
 		for k := range ac.header {
 			req.Header.Set(k, ac.header.Get(k))
 		}
+		for k, vals := range CustomHeaders(ctx) {
+			name := http.CanonicalHeaderKey(k)
+			if _, reserved := ReservedHeaders[name]; reserved {
```
sondavidb: nit: `reserved` -> `ok` since that's what I tend to see. I think it's clear enough what it's checking for that using a generic name is fine.

internal/http/headers.go
sondavidb: I think it's fair yeah. This seems like a good way to prohibit such behaviors from mutating the request in general.

I think there's a bit of risk here in that if we aren't super comprehensive, we can potentially send an arbitrary HTTP request, my request from before was an attempt to think about if there's a good way to make sure we cover all use cases.

internal/http/headers_test.go
```diff
@@ -0,0 +1,170 @@
...
+		t.Fatalf("ReservedHeaders has %d entries, want %d: %v", len(ReservedHeaders), len(want), ReservedHeaders)
+	}
+	for _, name := range want {
+		if _, ok := ReservedHeaders[http.CanonicalHeaderKey(name)]; !ok {
+			t.Errorf("expected %q to be reserved", name)
+		}
+	}
+}
+
+func TestCustomHeadersAttachedPerRequest(t *testing.T) {
```
sondavidb: Is this test (and other tests below) different from the `resolver_test.go` tests?

internal/http/headers_test.go
```diff
@@ -0,0 +1,170 @@
...
+			if err != nil {
+				errs <- fmt.Errorf("goroutine %d: %w", i, err)
+				return
+			}
+			if got := resp.Header.Get("x-seen-id"); got != id {
+				errs <- fmt.Errorf("goroutine %d saw request id %q, want %q", i, got, id)
+			}
+		}(i)
+	}
+	wg.Wait()
```
sondavidb: Let's make sure we don't hang forever for whatever reason via a timeout, see [here](https://github.com/awslabs/soci-snapshotter/blob/be07ef0ad13821a33e3b5a830acaf12bfef3e8a3/fs/layer/layer_test.go#L119-L127) for an example

sondavidb (commented):
A lot of nits but overall seems fine to me, would like a few things addressed

fs/remote/resolver_test.go
raduConstantinescu: sure, we can do some table testing here and reduce nr of tests

fs/source/source.go
```diff
@@ -90,8 +94,55 @@ const (
 
 	// HasSociIndexDigest is a label that tells if the layer was pulled with a SOCI index.
 	HasSociIndexDigest = "containerd.io/snapshot/remote/has.soci.index.digest"
+
+	// CustomHeaderLabelPrefix marks a snapshot label that carries a custom HTTP
+	// header for the snapshotter's registry requests. The header name is the label
+	// key after the prefix. The header value is the label value. For example:
+	//	containerd.io/snapshot/remote/soci.header.x-request-id = "<value>"
+	CustomHeaderLabelPrefix = "containerd.io/snapshot/remote/soci.header."
```
ginglis13, line 102: Since the intent of this change is for soci clients to control these headers, could we move the prefix out of `containerd.io/snapshot/` ?


```suggestion
	CustomHeaderLabelPrefix = "soci.header."
```

otherwise the header could be inherited from a pulled image's manifest

raduConstantinescu: sure it was part one the initial RFC in the PR regarding the label prefix value, it makes sense to not be associated with containerd

ginglis13 (approved):

smandhada16 (dismissed):
Looks good.

sondavidb (dismissed):
LGTM, though there are merge conflicts

sondavidb (approved):

sondavidb (approved):

shashank-boyapally (approved):

fs/source/source.go
sondavidb: Worth pointing out that this also means that this is client-controlled, not image-controlled (removing the prefix stops a malicious image from forwarding a label to the containerd client, since originally it would be forwarded from the image to the snapshotter, and now it can only be set by the containerd client). However, as it stands this can only be used by folks who are using a custom containerd client, which is not many people and makes it restrictive in its use cases.

I think we can expand upon this by allowing labels to be set via the prefix (passed by client) or via the image with an explicit allowlist set by SOCI. This way arbitrary headers wouldn't be passed without being explicitly allowed by SOCI client. I'll open an issue to track this.

sondavidb: https://github.com/awslabs/soci-snapshotter/issues/2085

sondavidb:
Pointing out my [comment](https://github.com/awslabs/soci-snapshotter/pull/2078#discussion_r4020858599), there's some followup AIs and some considerations but as-is is fine so merging.

PR #2082, opened 2026-09-10
parallel-pull: improve concurrency and fix daemon crash

### LINKED ISSUE #2036: [Bug] parallel_pull_unpack: concurrent duplicate pull of the same image removes the in-flight image job; nil dereference can crash the daemon

sondavidb:
**Issue #, if available:**
Fixes #2036 
Closes #2076 as an alternative

**Description of changes:**
This is a slight redesign of the parallel pull workflow, where the crux of the issue lies in a nil dereference between when an image is getting ready for a pull operation, and when it starts to premount. As there's no lock that stays in place between premounting and getting an image layer job, there was a race that would cause it to panic.

This specific fix is solvable enough via better locking mechanisms, which this does fix, however there's a lot of messy contention here, in particular with pulling the same image. We operate off the assumption that the same diffID will never be in-flight in SOCI (CRI has an option for this), but it doesn't stop separate requests from being de-duplicated, which is why we have a layerJob array instead of just a singular one per image pull. Solving this means separating out contexts to each request instead of just the one per image job, which required a whole restructuring. This should hopefully allow concurrent operations to both attempt an image pull, instead of one failure killing the whole image operation.

**Testing performed:**

### DIFF
--- a/fs/adaptive_fetch_image_layers.go
+++ b/fs/adaptive_fetch_image_layers.go
@@ -178,13 +178,10 @@ func checkParallelPullUnpack(cfg *config.Parallel) error {
 	return err
 }
 
-// GetOrAddImageJob adds the requisite image job to unpackJobs.
-// If the job already exists, return nil.
+// getOrAddImageJob adds the requisite image job to unpackJobs.
+// If the job already exists, return it.
 // Else, return the newly created imageUnpackJob.
-func (jobs *unpackJobs) GetOrAddImageJob(imageDigest string, cancel context.CancelCauseFunc) *imageUnpackJob {
-	jobs.mu.Lock()
-	defer jobs.mu.Unlock()
-
+func (jobs *unpackJobs) getOrAddImageJob(imageDigest string) *imageUnpackJob {
 	if jobs.imageExists(imageDigest) {
 		return jobs.images[imageDigest]
 	}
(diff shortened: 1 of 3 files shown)


coderbirju (approved):

ginglis13 (approved):

smandhada16 (approved):
lgtm

PR #2086, opened 2026-09-17
[Feature] Allowlist custom request headers and update label prefix

### LINKED ISSUE #2085: [FEATURE] Add allowlist in soci-snapshotter-grpc for custom header requests
### Description

#2078 added support for custom HTTP headers in SOCI content requests. Due to the nature of the feature, it's extremely restrictive, to the point where you need a custom containerd client to use this feature at all. This is intentional as we don't want a malicious actor to be able to send arbitrary HTTP request headers. However, we would like to make this more readily usable from other folks.

This can probably be achieved via an allowlist in the config that would only allow certain headers to be passed through the snapshotter. Any other requests mutating any other field would be rejected. (And we should probably still reject reserved headers)

We could keep it as-is, and i wouldn't be horribly upset at it, however I think that other folks who may find this feature useful would be discouraged from using it given how difficult it is to use in its current state.

_No response_

ZephirFXEC:
Containerd filters the custom-header labels introduced in #2078. Use `containerd.io/snapshot/remote/soci.header.` so those labels reach SOCI, and allow only HTTP header names configured through `[resolver] custom_headers`.

Custom headers are disabled by default. Invalid or reserved names fail configuration loading, and cloned clients retain the allowlist. The generated default config and docs are updated.

Validation: five affected library packages, daemon command tests, CLI configuration checks, lint, and generated-config consistency all passed.

Fixes #2085


config/resolver.go
```diff
@@ -56,6 +56,14 @@ type ResolverConfig struct {
...
 	// its own auth client and token exchange.
 	EnableAuthClientSharing bool `toml:"enable_auth_client_sharing"`
+
+	// AllowedRequestHeaders is an allowlist of custom request-header NAMES the
+	// snapshotter may attach to its registry fetches. Custom headers arrive as
+	// snapshot labels which any image can also set (see containerd
+	// FilterInheritedLabels), so this list is the trust boundary. Only these
+	// names pass; when unset or empty, no custom header is attached. Reserved
+	// headers are always dropped. Case-insensitive.
+	AllowedRequestHeaders []string `toml:"allowed_request_headers"`
```
sondavidb: nit: `CustomHeaders` might be clearer, since `AllowedRequestHeaders` could cause some confusion on if this would block certain required headers.

fs/source/source.go
```diff
@@ -98,8 +98,16 @@ const (
...
+	//	containerd.io/snapshot/soci.header.x-request-id = "<value>"
+	// sends "x-request-id: <value>".
+	//
+	// The prefix lives under containerd.io/snapshot/ because containerd only
+	// forwards labels with that prefix to a proxy snapshotter's Prepare (see
+	// containerd's snapshots.FilterInheritedLabels); any other label is dropped
+	// before the snapshotter sees it. An image can carry such labels too, so the
+	// snapshotter only forwards the header names listed in the resolver's
+	// allowed_request_headers config; see [socihttp.AuthClient].
+	CustomHeaderLabelPrefix = "containerd.io/snapshot/soci.header."
```
sondavidb: To stick with convention let's use `containerd.io/snapshot/remote/soci.header.`

config/resolver.go
sondavidb: Let's also add a `parseResolverConfig` or similar and add it to the list [here](https://github.com/awslabs/soci-snapshotter/blob/0479e9dcfe3ccdb08dac117959d517ec6284cd14/config/config.go#L95). We should vet the headers listed, make sure they are sensical (similar check to [here](https://github.com/awslabs/soci-snapshotter/blob/0479e9dcfe3ccdb08dac117959d517ec6284cd14/fs/source/source.go#L133-L140)), and log an error or fail if it is (I think we should hard-fail, but happy to discuss otherwise)

sondavidb (commented):
Implementation generally looks good, simple straightforward approach

I'll run tests in a moment, but a heads up that you will have to regenerate the default config via `make config`

config/config_test.go
```diff
@@ -454,3 +456,79 @@ func TestSizeParser(t *testing.T) {
...
+		"User-Agent",
+		"Referer",
+	} {
+		t.Run(name, func(t *testing.T) {
+			path := filepath.Join(t.TempDir(), "config.toml")
+			content := fmt.Sprintf("[resolver]\ncustom_headers = [%q]\n", name)
+			if err := os.WriteFile(path, []byte(content), 0644); err != nil {
+				t.Fatal(err)
+			}
+			if _, err := NewConfigFromToml(path); err == nil || !strings.Contains(err.Error(), "resolver.custom_headers") {
```
sondavidb, line 529: nit: Could consider separating out the error in the parser and doing `errors.Is` on that, but I think this shouldn't change that much so probably fine to keep it

fs/source/source_test.go
```diff
@@ -129,12 +129,12 @@ func TestHeadersFromLabels(t *testing.T) {
 		strings.ToLower(socihttp.HeaderRange),
 	}
 	labels := map[string]string{
-		CustomHeaderLabelPrefix + "x-request-id": "caller-123",
-		CustomHeaderLabelPrefix + "x with space": "v",
-		CustomHeaderLabelPrefix + "x-inject":     "v\r\nX-Evil: 1",
-		CustomHeaderLabelPrefix + "x-empty":      "",
-		CustomHeaderLabelPrefix:                  "no-name",
-		"unrelated":                              "nope",
+		"containerd.io/snapshot/remote/soci.header.x-request-id": "caller-123",
```
sondavidb, line 132: nit: any reason this is fully typed out?

sondavidb (approved):
Small nits but LGTM pending CI

fs/source/source_test.go
ZephirFXEC: So that accidentally changing`CustomHeaderLabelPrefix` fails the test

smandhada16 (approved):
ltgm

